PR #36, opened 2012-10-24
Implemented contains operator

stroan:
The expression 'a contains b' will produce true if b is completely contained within a. 
- A string B is contained in a string A is B is a substring of A. 
- An array B is contained in an array A if all elements in B are contained in some element in A. 
- An object B is contained in  object A if all of the values in B are contained in the value in A with the same key. 
- All other types are assumed to be contained in each other if they are equal.

### DIFF
--- a/parser.gen.c
+++ b/parser.gen.c
@@ -573,8 +573,8 @@ static const yytype_uint16 yyprhs[] =
        0,     0,     3,     5,     7,     8,    11,    14,    21,    27,
       31,    35,    39,    43,    47,    51,    55,    59,    63,    67,
       71,    75,    79,    83,    87,    91,    95,    99,   103,   107,
-     111,   115,   119,   123,   125,   127,   131,   137,   146,   147,
-     150,   155,   161,   165,   169,   171,   173,   177,   180,   185,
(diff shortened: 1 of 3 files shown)


dsikawat:
`$ jq '.[]|select(.key1|contains("bar")).key2' foo.json`
`jq: error (at foo.json:0): null (null) and string ("bar") cannot have their containment checked`

I'd expect the item with `key1=null` to return false and not get selected. Is contains not the right operator to use in this case?

pkoppstein:
Since you haven't shown the contents of foo.json it's difficult to be sure, but it looks like you might want to use the postfix "?" operator:

```
$ jq '.[]|select(.key1|contains("bar")).key2' <<< '[{"key1":0}]'
jq: error (at <stdin>:1): number (0) and string ("bar") cannot have their containment checked

$ jq '.[]|select(.key1|contains("bar")?).key2' <<< '[{"key1":0}]'
$
```

dsikawat:
This worked. Thanks.

sakshan:
Do we have a keyword to match only if exact string? contains is matching if it has partial match.

Issue #218, opened 2013-11-23
bignum/bigreal number representation needed


jmblog:
There are cases that jq converts extra large numbers to ones with scientific (exponent) notation.

``` text
$ cat test.json
{"income":10000000,"total":11111111}

$ cat test.json | jq '.'
{
  "total": 11111111,
  "income": 1e+07
}
```

kutzi:
Would be really really nice to have some progress here. I've just been bitten by this bug and it's really hard to catch as the numbers jq spits out looks totally legid - i.e. within the same magnitude

nicowilliams:
We have a PR (that I need to find time to finish) for 64-bit integer support (in addition to IEEE754 doubles).  We're not likely to add any kind of bignums unless someone submits a PR.  If you or anyone else wants to work on bignum support for jq, you'll need to be aware of a couple of things:

 - no GPL, no LGPL welcomed
 - jq's source code does not always `jv_free()` values known to be numbers, and valgrind can't check for that anyways given that numbers today are never allocated

I'd start by adding a compile-time option to use allocated numbers so that numeric `jv`s point to a `malloc()`ed `double`, that way failures to `jv_free()` numbers can be caught by valgrind and fixed.

Next I'd look for a suitable bignum library.  There are quite a number of them, but they'd have to be a) C-coded or otherwise have a C API, b) licensed in a way that's friendly to jq's license and jq's users.

Lastly, I'd integrate such a bignum library much like Oniguruma: as a [git] submodule that is used if `./configure` can't find it installed or if the user wants the submodule used.

gcsfred2:
Any updates?

mjustin:
Would this be the right issue to ask about having jq preserve zeros after the decimal point as well, e.g. not converting `5.0` to `5`?  [JSON is agnostic about the semantics of numbers](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf), but the programs using the JSON may very well care to differentiate between `500`, `500.0`, and `5e2` (it might be doing precision-based calculations, for instance).

As a concrete example Java's [`BigDecimal`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/math/BigDecimal.html) class keeps track of both the unscaled value and the scale.  So these could be reasonably be differentiated as different values when read by a Java program:

```java
        System.out.println(new BigDecimal("500"));
        System.out.println(new BigDecimal("500.0"));
[...]

Issue #143, opened 2013-06-05
JSON does allow better than IEEE 754 numbers


nicowilliams:
The issue was raised at the IETF JSON WG this week: does JSON permit arbitrary precision numbers?  The syntax certainly does.  Neither RFC4627 nor ECMAScript 6 (draft) preclude it, and many implementations use whatever the native language run-time, OS, and/or machine architecture provide -- often not limited to IEEE 754 64-bit real numbers.

jq currently only supports IEEE 754 64-bit numbers.  It could probably do better: also 64-bit signed integers using C int64_t (perhaps with much care regarding overflows, or perhaps not).  It could even use bignum libraries.  It's all probably too much, but I'm filing (and closing) this issue just to record this.

rsrsps:
>>I'm not sure I see any use at all for 64-bit integers. Arbitrary-sized bigints might well be useful, but I see no reason to want int64.<<

Concrete values that only make sense as a number like Unix epoch nanotime already exceed the resolution available in the implementations that use floats for large integers.  This is genuinely annoying as the Javascript implementations quietly truncate the values.

(just adding)

tomwhoiscontrary:
I just tripped myself up using jq with nanosecond-scale unix timestamps:

```
$ echo '{"timestamp":1499816949527975237}' | jq .timestamp
1499816949527975200
```

I appreciate that there are some real headaches around using different number implementations according to the value. But there are also headaches around silently changing the values in the data.

If support for numbers not exactly representable as 64-bit floats isn't on the cards, how about an error exit rather than approximation?

lfairy:
Just got bitten by this today.

The API we use at work associates each entity with a 64-bit integer ID. I started to question the output of jq when some of these IDs were showing up as equal.

leonid-s-usov:
Please try this branch #1752

PR #229, opened 2013-12-05
Make numbers use their original string representation if possible

airfrog:
Doing any math (e.g. num+0) will change the representation.

This also fixes a bunch of jv_number memory management issues that matter now that numbers actually get allocated sometimes.

See issue #218 and #143


dtolnay:
As discussed here and on many related tickets, bignum would be a better solution for this.

Issue #238, opened 2013-12-10
A REPL would be nice!


nicowilliams:
A read-eval-print loop for jq would be awesome.

dtolnay:
There is a quick prototype at #284. Parts of it may be useful to someone working on a REPL for 2.0.

fiatjaf:
There is https://github.com/charlesetc/jqr. Not very usable, but works.

nicowilliams:
@fiatjaf Yeah, I want something that is properly integrated:

 - uses libeditline
 - finds the end of jq programs / `def`s
 - stores all `def`s for subsequent `def`s/programs to see
 - lets you get back your current `def`s' definitions
 - handles `import`/`include` directives similarly

I fear that's a significant effort.

fiatjaf:
@nicowilliams, please tell me what you think of https://github.com/fiatjaf/ijq

nicowilliams:
@fiatjaf That's very interesting!  I agree that a lot can be done without changing jq at all, and maybe that's the best way for now...

creallfluharty:
For anyone else who stumbles here from google, [jiq](https://github.com/fiatjaf/jiq) (which is a fork of [jid](https://github.com/simeji/jid)) is a really convenient jq repl that has enough stars and recent commits to seem like it might last.

([it's also packaged for arch](https://aur.archlinux.org/packages/jiq/))

Freed-Wu:
`fq -i` is also a REPL of jq :smile:

PR #284, opened 2014-02-04
REPL. very first impl. lots from Python3.3.2 source

bgnori:
Hi. I tried very naive REPL implementation. 
I bet more work needed, but I have know what you want.

1) Coding style.
2) placing functions, such as "whether be jv_fgest and jv_readline in jv_file.c?" etc.
3) options "z" and "--REPL" is nice for you?
4) Scoping of "def". I expect  when using def in first line of repl, I see that function in second line.
 How do you want it to be achieved?

### DIFF
--- a/main.c
+++ b/main.c
@@ -55,6 +55,7 @@ enum {
   COLOUR_OUTPUT = 64,
   NO_COLOUR_OUTPUT = 128,
   SORTED_OUTPUT = 256,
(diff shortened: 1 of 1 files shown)


dtolnay:
I linked to this PR from the main REPL ticket #238. Closing for now, but this code may be useful to anyone else working on a REPL.

Issue #164, opened 2013-07-11
Feature request: regex (or minimally, startswith/endswith) string tests


gojomo:
String `contains()` is nice, but full regular-expression matching would be great. Or perhaps as smaller step: startswith/endswith or positional string-matching tests?

PR #431, opened 2014-06-18
Added regex support as per issue #164.

wtlangford:
jq now depends on oniguruma for regex support.
Modified configure.ac accordingly.

Added valgrind suppression file for oniguruma to prevent one-time and bounded leaks from causing tests to fail.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1,6 +1,7 @@
 #include <assert.h>
 #include <limits.h>
 #include <math.h>
+#include <oniguruma.h>
 #include <stdlib.h>
 #include <string.h>
 #include "builtin.h"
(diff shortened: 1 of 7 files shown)


guruprasannaps:
I want to validate the filename. If the file name contains the \ / : * ? " < > | I should get the result as false. I tried many things but unable to make it work. CVOuld you help me please on this. how I can validate?
test("\ / : * ? " < > |*"; "gx")] is throing an compilation error .Please help me

pkoppstein:
@guruprasannaps - You've already asked this question (#432). Please do not post the same question more than once on the same website, and please ask usage questions at https://stackoverflow.com/questions/tagged/jq

Issue #432, opened 2014-06-19
enhancement request: regex filters


pkoppstein:
jq now has basic regex support in the form of `match` and `test` (see #431 and #164) but I believe still lacks "builtin" support for the following:

```
split(REGEX)
sub(REGEX, STRING)
gsub(REGEX, STRING) 
scan(REGEX)
```

guruprasannaps:
I want to validate the filename. If the file name contains the \ / : * ? " < > | I should get the result as false. I tried many things but unable to make it work. CVOuld you help me please on this. how I can validate?
test("\ / : * ? " < > |*"; "gx")] is throing an compilation error .Please help me

pkoppstein:
@guruprasannaps - One answer:

    jq -R 'test("[\\/:*?\"<>|]") | not'

In short, the regex must be a JSON string.

In future, please ask usage questions at stackoverflow.com with the jq tag: https://stackoverflow.com/questions/tagged/jq

Thank you.

guruprasannaps:
Thank you very much for your quick support, I will ask the new question in the above link.

PR #522, opened 2014-08-01
regex filters (#432): scan, splits, split, sub, gsub

pkoppstein:
split is for backward compatibility; splits is a generator.

nwise is a helper function but it seems useful enough so it is at the top-level.

Documentation and tests have been updated, but the documentation in particular might need tweaking.


dtolnay:
As far as I can tell, these all made it into master in other commits. Please let me know if I am mistaken.

guruprasannaps:
I want to validate the filename. If the file name contains the \ / : * ? \" < > |  I should get the result as false. I tried many things but unable to make it work. CVOuld you help me please on this. how I can validate?

pkoppstein:
@guruprasannaps - You've already asked this question (#432). Please do not post the same question more than once on the same website, and please ask usage questions at https://stackoverflow.com/questions/tagged/jq

PR #530, opened 2014-08-05
MinGW doesn't contain mkstemp(), fixing.

wtlangford:
Handle this generically in case _any_ platform doesn't contain it.


nicowilliams:
It'd have been convenient for me to have noticed this earlier...

PR #559, opened 2014-08-21
Add "map_values" builtin

dreiss:
This would be useful in a script I'm writing.  Let me know if you think it is a good addition and I can add it to the docs.


dreiss:
Looks like this was resolved by https://github.com/stedolan/jq/commit/84d3203dd1fe741a76c4ecdd1ca3d8fe7c1a0268

PR #598, opened 2014-10-17
Updated v1.3 filenames for Windows

trhall:
The jq 1.3 filenames for Windows are jq-1.3.exe instead of jq.exe. Updated these so they no longer return a 404.


dtolnay:
Cherry-picked onto master in 3a43e2f. Thanks!

PR #629, opened 2014-11-25
Implement flatten\0 in terms of flatten\1

slapresta:
It's a small change but I was reviewing my implementation and having two almost identical copies of the same function side by side made me cringe. This feels better.


MrTomWhite:
It looks like this discussion and PR have gone stale since @pkoppstein 's last comment. Could we close this PR if we're not going to merge it?

nicowilliams:
Any jq function that visits potentially every node in an arbitrarily-deep structure has to be recursive and can't benefit from TCO.  Or it has to use `reduce` or `foreach` and keep an explicit stack.  Implementing `flatten/0` in terms of `flatten/1` means marginally less work later if we want to make `flatten/*` avoid recursion, so I'll take it.

nicowilliams:
Except that @slapresta needs to resubmit with no errors.

First, `flatten/0` is missing a terminating semi-colon.  Second, I get this failure:

    Testing 'flatten(-1,3,2,1)' at line number 331
    *** Expected [0,[1],[[2]],[[[3]]]], but got [0,1,2,3] for test at line number 333: flatten(-1,3,2,1)

Whether the original was wrong, the docs are wrong, or the new `flatten/1` is wrong, an update is needed.

slapresta:
Well, shit. That's indeed quite sloppy. I'll look at it and find out what's going on.

nicowilliams:
Heh.  Just always make check.  If you're only touching jq-coded functions
then you can--disable-valgrind and the tests will be very fast.

slapresta:
Right. I rebuilt the commit on top of master and fixed the missing semicolon. About the broken test, I'd argue that the expected behavior of `flatten/1` when its argument is less than 0 was never defined as a meaningful operation, and this does redefine it to something else. Of course, it this retconning won't fly, it could be implemented differently.

dtolnay:
Rebased and merged in 3d49fc5 with minor changes. I added an error message for flatten(-1) instead of defining some non-obvious behavior, and I added unit tests.

MrTomWhite:
Awesome! :sunglasses: :+1:

PR #669, opened 2015-01-14
Empty arrays join/1 to an empty string, fixes #668 bug introduced by 9760245

### LINKED ISSUE #668: join/1 should respect empty strings in input
The `join/1` builtin needs help:

    -def join($x): reduce .[] as $i (""; . + (if . == "" then $i else $x + $i end));
    +def join($x): reduce .[] as $i (null; (.//"") + (if . == null then $i else $x + $i end));

See also comments on #552.

joelpurra:
Commit 9760245 would return `null` when joining an empty array.


joelpurra:
@nicowilliams: thanks!

Issue #672, opened 2015-01-19
Module system is a mess


nicowilliams:
@joelpurra @pkoppstein @wtlangford 

It's been a few weeks since I've given this any thought.  Here's an attempt to look at the whole jq module system from a fresh perspective.

jq has no build artifacts at this time, no `.jqc` files.  This means that .jq files (sources) cannot contain "absolute" references to dependencies: relative paths are OK, absolute paths are not OK, minimum version numbers are OK, exact version numbers (and/or commit hashes, etcetera) are not.

Absolute locators/bindings have to be in an external location, in a `.jqc` or in a `.json` or some such file.  Or else we must do without absolute locators/bindings, using dependent- and/or universe-relative locators only, leaving absolute binding to an external link-editor of sorts (a/the packaging system) that can use symlinks to resolve relative bindings.

In general absolute bindings at run-time is the best answer for semantic reasons (absolute bindings mean repeatable results) and for performance reasons: the work of link-editing can be done at build- (or, libtool-like, at intstall-) time.  We don't have this luxury at this time, but we shouldn't want to close the door on a future `.jqc` world either (nor, in such a future, would we want to force all jq programs and modules to be compiled).

For user-friendliness it really helps for a language like jq (which has not build step) to support running "in place", meaning: without having to `make install` then run from the install location.

On thing that becomes obvious is that we shouldn't really have a "path" (relative or otherwise) in `import` directives in `.jq` "source" files: it's a locator, even if relative, when all we need is a "name" -- an identifier.  Ideally we'd have:

    import as foo {uri: "...", <perhaps other metadata>};

If you want to use whatever that module is, your pkg system has to fetch and install it (if it isn't already installed) relative to the install location of your dependent so you can run it.

But then we absolutely need a separate file, built by the pkg mgr (really, now functioning as a link-editor) in which to store resolved dependency information that *jq* will read.  And then we need a build step, which I think we don't and shouldn't want to _require_.

I think that's a fair synthesis of the problem we're facing.  Would you agree?

So now what?

The problem that arises is that we need to have two somewhat contradictory options work with the same source code for the dependent:

 - external file with absolute locators
 - run in-place must work

Which means we need to distinguish at run-time (since that's all we have right now) whether a dependency is intended to be run in-place or otherwise, and we need the `import` directive to provide enough information for both: the external link-edit/pkg mgr case, and the run in-place case.

[...]

PR #673, opened 2015-01-20
Working module/package system

joelpurra:
The current package system is not a mess - this should prove things work. I'll read #672 soon.

See also https://github.com/stedolan/jq/issues/659#issuecomment-70423916. If you haven't tried [`jqnpm`](https://github.com/joelpurra/jqnpm) yet, please clone it right away. There is no need for me to try to convince you if you won't even try the POC of a working system.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -2754,6 +2754,10 @@ sections:
       For paths starting with "$ORIGIN/", the path of the jq executable
       is substituted for "$ORIGIN".
 
+      For paths starting with "$PACKAGEROOT/", the path of the current or nearest
+      parent directory containing a subdirectory ".jq/" is substituted for
+      "$PACKAGEROOT".
+
       For paths starting with "./" or paths that are ".", the path of
       the including file is substituted for ".".  For top-level programs
       given on the command-line, the current directory is used.
@@ -2763,7 +2767,7 @@ sections:
 
       The default search path is the search path given to the `-L`
       command-line option, else `["~/.jq", "$ORIGIN/../lib/jq",
-      "$ORIGIN/../lib"]`.
+      "$ORIGIN/../lib", "$PACKAGEROOT/.jq/packages"]`.
 
       Null and empty string path elements terminate search path
       processing.
--- a/linker.c
+++ b/linker.c
@@ -46,6 +46,16 @@ static jv build_lib_search_chain(jq_state *jq, jv search_path, jv jq_origin, jv 
       expanded_elt = jv_string_fmt("%s/%s",
                                jv_string_value(jq_origin),
                                jv_string_value(path) + sizeof ("$ORIGIN/") - 1);
+    } else if (jv_get_kind(lib_origin) == JV_KIND_STRING &&
+               strncmp("$PACKAGEROOT/",jv_string_value(path),sizeof("$PACKAGEROOT/")-1) == 0) {
+      // Could also use non-string package_root values as error message output.
+      jv package_root = jq_find_package_root(jv_copy(lib_origin));
+      if (jv_get_kind(package_root) == JV_KIND_STRING) {
+        expanded_elt = jv_string_fmt("%s/%s",
+                                 jv_string_value(package_root),
+                                 jv_string_value(path) + sizeof ("$PACKAGEROOT/") - 1);
+      }
+      jv_free(package_root);
     } else {
       expanded_elt = expand_path(path);
       if (!jv_is_valid(expanded_elt)) {
--- a/linker.c
+++ b/linker.c
@@ -25,6 +25,29 @@ static int load_library(jq_state *jq, jv lib_path, int is_data, int raw,
                         const char *as, block *out_block,
                         struct lib_loading_state *lib_state);
 
+static jv get_package_root_search_start(jq_state *jq, jv lib_origin) {
+  // PROGRAM_ORIGIN is $PWD if PROGRAM_ORIGIN_FILE is not a PROGRAM_FROM_ACTUAL_FILE.
+  // Use PROGRAM_ORIGIN when getting the package root for PROGRAM_ORIGIN_FILE instead
+  // of lib_origin, but only in that specific case.
+  // file (more specifically a file in the same directory as the PROGRAM_ORIGIN file).
+  // TODO: pass the full file path, not just the directory path "lib_origin" to this
+  // function, then use PROGRAM_ORIGIN_FILE for comparison.
+  jv ret;
+  jv from_file_p = jq_get_attr(jq, jv_string("PROGRAM_FROM_ACTUAL_FILE"));
+  jv program_origin = jq_get_attr(jq, jv_string("PROGRAM_ORIGIN"));
+  if (jv_get_kind(from_file_p) == JV_KIND_FALSE &&
+      strcmp(jv_string_value(program_origin), jv_string_value(lib_origin)) == 0) {
+    // It's a (named/anonymous) FIFO pipe; use PROGRAM_ORIGIN (which is $PWD) instead.
+    ret = jq_get_prog_origin(jq);
+  } else {
+    ret = jv_copy(lib_origin);
+  }
+  jv_free(program_origin);
+  jv_free(from_file_p);
+  jv_free(lib_origin);
+  return ret;
+}
+
 
 // Given a lib_path to search first, creates a chain of search paths
 // in the following order:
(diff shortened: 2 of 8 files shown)


joelpurra:
In this POC I haven't fixed importing relative modules `./other-file.jq`. Nor do I use `$PACKAGEROOT/jq.json` for package `main` file path lookups yet. But they're on their way,

Zopieux:
If this idea is to be implemented upstream one day, please consider choosing another name for the `PACKAGEROOT` env var. It is way too generic as it does not include `JQ`. Env var are messy enough by themselves, would be great if library developers could choose the least-conflicting names possible!

I would suggest `JQPKGPATH` or just `JQPATH`, because it's more like a search path than a root, isn't it?
You could reuse the `PATH` syntax used by plenty of well known programs, eg.:

    export JQPATH="/usr/share/lib/jq;$HOME/.jqpkg"

joelpurra:
> If this idea is to be implemented upstream one day, please consider choosing another name for the PACKAGEROOT env var. It is way too generic as it does not include JQ.

No worries -- it's not an env var, even though it looks like one. It is [used internally when resolving context-dependent paths](https://github.com/joelpurra/jq/commit/42d20efc4d719c57d63fa031a1b3ab56cd32af8d); as is `$ORIGIN`. It is also mostly meant to be used internally in the default module search path.

That said, these strings *could* also be prefixed internally -- although I'm not sure it's necessary.

Zopieux:
My bad, thought that was part of the finding-lib-directories stuff.

joelpurra:
@Zopieux: it *is* part of finding lib dirs for the package system in this pull request -- unless we're accidentally talking about different things -- but `$PACKAGEROOT` is never resolved against any env vars.

joelpurra:
Code comments by @nicowilliams were lost during rebase; here are the direct commit links:

> https://github.com/joelpurra/jq/commit/42d20efc4d719c57d63fa031a1b3ab56cd32af8d Find package root, look for packages under 'P/.jq/packages' (proof-of-concept)


> https://github.com/joelpurra/jq/commit/1218024b736abe1f1ec97c3862cf4f5dc000ce7d Save if jq was started with -f (a filter file, a filter fifo file/pipe) or an inline filter

joelpurra:
@nicowilliams: pushed updated code, hopefully addressing your comments.

Further things to work on:

1. Detect the path to the main jq file in a package through `.main` in `jq.json` (default is `P/jq/main.jq`).
1. Automatically use the main jq file if no `jq -f code.jq` or `jq '.some.inline-filter'` has been specified.
1. Test (or fix jqnpm tests for) relative module paths work (starting with `./`, such as `./other.jq`).

joelpurra:
The default file path `P/jq/main.jq` is now used of no `-f` nor direct filter has been specified.

nicowilliams:
You can see my fixes in the `package-root` branch of my fork of jq.  Let me know what you think.

joelpurra:
@nicowilliams: great! I'll look at https://github.com/nicowilliams/jq/commit/3cd687ccd68176d8d152329c39fc6bd22ee58ba2 and https://github.com/nicowilliams/jq/commit/2afc83a867dac3498a59baca0d1550eb2d5b6d59.

joelpurra:
@nicowilliams: regarding comment in https://github.com/joelpurra/jq/commit/784fbc37ba6c42a28f2ac358ceeba08b08a3b5db for "Use P/jq/main.jq if no -f and no direct filter"; this is the old packages/projects difference confusion.

Packages (only functions, only imported as modules) and projects (non-functions are allowed, can't be imported, are used directly) are very similar and they're treated the same by jqnpm in terms of installation/management. Guess projects can be seen as "entry-points" -- this is where you'd write code, while depending on (other people's) packages.

You won't execute `jq` inside a package, but inside a project. The use case is `cat input.json | jq` which would automatically execute all of the code in `P/jq/main.jq` with dependencies. As projects are "entry-points," it's inside a project folder your ordinary jq user would write and execute code.

"Use P/jq/main.jq if no -f and no direct filter" is not a critical feature and I'm open to comments. It's partially connected to the (current lack of) `P/jq.json` handling in `jq`, as users shouldn't have to type the `.main` path set in `P/jq.json` every time they want to execute the project.

This may be superseded by `jqnpm install` automatically creating/linking projects' executables to `~/.jq/bin` (expected to be in `$PATH`) which would trigger the same functionality by pointing `jq` to `P/`.

```bash
#!/usr/bin/env jq --SOMEFLAG /path/to/P
```

nicowilliams:
@joelpurra But deciding on such things based on `$PWD` is very different from just about all other interpreters, and to me, asking for much trouble and very wrong.  I don't mind forcing users to use `#!jq -f` and I don't mind forcing users to use `-f` on the command-line.  Why is using `-f` undesirable?

(The default program, when we have a tty, is now `.`, FYI.)

joelpurra:
@nicowilliams: doesn't mean just about every other interpreter is right. I'll drop the commit though. It was meant to come later anyways, but it was easier to implement than reading `jq.json` from inside `jq` (in a smart way).

nicowilliams:
And BTW, the tests pass for me on Linux.  There was an infinite loop in the package dir search code.  You saw I removed all the logic for detecting that we've reached the root directory and replaced it with checking when `dirname()` no longer shortens `current_path` -- this is portable and safer, and outsources the dirty work to libc.

joelpurra:
@nicowilliams: checking the length is much smarter than my solution, thanks! All that should be merged.

nicowilliams:
Thanks.  I'll do a final review the next couple of days.

joelpurra:
@nicowilliams: the tests in jqnpm are more evolved (emulating remote git server and the local cache, etcetera), so please check them out. I'll keep the `-f` emulation in `jqnpm execute` for now.

I'll look at `import` relative to the currently parsed file (instead of the initial `-f` or `$PWD`) as well. Not sure I know how to access `jq.json` to get `.main` for a non-default `P/jq/main.jq` path from inside an executing `jq` though. Suggestions to both work items are welcome!

nicowilliams:
@joelpurra Import relative to module doing the import is already supported.

As to the `main.jq` thing, for now, hardcode it, but if you can read a .json file from C using `jv_load_file()`.  Though frankly I prefer that the main file be named after the module's name anyways.

joelpurra:
@dtolnay: haven't rebased my jq fork since May, so it looks like this pull-request is a bit outdated. The last thing I did was incorporate @nicowilliams' suggested changes (https://github.com/stedolan/jq/pull/673#issuecomment-105561222). Feature-wise this is a good start for a module system; I've also [implemented `jqnpm` plus bunch of packages for both testing and actual use](https://github.com/joelpurra/jqnpm/wiki).

From the first comment (https://github.com/stedolan/jq/pull/673#issuecomment-70773687):

> In this POC I haven't fixed importing relative modules ./other-file.jq. Nor do I use $PACKAGEROOT/jq.json for package main file path lookups yet. But they're on their way,

@nicowilliams says (in the previous comment https://github.com/stedolan/jq/pull/673#issuecomment-106107187) relative paths are supported, but I haven't confirmed in jqnpm myself. Barring any conflicts since I last ran the test suite in `jqnpm` this should be good to go. Please use the [joelpurra/jqnpm@jq-package-root](https://github.com/joelpurra/jqnpm/tree/jq-package-root) branch when testing this pull-request.

@dtolnay: you seem to be on a roll -- care to look at the merge conflicts in this pull request? C isn't my forte, but you seem fluent =)

By the way: gave core jq developers/contributors collaborator access to [joelpurra/jqnpm](https://github.com/joelpurra/jqnpm), so as to not hold up v1.5 more than necessary -- go ahead and implement minor fixes.

joelpurra:
@nicowilliams, @dtolnay: rebased and pushed. Please test, lint etcetera to make sure the C code is alright.

@pkoppstein, @slapresta, @wtlangford, @agordon, others interested: please test this pull request against [joelpurra/jqnpm@jq-package-root](https://github.com/joelpurra/jqnpm/tree/jq-package-root).

Testing jqnpm's with `tests/all.sh` (with `$PATH` set to your jq build and jqnpm clone) would be a good start, creating and publishing your own package would be wonderful!

joelpurra:
Any news on this?

nicowilliams:
Re-reviewing this and my fixups, I think the only part we should take is the `.jq` directory search (and the `/packages/`).  The search should be up from the directory containing the calling jq program file (or `$PWD` if the program is not in a file) up to `$HOME` (if available) or the root.

I don't really like having to hard-code things like `/proc/` and `/dev/` for detecting ways to name a program file that isn't a regular file (but where `stat()` may say that the thing is a regular file anyways).  Making users have to specify an option for saying "search from `$PWD`" seems fine to me.

joelpurra:
Haven't updated my forked `jq` branch [`joelpurra:package-root`](https://github.com/joelpurra/jq/tree/package-root) since July 2015, so it's 100+ commits behind `jq`'s `HEAD` (and I have not been tracking relevant/related changes). Either way, made it easier to `brew install` this particular `--devel` fork/branch to be able to use [`jqnpm`](https://github.com/joelpurra/jqnpm) more easily. Just an FYI.

```bash
brew tap joelpurra/joelpurra
brew unlink jqnpm
brew unlink jq
brew install jqnpm --devel
```

```bash
jq -f file/with/deep/imports.jq
```

xximwon (approved):

xximwon (approved):

Issue #630, opened 2014-11-26
Document when each feature was added


nicowilliams:
The following lousy bit of shell gives an approximation.  Some builtins appear more than once, while anything known to `parser.y` instead of `builtin.c` does not appear at all:

        prev=master;
        for i in $(seq 0 4|tac) 5;
        do
            ((n=i+1))
            v=jq-1.$i
            [[ $v = jq-1.5 ]] && v=master
            #echo BETWEEN $prev and $v;
            git log -p ${v}..$prev builtin.c |
              grep '^[+] *"def' |
            cut -d: -f1 |
            sed -e 's/^[+]  "def //' \
              -e 's/[(][^;]*;[^;]*;[^;]*;*[)].*$/\/4/' \
              -e 's/[(][^;]*;[^;]*;*[)].*$/\/3/' \
              -e 's/[(][^;]*;*[)].*$/\/2/' \
              -e 's/[(][^;][)]*$/\/1/' \
              -e 's/$/ jq-1.'"${n}/"
            prev=$v;
        done | sort -u

wtlangford:
@nicowilliams I'd like to make this coincide with the 1.5 release, so we have a new, clean version of the docs to bundle with the release.  I'm marking this as a 1.5 milestone so we don't miss it and I'll start working on it again in force.

nicowilliams:
Thanks!

pkoppstein:
@wtlangford - As you may already have noticed, there is currently a discrepancy between the current version of manual.yml and http://stedolan.github.io/jq/manual: the description of `split` and `splits` is correct in the former, but the latter describes `split2(re)`, which I doubt ever existed.  I mention this partly in the hope that the online manual can be corrected, and also in case there's any doubt about the reason for preferring the pair of names split/split2 over split/split2: it is simply that splits(re) produces a stream, whereas split(sep) produces an array. Thanks.

chiefy:
:+1: just ran into this using 1.4 and trying to use `jq -j` which I really need :santa:

dtolnay:
I am keeping this marked for the 1.5 milestone, but I think we should release with or without it. #866 will alleviate some of the pain.

nicowilliams:
While this would be nice, it is now not a show-stopper for 1.5.

dtolnay:
I think #866 has addressed the same needs as this, perhaps even in a better way. Let me know if you disagree.

tianon:
[...]

Issue #558, opened 2014-08-21
Use semantic versioning for jq releases


joelpurra:
jq is being released too seldom. I think it's because there is no particular way you can tell if changes that are introduced are breaking compatibility, so everything has to wait until a _really_ breaking release comes along. All possible minor improvements, bug fixes and additions are hindered by development of these large or breaking changes.

Using semantic versioning would allow jq to release more often, while still letting users know that it wouldn't break their jq scripts.
https://semver.org/

davidjeddy:
I would like to add my support to this imitative. X.Y.Z format release tagging. 

Ran into an issue recently where `1.6` is eval'd as a float, broke the application build pipeline. Of all the tools used to build the target application only jq has to have its version number encapsulated with quotes. :/

Excerpt:

```yml
versions:
  ...
  HELM_VER: 2.16.1
  JQ_VER: '1.6' # w/o quotes value is eval'd as a float
  TERRAFORM_VER: 0.12.18
  VAULT_VER: 1.1.0
  YQ_VER: 2.4.1
  ...
```

SemVer would bring jq inline with the very widely adopted tagging nomenclature of MAJOR.MINOR.PATCH.

nicowilliams:
@davidjeddy I hope you filed a bug about that though!  :)

I suppose for 1.7 we might call it 1.7.0, but honestly, if people packaging jq want to do that and we just call it 1.7, that's fine too.

davidjeddy:
@nicowilliams TY for the response. According to the YAML wikipedia page YAML supports float data types as part of the spec. https://en.wikipedia.org/wiki/YAML . 

I agree down stream providers could very easily add the additional patch SemVer indicator very easily. However, this would not correct the issue in situations where YQ is obtained directed from this project; as is the case in the situation that encountered the issue originally.

joelpurra:
@davidjeddy: from v1.8.0 onwards jq uses a version number with three parts.

Given the breaking changes in v1.8.0, without a bump to v2.0.0, the version number format change does not imply semantic versioning though.

- https://github.com/jqlang/jq/releases/tag/jq-1.8.0
- https://github.com/jqlang/jq/issues/2999

PR #678, opened 2015-01-27
readability

isomorphisms:
Easier to read dashes than hyphens imho.

### DIFF
--- a/docs/content/1.tutorial/default.yml
+++ b/docs/content/1.tutorial/default.yml
@@ -150,7 +150,7 @@ body:
 
   - text: |
 
-      For the rest of the examples, I'll leave out the `curl` command - it's not going to change.
+      For the rest of the examples, I'll leave out the `curl` command—it's not going to change.
 
       There's still a lot of info we don't care about there, so we'll
       restrict it down to the most interesting fields.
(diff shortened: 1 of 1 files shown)


joelpurra:
@isomorphisms: I disagree. Spaces mean the words are separate.

wtlangford:
@joelpurra - The PR uses em dashes instead of hypens, so they're longer and do _technically_ have different meaning, though I think many(most?) people won't pick up on the difference without it being explicitly pointed out.

I tried out this version of the docs, and I do like the em dashes, personally.  I'm going to go ahead and merge this in.

joelpurra:
@isomorphisms: didn't see that they're em dashes.

@wtlangford: still prefer spaces around em dashes.

Oh well.

wtlangford:
@joelpurra:  I played around with the different combinations (em dash, hyphen)x(spaces, none) and decided I liked (em dash,spaces) best.  Do not fear, though.  I'm looking towards a rebuild of the docs to coincide with the 1.5 release (which may mean they get removed again, though it is unlikely). Basically, I want whatever we end up doing wrt packages/modules/etc to end up in the documentation for it.  I'm also going to add version info to core functions/syntax (introduced in, deprecated/removed in, semantics/behavior changed in, etc).  This should also help insulate against another leakage of future functionality into the current release docs.  #630 should have the current status on that, though I haven't really been up to date in there.

The other major goal with that is to try to slim down the amount of dependencies on docs building.  Most of those exist to support serving the docs (`rake serve` in the docs folder).  I think there's a way to allow manpage building without needing the serving dependencies, but I'll have to play with it.

joelpurra:
@wtlangford: sounds good! I'm pushing for #558, to have smaller releases more often, in a structured way. It could/would/should also mean more active use of marking functionality as introduced, deprecated etcetera.

PR #692, opened 2015-02-10
Enhance from_entries to better deal with Amazon AWS Tags

kim-toms:
AWS Tags use a structure of { "Name" : "key", "Value" : "value" }

This change enhances from_entries to allow Name as the key name.  Also included is an enhancement to the documentation to better describe from_entries, and a test to ensure it works.


nicowilliams:
Thanks!

PR #699, opened 2015-02-16
Fix oniguruma detection logic

hroptatyr:
At the moment, a failed oniguruma header check would leave
HAVE_ONIGURUMA set to 1 still, resulting in a compiler error
in builtin.c.


nicowilliams:
Thanks!

PR #709, opened 2015-02-23
adds explicit bison version requirement

rubicks:


nicowilliams:
I guess this can't hurt.  But don't we already do enough in configure.ac?

rubicks:
You know, [that](https://github.com/stedolan/jq/blob/master/configure.ac#L36) is an excellent point. I don't know why or how it didn't work on my weird machine (RedHat 6.5 something something), but I strongly suspect an out-of-date `autoconf-archive`. In any case, I think we're agreed---it makes little sense to check the version of one executable in two places. Sorry for taking your time on this.

nicowilliams:
@rubicks  :)

If this becomes necessary, I'll re-open and merge this.  Thanks!

PR #714, opened 2015-03-02
Check output errors

agordon:
Hello,

Thanks for writing 'jq' - great program, I use it daily.

Attached two suggestions for improvements: the first detects output errors, making jq more robust in automated environments.
the second improves error messages with IN_PLACE.

regards,
 - assaf


agordon:
With "IN_PLACE" feature removed, this is out-dated. replaced by #719 .

PR #717, opened 2015-03-05
Fix #716: Adds wrapping and clamping to jv_array_slice.

### LINKED ISSUE #716: Make jv_array_slice indices consistent with jv_string_slice
`jv_string_slice` has this nice little bit to allow slicing by negative indices:

    if (start < 0) start = len + start;
    if (end < 0) end = len + end;

It also clamps indices to be within `[0, jv_string_length]`.

`jv_array_slice`, despite looking very similar, does not have the same behavior. It takes the simpler path of `assert`ing.

    assert(0 <= start && start <= end);
    assert(end <= jvp_array_length(a));

This seems inconsistent.

tgockel:
Now `jv_array_slice` and `jv_string_slice` behave the same with negative indices.

### DIFF
--- a/jv.c
+++ b/jv.c
@@ -276,11 +276,33 @@ static int jvp_array_equal(jv a, jv b) {
   return 1;
 }
 
+static int jvp_clamp_slice_params(int len, int *pstart, int *pend)
+{
+  if (*pstart < 0) *pstart = len + *pstart;
+  if (*pend < 0) *pend = len + *pend;
+
+  if (*pstart < 0) *pstart = 0;
+  if (*pstart > len) *pstart = len;
+  if (*pend > len) *pend = len;
+  if (*pend < *pstart) *pend = *pstart;
+  
+  return !(*pstart < 0 || *pstart > *pend || *pend > len);
+}
+
 static jv jvp_array_slice(jv a, int start, int end) {
   assert(jv_get_kind(a) == JV_KIND_ARRAY);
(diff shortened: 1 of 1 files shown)


jv.c
```diff
@@ -728,15 +750,10 @@ jv jv_string_slice(jv j, int start, int end) {
...
-  if (end < 0) end = len + end;
-
-  if (start < 0) start = 0;
-  if (start > len) start = len;
-  if (end > len) end = len;
-  if (end < start) end = start;
-  if (start < 0 || start > end || end > len)
+  if (!jvp_clamp_slice_params(len, &start, &end)) {
+    jv_free(j);
     return jv_invalid_with_msg(jv_string("Invalid string slice indices"));
```
tgockel: I noticed this while I was in here -- I believe returning here without calling `jv_free(j)` first is a leak.

tgockel:
I thought about taking out some logic there, but I believe it would change the behavior of string slicing. The `len` in `jv_string_slice` comes from `jv_string_length_bytes`, while `len` in `parse_slice` comes from `jv_string_length_codepoints`.

That said, it makes more sense to me to have `jv_string_slice` wrap in terms of code points.

nicowilliams:
Thanks.  I'm traveling today, I'll try to get to this and other pending PRs
tomorrow.

nicowilliams:
Thanks!

PR #719, opened 2015-03-06
detect and report output writing errors

agordon:
Detect output errors when the program exits.

Currently:
    $ echo '{}' | jq . > /dev/full && echo ok
    ok

with the patch:
    $ echo '{}' | jq . > /dev/full && echo ok
    Error: writing output failed: No space left on device

also apply to hardware/network/other I/O errors.

(replaces #714)

### DIFF
--- a/main.c
+++ b/main.c
@@ -476,6 +476,11 @@ int main(int argc, char* argv[]) {
   if (jq_util_input_open_errors(input_state) != 0 && ret == 0 && (options & EXIT_STATUS))
     ret = 2;
 
+  if (fclose(stdout)!=0) {
+    fprintf(stderr,"Error: writing output failed: %s\n", strerror(errno));
+    ret = 2;
+  }
+
   if (ret != 0)
     goto out;
 


nicowilliams:
Thanks.  This is clever.  Changing all the `jv_princ.c` code to handle stdio errors would be more work, though I think that would be better (do we want to try to output 10GB of data *then* report the failure on close, or would we rather fail as soon as `ENOSPC` results?  well, probably the latter, unless that meant making the common case go slower, but jq has lots of opportunities for performance improvement, so I say the latter).

The other problem is that I'm averse to closing what might be `stdout` and not opening something as `stdout` again, but here it's pretty clear that we're about to exit.  I'd propose using `freopen()`, but just glancing at the man page tells me it's not clear that we'll get an error about the close.  For now I'm inclined to take this patch and re-work it later, though perhaps moving the `fclose()` to the very last operation before returning, just in case.

(Eventually, when we add I/O builtins, we should make sure that write failures are reported immediately.)

nicowilliams:
Here's my version of your patch:

    diff --git a/main.c b/main.c
    index adc9409..d12424e 100644
    --- a/main.c
    +++ b/main.c
    @@ -477,15 +477,17 @@ int main(int argc, char* argv[]) {
       if (jq_util_input_open_errors(input_state) != 0 && ret == 0 && (options & EXIT_STATUS))
         ret = 2;

    -  if (ret != 0)
    -    goto out;
    -
     out:
       jq_util_input_free(&input_state);
       jq_teardown(&jq);
       if (ret >= 10 && (options & EXIT_STATUS))
    -    return ret - 10;
    -  if (ret >= 10)
    -    return 0;
    +    ret -= 10;
    +  else if (ret >= 10)
    +    ret = 0;
    +  /* XXX We really should catch ENOSPC and such errors in jv_dumpf()! */
    +  if (fclose(stdout) != 0) {
    +    fprintf(stderr,"Error: writing output failed: %s\n", strerror(errno));
    +    return 2;
    +  }
       return ret;
     }

What do you think?

agordon:
Hello,

The change looks good. If I understand the implications correctly, it means you'll report output errors even if other errors already encountered.

Note that even if error checking is added to `jv_dumps` (and also to the `printf` and `fflush` in `main.c:process()`) - closing stdout and checking the returned value from `fclose` is really the only standard way to verify the stream was successfully written.
for example, closing stdout is a common practice in gnu code (http://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/closeout.c) though that code handles some more complicated cases.

Thanks!
  gordon

nicowilliams:
Yes, `fclose(stdout)` has to happen no matter what, I agree (does sed do
this? does ex?).

agordon:
At least for gnu sed and other gnu utilities - yes: they all close stdout and explicitly check the result (sometimes by using `atexit()` to ensure it will happen regardless of how the program terminates).

examples:
sort: http://lingrok.org/xref/coreutils/src/sort.c#829
cat: http://lingrok.org/xref/coreutils/src/cat.c#567
tr: http://lingrok.org/xref/coreutils/src/tr.c#1741
uniq: http://lingrok.org/xref/coreutils/src/uniq.c#498
sed: http://git.savannah.gnu.org/cgit/sed.git/tree/sed/utils.c#n283

nicowilliams:
Thanks for checking!

nicowilliams:
Oy, I picked your version and pushed it.

nicowilliams:
Thanks!

PR #724, opened 2015-03-10
always propagate input errors to exit code

agordon:
Improve robustness in automated system when using exit code in shell scripts,
by exiting with code 2 if there was any input error (even overriding other
possible error exit codes).
Exit code 2 is already used to indicate system errors.

Without the patch:
   $ jq . no-such-file ; echo $?
   jq: no-such-file: No such file or directory
   0

With the patch:
   $ jq . no-such-file ; echo $?
   jq: no-such-file: No such file or directory
   2

### DIFF
--- a/main.c
+++ b/main.c
@@ -474,7 +474,7 @@ int main(int argc, char* argv[]) {
     }
   }
 
-  if (jq_util_input_open_errors(input_state) != 0 && ret == 0 && (options & EXIT_STATUS))
+  if (jq_util_input_open_errors(input_state) != 0)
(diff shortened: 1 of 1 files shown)


agordon:
A side note: I'm not sure if the removed code ( `&& ret == 0 && (options & EXIT_STATUS)`) has side effects which were desired (some cases were input errors should not be reported??).
But it seems that in most unix programs, input errors trump other runtime errors.

nicowilliams:
Picked onto master.  Thanks!

PR #727, opened 2015-03-12
mark functions which require 1.5

JuanCanham:
The online docs are the main source for a lot of people (e.g me) so merging this will prevent people (e.g me) banging their heads when the docs don't work.

### DIFF
--- a/manual/index.html
+++ b/manual/index.html
@@ -2095,7 +2095,7 @@ map(foo) | from_entries</code>, useful for doing some operation to all keys and 
               
                 <section id="flattenflattendepth">
                 <h3>
-                  <p><code>flatten</code>, <code>flatten(depth)</code></p>
+                  <p>[Requires 1.5] <code>flatten</code>, <code>flatten(depth)</code></p>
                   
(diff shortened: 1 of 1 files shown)


wtlangford:
Hey! You've started the process I've been putting off forever! The issue,
though, is that the changes need to be made to the .yml file and the script
to generate the html and man pages needs to be updated.

pkoppstein:
In the interim, would it be possible to revert the online manual so that it is in-sync with jq 1.4?

nicowilliams:
Thanks!

JuanCanham:
@wtlangford, I was worried you might say that. As this temp fix will break as soon as you update the docs  I'll try and get it done this week (including Sunday).

wtlangford:
In the meantime, @nicowilliams has taken your PR and merged it into a new
branch to use for the docs.

pkoppstein:
It would be great if the patch (or patches) could correct the references to `split2`. They should all be changed to `splits` (as in "split-with-stream-orientation").  Thank you!

PR #745, opened 2015-04-03
manual.yml: Clarify how to specify keys with ":" and special chars.

TomOnTime:


slapresta:
Thanks for taking the time to do this! Pre-emptive :+1: to making this clearer, since lots of issues are constantly opened about this.

docs/content/3.manual/manual.yml
```diff
@@ -301,6 +301,12 @@ sections:
           to the last element, -2 referring to the next to last element,
           and so on.
 
+          The `.foo` syntax only works for simply keys i.e. keys that
```
slapresta: Maybe you meant "simple", not "simply"?

docs/content/3.manual/manual.yml
```diff
@@ -301,6 +301,12 @@ sections:
...
+          are all alphanumeric characters. `.[<string>]` works with
+          keys that contain special charactors such as colons and dots.
+          For example `.["foo::bar"]` and `.["foo.bar"]` work while
+          `.foo::bar` and `.foo.bar` would not.
```
slapresta: It's weird to say that `.foo.bar` doesn't work; it does select the key `bar` in `foo`. Perhaps qualifying that "would not", as in "would not select a key named `foo.bar`", would make it clearer?

TomOnTime:
All good points.  I've updated the PR based on your suggestions.

docs/content/3.manual/manual.yml
pkoppstein: The current version of the paragraph about .foo, as well as the variants that have so far been proposed, all have their problems.  I believe the following version addresses all the inaccuracies:
```
.foo, .foo.bar

Given a JSON object, the filter .foo produces the value associated with the key named "foo" if there is one;
otherwise it produces the value null.  

The .key syntax can be used provided the key name begins with an alphabetic character or an underscore, and only contains alphanumeric characters, digits, and underscores. 

If the key contains special characters, you can surround it with double quotes (e.g. ."foo$") or use the form .[s] (e.g. .["foo$"]). If s is a string-valued variable or parameter, use .[s], like this: .[$x]
```

nicowilliams:
Cherry-picked onto master and pushed.  Thanks!

pkoppstein:
@nicowilliams - The commit message says: 
> Signed-off-by: Nicolas Williams <nico@cryptonector.com>

but the text in the version you installed is simply incorrect.  It is a step in the right direction, but here is text which I invite you to compare with what's there:

```
.foo, .foo.bar

Given a JSON object, the filter .foo produces the value associated with the key named "foo" if there is one;
otherwise it produces the value null.  

The .key syntax can be used provided the key name begins with an alphabetic character or an underscore, and only contains alphanumeric characters, digits, and underscores. 

If the key contains special characters, you can surround it with double quotes (e.g. ."foo$") or use the form .[s] (e.g. .["foo$"]). If s is a string-valued variable or parameter, use .[s], like this: .[$x]
```

Issue #729, opened 2015-03-15
Escaping characters with --raw-output


agordon:
Hello,

I'm wondering if there's a way to escape printed strings (such as '\t' and '\n') when using `--raw-output`,
and if not, would such an option be useful?

A example to illustrate the issue:
```
# When not using raw-output, strings are escaped:
$ echo '{"name":"John\tDoe","id":"hello\nworld"}' | jq '[ .name, .id ]'
[
  "John\tDoe",
  "hello\nworld"
]
$ echo '{"name":"John\tDoe","id":"hello\nworld"}' | jq '.id'
"hello\nworld"
$ echo '{"name":"John\tDoe","id":"hello\nworld"}' | jq '.name'
"John\tDoe"

# When using raw-output (and the output is one string), strings are not escaped:
$ echo '{"name":"John\tDoe","id":"hello\nworld"}' | jq -r '.name' 
John     Doe
$ echo '{"name":"John\tDoe","id":"hello\nworld"}' | jq -r '.id'
hello
world
```

This can be unexpected, as the manual tells that `--raw-output` can be used when one wants to produce non JSON, especially tabular output:
```
$ echo '{"name":"John\tDoe","id":"hello\nworld"}' | jq -r '[ .name, .id ] | join("\t") '
John     Doe       hello
world
```

Is there a way to ensure the strings are escaped even with raw-output?
or another way to safely output strings in non-json format?

Ideally the output would've been:
```
John\tDoe         hello\nworld
```

Thanks,
 gordon

pkoppstein:
If you have access to a version of jq with gsub, you could use that, as illustrated by the following:
```
$ jq -r -n '"a\tb\nc" | gsub( "\t"; "\\t") | gsub( "\n"; "\\n")'
a\tb\nc
```

joelpurra:
This works for jq 1.4:

```jq
def replace(a; b):
	split(a)
	| join(b);
```

slapresta:
The `@tsv` format filter was recently added on master. Sounds like what you want.

nicowilliams:
Yes, do try `@tsv`.

PR #751, opened 2015-04-15
@tsv: escape \r, \n, \\

agordon:
When using '@tsv' output format with '-r' option,
escape \r, \n and \\ in addition to \t.

Example:
    $ printf '{"id":"hello\\ttab\\nworld","x":43 }' | jq .
    {
      "id": "hello\ttab\nworld",
      "x": 43
    }

Before: newlines are not escaped, generating invalid TSV output:

    $ printf '{"id":"hello\\ttab\\nworld","x":43 }' \
        | ./jq-old -r '[.id,.x] | @tsv'
    hello\ttab
    world	43

After: newlines are properly escaped:

    $ printf '{"id":"hello\\ttab\\nworld","x":43 }' \
        | ./jq-new -r '[.id, .x] | @tsv'
    hello\ttab\nworld	43

Before: backslashes themselves are not escaped, so there's no way to
distinguish between escaped characters and 'real' backslashes
(in the example below, there is should not be newline, despite the
output containing "\n".

    $ printf '{"x":"hello\\ttab\\\\new world"}' \
        | ./jq-old -r '[.x]|@tsv'
    hello\ttab\new world

After: backslashes are escaped:

    $ printf '{"x":"hello\\ttab\\\\new world"}' \
        | ./jq-new -r '[.x]|@tsv'
    hello\ttab\\new world

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -383,7 +383,7 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
       assert(!strcmp(fmt_s, "tsv"));
       quotes = "";
       sep = "\t";
-      escapings = "\t\\t\0";
+      escapings = "\t\\t\0\r\\r\0\n\\n\0\\\\\\\0";
     }
     jv_free(fmt);
     if (jv_get_kind(input) != JV_KIND_ARRAY)
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1562,7 +1562,12 @@ sections:
           * `@tsv`:
 
             The input must be an array, and it is rendered as TSV
-            with tabs escaped as `\t`.
+            (tab-separated values). Each input array will be printed as
+            a single line. Fields are separated by a single
+            tab (ascii `0x09`). Input characters line-feed (ascii `0x0a`),
+            carriage-return (ascii `0x0d`), tab (ascii `0x09`) and
+            backslash (ascii `0x5c`) will be output as escape sequences
+            `\n`, `\r`, `\t`, `\\` respectively.
 
           * `@sh`:
 


agordon:
This is also relevant to issue #729 .
- gordon

pkoppstein:
@agordon -- Indeed it would be preferable if the combination of -r and @tsv would result in valid .tsv.  It follows that the current behavior with respect to \n and probably \r should be changed, but it seems to me that the current behavior with respect to \t and \\ is exactly what it should be:
```
$ echo '{"id":"tab:\t,backslash:\\","x":43 }' | jq -r  '[.id, .x] | @tsv'
tab:\t,backslash:\	43
```

The tsv "standard" (http://www.iana.org/assignments/media-types/text/tab-separated-values) only singles out TAB and EOL as special characters, so \n, \r, and \t must be handled specially.  Perhaps I'm missing something, but it seems to me that everything else should be left alone.

agordon:
@pkoppstein - my patch does not change anything about the behavior of \t - it stays the same as before.
regarding backslash - I think the current state (and your example) is not ideal.
As a demonstration, add the character "t" after the second backslash, and you'll get this:

    $ echo '{"id":"tab:\t,backslash:\\t","x":43 }' | ./jq-old -r  '[.id, .x] | @tsv'
    tab:\t,backslash:\t	43

The output prints "\t" for both instances - the first should be a tab (and thus escaped), but the second should represent two characters: backslash (which is not escaped) and "t". The output conflates these two different instances.

With my patch, the output because unambiguous: 

    $ echo '{"id":"tab:\t,backslash:\\t","x":43 }' | ./jq-new -r  '[.id, .x] | @tsv'
    tab:\t,backslash:\\t	43

pkoppstein:
@agordon wrote:

> With my patch, the output because unambiguous:
```
tab:\t,backslash:\\t    43
```
Since the tsv "standard" is quiet about the interpretation of "\", the string "\\t" has no special meaning in tsv other than the literal meaning.  You are of course free to adopt a convention of your choosing, but someone else is just as free to adopt a different convention.  In the tsv context, since only TAB and EOL have special significance, an argument could be made that the only special two-character combinations should be "\t", "\n" and "\r".

Since ```@tsv``` is not in jq 1.4, there is still some room for significant changes in its specification. What do you think of changing the specification to conform to the constraint that the combination of -r and @tsv must result in "tab-separated JSON values"?  E.g. 
```
echo '["tab:\t,backslash:\\", 43 ]' | jq_hypothetical -r  '@tsv'
```
would produce the following, where <TAB> signifies a tab character:
```
"tab:\t,backslash:\\"<TAB>43
```

agordon:
@pkoppstein - for my purposes (and the many pipelines I use jq for) this would not be useful, and would just require another step or removing the quoting characters.
I don't see the use or need to add additional constraints.

The rational is this:
1. In `--raw-output` + `@tsv`, Tab must not appear as content of a field - therefore tab must be escaped.
2. Writing newlines (`\n` and `\r` to cover linux/mac/windows) as-is will break lines, rendering downstream processing of the tabular output very complicated. Notice that in the iana document you've linked to, the text says: "Each record is represented as a single line." - therefore newlines should be escaped as well.
3. Because backslash is used as the escaping character (which is not codified in the standard, but in the unix/C world this is extremely commonly accepted) - it *must* be escaped as well. otherwise decoding text such as "\t" is ambiguous.

That's it - all other encoding/escaping is not needed, or at least not required, and certainly not standardized - and I'd recommend against it.

Adding quotes can make (down-stream) parser very complicated.
For example, one purpose of quotes is to disable the special meaning of characters.
That's why in `@csv` you can have  `A,"B,C",D` which is three fields, not four - because the quotes make the the comma inside them lose its special meaning.
It gets worse as at least some CSVs actually contain newlines inside the fields, such as:

    A,"B
    C",D

Is also a valid CSV record (just one record) with three fields, where the second field contains a newline.
I think we should keep the `@tsv` output as simple as possible.

pkoppstein:
Your proposal is reasonable -- it closely parallels http://dataprotocols.org/linear-tsv --
but I believe the assertion that:
> [backslash] must be escaped as well. otherwise decoding text such as "\t" is ambiguous.

is technically incorrect.  That is, if one adopts the convention that "\t" always means TAB and similarly for "\n and "\r" but no other two-character sequence, then there is logically no ambiguity; e.g. "a\tb", "a\\tb", "a\\\tb", and "a\b" are all unambiguous.  

So the problem comes back to the absence (at least to my knowledge) of a generally accepted unambiguous standard for tsv that supports the encoding of tabs. Even the above-mentioned linear-tsv proposal is marked as "beta".

agordon:
If I understand your approach, it means this:
1. `a\tb` means: `a`, tab, `b`.
2. `a\\tb` means: `a`, backslash, `t`, `b`
3. `a\b` means: `a`, backslash, `b` - without ambiguity because `\b` is not valid escape sequence in our definition of tsv.

While it's technically valid, it makes it more complicated both for `jq` and for any downstream parser:
because for this `jq` now needs to look-ahead for any character after a backslash in the input value, and escape it only if it's not valid. And any downstream parser will also need to look-ahead and determine if the escape sequence is valid (e.g. `\t` - generate one character: tab) or not valid (e.g. `\b` - generate two characters: `\` and `b`).

It's valid approach, but I highly recommend against it. Always escaping the backslash makes life easier for all programs. note that even in the "linear-tsv" you've mentioned they say that backslash must be escaped.

pkoppstein:
@agordon - I see that there's a python package supporting linear-tsv (https://pypi.python.org/pypi/linear-tsv/0.0.0).  A draft spec plus an implementation is certainly better than either alone.  If your PR conforms to linear-tsv then I'd suggest you mention that in the docs file. If it doesn't then, is there a good reason not to change it?  

(Btw, under the "minimalist" approach, "a\\tb" would mean: a, backslash, tab, b.)

Thanks!

agordon:
I'm not sure what any python package has to do with this...
regardless of what decision is taken about backslash escaping - the current output from 'jq' is definitely wrong with regard to newlines and backslash, and this patch fixes it.

The suggested patch does seem to agree with the 'linear-tsv' - which documents need to be adjusted?

Regarding your example of `a\\tb`: I don't know under which approach this would be interpreted as a,backslash, tab, b - seems very strange to me.
Unix standard and C standard would (and should) interpret this as: `a`, backslash, `t`, `b`:

    $ printf 'a\\tb' | od -a
    0000000   a   \   t   b
    0000004

agordon:
To expand on what I mean by unix/c, see the following C code:

    #include <stdio.h>
    void main()
    {
      char *c = "a\\tb";
      char *p = c;
      while (*p){
        printf("character '%c' ascii '%d'\n", *p, *p);
        p++;
      }
    }
  
And the output is:

    character 'a' ascii '97'
    character '\' ascii '92'
    character 't' ascii '116'
    character 'b' ascii '98'

Meaning the C parser/compiler took the string "a\\tb" and understood it to be `a`, backslash, `t`, `b`.
not with a tab.

pkoppstein:
> which documents need to be adjusted?

@tsv is documented in docs/content/3.manual/manual.yml

agordon:
@pkoppstein - added my attempt at expanding the `@tsv` documentation.
Sadly I can't test it, the ruby/rvm/bundle doesn't work for me (not a ruby expect, I can't troubleshoot it at the moment).

pkoppstein:
@agordon - I noticed you've defined:
```
escapings = "\t\\t\0\r\\r\0\n\\n\0\\\\\\\0";
```
Are any of the occurrences of "\0" beside the last one needed?

Also, while you're there, could you please change the two occurrences of:
>"The input must be an array,"

 to:  "The input must be an array of scalars,"?

Thanks again.

agordon:
Regarding the `\0` - yes. That's the way `builtin.c:escape_string()` works:
lines 339-344 build the lookup table, from a single character ( `\t` ) to a replacement string ( `\\t\0` ).
the actual replacement is in line 354.
So this encodes 4 null-terminated replacement look-up strings.

regarding rewording: I'll send a better patch tomorrow.

Thanks!

nicowilliams:

Issue #743, opened 2015-04-02
output filenames when filtering multiple input files


sparr:
When I do `jq 'whatever' *.json` I can't easily identify which lines of the output were produced by each input file. I'd like to see an option (possibly on by default) to add something like "file1.json:" before each section.

nicowilliams:
Perhaps we could have `$source` variable that names the source of the
current input.

pkoppstein:
The workaround (in bash, for example) is to proceed like so:
```
for f in *.json
do
  jq --arg file "$f" 'whatever' "$f"
done
```
where references to $file in 'whatever' will have the relevant file name.

In fact, I think this workaround is so good that if an "enhancement" is needed, it should wait until after jq 1.5 is released.

dtolnay:
Fixed in #753.

vito-c:
@pkoppstein the problem with the workaround you illustrated is what happens when I am trying to actually run a filter across a group of json files. So for example aggregating objects into an array and I want to know where each object came from.

A way to handle this could be to create an interim json object with the filename added.
```bash
alljson=""
for f in *.json
do
  alljson+=$(jq --arg file "$f" '{"data":.,"file":$file}' "$f")
done
```

vito-c:
I'm really not sure how to use `$__loc__` to get the file and line number.

pkoppstein:
@vito-c - I think `input_filename` is more relevant here.  It's in jq 1.5 and is easy to use.  (See e.g. https://stedolan.github.io/jq/manual).

vito-c:
@pkoppstein I didn't see that one! so good :+1:

iamwhil:
I'm trying to do the same thing ended up with this: 

```
for f in *.json
do
  jq --arg file "$f" '<first bit of query / select statement> | input_filename' "$f"
done
```

emanuele6:
@iamwhil that `--arg file "$f"` seems useless, why do you need to pass it to `jq`?

Also why are you even using a shell loop at all instead of just passing multiple files to `jq`?

```
$ tail -n +1 *.json
==> file1.json <==
{
    "a": "x",
    "b": "d"
}

==> file2.json <==
{
    "b": "d",
    "d": "x"
}

$ jq 'select(has("a"))|input_filename' -- *.json
"file1.json"

$ jq 'select(has("b"))|input_filename' -- *.json
"file1.json"
"file2.json"

$ jq 'select(has("d"))|input_filename' -- *.json
"file1.json"
```

[...]

Issue #677, opened 2015-01-27
Is it possible to add a random selector?


ashleycoker:
Take this as an example:

`
curl 'https://query.yahooapis.com/v1/public/yql?q=SELECT%20*%20FROM%20xml%20WHERE%20url%3D%22http%3A%2F%2Fopenmenu.com%2Fmenu%2F21573c8e-15bb-11e0-b40e-0018512e6b26%22&format=json' | jq '[.query.results.omf.menus.menu.menu_groups.menu_group[].menu_items.menu_item[] | {title: .menu_item_name, price: .menu_item_price, desc: .menu_item_description, type: "food"}]'
`

I would like to be able to add another field to the resulting object "related" which is a random number of random selections of the other titles.  For example a single object in the resulting array may look like:

    {
       "title": "FRESH SNOWPEAS",
       "price": "123",
       "desc": "a desc",
       "type": "food",
       "related": [{"title": "SAUTEED SPINACH"}, {"title": "CRIMINI MUSHROOMS"}]
    }

The related field must not contain itself, and must contain between say 1 and 5 other items.

I have read ALL the docs and cant find any way to select at random?  Surely it must be possible?
Thanks in advance

pkoppstein:
@ashleycoker  asked:
> Surely it must be possible?

I have also asked for a PRNG, and it is likely to happen (eventually), but in the meantime, you might like to use an existing PRNG for jq (see http://rosettacode.org/wiki/Linear_congruential_generator#jq), or write your own.  Perhaps http://rosettacode.org/mw/index.php?title=Van_der_Corput_sequence#jq may also be of interest.

joelpurra:
@pkoppstein wrote:
> I have also asked for a PRNG, and it is likely to happen (eventually), but in the meantime, you might like to use an existing PRNG for jq (see http://rosettacode.org/wiki/Linear_congruential_generator#jq), or write your own. Perhaps http://rosettacode.org/mw/index.php?title=Van_der_Corput_sequence#jq may also be of interest.

Very nice! Until there is a proper internal implementation, would you mind `jqnpm generate`-ing packages for these algorithms? Otherwise I could package them.

pkoppstein:
@joelpurra wrote:
> Otherwise I could package them.

Please feel free to do so. I was wondering whether it would be possible/easy/sensible to arrange for these snippets from rosettacode.org to be grouped together under "rosettacode" (or just "rc").  

[...]

Issue #650, opened 2014-12-18
ER: curl


pkoppstein:
There has been some discussion about endowing jq with curl-like
support for retrieving information (and especially JSON) from remote
resources. In order to expedite enhancements in this direction (as
well as to provide support for reading from LOCAL files), I would like
to propose that jq use the "easy interface" of libcurl.  To this end,
I am appending a stand-alone C program, parts of which could be used 
for integrating jq and libcurl to support synchronous retrieval.

I realize that libcurl may be overkill, so if there are better
alternatives that would enable similar support for synchronous
retrieval of JSON documents (both locally and remotely) to be provided
expeditiously, that is fine.  However, I'd also like to point out that
libcurl could be used in the short-run without without making a
long-term commitment to using libcurl: stability is only needed for
the jq filters.

Still, libcurl has its advantages.  It is widely used and will
presumably be supported indefinitely.  It is sophisticated and will
provide a path for jq to follow as its own capabilities become more
sophisticated.

### Motivation
One of the reasons for requesting this enhancement is that I am using
a RESTful collection of tens of thousands of JSON documents, linked
together by ids in a kind of "graph database".  For example, an id
field in one document might be "1622", referring to another JSON
document at

http://api.legiscan.com/?key=InsertKeyHere&op=getSponsor&id=1622

Similarly, there are references within this "graph database" to entities
available elsewhere on the web as JSON objects.

The ability to query local files is also a major motivation.

### Non-JSON resources
For the sake of simplicity, the following assumes that the resource being
queried will return a single JSON entity.

### Specification
In order to accommodate future enhancements, I would tentatively
propose that all parameters except the URL and timeout be passed in
via a JSON object, e.g.
```
 { "username": "jqUser", "password": "secret", 
   "headers": { "User-Agent": "jq" } }
```
One possibility along these lines would be for jq initially to support
two jq filters:
```
def curl(obj; timeout): # timeout in seconds; input is a string specifying the URL with query parameters

def curl(obj): curl(obj;10);
```
These would of course either fail (or return null) or return a JSON entity.

### Questions

What should the name of the jq filter for retrieving JSON entities be?

How should non-JSON resources be supported?

### First Steps
The following is a standalone C program with functions that could be used to integrate jq with libcurl. Please feel free to use it.
```
// The following is based on http://curl.haxx.se/libcurl/c/getinmemory.html

// gcc -lcurl curl.c

#include <stdio.h>
#include <stdlib.h>
#include <string.h>
[...]

PR #753, opened 2015-04-17
Current file function

agordon:
based on previous patch (#752), this adds 'filename' and 'line' built-in
functions to jq (discussed in #743).

Example:

    $ printf '{"a":1}\n{"a":2}\n' > 4.json
    $ printf '{"a":"hello"}\n' > 5.json
    $ ./jq '{ "file":filename, "line":line, "value":.a }' 4.json 5.json
    {
      "file": "4.json",
      "line": 1,
      "value": 1
    }
    {
      "file": "4.json",
      "line": 2,
      "value": 2
    }
    {
      "file": "5.json",
      "line": 1,
      "value": "hello"
    }

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1122,6 +1122,12 @@ static jv f_now(jq_state *jq, jv a) {
 }
 #endif
 
+static jv f_current_filename(jq_state *jq) {
+  return jq_util_input_get_current_filename(jq);
+}
+static jv f_current_line(jq_state *jq) {
+  return jq_util_input_get_current_line(jq);
+}
 
 #define LIBM_DD(name) \
   {(cfunction_ptr)f_ ## name, "_" #name, 1},
@@ -1186,6 +1192,8 @@ static const struct cfunction function_list[] = {
   {(cfunction_ptr)f_mktime, "mktime", 1},
   {(cfunction_ptr)f_gmtime, "gmtime", 1},
   {(cfunction_ptr)f_now, "now", 1},
+  {(cfunction_ptr)f_current_filename, "filename", 1},
+  {(cfunction_ptr)f_current_line, "line", 1},
 };
 #undef LIBM_DD
 
--- a/jq.h
+++ b/jq.h
@@ -47,5 +47,6 @@ int jq_util_input_open_errors(jq_util_input_state);
 int jq_util_input_read_more(jq_util_input_state);
 jv jq_util_input_next_input(jq_util_input_state);
 jv jq_util_input_next_input_cb(jq_state *, void *);
+jv jq_util_input_get_position(jq_state*);
 
 #endif /* !_JQ_H_ */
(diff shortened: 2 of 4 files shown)


nicowilliams:
It would be nice too if we could get run-time errors to include jq source location.  That would be tricky, as I've noted elsewhere (I think), because we don't have that information left in the bytecode.  We'd need something like DWARF for jq :(

nicowilliams:
I'll review.

pkoppstein:
@agordon - This tweak and PR752 are both very welcome!

With respect to the choice of names for what you have called "filename" and "line", I was thinking that to minimize the chance of conflicts with existing code, it might be better to choose names such as __filename__ and __line__ (or in C/Ruby-style __FILE__ and __LINE__).  What do you think?

Also, while you're on a roll, or whenever you have the time and energy, you might like to have a crack at these ERs:

PRNG #677
curl #650

Thanks!

agordon:
@pkoppstein - to be honest - I don't have much free time.
The patches I sent are directly helping my work when I process huge amounts of JSON data and need to troubleshoot weird errors.

Looking at #677 - it might be useful in some circumstances, but implementation a reliable, truly-random and yet portable PRNG is not trivial at all (adding a crappy one is easy, though :) ).

and for #650 - I don't understand why anyone would want that - JQ is supposed to be a stream-processing (i.e. like a unix filter program) - adding the complexity of web access with it's numerous problems and flow control sounds counter-productive. For example: how do you handle retries? or redirections (IIRC curl doesn't handle redirections automatically). So many related options that will need to be implemented - doesn't sound like it's worth it.
Similarly to the request in #650, I also download large amount of JSON data, but I wouldn't do it in JQ itself - I prefer a different programming language to handle all the different cases.
JQ fits so nicely in the unix paradigm of a 'text filter' that I would wan't to make it into something else.

pkoppstein:
@agordon wrote:
>  I don't have much free time. ...

Understood.  We appreciate your taking the time to enhance jq, even if it is for selfish reasons! :-)

Regarding #677 - yes, portability (across the jq target platforms) is required, but I suspect that those who have expressed an interest in having a PRNG in jq would not require anything like "true-randomness". Perhaps most would prefer even a crappy PRNG to none, provided it's properly documented. Once a working PRNG is in place, non-jq experts could more easily offer enhancements. 

Regarding #650 - yes, jq and curl often work nicely together, but in certain contexts, the file or URL one needs is sometimes (or often) determined dynamically. Please also note that #650 envisions a very simple approach.  Hopefully, both file: and remote URLs will be supported.

nicowilliams:
Thanks!

nicowilliams:
There seems to be a heisenbug that causes the tests to fail (jq --run-tests exits with 1 but that makes no sense because it says all tests pass and none are malformed, unless... could `fclose(stdout)` be failing?).

richardbarrell-calvium:
For anyone else who landed here from a google search: the filename/0 function was later renamed to input_filename/0. Likewise line/0 was renamed to input_line/0.

PR #754, opened 2015-04-17
print offending object in runtime error messages

agordon:
When reporting an error to the user, add information about the offending
object/value (possibly truncated).
The goal is to give a user some context regarding which input object
caused the runtime error.

* jv_print.c:
   strncpyz(): helper function
   jv_dump_string_for_errmsg() - return 'char*' of the object's string
                                 representation, truncate if needed.
* builtin.c:
   type_error(),type_error2(): use jv_dump_string_for_errmsg.
* execute.c:
   jq_next(): use jv_dump_string_for_errmsg.
* jv.h:
   declare jv_dump_string_for_errmsg.

Examples:

    $ echo '"hello"' | ./jq '-.'
    jq: error: string ("hello") cannot be negated

    $ echo '"very-long-string"' | ./jq '-.'
    jq: error: string ("very-long-...) cannot be negated

    $ echo '["1",2]' | ./jq '.|join("\t")'
    jq: error: string ("\t") and number (2) cannot be added

    $ echo '["1","2",{"a":{"b":{"c":33}}}]' | ./jq '.|join(",")'
    jq: error: string (",") and object ({"a":{"b":{...) cannot be added

    $ echo '{"a":{"b":{"c":33}}}' | ./jq '.a | @tsv'
    jq: error: object ({"b":{"c":33}}) cannot be csv-formatted, only array

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -22,18 +22,25 @@
 
 
 static jv type_error(jv bad, const char* msg) {
-  jv err = jv_invalid_with_msg(jv_string_fmt("%s %s",
+  char errbuf[15];
+  jv err = jv_invalid_with_msg(jv_string_fmt("%s (%s) %s",
                                              jv_kind_name(jv_get_kind(bad)),
-                                             msg));
+                                             jv_dump_string_for_errmsg(bad,errbuf,sizeof(errbuf)),
+                                             msg
+                                             ));
   jv_free(bad);
   return err;
 }
 
 static jv type_error2(jv bad1, jv bad2, const char* msg) {
-  jv err = jv_invalid_with_msg(jv_string_fmt("%s and %s %s",
+  char errbuf1[15],errbuf2[15];
+  jv err = jv_invalid_with_msg(jv_string_fmt("%s (%s) and %s (%s) %s",
                                              jv_kind_name(jv_get_kind(bad1)),
+                                             jv_dump_string_for_errmsg(bad1,errbuf1,sizeof(errbuf1)),
                                              jv_kind_name(jv_get_kind(bad2)),
-                                             msg));
+                                             jv_dump_string_for_errmsg(bad2,errbuf2,sizeof(errbuf2)),
+                                             msg
+                                             ));
   jv_free(bad1);
   jv_free(bad2);
   return err;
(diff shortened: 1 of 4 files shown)


agordon:
Another error-related patch.
this is orthogonal to the other one (#752 of printing the input file/line) - but together they'll make life much easier...

agordon:
BTW, the one annoying case this patch doesn't solve is 'iterate over null':

    $ echo '{"a":{"b":{"c":33}}}' | ./jq '.foo[] | .a'
    jq: error: Cannot iterate over null (null)

Because the offending object itself is null. troubleshooting that might require some intricate tweaking of the vm.

nicowilliams:
@agordon What I'm thinking of is this: in `compile.c:compile()`, for each top-level block and each function, record the location from the block in `bc->debuginfo`, then for each opcode (or, better, each opcode that can raise an error, which would require extending the `OP()` macro), add the line/column location to a locations `jv` array indexed by the opcode's position in the bytecode, then add this to the result with `bc->debuginfo = jv_object_set(bc->debuginfo, locations);`.  Finally, in `execute.c:set_error()`, we'd use the current opcode's position to find the location in the bytecode, and then format that into the error message.

nicowilliams:
I.e., the `debuginfo` field of `struct bytecode` is where jq puts debug info (obviously), and which we can extend to include file location information.

agordon:
@nicowilliams - is your comment referring to reporting the input file's name/line (#752) or to reporting the offending object (#754), or to reporting the location of the JQ code which triggered the error?

I must admit I haven't yet delved into the VM internals, so I can't much help there.

nicowilliams:
@agordon The location of the offending **jq** code.

nicowilliams:
For the "cannot iterate over null" case.

nicowilliams:
Don't worry about it.  I'll add that at some point.  I have added `$__loc__`, which produces the location in the **jq** code where it occurs, which helps in debugging.

Longer term I want all run-time errors to include information about the jq location that raised them, and perhaps it'd be nice to be able to get complete and partial stack traces, but again, that's longer term.

nicowilliams:
Thanks!

PR #755, opened 2015-04-18
string escape: skip NULs in output instead of asserting.

agordon:
While NULs are not expected in C/ASCII strings, they could be encoded
with unicode:

    $ echo '"a\u0000X"' | ./jq -r '[.]|@tsv' | od -tx1a
    jq: builtin.c:352: escape_string: Assertion `c > 0' failed.

As '\u0000' is a valid unicode symbol. The culprit is
`builtin.c:escape_string()` using `jvp_utf8_next` to iterate the string,
and it returns ascii 0 for '\u0000' - which is valid, but trigger the
assertion.

With this patch, encoded NULLs in the output are silently discarded when
escaping the string with '@tsv' or '@csv':

    $ echo '"a\u0000X"' | ./jq -r '[.]|@tsv' | od -tx1a
    0000000  61  58  0a
              a   X  nl

Note the encoded NULL is kept when not using @tsv/@csv:

    $ echo '"a\u0000X"' | ./jq '[.]'
    [
      "a\u0000X"
    ]

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -349,7 +349,8 @@ static jv escape_string(jv input, const char* escapings) {
   const char* cstart;
   int c = 0;
   while ((i = jvp_utf8_next((cstart = i), end, &c))) {
-    assert(c > 0);
+    if (c==0)
+      continue;
     if (c < 128 && lookup[c]) {
       ret = jv_string_append_str(ret, lookup[c]);
     } else {


agordon:
I've (sadly) encountered this when processing messy JSONs from another source.
Not at all sure this is the correct solution - just a possible one.

also note that it's not related to the recent escape-string change #751 - it existed before.

agordon:
Follow-up: this is not (yet) a complete fix, as it outputs a NUL when using raw-output without '@tsv':

    $ echo '"a\u0000X"' | ./jq -r '.' | od -tx1a
    0000000  61  00  58  0a
              a nul   X  nl

Which depending on your POV, could be better or worse than asserting...
(but for me, at the moment, the '@tsv' fix is sufficient).

pkoppstein:
@agordon wrote:
> NULLs in the output are silently discarded when escaping the string with '@tsv' or '@csv'

Although U+0000 is named the NULL character, it is nevertheless a "Character", so quietly dropping it seems like the wrong thing to do, at least from the JSON perspective.

In the case at least of ```@tsv``` with the -r option, wouldn't it be better either to leave them alone (in accordance apparently with the IANA spec for tsv), or to emit them as "\0" (in accordance with the jq handling of tab, return and newline)?  

A quick perusal of rfc4180.txt (CSV) suggests it is also quiet on the subject of NUL/NULL, but CSV is a mess anyway, so I am less concerned about what jq does with @csv.

agordon:
@pkoppstein - encoding of NULL in various ways (e.g. `\u0000`) might be acceptable, but I argue that NUL as ascii `0x00` should *never* appear in a standard ascii/utf-8 text file (unless the user explicitly asked for it, e.g. with `find -print0` or gnu's `sort -z` or `perl -0` etc.).

This will break every C program which does not expect NUL, including jq itself:

    $ printf 'hello no-null world\nhello world' | ./jq -R '.'
    "hello no-null world"
    "hello world"
    $ printf 'hello \0null world\nhello world' | ./jq -R '.'
    "hello hello world"

Definitions of `tsv` could vary and fluctuate, but a text file is long considered not to contain NUL, e.g. from POSIX:
http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_392
(although that definition also limits the line length, which is commonly not adhered to).

Similarly, UTF-8 (which I assume is a common encoding for JSON) is specifically designed never to contain '0x00' in valid input - exactly so it won't break C/Unix programs.

Whatever decision is made (silently discarding, or not silently, or encoding it somehow) - outputting the ascii value `0x00` should not be done (unless the output encoding explicitly allow it, such as UTF-16).

Interestingly, some JSON parsers simply disallow NULL because of this reason:
https://jansson.readthedocs.org/en/2.5/conformance.html

But at least in my case, the JSON data comes from an external source, and I can't control what's in it.
For me it's better if jq does not reject such input, and doesn't assert on it.

slapresta:
UTF-8 is *not* designed to never contain 0x00. U+0000 (that is, the NULL character) is encoded as 0x00 in UTF-8.

It *is* true, however, that multi-byte UTF-8 characters are always encoded with bytes on the 0x80-0xff range.

JSON documents should never contain 0x00, but, even if JSON string literals explicitly forbid control characters, they can encode the entirety of Unicode, including the NULL character (which is escaped as `\u0000`. In my opinion, `jq` should be able to handle those correctly in all circumstances, including `-r`.

pkoppstein:
@slapresta -- Thanks for chiming in!  In the context of @tsv and the -r option, can I infer from your comments that you would be in favor of "\0"?  That would be trivial to implement, and as I mentioned above, it would fit in with the general direction that @tsv with -r has taken.

agordon:
@slapresta - I'll rephrase: UTF-8 was designed never to produce `0x00` as part of a text stream which encodes any non-null characters - explicitly so that C programs will not encounter an unexpected NUL.
of course if someone puts a NUL in the input on purpose, it might be in the input (or cause some unintended consequences).

I think the definition what is "correctly" is exactly what we're discussing here :)

slapresta:
@pkoppstein Sorry, when I said `-r` I actually meant `-R`, referring to @agordon's previous example. That's what I'm most worried about.

On the topic of `@tsv`, it probably makes sense to go with whatever option maximises correct parsing and minimises breakage on external applications, which is probably `\0`.

@agordon You're completely right: "correctly" is completely meaningless: no one is proposing to treat NULL characters *incorrectly*. It was quite rude for me to say so, and I apologise.

I'll rephrase, too: I believe we should treat the NULL character as any regular character (or, at least, as any control character) is treated; that is, by escaping it when inside JSON string literals, by converting to its UTF-8 representation when on `-r` and by converting from it's UTF-8 representation when on `-R`.

On the specific case of `@tsv`, I believe we should convert it to its appropriate representation on TSV, whatever that is; if the standard doesn't specify, let's go by the "living standard", that is, what other TSV-using applications understand or what makes them crash or behave incorrectly.

agordon:
PR #757  is suggestion for escaping it as `\0` in both tsv and `--raw-output`.

nicowilliams:
The IANA MIME registry entry for tsv says nothing about escapings, but obviously we want some (e.g., tabs and newlines, certainly).  I'm not sure that I want NULs escaped, but if that's at all common elsewhere, then that's probably OK.  I'd like to think about whether we want raw-output mode to skip or escape NULs, or to output them as NULs, or whether we want to add a `--raw-output0`, or what.

Right now I'm not keen `--raw-output` escaping NULs as I think that is surprising, but I don't mind tsv escaping NULs.

slapresta:
@nicowilliams: :+1: It's OK if `@tsv` has to escape nulls for compatibility or any other reason, as `@sh` and others might need as well; but `--raw-*` probably shouldn't.

agordon:
Thanks all for the interesting discussion. I'm closing this, but summarized it in #759.

PR #757, opened 2015-04-18
escape input NULLs as '\0' in raw/tsv/cvs output.

agordon:
Input NULLs can be specified with "\u0000".

* builtin.c: escape_string(): always encode NUL (ascii 0x00) as '\0'.
  NOTE: this can't be specified in 'f_format()' in 'escapings' variable
  because the NUL itself will terminate the lookup loop in
  builtin.c:340.
* jv_print.c: new function 'jv_string_write_raw()' to print a JV_STRING
  to I/O stream (e.g. stderr/stdout) while escaping NUL to '\0'.
* jv.h: mention 'jv_string_write_raw()'.
* main.c: use 'jv_string_write_raw' in 'process()'.

Examples:

JSON (non-raw) output, same as before, NULs are encoded:

    $ echo '"a\u0000b"' | ./jq .
   "a\u0000b"

raw output without tsv/cvs, uses 'jv_string_write_raw':

    $ echo '"a\u0000b"' | ./jq -r .
    a\0b

raw output with tsv/cvs, uses 'escape_string':

    $ echo '"a\u0000b"' | ./jq -r '[.]|@tsv'
    a\0b

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -336,6 +336,7 @@ static jv escape_string(jv input, const char* escapings) {
   assert(jv_get_kind(input) == JV_KIND_STRING);
   const char* lookup[128] = {0};
   const char* p = escapings;
+  lookup[0] = "\\0";
   while (*p) {
     lookup[(int)*p] = p+1;
     p++;
@@ -349,7 +350,6 @@ static jv escape_string(jv input, const char* escapings) {
   const char* cstart;
   int c = 0;
   while ((i = jvp_utf8_next((cstart = i), end, &c))) {
-    assert(c > 0);
     if (c < 128 && lookup[c]) {
       ret = jv_string_append_str(ret, lookup[c]);
     } else {
--- a/jv.h
+++ b/jv.h
@@ -119,6 +119,7 @@ jv jv_string_append_str(jv a, const char* str);
 jv jv_string_split(jv j, jv sep);
 jv jv_string_explode(jv j);
 jv jv_string_implode(jv j);
+int jv_string_write_raw(jv v, FILE* stream);
 
 jv jv_object(void);
 jv jv_object_get(jv object, jv key);
--- a/jv_print.c
+++ b/jv_print.c
@@ -313,3 +313,41 @@ jv jv_dump_string(jv x, int flags) {
   jvp_dtoa_context_free(&C);
   return s;
 }
+
+/* A wrapper for fwrite, writing the content of the string
+   to a standard stream, EXCEPT escaping NUL as "\0".
+   other characters are printed as-is.
+
+   returns:
+   ZERO if all data was written successfully.
+   NON-ZERO if there was an I/O error with 'fwrite/fputc',
+            use ferror/errno to find the error.
+*/
+int jv_string_write_raw(jv v, FILE* stream)
+{
+  size_t len = jv_string_length_bytes(jv_copy(v));
+  const char* buf = jv_string_value(v);
+  void*  pnul = memchr(buf, 0, len);
+  while (pnul) {
+    /* null found, write around it */
+    const size_t print_len = pnul - (void*)buf;
+    if (fwrite(buf,1,print_len,stream) != print_len)
+      return -1;
+    //Write the NUL escape sequence.
+    //TODO: change this to 'N' to change the escape sequence to '\N'
+    if (fputc('\\',stream)==EOF)
+      return -1;
+    if (fputc('0',stream)==EOF)
+      return -1;
+    // find next null (if any)
+    buf += print_len + 1;
+    len -= print_len + 1;
+    pnul = memchr(buf, 0, len);
+  }
+  /* write the remaining string */
+  if (fwrite(buf,1,len,stream) != len)
+    return -1;
+
+  return 0;
+}
+
--- a/main.c
+++ b/main.c
@@ -106,7 +106,7 @@ static int process(jq_state *jq, jv value, int flags, int dumpopts) {
   jv result;
   while (jv_is_valid(result = jq_next(jq))) {
     if ((options & RAW_OUTPUT) && jv_get_kind(result) == JV_KIND_STRING) {
-      fwrite(jv_string_value(result), 1, jv_string_length_bytes(jv_copy(result)), stdout);
+      jv_string_write_raw(result, stdout);
       ret = 0;
       jv_free(result);
     } else {


agordon:
It's never that easy :)

In the weird combination of `@tsv` but not `--raw-output`, the string is escaped twice: once the NUL is escaped to `\0`, and then the backslash is escaped to `\\`:

    $ echo '"a\u0000b"' | ./jq '[.]|@tsv'
    "a\\0b"

Not sure why would anyone use `@tsv/@csv` without `--raw-output`, but still...

pkoppstein:
@agordon @nicowilliams @slapresta @wtlangford - I've tested PR757 and think the behavior with respect to `@tsv` with the -r option is excellent.

However, there are some problems elsewhere, as discussed
in the # COMMENTS below:
```
~/github/agordon:jq$ ./jq --version
jq-1.5rc1-55-gf09a55b

$ /usr/local/bin/jq --version
jq-1.4

~/github/agordon:jq$ ./jq -n '"x\u0000y\tz"'
"x\u0000y\tz"
# EXCELLENT

~/github/agordon:jq$ ./jq -n -r '"x\u0000y\tz"'
x\0y	z
# This represents a major change w.r.t. the handling of NUL.
# The raw version of "\t" is the tab character, so why should the 
# raw version of "u0000" be "\0"?

~/github/agordon:jq$ ./jq -n -r '"x\u0000y\tz" | [.] | @tsv'
x\0y\tz
# EXCELLENT!

$ /usr/local/bin/jq -n -r '"x\u0000y\tz" | [.] | @csv'
"xy	z"

~/github/agordon:jq$ ./jq -n -r '"x\u0000y\tz" | [.] | @csv'
"x\0y	z"
# The 1.4 behavior was fine by me, but CSV is a free-for-all, so maybe
# this change is OK, but it should be documented.

~/github/agordon:jq$ ./jq -n '"x\u0000y\tz\u0001w" | [.] | @tsv'
"x\\0y\\tz\u0001w"
# The line above highlights the anomalous treatment of "\u0000" 
# in this new version. Since jq is JSON-oriented, I would rather see
# "\u0000" in the line above.
```

agordon:
@pkoppstein - thanks for testing! you raise good points. Especially regarding 'if `\t` is not escaped, why does NUL' ?

It could boil down to a matter of judgement call, which we just need to make.
I can think of few output modes, I'll try to summarize them, please add as you see fit:

1. standard JSON output (meaning no `--raw-output` and no format specifiers like `@tsv/@csv`): all characters as escaped, because JSON does not allow embedded control characters (ASCII 0x00 -> 0x1f).  Tab (ASCII `0x09`) is escaped as `\t`. NUL (ASCII `0x00`) is escaped as `\u0000`.

        $ echo '"hello\tworld"' | ./jq '.'
        "hello\tworld"
        $ echo '"hello\u0009world"' | ./jq '.'
        "hello\tworld"

2. Raw output `--raw-output` without format specifiers: this needs to be decided.
    tab, carriage-returns,line-feeds are currently written as-is. Perhaps NUL should be as well?
    this will be consistent with the accepted unix behavior:

        $ echo '"hello\tfoobar\nworld"' | ./jq --raw-output '.'
        hello	foobar
        world
        $ printf "hello\tfoobar\nworld"
        hello	foobar
        world

        $ echo '"foo\u0000bar"' | jq --raw-output '.' | od -a
        0000000   f   o   o nul   b   a   r  nl

        $ printf 'foo\u0000bar' | od -a
        0000000   f   o   o nul   b   a   r

    And we can decide that if someone is using raw-output, he should be aware that NULs in the input will be NULs in the output. One small thing that worries me, is that this opens up a way to really mess-up unsuspecting programs. Imagine someone using jq (as executable or the library) and having to process JSON input form external sources. A malicious user could easily send `\u0000`, and the output will be NUL, which could cause lots of damage down-stream. Perhaps a warning that `--raw-output` could produce such things is sufficient.

3. Using format-specifiers (e.g. `@tsv/@csv`) without `--raw-output`. I must admit I don't see the usefulness of this thing, but it's technically possible. Not sure what should be the escaping policy.

4. Using format-specifier `@tsv` with `--raw-output` - Here I think all control characters must be escaped. A tab (ascii `0x09`) in the output should only be used for field separators, and a tab in the input must be output as two characters `\t` - same for NUL and Newlines, and all control characters (ASCII `0x00` to `0x1f`).

        $ echo '"hello\tfoo\u0000bar\nworld"' | ./jq --raw-output '[.]|@tsv'
        hello\tfoo\0bar\nworld

5. Other format-specifiers: not sure... could depend on the actual format, and the internal with/without `--raw-output`.

WDYT?

nicowilliams:
@agordon There is no connection between `@tsv`/`@csv` and jq output mode.

JSON does escape in *encoded* strings, but JSON strings can notionally contain any Unicode codepoints (there may be a couple of exceptions, but let's not go there), and the data model for *parsed* strings is that they may contain any Unicode codepoints.

Raw output mode does not escape anything.

There is an out though: write a jq-coded function to escape NULs in strings, then you can write

    @tsv "\(...)" | escape_nuls

How about that?

pkoppstein:
@agordon -- I am not sure whether you have fully taken into account that in an important sense, jq is "frozen" with respect to its 1.4 behavior.  Of course, since @tsv is not in 1.4 at all, there is still time to tweak it and its interactions with "-r".

In the short term, it would also be helpful to focus on @tsv and its interactions for another reason: there are many of us who would like to see the release of jq 1.5 ASAP, so the fewer "entanglements" the better.  

With respect to the combination of @tsv and -r, I would like to stay on the path that we were on (the "linear-tsv" path), with the understanding that NUL should be added to the list of characters that are necessarily special to the tsv format.  That is, the only control characters that would be backslash-escaped would be TAB NEWLINE RETURN and NUL.  That's not to say that there couldn't be another filter or command-line option to modify the handling of control characters, but in practice I suspect that `gsub` will usually suffice.

With respect to new filters and new command-line options, I was thinking that since -r is largely frozen, it might be appropriate to add an output option such as --utf8 that would authorize jq to emit non-JSON output in a way that is different from -r.

nicowilliams:
@pkoppstein You're complicating things too much.  The obvious thing to do is to add a function that escapes NULs, so that one can jus `jq -r '@tsv | escape_nul'`.

pkoppstein:
@nicowilliams wrote:
> You're complicating things too much.

I'm surprised and disappointed that you have that impression. In fact, I'm trying to keep things as simple as possible for jq 1.5.

In fact, the only change (relative to the current jq (jq-1.5rc1-60-gb6cc00f)) that I think is needed for jq 1.5 is that the @tsv/-r combination should emit \0 for NUL (instead of skipping it).  That's a trivial change to implement, and makes it very easy for users to write their own jq filters if required.

slapresta:
@pkoppstein:

> With respect to new filters and new command-line options, I was thinking that since -r is largely frozen, it might be appropriate to add an output option such as --utf8 that would authorize jq to emit non-JSON output in a way that is different from -r.

I agree with the idea on principle, but `--utf8` is probably not a good name. If I'm not mistaken, `-r`'s current behaviour is to emit unescaped, perfectly valid UTF-8 strings. We're probably looking for something like `--escape-control-characters`. But, uh, maybe shorter.

@nicowilliams:

> @pkoppstein You're complicating things too much. The obvious thing to do is to add a function that escapes NULs, so that one can jus `jq -r '@tsv | escape_nul'.`

I think part of this comes down to the original question: is escaping NULL characters a part of the TSV standard (or an expectation of popular implementations) or is it a thing people want to do the side to `-r` in general, because of broken external applications?

Also, as mentioned earlier, this is not just about NULL: lots of control characters have weird effects on other applications. How about something like `escape/1` being a function that takes a string of codepoints to escape and recursively escaping them on all strings on `.`, and `escape/0` being an alias for feeding `escape/1` all control characters? Granted, I haven't given much thought to this, but it might work, provided this is something we want to do as a filter and not as a command-line argument.

agordon:
@nicowilliams - what do you think about the minimal patch in #758 - does it fit your idea? in it, raw-output can output NULs.

@slapresta - while I agree other control-characters are also messy, NUL is special because it will disrupt other C programs. Other control characters usually don't, except perhaps messing up the display.

nicowilliams:
On Mon, Apr 20, 2015 at 01:43:24PM -0700, pkoppstein wrote:
> @nicowilliams wrote:
> > You're complicating things too much.
> 
> I'm surprised and disappointed that you have that impression. In fact,
> I'm trying to keep things as simple as possible for jq 1.5.
> 
> In fact, the only change (relative to the current jq
> (jq-1.5rc1-60-gb6cc00f)) that I think is needed for jq 1.5 is that the
> @tsv/-r combination should emit \0 for NUL (instead of skipping it).

Speaking of "the @tsv/-r combination" is confusing to me.

There is no connection between `@tsv` and `-r`.  The jq program being interpreted has *NO* idea what the *C* program running it will be doing with the jq program's output values.  And this will continue to be so.

(Some day, perhaps, we'll modify jq so that it always emits "raw" output, with the non-`-r` case modifying the jq program to add ` | @json`.  But even then, the same observation applies: the expression "to the left of" ` | @json` can't know what is being done with its outputs.)

So what you must mean is that `@tsv` escapes NULs.  If you mean something else, be specific.

Personally, I want an `escape_nuls` function simply because this won't be the last time we'll need such a thing.  Ditto `filter_nuls`.

I should really finish the code in the "c-coded-generators" branch of my clone, so I can code up a fast C-coded builtin like `explode` but outputting a stream.  A fast imploder would also be nice.

nicowilliams:
On Mon, Apr 20, 2015 at 01:58:29PM -0700, A. Gordon wrote:
> @nicowilliams - what do you think about the minimal patch in #758 - does it fit your idea? in it, raw-output can output NULs.

Yes, that's better.  See my comment there.

> @slapresta - while I agree other control-characters are also messy, NUL is special because it will disrupt other C programs. Other control characters usually don't, except perhaps messing up the display.

But then, consider things like `find . -print0`.  It can be quite useful to be able to emit NULs!

nicowilliams:
@slapresta The tsv "standard" is just an IANA MIME type registration, and it says nothing about escaping.

nicowilliams:
@slapresta I like your `escape/0` and `escape/1` ideas.

agordon:
Thanks all for the interesting discussion. I'm closing this, but summarized it in #759.

PR #762, opened 2015-04-21
Fixes manual generation with psych

AndrewO:
When running `make` I ran into a couple of problems building the manual. While
I'm not entirely sure that this is the root cause, it appears to have been
related to the fact that ruby 2.0 dropped syck completely in favor of psych
(which was introduced in 1.9.2) for YAML processing. I'm currently using ruby
2.1.0p0.

I'm assuming that the fact that the YAML engine was explicitly set to syck in
the Rakefile was an attempt to work around some incompatibility between the two
libraries, so I looked into what would be necessary to get it to work with the
newer one. The changes to `manual.yml` ended up being pretty minor: I ran it
through `iconv` to convert some ISO-8859-1 characters to UTF-8 and added some
quotes in places (apparently you can't start a string value with '`').

### DIFF
--- a/docs/Rakefile
+++ b/docs/Rakefile
@@ -87,7 +87,6 @@ task :binaries => ["output/download"] do
 end
 
 def load_manual
-  YAML::ENGINE.yamler = 'syck'
   YAML::load(File.open("content/3.manual/manual.yml"))
 end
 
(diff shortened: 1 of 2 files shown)


nicowilliams:

PR #763, opened 2015-04-22
regex functions: report informative error if not available.

agordon:
When trying to use the new regex functions (match/test/sub/capture/etc)
in a JQ that was compiled without the ONIGURAMA regex library,
report an informative error message instead of a 'not defined' error.

Before:

    $ echo '"foo"' | ./jq-old 'test("foo")'
    jq: error: test/1 is not defined at <top-level>, line 1:
    test("foo")
    jq: 1 compile error

After:

    $ echo '"foo"' | ./jq 'test("foo")'
    jq: error: jq was compiled without ONIGURAMA regex libary. match/test/sub and related functions are not available.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -757,6 +757,10 @@ static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
   jv_free(regex);
   return result;
 }
+#else /* ! HAVE_ONIGURUMA */
+static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
+  return jv_invalid_with_msg(jv_string("jq was compiled without ONIGURAMA regex libary. match/test/sub and related functions are not available."));
+}
 #endif /* HAVE_ONIGURUMA */
 
 static jv minmax_by(jv values, jv keys, int is_min) {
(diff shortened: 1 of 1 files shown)


agordon:
Hi,

This is bitten me more than once - very confusing since I was sure I was using the latest 1.5rc1 from git.
Hopefully it will save some false bug reports.

nicowilliams:

wtlangford:
This is great.  I'm more concerned with how you got a 1.5rc1 version to compile without oniguruma in the first place.  It is supposed to be required for ./configure to succeed.

agordon:
@wtlangford - there's nothing in `configure.ac` that requires it (i.e. fails if it is not found).

See https://github.com/stedolan/jq/blob/master/configure.ac#L70 .
If it is not found, it is only a message that's printed (`AC_MSG_NOTICE`), not an error.

and `builtin.c` is designed to be compiled without using `#ifdef HAVE_ONIGURAMA`.

agordon:
Which is a good reminder to whomever packages 'jq' for Debian/Ubuntu/CentOS/HomeBrew - tell them onigurama is 'highly suggested' during build time.

nicowilliams:
@wtlangford I made it optional a while back.

wtlangford:
Oh wow, I missed that then.

nicowilliams:
FYI, I did that to make it easier to build [pre-]release binaries for
various platforms...

wtlangford:
Ahh, I was quite busy when you started that process, which explains it. I
like that oniguruma is optional now, to be honest.

guruprasannaps:
I want to validate the filename. If the file name contains the \ / : * ? " < > | I should get the result as false. I tried many things but unable to make it work. CVOuld you help me please on this. how I can validate?
test("\ / : * ? " < > |*"; "gx")] is throing an compilation error .Please help me

nicowilliams:
@guruprasannaps Please open a new issue, or better yet, go to stackoverflow and ask a question there using the `jq` tag.  We also have an IRC channel, `#jq`, on Freenode.

Issue #618, opened 2014-11-13
reduce is a dog (even in jq>1.4)


pkoppstein:
The following function uses "reduce" to emit the "equilibrium indices" of the input array.  An equivalent version written using recurse or foreach runs nicely in jq>1.4 (i.e. those versions that have TCO and foreach), but in both jq 1.4 and jq>1.4, the version using reduce has roughly exponentially worse performance in the length of the input array.  

Here are some gross timings using jq>1.4 (i.e. based simply on "time jq ...") for the following query:  

```
array(N;0) | equilibrium_indices | length
```

```
# N     (u+s in seconds)
#1000  0.19
# 1e4   0.43
# 2e4   1.54
# 3e4   3.43
# 4e4   6.2
# 5e4  10.4
# 6e4  16.5
```

Since the `recurse` and `foreach` versions of this program are fine in jq>1,4, the problem is evidently with the implementation of "reduce".  Hopefully a skilled eye will be able to identify it; if not, then maybe "reduce" should be implemented using "foreach"! 

```
def equilibrium_indices:
  def indices(a):
    reduce range(0;a|length) as $i 
      # state: [ headsum, tailsum, answers]
      ( [ 0, add, [] ];
        .[0] as $h
        | (.[1] - a[$i]) as $t
        | (if $h == $t then .[2] + [$i] else .[2] end) as $ans
        | [ $h + a[$i], $t, $ans ] )
    | .[2];

  . as $in | indices($in);

def array(n;init): reduce range(0;n) as $i ([]; . + [init]);
```

nicowilliams:
OK, so here's the last word on this bug.

The problem was that `FORK` pushes a forkpoint and the `DUP` that follows the `FORK` for the loops in `reduce` and `foreach` wants to pop the value that was on the top of the stack before the `FORK` -- that value is now not on top of the stack, so it can't be popped.

The better fix for this issue is not to add `DUPN` as I did, but to add a variant of `FORK` that pops a value, pushes a forkpoint, then pushes the value popped earlier.  Much simpler and easier to understand.

(A generalization would be to add an operand to `FORK` indicating how many values to pop and re-push after pushing the forkpoint, but since this operand today would only ever be zero or one, it's easier to add a new opcode instead.)

leonid-s-usov:
Hi. 
Came here via commit history, cause I'm sniffing around the place where DUPN is added and this instruction is standing in a way of generalising var destructure.

[...]

PR #806, opened 2015-06-08
array and object destructuring (fix #533)

### LINKED ISSUE #533: `. as [$foo, $bar, $baz]` array destructuring?

dtolnay:
`. as [$i, $j, $k] | ...`
`. as {a: $i, b: $j} | ...`
`. as [[[$i]], {a: $j}] | ...`

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -675,15 +675,25 @@ block gen_collect(block expr) {
                gen_op_bound(LOADVN, array_var));
 }
 
-block gen_reduce(const char* varname, block source, block init, block body) {
+static block bind_matcher(block matcher, block body) {
+  // cannot call block_bind(matcher, body) because that requires
+  // block_has_only_binders(matcher), which is not true here as matchers
+  // may also contain code to extract the correct elements
+  for (inst* i = matcher.first; i; i = i->next) {
+    if (i->op == STOREV && !i->bound_by)
+      block_bind_subblock(inst_block(i), body, OP_HAS_VARIABLE, 0);
+  }
+  return BLOCK(matcher, body);
+}
+
+block gen_reduce(block source, block matcher, block init, block body) {
   block res_var = gen_op_var_fresh(STOREV, "reduce");
   block loop = BLOCK(gen_op_simple(DUPN),
                      source,
-                     block_bind(gen_op_unbound(STOREV, varname),
-                                BLOCK(gen_op_bound(LOADVN, res_var),
-                                      body,
-                                      gen_op_bound(STOREV, res_var)),
-                                OP_HAS_VARIABLE),
+                     bind_matcher(matcher,
+                                  BLOCK(gen_op_bound(LOADVN, res_var),
+                                        body,
+                                        gen_op_bound(STOREV, res_var))),
                      gen_op_simple(BACKTRACK));
   return BLOCK(gen_op_simple(DUP),
                init,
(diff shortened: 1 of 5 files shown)


dtolnay:
I have not changed reduce and foreach in this commit because first I want to better understand issue #618 and the need for DUPN.

nicowilliams:
@dtolnay  About `DUPN`, take a look at `stack_pop()`: it does a `jv_copy()` if `!stack_pop_will_free(...)`, which is when the value is not at the top of the stack, which is when we've pushed a fork point, say.  But we know the value won't be used again here (since we're generating the `DUPN` or `LOADVN`), so we just replace the value with a `null` instead of incrementing its reference count.  Perhaps I should instead have written a check that the only thing between the value and the top is a fork point, and for all I remember right now maybe I tried that and decided for `DUPN` instead for some reason I don't recall.  I appreciate more minds looking at this, as this is tricky business!

nicowilliams:
Interesting thought though to allow destructuring in the `reduce`/`foreach` syntax...  I like it.

dtolnay:
Addressed feedback (added comment for block_bind_subblock and renamed prior/pattern to left/curr) and finished reduce and foreach.

pkoppstein:
@dtolnay - This is great stuff!  It is especially helpful, for example, when writing 0-arity functions that stuff a lot of baggage into the "input".  One thing that is perhaps not so obvious is that the match-making is quite liberal.  Since this is evidently intended, I think it would be a good idea to signal as much e.g. by adding a test case like this:
```
reduce [[1,2,10], [3,4,10]][] as [$i,$j] (0; . + $i * $j)
null
14
```


@nicowilliams - Any chance this can make it into 1.5rc2?

dtolnay:
It is intended. `. as [$i, $j]` should be shorthand for `.[0] as $i | .[1] as $j`, and similarly for objects. I have added your suggested test case. I had this similar one but not for reduce: `[[1], [1, 2, 3]] as [$a, $b] | [$b, $a]`.

nicowilliams:
@pkoppstein  Right, the destructuring patterns need not completely match the input -- that would be obnoxious, though perhaps an option for that would be nice (but let's leave that to another day).  Also, array destructuring has to be anchored on the left or right end of the array given this, and the left it is.  Finally, it might be nice to be able to bind a `$varnme` to "the rest of the array", and or to destructure out of order.  I'm thinking of something like `. as [1->$second, 0->first, rest->$rest] | ...`, but nothing in @dtolnay's work precludes us adding that later, so I think we're good (not that you're suggesting that we do this now).  But you're right that the docs need a clarification.

nicowilliams:
I've pushed @dtolnay's commit with one change: I git added the `parser.c` generated by bison.

I also pushed a clarification to the manual about array destructuring patterns.

nicowilliams:
Closed with d3343d5 (cherry-picked from c22543a, with an updated `parser.c` squashed in).

Thanks!

nicowilliams:
@dtolnay I finally cleared everything up (I hope) w.r.t. #618.  I may push a better fix for that.

nicowilliams:
Another destructuring feature that might be nice would be syntax like this:

    # for each key/value pair in ., produce . and
    # assign the key to $k and the value to $v
    . as .{$k: $v}

and

    # for each value in . (array or object), produce . and
    # assign the value to $v
    . as .[$v]

Then one could write expressions like:

    . as .objects[].{object: $obj}.relations[].{relation: $rel} | ...

which is useful when dealing with perverse :) schemas like

    {
      "objects": [
        {
          "object": {
            ...,
            "relations": [
              {
                "relation": { ... }
              },
              ...
            ]
          }
        },
        ...
      ]
    }

OK, this is me casting about for a hammer for that ugly nail.  I really do have such JSON to work with, so I often write this sort of expression:

    .objects[].object as $o | .relations[].relation as $r | ... do things with $o and $r ...

This is probably a terrible idea, so don't go off and implement.  It should stew for a while.

PR #809, opened 2015-06-09
Add Missing Makefile Dependency

apeschel:
main.c requires version.h, but there is no logic marking this dependency
in the Makefile. This commit adds the dependency to the Makefile
template.


nicowilliams:
Pushed.  Thanks!

PR #810, opened 2015-06-10
Proposal: combinations/0 and combinations/1

slapresta:
This is an implementation of Python's [`itertools.product()`](https://docs.python.org/2/library/itertools.html#itertools.product) function.

The filter `product/0` takes an array of arrays as input and generates the product combination of those arrays. The filter `product/1` takes an array as input and a number `n` as argument and generates the product combination of `n` copies of the same array. Some samples:

```
jq -n '[["a", "b"], ["c", "d"]] | product | map(join(""))'
[
  "ac",
  "ad",
  "bc",
  "bd"
]
```

```
jq -n '["0", "1"] | product(2) | map(join(""))'
[
  "00",
  "01",
  "10",
  "11"
]
```

The documentation is not there, so it is not ready to merge; but if there's interest on having this merged, I will work on the documentation.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1424,19 +1424,16 @@ static const char* const jq_builtins[] = {
   "def first: .[0];",
   "def last: .[-1];",
   "def nth($n): .[$n];",
-  "def product:"
-  "    if any(length == 0) then [] else"
-  "        reverse |"
-  "        reduce .[] as $i ("
-  "            [];"
-  "            [{x: $i[], xs: (.[] // null)}]"
-  "        ) |"
-  "        map([recurse(.xs?).x?])"
+  "def combinations:"
+  "    if length == 0 or any(length == 0) then empty else"
+  "        .[0][] as $x |"
+  "        (.[1:] | combinations) // [] as $y |"
+  "        [$x] + $y"
   "    end;",
-  "def product(n):"
+  "def combinations(n):"
   "    . as $dot |"
   "    [range(n) | $dot] |"
-  "    product;",
+  "    combinations;",
   // # transpose a possibly jagged matrix, quickly;
   // # rows are padded with nulls so the result is always rectangular.
   "def transpose:"
--- a/builtin.c
+++ b/builtin.c
@@ -1425,12 +1425,14 @@ static const char* const jq_builtins[] = {
   "def last: .[-1];",
   "def nth($n): .[$n];",
   "def product:"
-  "    reverse |"
-  "    reduce .[] as $i ("
-  "         [];"
-  "         [{x: $i[], xs: (.[] // null)}]"
-  "    ) |"
-  "    map([recurse(.xs?).x?]);",
+  "    if any(length == 0) then [] else"
+  "        reverse |"
+  "        reduce .[] as $i ("
+  "            [];"
+  "            [{x: $i[], xs: (.[] // null)}]"
+  "        ) |"
+  "        map([recurse(.xs?).x?])"
+  "    end;",
   "def product(n):"
   "    . as $dot |"
   "    [range(n) | $dot] |"
--- a/builtin.c
+++ b/builtin.c
@@ -1426,11 +1426,9 @@ static const char* const jq_builtins[] = {
   "def nth($n): .[$n];",
   "def combinations:"
   "    if length == 0 then [] else"
-  "        if any(length == 0) then empty else"
-  "            .[0][] as $x"
-  "              | (.[1:] | combinations) as $y"
-  "              | [$x] + $y"
-  "        end"
+  "        .[0][] as $x"
+  "          | (.[1:] | combinations) as $y"
+  "          | [$x] + $y"
   "    end;",
   "def combinations(n):"
   "    . as $dot"
--- a/builtin.c
+++ b/builtin.c
@@ -1425,10 +1425,12 @@ static const char* const jq_builtins[] = {
   "def last: .[-1];",
   "def nth($n): .[$n];",
   "def combinations:"
-  "    if length == 0 or any(length == 0) then [] else"
-  "        .[0][] as $x"
-  "          | (.[1:] | combinations) as $y"
-  "          | [$x] + $y"
+  "    if length == 0 then [] else"
+  "        if any(length == 0) then empty else"
+  "            .[0][] as $x"
+  "              | (.[1:] | combinations) as $y"
+  "              | [$x] + $y"
+  "        end"
   "    end;",
   "def combinations(n):"
   "    . as $dot"
--- a/builtin.c
+++ b/builtin.c
@@ -1425,15 +1425,15 @@ static const char* const jq_builtins[] = {
   "def last: .[-1];",
   "def nth($n): .[$n];",
   "def combinations:"
-  "    if length == 0 or any(length == 0) then empty else"
-  "        .[0][] as $x |"
-  "        (.[1:] | combinations) // [] as $y |"
-  "        [$x] + $y"
+  "    if length == 0 or any(length == 0) then [] else"
+  "        .[0][] as $x"
+  "          | (.[1:] | combinations) as $y"
+  "          | [$x] + $y"
   "    end;",
   "def combinations(n):"
-  "    . as $dot |"
-  "    [range(n) | $dot] |"
-  "    combinations;",
+  "    . as $dot"
+  "      | [range(n) | $dot]"
+  "      | combinations;",
   // # transpose a possibly jagged matrix, quickly;
   // # rows are padded with nulls so the result is always rectangular.
   "def transpose:"
--- a/builtin.c
+++ b/builtin.c
@@ -1424,7 +1424,18 @@ static const char* const jq_builtins[] = {
   "def first: .[0];",
   "def last: .[-1];",
   "def nth($n): .[$n];",
-  // # transpose a possibly jagged matrix, quickly; 
+  "def product:"
+  "    reverse |"
+  "    reduce .[] as $i ("
+  "         [];"
+  "         [{x: $i[], xs: (.[] // null)}]"
+  "    ) |"
+  "    map([recurse(.xs?).x?]);",
+  "def product(n):"
+  "    . as $dot |"
+  "    [range(n) | $dot] |"
+  "    product;",
+  // # transpose a possibly jagged matrix, quickly;
   // # rows are padded with nulls so the result is always rectangular.
   "def transpose:"
   "  if . == [] then []"
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1279,6 +1279,21 @@ sections:
             input: '["foobar", "barfoo"]'
             output: ['[false, true]']
 
+      - title: "`combinations`, `combinations(n)`"
+        body: |
+
+          Outputs all combinations of the elements of the arrays in the
+          input array. If given an argument `n`, it outputs all combinations
+          of `n` repetitions of the input array.
+
+        examples:
+          - program: 'combinations'
+            input: '[[1,2], [3, 4]]'
+            output: ['[1, 3]', '[1, 4]', '[2, 3]', '[2, 4]']
+          - program: 'combinations(2)'
+            input: '[0, 1]'
+            output: ['[0, 0]', '[0, 1]', '[1, 0]', '[1, 1]']
+
       - title: "`ltrimstr(str)`"
         body: |
 


pkoppstein:
There are, I think, several points worth discussing:

1. naming
2. alternative specification
3. Cartesian-product implementation 

(1) Naming

There are many kinds of "products", so a more specific name, or
perhaps a module prefix would be appropriate.  Pursuing the first of
these alternatives: since the "product" here is a kind of "outer
product" the name "outer_product" suggests itself. Pursuing the second
of these alternatives: a module name such as Itertools or Algebra or
some such would be appropriate.

I realize that to date, no "builtin" function has a module prefix, but
it's definitely possible.  Alternatively, the "product" could be part
of an optional module. In similar discussions, the view that functions
(such as matrix multiplication, correlation coefficients, etc) belong
in optional modules.

(2) Alternative Specification

Instead of defining product/0 as the core function, there is much to
be said for defining product(f) to behave like "product | map(f)".
This would I believe not only typically be more efficient and elegant,
but also more "jq-esque".  Your product/1 would then become a filter
of arity 2, so, for example, we'd end up with
outer_product(f) and outer_product(f;n).


(3) Cartesian-product implementation
Since jq already has a kind of Cartesian product semantics,
I was wondering whether the following definition, which produces equivalent
results using your examples, wouldn't suffice:
```
def product:
  . as $in
  | [ $in[0][] as $x | $in[1][] as $y | [$x, $y]];
```

Or, combining (1), (2) and (3) we'd have something like this:
```
def outer_product(f):
  . as $in
  | [ $in[0][] as $x | $in[1][] as $y | [$x, $y] | f];
```

slapresta:
**On naming:**

I get why `product` on its own would be problematic (although considering jq calls its `reduce(+)` filter `add` and not `sum`, then `reduce(*)` would aptly be called `multiply` and not `product`, so there wouldn't be a collision in theory) and I wouldn't be opposed to either `outer_product` or putting it in a module. It would be nice to put the entire standard library in modules for 2.0, wouldn't it?

**On the specification:**

I don't see the point, semantically, on aliasing `product/1` to `product | map(f)`; the filters that do this do so because of readability reasons (i.e. the use of `any(length)` on the definition of `product/0`) but I fail to grasp what the meaning of `product(foo)` would be.

The current semantics of `product/1` aim to imitate the `repeat=` named argument of Python's `itertools.product`. Wanting the product of something with itself feels like a relatively reasonable operation.

**On `[]`-based equivalents**

Unless I'm missing anything, the examples you posted compute the product of two lists. My function computes the product of as many lists as desired. For the case where the amount of lists is bounded, I would use `[{x: .[0][], y: .[1][]}]`, which works using the same behavior of the spread operator that your examples use (my implementation of `product/0` is based on that same behavior too)`

pkoppstein:
> On naming:

An important point here is that jq has already passed version 1.0, and thus what would make sense as a builtin for jq 1.0 might not be wise for jq now.  The question to ask of prospective names that are also English words is: how likely is it that the name is already being used in someone's library, but with a DIFFERENT meaning?

> It would be nice to put the entire standard library in modules for 2.0, wouldn't it?

Unfortunately, unless I'm mistaken, the import statement as currently implemented does not support importing into the current namespace; that is, if a function is imported, then to invoke it, a module name must be specified.  I hope that will change!

> On the specification:

> I don't see the point, semantically, on aliasing product/1 to product | map(f)

The point is more about efficiency (mainly space) and expressivity/readability/brevity (product(f) vs product | map(f)).  product/0, product(f) and
product(f;n) could all be supported (with product/0 defined as product(.), etc).


> On []-based equivalents:

> Unless I'm missing anything, the examples you posted compute the product of two lists.
> My function computes the product of as many lists as desired.

I was asking whether a cross-product wouldn't be sufficient as a builtin.
On further reflection, though, I think we can have our cake and eat it
by introducing:

```
def combinations:
  if length == 0 then []
  else
  .[0][] as $x
  | (.[1:] | combinations) as $y
  | [$x] +  $y
  end ;
```
This produces a stream of combinations, and is thus nicely jq-esque;
and your product is equivalent to [combinations].

Also, since combinations/1 does indeed produce a stream of "combinations",
the name is precisely apropos and therefore justifiable.

slapresta:
> The question to ask of prospective names that are also English words is: how likely is it that the name is already being used in someone's library, but with a DIFFERENT meaning?

I hadn't thought about it that way; that makes a lot of sense.

> I was asking whether a cross-product wouldn't be sufficient as a builtin.

Maybe it could have been, but I don't see the point, considering the language already has the cross-product of a fixed amount of elements as a builtin when spreading on object literal properties: `{x: $x[], y: $y[], z: $z[]}` computes the cross-product of `$x`, `$y` and `$z`.

> This produces a stream of combinations, and is thus nicely jq-esque;
and your product is equivalent to [combinations].

Love it! We should definitely go with that implementation. I really like the name, too.

> Also, since combinations/1 does indeed produce a stream of "combinations",
the name is precisely apropos and therefore justifiable.

Not sure what you mean here by combinations**/1**? If you mean combinations/0, I agree, the name is perfect.

Speaking of which, since having `combinations/1` be a pipe-to-map shortcut wouldn't be necessary given that the combinations are already spread (that is, if you want to map over them, just pipe) then maybe we could keep the use of the argument as in `product/1`, that is, as the "repeat" argument from Python's itertools.product?

I will update the pull request with these changes; ping me if something doesn't feel right!

slapresta:
I changed the name and added your simplified implementation. I kept `combinations/1`'s behaviour identical to `product/1`'s, as per the previous comment. I made two small changes to the implementation you provided:

First, it now outputs `empty` instead of `[]` on zero length, and uses the `//` operator to catch it on `$y` and turn it back into a `[]`. This is so that `[] | [combinations]` returns `[]` (no combinations) instead of `[[]]` (a combination of zero elements; what does that even mean?)

Then, it now checks for `any(length == 0)` along the existing check for `length == 0`. This is so that `[[1, 2], [], [3, 4]] | [combinations]` returns `[]` instead of `[[1], [2]]`, which was definitely incorrect (my intuition argues that `[[1, 3], [1, 4], [2, 3], [2, 4]]` is the correct answer, but it seems like it's not)

These fixes comply with the behaviour of Python's `itertools.product`.

pkoppstein:
@slapresta - Thanks for considering "combinations".  Yes, I meant combinations/0, and yes, I have no problems with combinations(n).

With regard to the edge cases, I would agree that consistency with other "standards" (including Python) is a consideration, but the most important considerations are (1) mathematical correctness and (2) "internal consistency" (by which I mean something like "letting jq speak for itself"). As I hope to convince you, both (1) and (2) imply that `[] | combinations` should yield precisely 1 result, and `[[1,2, ... k]] | combinations` should yield k results.

Let K = [1,2, ... k]. Then clearly we want:

 [K,K,K] | combinations to yield k^3 results
 [K,K] | combinations to yield k^2 results
 [K] | combinations to yield k^1 results
 [] | combinations to yield k^0 results

(And of course k^0 is 1.)

As for (2), the key is to focus on the simplest possible implementation that meets the main functional requirements.  In this case, if you compare the original implementation of `combinations` with your implementation, I think you'll see that the former is just about the simplest possible implementation that satisfies the above k^n requirements for n>1, whereas the revised implementation adds "special cases" as well as complexity.

pkoppstein:
@slapresta - By the way, I'd like to try to persuade you that pipes in multi-line pipelines should
be placed at the start of the line rather than at the end.  When the pipes are vertically aligned properly,
the result is highly readable, and sometimes even artistic, as in the design of a book on Elixir:

image

Here is another reference in the context of shells: http://google.github.io/styleguide/shell.xml
```
"If a pipeline all fits on one line, it should be on one line.

"If not, it should be split at one pipe segment per line with the pipe on the newline 
and a 2 space indent for the next section of the pipe. ...

# All fits on one line
command1 | command2

# Long commands
command1 \
  | command2 \
  | command3 \
  | command4
```

This is similar to the "method chaining" style, commonly used, for example, with javascript:
```
$('#my-div')
  .css('background', 'blue')
  .height(100)
  .fadeIn(200);
```

slapresta:
> With regard to the edge cases

@pkoppstein I'll admit I don't know enough algebra to be up to this task; I definitely think the case where `[[1, 2], [], [3, 4]] | [combinations]` was `[[1], [2]]` is a bug that needed to be fixed, but I don't feel strongly about the case for `[] | [combinations]` being `[[]]`.

Your argument makes sense to me; I would argue for the other fix that `[$a, $b, $c] | [combinations] | length` should equal the product of the lengths `$a`, `$b` and `$c`, and thus if one of those has zero length there should be zero combinations. From that idea, then, stems the very foundation of *your* argument, for the product of zero elements is usually considered to be 1; and, in fact, there exists one combination of zero elements, that is, the unit tuple (or, guessing, in jq's parlance, the empty list).

So, while it feels mind-blowing to me that `[] | [combinations] == [[]]` and `[[]] | [combinations] == []`, I have to admit it makes mathematical sense. I will revert that change to your original implementation.

> I'd like to try to persuade you that pipes in multi-line pipelines should be placed at the start of the line rather than at the end. 

Noted! I don't have a strong opinion on this in either way, so I will correct that too.

slapresta:
Feels done to me! Still missing documentation, though. Please verify @pkoppstein @nicowilliams :)

pkoppstein:
@slapresta wrote:
> I definitely think the case where [[1, 2], [], [3, 4]] | [combinations] was [[1], [2]] is a bug that needed to be fixed,

I certainly agree that  `[[1, 2], [], [3, 4]] | [combinations]` should produce [], and that is precisely what the original `combinations` does (see below).  I also agree that your revision does the same, but at the expense of efficiency (and complexity).

Here's a little demonstration of the differences between (1) `product` using the original definition; (2) `[combinations]` using the original definition of `combinations`; and (3) `[combinations]` using your latest version.  The examples can be adapted for "tests/all.test":
```
def product:
  if any(length == 0) then [] else
    reverse |
    reduce .[] as $i (
      [];
      [{x: $i[], xs: (.[] // null)}]
     ) |
     map([recurse(.xs?).x?])
  end;

# Input: an array of 0 or more arrays.
# Output: a stream of arrays, with output[i] being drawn from input[i] in every possible way.
# If $input[i] is [] for any i, then the result is a singleton stream consisting of [];
# otherwise, if the input arrays have lengths l0, l1, ..., then the output
# stream will have l0 * l1 * ... items.
#
def combinations:
  if length == 0 then []
  else
  .[0][] as $x
  | (.[1:] | combinations) as $y
  | [$x] +  $y
  end ;

def revised_combinations:
  if length == 0 then [] else
    if any(length == 0) then empty else
      .[0][] as $x
      | (.[1:] | revised_combinations) as $y
      | [$x] + $y
    end
  end;

"product: \(product)",
"original: \([combinations])",
"revised: \([revised_combinations])"
```

```
$ jq -c -f pc.jq
[]
"product: []"
"original: [[]]"
"revised: [[]]"

[[]]
"product: []"
"original: []"
"revised: []"

[[1],[2]]
"product: [[1,2]]"
"original: [[1,2]]"
"revised: [[1,2]]"

[[1],[],[2]]
"product: []"
"original: []"
"revised: []"

[[1,2],[3],[4,5]]
"product: [[1,3,4],[1,3,5],[2,3,4],[2,3,5]]"
"original: [[1,3,4],[1,3,5],[2,3,4],[2,3,5]]"
"revised: [[1,3,4],[1,3,5],[2,3,4],[2,3,5]]"
```

slapresta:
Huh?! I was convinced that it didn't... I must have tested it wrong. I'll revert it to the original `combinations`, then. Thank you for your patience! :sweat_smile:

pkoppstein:
@slapresta - And thank you for your willingness to consider alternatives, not to mention all your other contributions!

nicowilliams:
Docs?

I'm not convinced as to style.  I want readers to read to the end of the line.  Maybe that's not desirable (i.e., let them skip ahead, noting only that which must be noted as they do).

pkoppstein:
@slapresta - There's talk of rc2 being released soon ... could you add some documentation?  Feel free to borrow from my # comments :-)

slapresta:
@nicowilliams @pkoppstein Docs are coming! I'm a bit short on time lately, but I will try to get them ASAP.

@nicowilliams About style: whatever you or @pkoppstein prefer is good to me. I find the syntax unusual enough as to not have a clear preference on how it should be formatted. I guess your preference depends on whether you consider `|` to be a statement terminator (as in C-like `;`) or a function chainer (as in ML-like `.`)

nicowilliams:
> @nicowilliams About style: whatever you or @pkoppstein prefer is good to me. I find the syntax unusual enough as to not have a clear preference on how it should be formatted. I guess your preference depends on whether you consider `|` to be a statement terminator (as in C-like `;`) or a function chainer (as in ML-like `.`)

Leave it as is for now.  I'll think about style later.

nicowilliams:
Where are the docs?

nicowilliams:
Ah, there they are.

nicowilliams:
Closed with ec7647c.  Thanks!

PR #812, opened 2015-06-14
inline strings (fix #506)

### LINKED ISSUE #506: Inline strings
We could fit strings from zero to 14 bytes long (15, including the C null terminator) in `jv` values.  It's probably worth doing.

dtolnay:
A jv is 16 bytes, so strings up to 14 utf8 characters fit
inline. The other two bytes are kind_flags at the beginning
and null terminator at the end.

The `kind` of an inline string is still JV_KIND_STRING. The
`flags`, which is the high 4 bits of kind_flags, is the
length plus 1. The `flags` of a non-inline string is 0.

### DIFF
--- a/jv.c
+++ b/jv.c
@@ -37,10 +37,11 @@ static int jvp_refcnt_unshared(jv_refcnt* c) {
 }
 
 /*
- * Simple values (true, false, null)
+ * Kinds and flags
  */
 
-#define KIND_MASK 0xf
+#define KIND_MASK  0x0f
+#define FLAGS_MASK 0xf0
 
 jv_kind jv_get_kind(jv x) {
   return x.kind_flags & KIND_MASK;
@@ -60,6 +61,14 @@ const char* jv_kind_name(jv_kind k) {
   assert(0 && "invalid kind");
   return "<unknown>";
 }
+
+static uint8_t jvp_get_flags(jv x) {
+  return x.kind_flags & FLAGS_MASK;
+}
+
+/*
+ * Simple values (true, false, null)
+ */
 
 static const jv JV_NULL = {JV_KIND_NULL, 0, 0, 0, {0}};
 static const jv JV_INVALID = {JV_KIND_INVALID, 0, 0, 0, {0}};
@@ -434,9 +443,35 @@ typedef struct {
   char data[];
 } jvp_string;
 
+// -2 for kind_flags and null terminator
+#define JV_STRING_INLINE_MAX (sizeof(jv)-2)
+
+static int jvp_string_is_inline(jv j) {
+  assert(jv_get_kind(j) == JV_KIND_STRING);
+  return jvp_get_flags(j) != 0;
+}
+
+static const char* jvp_inline_string_ptr(const jv* j) {
+  assert(jvp_string_is_inline(*j));
+  return 1 + (const char*) j;
+}
+
+static uint32_t jvp_inline_string_length(jv j) {
+  assert(jvp_string_is_inline(j));
+  return (jvp_get_flags(j) >> 4) - 1;
+}
+
 static jvp_string* jvp_string_ptr(jv a) {
-  assert(jv_get_kind(a) == JV_KIND_STRING);
+  assert(!jvp_string_is_inline(a));
   return (jvp_string*)a.u.ptr;
+}
+
+static char* jvp_string_value_mut(jv* j) {
+  assert(jv_get_kind(*j) == JV_KIND_STRING);
+  if (jvp_string_is_inline(*j))
+    return (char*) jvp_inline_string_ptr(j);
+  else
+    return jvp_string_ptr(*j)->data;
 }
 
 static jvp_string* jvp_string_alloc(uint32_t size) {
@@ -473,12 +508,22 @@ static jv jvp_string_copy_replace_bad(const char* data, uint32_t length) {
 
 /* Assumes valid UTF8 */
 static jv jvp_string_new(const char* data, uint32_t length) {
-  jvp_string* s = jvp_string_alloc(length);
-  s->length_hashed = length << 1;
-  memcpy(s->data, data, length);
-  s->data[length] = 0;
-  jv r = {JV_KIND_STRING, 0, 0, 0, {&s->refcnt}};
-  return r;
+  if (length <= JV_STRING_INLINE_MAX) {
+    jv r;
+    r.kind_flags = JV_KIND_STRING | ((length + 1) << 4);
+    memcpy(jvp_string_value_mut(&r), data, length);
+    jvp_string_value_mut(&r)[length] = 0;
+    assert(jvp_string_is_inline(r));
+    return r;
+  } else {
+    jvp_string* s = jvp_string_alloc(length);
+    s->length_hashed = length << 1;
+    memcpy(s->data, data, length);
+    s->data[length] = 0;
+    jv r = {JV_KIND_STRING, 0, 0, 0, {&s->refcnt}};
+    assert(!jvp_string_is_inline(r));
+    return r;
+  }
 }
 
 static jv jvp_string_empty_new(uint32_t length) {
@@ -491,9 +536,10 @@ static jv jvp_string_empty_new(uint32_t length) {
 
 
 static void jvp_string_free(jv js) {
-  jvp_string* s = jvp_string_ptr(js);
-  if (jvp_refcnt_dec(&s->refcnt)) {
-    jv_mem_free(s);
+  if (!jvp_string_is_inline(js)) {
+    jvp_string* s = jvp_string_ptr(js);
+    if (jvp_refcnt_dec(&s->refcnt))
+      jv_mem_free(s);
   }
 }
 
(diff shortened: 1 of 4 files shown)


dtolnay:
I benchmarked this using `seq 10000000 | time ./jq -R empty` across 10 runs.

Parent commit: 9.79, 9.59, 9.31, 9.37, 9.42, 9.40, 9.69, 9.96, 9.45, 9.91
This commit: 6.59, 7.04, 6.82, 6.78, 6.98, 6.56, 6.62, 6.65, 6.65, 6.60

So the speedup (in that contrived case) is about 30%.

nicowilliams:
This is great!

(I had a half-baked version waiting to be finished.  Now i don't have to
finish it :)

I was expecting the impact to be more noticeable.

wtlangford:
I'd imagine it could do a lot of good for object indexing, too.  Typically keys are short.  That being said, it _could_ slow it down, since this implementation can't store the hashes on inlined strings.  I'd need to construct a few more tests to see.

dtolnay:
We could store hashes for inline strings up to 10 characters, and either current behavior for 11-14 or not inline those. Can you suggest a good way to benchmark object indexing? It would depend on the workflow which one is better - many strings with the hashes used once vs few strings with the hashes used many times.

@nicowilliams this shouldn't merge unless you are happy with the impact. 30% is more than I was expecting but I come from Java where short-lived allocations are extremely cheap.

nicowilliams:
A clever thing to do for string hashing would be to have a simple hash function to compute on the fly for inline strings between 11 and 14 bytes so we don't have to give up 4 bytes for a hash.  The hash function doesn't have to be fabulous, just an addition of the "two 64-bit halves" of the jv might do.  Alternatively we could inline only strings up to 10 bytes as @wtlangford suggests.  Branches aren't free, but we're way past that here anyways (at least for the interpreted jq; for a libjq app that only uses the jv API it might be different).

I don't have a good benchmark.  We'd have to go look at various APIs (gihub, AWS, twitter are good candidates) and construct one.  The app I use this has object keys like 'entity' (6 bytes) but also 'relationships' (13 bytes), but just inlining strings up to 10 bytes would probably be a net win.

nicowilliams:
>
>
>
Another possibility is to use the trailing four bytes of  the jv in all
inline string cases, filling in with bytes from an actual hash when we have
the space for it.

nicowilliams:
Other small things can be inlined: one-element arrays (with a string up to
14 bytes long) and one-key objects with a key string of up to 6 bytes.

nicowilliams:
The performance numbers are not surprising after some thought.  Most use
cases Weil have lots of objects, arrays, and strings, and inline strings
only save a small portion of the parsing load.  And since string arithmetic
is not all that common, inlining strings won't change typical jq program
runtime overhead, just parser overhead.  Still, a consistent 10-20% win is
*very* nice.

dtolnay:
To summarize, the next steps for me are:
* Come up with a benchmark for object indexing.
* Implement hash cache for inline strings up to 10 characters.
* Compare performance of cached vs uncached hash; decide which to keep.

I will try to get these done this weekend.

dtolnay:
I put a second commit into this PR that stores a hash for inline strings, reducing the max length of an inline string to 10 characters.

I benchmarked on two computers:
- MacBook Pro running Yosemite and Apple LLVM 6.0 (based on LLVM 3.5)
- Lenovo T530 running Ubuntu 14.04 and GCC 4.8.2

For each commit (parent, inline, inline+hash) I ran:
- `for i in {1..10}; do seq 10000000 | time -f %U ./jq -R 'empty'; done`
   - 0000 TOP
   - 0001 CALL_JQ empty:0
   - 0005 RET
- `for i in {1..10}; do seq 10000000 | time -f %U ./jq -R '{}[.] | empty'; done`
   - 0000 TOP
   - 0001 SUBEXP_BEGIN
   - 0002 SUBEXP_END
   - 0003 LOADK {}
   - 0005 INDEX
   - 0006 CALL_JQ empty:0
   - 0010 RET

The +- is the standard deviation across the 10 runs.

   | parent | inline | inline+hash
--- | --- | --- | ---
**llvm** `'empty'` | 9.63+-0.20 | 6.83+-0.22 | 8.30+-0.21
**llvm** `'{}[.] | empty'` | 11.14+-0.33 | 8.61+-0.22 | 9.70+-0.13
**gcc** `'empty'` | 6.09+-0.06 | 5.63+-0.10 | 5.50+-0.11
**gcc** `'{}[.] | empty'` | 7.66+-0.04 | 7.64+-0.10 | 7.04+-0.09

Here is the raw data. The order of the rows matches the order of the cells in the previous table.

 | | | | | | | | | | 
--- | --- | --- | --- | --- | --- | --- | --- | --- | --- | ---
llvm `'empty'` parent | 9.77 | 9.81 | 9.40 | 9.82 | 9.42 | 9.97 | 9.52 | 9.43 | 9.60 | 9.60
llvm `'empty'` inline | 7.04 | 6.92 | 7.17 | 6.61 | 6.60 | 6.98 | 7.02 | 6.60 | 6.63 | 6.78
llvm `'empty'` i+h | 8.27 | 8.40 | 8.15 | 8.30 | 8.06 | 8.69 | 8.21 | 8.32 | 8.59 | 8.04
llvm `'{}[.]'` parent | 11.64 | 10.96 | 11.28 | 10.99 | 11.67 | 10.72 | 10.80 | 11.12 | 11.30 | 10.96
llvm `'{}[.]'` inline | 8.38 | 8.46 | 8.85 | 8.23 | 8.46 | 8.88 | 8.69 | 8.83 | 8.75 | 8.61
llvm `'{}[.]'` i+h | 9.47 | 9.90 | 9.87 | 9.79 | 9.64 | 9.59 | 9.68 | 9.71 | 9.65 | 9.70
gcc `'empty'` parent | 6.03 | 6.10 | 6.07 | 6.05 | 6.23 | 6.04 | 6.09 | 6.09 | 6.01 | 6.15
gcc `'empty'` inline | 5.71 | 5.65 | 5.64 | 5.73 | 5.76 | 5.60 | 5.56 | 5.50 | 5.49 | 5.74
gcc `'empty'` i+h | 5.44 | 5.42 | 5.69 | 5.65 | 5.48 | 5.42 | 5.41 | 5.40 | 5.58 | 5.56
gcc `'{}[.]'` parent | 7.73 | 7.63 | 7.70 | 7.68 | 7.62 | 7.62 | 7.70 | 7.68 | 7.65 | 7.59
gcc `'{}[.]'` inline | 7.63 | 7.53 | 7.61 | 7.85 | 7.49 | 7.72 | 7.63 | 7.64 | 7.67 | 7.64
gcc `'{}[.]'` i+h | 7.00 | 6.93 | 7.04 | 7.06 | 7.07 | 7.20 | 7.01 | 6.91 | 7.09 | 7.15

nicowilliams:
Interesting.  How about a test like this:

    for i in {1..10}; do seq 10000000 | time -f %U ./jq -R '{.: .}' > /dev/null

and/or

    for i in {1..10}; do seq 10000000 | time -f %U ./jq -Rn 'reduce inputs as $i ({}; .[$i] |= $i) | empty'

I.e., let's take advantage of the cached hash, since we're paying for computing it every time (instead of just at object insert time).

nicowilliams:
At least this doesn't hurt, so now we just need to know if it helps when we use the strings.

Mind you, we don't mutate strings often in jq programs, so there won't be a win from that.

nicowilliams:
Maybe we should go back to the first cut and just pay an allocation price on object key insertion.

dtolnay:
Yeah it is not yet clear to me whether neither, one, or both of these commits is worth merging. I will run more tests today.

dtolnay:
Here is what I got for your second suggestion.

- `seq 2000000 | time -f %U ./jq -Rn 'reduce inputs as $i ({}; .[$i] |= $i) | empty'`:

     | parent | inline | inline+hash
--- | --- | --- | ---
**llvm** | 7.34+-0.21 | 6.82+-0.16 | 6.95+-0.19
**gcc** | 5.67+-0.04 | 6.02+-0.07 | 5.69+-0.07

dtolnay:
I am leaning toward abandoning both commits... The performance difference doesn't seem to make up for the added complexity. If strings were created/destroyed more often in jq it may be a different story, but since they mostly all come from the parser, the speedup in the parser is minor compared to the execution time for any nontrivial program.

nicowilliams:
I agree.  The only other thing I'd look at is a faster (but maybe not so good) hash for the inline case.  Something like: cast the `jv` to an array of `uint32_t`s, multiply-and-add them, XOR one half of the resulting 64-bit number into the other, then XOR in a "seed".

nicowilliams:
If we hashed the entire jv that way we could go back to inlining up to 14 byte strings, but now we'll pay for hashing on every object index operation, which sucks because while we don't modify strings often, we do index objects a lot.  So, yeah, I think it's clear now.

nicowilliams:
Thanks for researching this!

nicowilliams:
(And sorry it didn't work out.  Though I suppose there are some good lessons here.)

nicowilliams:
BTW, it stands to reason now that the other possible inlinings (empty and one-element arrays, and empty and one-pair objects where the key is six bytes or shorter) too wouldn't help much.

I had hoped that inlining would also improve cache hit/miss ratios for typical applications.  That we did not measure.

PR #815, opened 2015-06-18
fix broken tests in manual.yml

dtolnay:
I noticed the incorrect `infinite, nan | type` test in @nicowilliams's commit from earlier today. Apparently many others were also incorrect. I had assumed these ran as part of `make check` but I had to run `rake mantests | jq --run-tests` to see the failures.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -696,12 +696,12 @@ sections:
           of `has`.
       
         examples:
-          - program: 'in({"foo": 42})'
-            input: '"foo", "bar"'
+          - program: '.[] | in({"foo": 42})'
+            input: '["foo", "bar"]'
             output: ['true', 'false']
           - program: 'map(in([0,1]))'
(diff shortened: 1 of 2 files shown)


nicowilliams:
Huh, yeah, I never run the tests in the manual except by hand when I add them.  We should just run them in `make check` when we have Ruby around.

nicowilliams:
Thanks!  Cherry-picked and pushed as 51a81c9.

nicowilliams:
I would really like github to have a cherry-pick / rebase onto button in addition to the merge button.

nicowilliams:
I should also split out of `tests/run` everything that isn't related to `tests/all.test`, then we can even have `test/man.test` generated from `rake mantests`.

nicowilliams:
I'll admit I'm not too familiar with the autotool make check details, which is why I've treated `tests/run` as a dumping ground...

dtolnay:
I can give it a shot.

PR #816, opened 2015-06-18
Add alloca() discovery

dougluce:
The build failed on FreeBSD as there is no alloca.h.  This patch is lifted from the autoconf documentation.

I've only tested this on FreeBSD 10.1-RELEASE-p6, Ubuntu 12.04.5 LTS, and Darwin 14.0.0.


nicowilliams:
Closed with 3cbefde.  Thanks!

PR #819, opened 2015-06-19
fix use after free in f_strptime

dtolnay:


nicowilliams:
Thanks.  Fixed with b04cd6a.

Issue #811, opened 2015-06-13
File names with non ASCII characters


svnpenn:
Running on master gives me this error:

~~~
$ jq . róisín.json
jq: error: Could not open file r�is�n.json: No such file or directory
~~~

Running a git bisect reveals this:

~~~
0d414471bb41f41373754c680ecc7f955a9ec2ad is the first bad commit
commit 0d414471bb41f41373754c680ecc7f955a9ec2ad
Author: Nicolas Williams <nico@cryptonector.com>
Date:   Sun Feb 1 18:35:42 2015 -0600
Refactor moar: move parts of main.c into libjq
~~~

I dont know how, but 0d41447 is the culprit. Just to confirm, I ran:

~~~
git checkout 0d41447
~~~

and it errors as expected, then ran:

~~~
git checkout 0d41447^
~~~

and it gives desired result:

~~~json
$ jq . róisín.json
{
  "alpha": "bravo"
}
~~~

Please look into this @nicowilliams , as I would like to be able to use non
ASCII characters sometimes.

nicowilliams:
I'm travelling but i think i know what did it.  You're not in a Unicode
locale, right?

svnpenn:
I am on Windows but everything should be UTF-8

~~~
$ locale
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_ALL=

$ chcp.com
Active code page: 65001
~~~

nicowilliams:
Do you have an od(1)?.  The key is: what encoding is the input method is
producing?

The change, IIRC, was to use jv arrays too keep track of the filenames.  If
the input isn't UTF-8...

svnpenn:
Upon more research, it appears the problem is much deeper. Before 0d41447 non ASCII characters worked on Windows, but only if they were in Windows-1252 code page. Any characters outside of that code page fail on all versions of JQ.

The solution to this would be to implement an `#if` statement that includes a call to `GetCommandLineW` as mentioned here

http://stackoverflow.com/questions/30832756/accept-non-ascii-characters

I found an implementation of this in the FFmpeg project if it helps

http://github.com/FFmpeg/FFmpeg/blob/ec5164a/cmdutils.c#L221-L281

[...]

Issue #262, opened 2014-01-03
characters encoding change


lazna:
downloaded json utf-8 encoded data under windows is displayed fine as a simple curl.exe output, but if piped into jq (......| jq.exe ".")  it displayed all output as expected but with crippled diacritic.

How can I preserve characters encoding?

Petr Laznovsky

nicowilliams:
Is this a dup of #824?

svnpenn:
#939

svnpenn:
#811

PR #824, opened 2015-06-21
Support non ASCII content on Windows

svnpenn:
With #811 being resolved, JQ now supports filenames with non ASCII characters on
Windows:

~~~json
$ jq . Ω.json
{
  "alpha": "bravo"
}
~~~

However it appears that no Windows version of JQ supports non ASCII content:

~~~json
$ cat alpha.json
{"alpha":"Ω"}

$ jq . alpha.json
{
  "alpha": "��"
}
~~~

Note this is with all Windows versions of JQ, including master:

~~~
$ jq --version
jq-1.5rc1-114-g7811ef1-dirty
~~~

even with correct code page:

~~~
$ chcp.com
Active code page: 65001
~~~

and the file is correctly encoded as UTF-8:

~~~
$ od -tx1c alpha.json
0000000  7b  22  61  6c  70  68  61  22  3a  22  ce  a9  22  7d  0a
          {   "   a   l   p   h   a   "   :   " 316 251   "   }  \n
~~~

### DIFF
--- a/jv_print.c
+++ b/jv_print.c
@@ -3,6 +3,9 @@
 #include <float.h>
 #include <string.h>
 #include <assert.h>
+#ifdef WIN32
+#include <windows.h>
+#endif
 
 #include "jv_dtoa.h"
 #include "jv_unicode.h"
@@ -25,7 +28,12 @@ static void put_buf(const char* s, int len, FILE* fout, jv* strout) {
   if (strout) {
     *strout = jv_string_append_buf(*strout, s, len);
   } else {
+#ifdef WIN32
+    DWORD written;
+    WriteFile(fout, s, len, &written, 0);
+#else
     fwrite(s, 1, len, fout);
+#endif
   }
 }
 
@@ -297,7 +305,11 @@ void jv_dumpf(jv x, FILE *f, int flags) {
 }
 
 void jv_dump(jv x, int flags) {
+#ifdef WIN32
+  jv_dumpf(x, GetStdHandle(STD_OUTPUT_HANDLE), flags);
+#else
   jv_dumpf(x, stdout, flags);
+#endif
 }
 
 /* This one is nice for use in debuggers */


svnpenn:
@nicowilliams please review commit

nicowilliams:
I think there's more to it than that.  In main.c we also write a newline
and RS, and buffering would be a problem.  Granted, no one will be using
--seq on the console, but still.  I'm thinking we should have a single
function to do all output.

svnpenn:
Did you want to write something or have me take a crack at it? As I said I am not a C expert

nicowilliams:
I'll write it, but probably tomorrow.

nicowilliams:
Using `WriteFile()` doesn't work when `stdout` isn't the console.  How about that.  I guess that means: use `fwrite()` if `!isatty(STDOUT_FILENO)`, use `WriteFile()` otherwise.

nicowilliams:
It's tempting to close this as `wontfix`...  Yes, I can write a fix.  But should we?  Will mull it over.

svnpenn:
@nicowilliams I feel this issue is important. Yes Windows sucks, but we shouldnt not do something because of that. They said on Stack Overflow that `WriteConsoleW` might be an option instead of WriteFIle. If need be I can try to write a patch

svnpenn:
http://stackoverflow.com/questions/30961728/fwrite-with-non-ascii-characters/30967777#comment49967673_30967777

nicowilliams:
I thought I could tell you to pipe to more, but Windows' `more` is busted too.  I know.  Imagine my surprise.  Anyways, here you go.

svnpenn:
This is great work, thanks. I see you also covered the case when stdout is a pipe. Do you/JQ accept donations? I checked your homepage and this repo and did not see anything

nicowilliams:
No donations, thanks, just your kind words.

nicowilliams:
:)

svnpenn:
I will say again that this is excellent - I can use these few patches for other projects that have poor Windows support, so this work goes beyond JQ project

nicowilliams:
(And if jq took donations, IMO they should go to @stedolan, the father of this baby.  He did the vast majority of the work.)

I'm glad you can reuse these patches.  You can make `priv_fwrite()` more generic by having it call `isatty(fileno(fout))` -- I only didn't because I figured that would be a bit costly, but if you really need a quick fix, that will do it.  You can't name this `fwrite()` because with static linking there's nothing like `dlsym(RTLD_NEXT, "fwrite")`, sadly.

svnpenn:
#939

svnpenn:
#262

svnpenn:
https://github.com/Martchus/tageditor/issues/15

PR #833, opened 2015-06-27
get Travis CI working

dtolnay:
Here is the build of the commit in the PR: https://travis-ci.org/dtolnay/jq/builds/68582926

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,44 @@ sudo: false
 
 language: c
 
-before_script:
-    - sudo apt-get install -qq valgrind
+addons:
+    apt:
+        packages:
+            - libonig-dev
+            - valgrind
+
+before_install:
+    - rm lexer.h lexer.c parser.h parser.c
+
+install:
     - bundle install --gemfile=docs/Gemfile
+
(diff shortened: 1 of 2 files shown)


dtolnay:
I put libonig-dev under packages for the apt addon to install, but it still needs to be whitelisted by travis. I filed a whitelist request here: https://github.com/travis-ci/travis-ci/issues/4123. Until then I am building against artifacts from debian. Similarly for bison - although it is whitelisted, the version in the repo is 2.5 and we need 3.0.

nicowilliams:
Did you mean to keep the `env PAGER=less ` bit?  Does valgrind use `$PAGER`?

nicowilliams:
Pushed as 815b4c9.  Thanks!

nicowilliams:
Oh, no!  You did want to set the `PAGER` because a mantest wants it.  *sigh*

nicowilliams:
Ugh, my fix commit should have mentioned that the braino was mine, not yours.

dtolnay:
Thanks for adding the comment. I'll take the blame for not having one in the first place.

PR #834, opened 2015-06-27
Enable code coverage / coveralls.io

dtolnay:
Here is the result of this branch: https://coveralls.io/builds/2919770

Here is the same build on Travis: https://travis-ci.org/dtolnay/jq/jobs/68623801

The coverage build runs in parallel to the existing gcc and clang valgrind builds: https://travis-ci.org/dtolnay/jq/builds/68623798

There are two commits included in this PR. The first adds --disable-valgrind, the second adds --enable-gcov, and the coveralls build uses both.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,15 @@ sudo: false
 
 language: c
 
+compiler:
+    - gcc
+    - clang
+
+matrix:
+    include:
+        - compiler: gcc
+          env: COVERAGE="--disable-valgrind --enable-gcov"
+
 addons:
     apt:
         packages:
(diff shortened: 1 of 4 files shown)


nicowilliams:
Thanks!  This is nice.  I've gotten used to just editing the test script (now tests/setup) to disable valgrind.  This will be much better.

And test coverage.  Should have been doing this.  Thanks for fixing it.  I see the numbers are not as atrocious as one could have expected, especially if we ignore `lexer.c`.  It's interesting to go through the report for `parser.y` and see all the things we're not testing (mostly error conditions).  Now that `jq --run-tests` can test compilation errors we can do something about that (e.g., we can add a test that `module 1+true; .` fails, as it should, because the metadata isn't constant (because the constant folder can't fold it; it's not that smart, but in this case it could say something more specific).

The state of `linker.c` test coverage is *very* concerning.  We need to test the heck out of the module system.  The state of `main.c` is also very concerning.  I don't know how we might test things like colorized output...  (We could strip out escape sequences and verify that the result parses equal to the original, and anyways, even just running it will allow valgrind to catch errors there.  Maybe a parser option to strip out escape sequences would be nice, though I hesitate to do visit violence on `jv_parse.c` than I already have.)  And so on.  And we need better testing of Unicode support.  We also need to test number parsing and formatting; coverage of `jv_dtoa.c` is pretty low.

A lot of these things can be improved with new jq-coded tests (always a nice thing).  Better testing of `main.c` (and `jv_print.c`) will require more shell-coded tests.

Thanks!!

dtolnay:
There is a flag to exclude specific source files from the total percentage. I think it is reasonable to exclude lexer.c. Should we also exclude jv_dtoa.c, or buckle down and test as many cases as we can?

Issue #443, opened 2014-06-25
enhancement request: pow(n)


pkoppstein:
jq currently has Bessel functions and many other math functions -- but not pow(n)!  The main feature of the following implementation is the preservation of numerical accuracy as much as jq allows.   

    # 0^0 => 1
    # Preserve accuracy for integers as much as jq allows.
    # This version is self-contained.
    def pow(n):
      if n == 0 then 1
      elif . == 0 then 0
      else ( (n | floor) == n) as $intp
           | ( (n % 2) | if . == 0 then 1 else -1 end ) as $sign
           | if . == -1 and $intp then $sign
             elif . < 0 then -(.) | pow(n) * $sign
             elif $intp and n > 0 then . * pow(n - 1)
             else log * n | exp
             end
      end;

Example:

     input: 2 | [ pow(3), (log*3|exp) ]
     output: [ 8, 7.999999999999998 ]

pkoppstein:
```
$ uname -a
Darwin wws 11.4.2 Darwin Kernel Version 11.4.2: Thu Aug 23 16:25:48 PDT 2012; root:xnu-1699.32.7~1/RELEASE_X86_64 x86_64

$ git log -1
commit 6e27de4f747bc1f048aa54f3ea854f7320419e62
Author: David Tolnay <dtolnay@gmail.com>
Date:   Tue Jun 30 08:24:26 2015 -0700
    strftime wrong day-of-week (fix #838)
$ autoreconf -i ; ./configure ; make
...
In file included from builtin.c:107:
libm.h: In function 'f_pow10':
libm.h:132: warning: implicit declaration of function 'pow10'
libm.h:132: warning: incompatible implicit declaration of built-in function 'pow10'

Undefined symbols for architecture x86_64:
  "_pow10", referenced from:
      _f_pow10 in libjq.a(builtin.o)
ld: symbol(s) not found for architecture x86_64
collect2: ld returned 1 exit status
make[1]: *** [jq] Error 1
make: *** [all] Error 2
```

nicowilliams:
What does config.log say about this?  Why did it find this function if it's
not there?

Is jq linking with libm?

Is there a manpage on pow10() on OS X?

pkoppstein:
More details:
```
$ ./configure
...
checking for pow... yes
checking for pow10... yes
...

# extract from config.log:

configure:18260: checking for pow
configure:18288: gcc -std=gnu99 -o conftest -g -O2   conftest.c     -lonig >&5
[...]

PR #836, opened 2015-06-27
gcov exclusions

dtolnay:
86.6% after excluding lexer.c, parser.c, jv_dtoa.c

https://coveralls.io/builds/2920453

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,11 @@ script:
     - make check -j4
 
 after_script:
-    - rm -rf .libs # don't care about coverage for libjq
-    - if [ -n "$COVERAGE" ]; then coveralls --gcov-options '\-lp'; fi
+    - |
+        if [ -n "$COVERAGE" ]; then
+            rm -rf .libs usr # don't care about coverage for libjq, bison, libonig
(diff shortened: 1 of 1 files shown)


nicowilliams:
I agree.  jv_dtoa.c I'd rather be able to test, but that's rather difficult.  Numbers don't necessarily round-trip.  I was able to raise the code coverage of jv_dtoa.c by a mere 1.4% with this:

    +# Basic numbers tests: integers, powers of two
    +[range(-52;52;1)] as $powers | [$powers[]|pow(2;.)|log2] == $powers
    +null
    +true
    +
    +[range(-99/2;99/2;1)] as $orig | [$orig[]|pow(2;.)|log2] as $back | ($orig|keys)[]|. as $k | (($orig|.[$k])-($back|.[$k]))|if . < 0 then . * -1 else . end|select(.>.00005)
    +null
    +
    +

nicowilliams:
I had to add `pow` for that.

pkoppstein:
@nicowilliams wrote:
> I had to add pow for that.

Now might be a good time to review #443

nicowilliams:
Good point.

PR #839, opened 2015-06-30
strftime wrong day-of-week (fix #838)

### LINKED ISSUE #838: strftime: wrong day-of-week
```
$ uname -a
Darwin Mac-mini.local 13.4.0 Darwin Kernel Version 13.4.0: Wed Mar 18 16:20:14 PDT 2015; root:xnu-2422.115.14~1/RELEASE_X86_64 x86_64

$ jq --version
jq-1.5rc1-137-ge5dafae

$ jq -n 'now | strftime("%A, %B %e, %Y")'
"Sunday, June 30, 2015"

$ date
Tue Jun 30 09:37:38 EDT 2015

```
Needless to say, this is not a locale or TZ issue :-)

dtolnay:


dtolnay:
We are still omitting the tm_isdst field.

> A flag that indicates whether daylight saving time is in effect at
> the time described.  The value is positive if daylight saving time
> is in effect, zero if it is not, and negative if  the  information
> is not available.

-1 sounds like the way to go, although I have to look into what uses is.

dtolnay:
No, this commit is fine. To get tm structs we call gettimeofday, and to convert timestamp to tm struct we call gmtime not locatime. All of these deal with UTC, and UTC never has DST. Let me add a comment.

nicowilliams:
I'll try to get to this tomorrow (I'm traveling).  ISTR one libc that had
some weirdness as to dst.  glibc, was it?

nicowilliams:
BTW, you're quite right that we shouldn't memset a struct like that, so we
can discover via valgrind that some libc has extensions that we need to
know about.  But it also goes to show how awful the POSIX datetime APIs
really are.  Also, this means it's time to start testing with DrMemory on
Windows.  The test framework needs to be able to run ex-build tree.
Cross-compiling is nice, but it's not enough.

nicowilliams:
(It was always time to use DrMemory, of course, but for me Win32 was a but
of a second class platform.  After putting in the effort to properly work
around the awfulness of the MSCRT and console, it's more off a first class
platform.)

nicowilliams:
For dynamic linking it is best to zero such structs though.  Hmm.

dtolnay:
When I wrote this commit I didn't realize different implementations of libc could decide to put additional members in this struct, but you are right: "The \<time.h\> header shall declare the tm structure, which shall include at least the following members." Tonight I will look for a different way that jv2tm could create a tm with all of the correct members populated.

nicowilliams:
When we link dynamically then we must either memset the thing, or the
standard must commit to making no use of new fields without a way to
indicate whether they are set.  The problem is _this_ bloody API, which has
an unclean history.

dtolnay:
I can't find a better way. :confused: I have put back the memset.

nicowilliams:
The only thing potentially better thing i can think of is a macro that
tells us whether we're going to link statically, then if we are disable
these memset.

dtolnay:
I pushed a second commit that takes out memset if statically linked. I can't find anything that recommends using these two different ways to initialize depending on whether it is dynamically linked. Can you point me to documentation, or explain why it makes sense?

nicowilliams:
I bet you will find nothing saying that.  My thought was that for this API,
when linking statically we could rely on valgrind (if we elide the
memset) to detect new features that require initialization of fields we
wouldn't otherwise know about.  But generally we should memset all such
structs.

PR #840, opened 2015-07-03
docs and diagram for exec_stack

dtolnay:


nicowilliams:
Thanks.  This is excellent.  Of particular interest, I think, is that `stack_pop_block()` doesn't actually pop (i.e., move the `limit`) when the given `p` is not at the top of the stack, but it does always return the `next` block's pointer.

PR #841, opened 2015-07-04
Make jq.h usable from C++

dtolnay:
Previously, with clang++:
jq.h:46:37: error: typedef redefinition with different
      types ('struct jq_util_input_state *' vs 'jq_util_input_state')

With g++:
jq.h:46:37: error: conflicting declaration
      ‘typedef struct jq_util_input_state* jq_util_input_state’

This typedef was added to libjq by commit 0d41447 which was
after the 1.4 release, so although it is a public API, this
is not a backcompat break because it has never been in a
release.

Specifying the "*" at all uses of jq_util_input_state is
slightly tedious, but jq_state already works that way, so at
least it will be consistent.

### DIFF
--- a/jq.h
+++ b/jq.h
@@ -43,15 +43,15 @@ jv jq_get_attr(jq_state *, jv);
  * whereas jv string values must be in UTF-8.  This way the caller
  * doesn't have to perform any codeset conversions.
  */
-typedef struct jq_util_input_state *jq_util_input_state;
+typedef struct jq_util_input_state jq_util_input_state;
 typedef void (*jq_util_msg_cb)(void *, const char *);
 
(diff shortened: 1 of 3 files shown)


nicowilliams:
Unlike others, i think it's best to put the pointer-ness of a type in its
name, and we can still do it, provided we use a different name from the
struct.  Anyways, thanks!.  I've been meaning to write a simplified version
of main.c as a demo and API test, and this reminds me: we should build it
as a C and C++ program.

Also, the API is not documented (just in the wiki), so source compatibility
breaks are fine -- and would be even if the API were public, but the ABI
we'd have to commit to, or commit to breaking in a safe way.

PR #842, opened 2015-07-07
Dockerfile improvements

jamesandariese:
* Remove excess layer creation
  This is an antipattern in Docker.
* Remove build tools once build is complete
  Leaving them around is not necessary and expands the image size many times
* Make check works with libonig installed
* Compiles statically (needs libonig to be built rather than installed)
* Valgrind doesn't work with the use of pthread TLS in jq so it's disabled -- this
  might be a FIXME situation; I'm not familiar enough with valgrind to
  say whether this is expected
* Make entrypoint be the jq binary so that the image can also be used
  to run jq in environments where you don't want to or can't install
  jq (such as CoreOS).

See https://registry.hub.docker.com/u/jamesandariese/jq/ for a ready to use docker image with jq.

One can either get the static bin out of the docker container as suggested in the original Dockerfile pull request or one can use jq directly with docker like so:

    echo '"hi"' | docker run -i jamesandariese/jq .
    "hi"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,50 @@
-FROM debian:wheezy
-
-# get dependencies
-
-RUN apt-get update
-
-RUN apt-get install -y build-essential
-RUN apt-get install -y autoconf
-RUN apt-get install -y libonig2
-RUN apt-get install -y libtool
-RUN apt-get install -y git
-RUN apt-get install -y valgrind
-RUN apt-get install -y bison
-RUN apt-get install -y flex
-RUN apt-get install -y ruby1.9.3
-
-RUN gem install bundler
-
-# get docs dependencies
-
-COPY ./docs/Gemfile /app/docs/Gemfile
-COPY ./docs/Gemfile.lock /app/docs/Gemfile.lock
-
-WORKDIR /app/docs
-
-RUN bundle install
-
-# copy files
-
-WORKDIR /app
+FROM debian:8
 
 COPY . /app
 
-# build
+# get dependencies, build, and remove anything we don't need for running jq.
+# valgrind seems to have trouble with pthreads TLS so it's off.
 
-RUN autoreconf -i
-RUN ./configure
-RUN make -j8
-RUN make check
+RUN apt-get update && \
+    apt-get install -y \
+        build-essential \
+	autoconf \
+	libtool \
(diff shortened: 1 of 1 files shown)


dtolnay:
I don't know enough about Docker to review - perhaps @slapresta could take a look?

slapresta:
Thanks for improving on the Dockerfile! It was a quick hack and I'm glad to see someone else building on it.

> Remove excess layer creation This is an antipattern in Docker.

Agreed, on the general principle of not having one layer for every operation; Dockerfiles are not shell script. This is something I learned after building this Dockerfile. The eight `apt-get` layers the Dockerfile had are specially egregious.

That said, having everything in the same layer is probably not a good idea either. Notably, I don't want the dependencies to be re-installed every time I build jq, just because I changed one completely unrelated file. I'd rather not `COPY .` until after having installed all dependencies and downloaded Onig, for example. While I agree on removing layers, I think splitting the process into layers on its natural boundaries (get dependencies, get code, build code and manual) makes sense and reduces rebuilding time significantly.

Also, if the Docker script is going to get this long, we should probably just have it as a separate shell script file.

> Remove build tools once build is complete Leaving them around is not necessary and expands the image size many times

:+1: 

> Make check works with libonig installed
> Compiles statically (needs libonig to be built rather than installed)

:+1: :+1: :+1: :100: 

> Valgrind doesn't work with the use of pthread TLS in jq so it's disabled -- this might be a FIXME situation; I'm not familiar enough with valgrind to say whether this is expected

No idea on this; pinging the C overlords here. @nicowilliams @pkoppstein 

> Make entrypoint be the jq binary so that the image can also be used to run jq in environments where you don't want to or can't install jq (such as CoreOS).

No idea why I didn't do this in the first place. :+1: 

---

So, I'd say definitely merge it. While I'm not a fan of having everything on one big layer, there's an impressive amount of good work here. Bikes can be shed later on.

dtolnay:
Rebased and merged in f9f4d2d. Thanks! I filed #845 to follow up on the remaining cleanup tasks.

@jamesandariese, make sure the email you used for the commit is registered in your github profile so it is properly attributed to you - I notice it is different from the primary one on your account.

nicowilliams:
Thanks!

PR #844, opened 2015-07-09
Fixing missing D in LIBM_DDD_NO macro name

fletch:
Had problems compiling because this macro looked to be misnamed.  Added another D and it seemed to be happy.  Didn't look too much further into (hey, it compiled :) it so it might warrant further investigation.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1233,7 +1233,7 @@ static jv f_current_line(jq_state *jq) {
 
 #define LIBM_DDD(name) \
   {(cfunction_ptr)f_ ## name, "_" #name, 3},
-#define LIBM_DD_NO(name)
+#define LIBM_DDD_NO(name)
    
 static const struct cfunction function_list[] = {
 #include "libm.h"


dtolnay:
Yep it looks like this is what @nicowilliams intended in 6366cfd. Thanks!

nicowilliams:
Indeed, that was my intention.  Thanks!

pkoppstein:
@nicowilliams - Does this resolve the pow10 issue (#443)?

nicowilliams:
@pkoppstein Can you test it on OS X?

pkoppstein:
@nicowilliams - 

The two Macs that I use have different versions of OSX.  There was never a problem on the machine for which "checking for pow10... no", and that continues to be the case (i.e. after re-instating the pow10 line in configure.ac).

On the other machine (Darwin Kernel Version 11.4.2), I installed the DDD update, re-instated the line in configure.ac, and rebuilt jq successfully (autoreconf -i ; ./configure ; make). However, although the message "checking for pow10... yes" appeared, jq tells me that neither pow10/0 nor pow10/1 is defined.

On a different tangent, here are two tweaks for manual.yml:

(1)
      "Paths in the a search path are subject to various substitutions."
should read:
      "Paths in the search path are subject to various substitutions."

(2) Documentation for `input_filename` should explain that it can evaluate to empty.  (Wouldn't it be better to have it evaluate to null?)

nicowilliams:
Re: pow10, see this:

    commit 1c350c67d76444e6d169a782dd53a5b2a74035d9
    Date:   Fri Jul 3 21:13:13 2015 -0500

        Remove `pow10` for now; CHECK_MATH_FUNC needs work

        The CHECK_MATH_FUNC() m4 macro needs work.  It should use
        AC_RUN_IFELSE(), not just AC_LINK_IFELSE(), and it should #define
        _GNU_SOURCE and/or other such feature macros in the prologue to get
        non-standard math functions (alternatvely jq should only support
        standard math functions).

nicowilliams:
Re: input_filename, er, how do you get it to empty?

pkoppstein:
```
$ jq -n input_filename
$ jq -n 'input_filename // "foo"'
"foo"
```

pkoppstein:
@nicowilliams wrote:
> alternatively jq should only support standard math functions 

One of the very nice attractions of jq is (was?) the portability of jq programs across supported platforms.  The best way to uphold that goal, it seems to me, is to make exceptions dependent on exceptional settings.  In the present case, we can have our cake and eat it by ensuring that the default configuration is very strongly biased in favor of "standard math functions", as you suggest.

nicowilliams:
Hmm, `input_filename` should never empty, no.  That's an easy one to fix.

Re: math, we're near 1.5, so I'm going to leave it as-is.  jq currently does define functions that may not exist on the platform, and which then produce an error instead of any result values.  I'm not going to re-add pow10 for now, see that commit.

nicowilliams:
Eventually i suspect we'll need compile-time conditionals...

pkoppstein:
@nicowilliams wrote:
> Re: math, we're near 1.5

There's one jq issue that continues to keep me awake at night, especially since freezing jq in its current state will surely make it even harder to resolve satisfactorily in the future.

The issue concerns the quantum-like collapse of the "infinite" and "nan" values, which have a kind of hidden nano-existence until observed, when they seem to become something else, but don't quite.

It seems to me that EITHER (1) jq should strictly adhere to the principle that a jq expression that does not raise an error should yield a stream of JSON expressions, OR (2) if there are to be some exceptions, then jq should frankly extend the JSON type as required, e.g. by supporting unquoted Inf and Nan (analogously to true/false/null).

Following the first path, since 1/0 now yields an error anyway, I think it would be fine to follow the raise-an-error approach.  Alternatively, if we really want 0/0 to yield null, that would be OK too. If we want a JSON value for Inf, then I'd propose a string such as "Inf", it being understood that the implementation of the arithmetic operators would have to be adjusted accordingly.

Of course, one could try to have one's cake and eat it by having an "allow_nan" flag (see
e.g. https://docs.python.org/2/library/json.html).  Even a blended approach would be preferable to the current mishmash.

nicowilliams:
No JSON extensions, full stop, at least not for now and never by default on
output.  With the isnan and isinfinite (and friend) builtins you get to
handle exceptional conditions.  Raising an exception in all NaN and
infinity cases instead would make it impossible to carry around NaNs and
infinities in the middle of a jq program -- yes these have to become
something else in the end (null), but in the middle they can still be
useful.

Of course, @stedolan has the last say, and anyways, nothing really stupid
us from later making functions raise when they would output a NaN or
infinity, as long as we retain a way to produce NaNs and infinities.

nicowilliams:
We should document that in the future jq might raise errors where it now
produces NaNs an/or infinities.  I don't think a global setting for that is
a good idea.  I suspect it's better to produce and consumer NaNs and
infinities without raising exceptions, but i admit that I'm not certain of
that.

pkoppstein:
@nicowilliams wrote:
> We should document that in the future jq might raise errors ...

That would have been sufficient until 'infinite' and 'nan' were introduced.  That's why I think that for jq 1.5, we either need to back-pedal a bit, or resolve the issue.

Back-pedaling could probably be achieved simply by removing 'infinite' and 'nan'. 

But a resolution that addresses your main concerns is, I believe, within easy reach.  Since we want to have our cake (NaN and Inf) and eat it (consistency with JSON), we just need a JSON representation of NaN and Inf (both on input and output).  An obvious solution would be to use strings, e.g. "NaN" and "Inf".  Thus, for example, 1e1000 would internally evaluate to Inf and would externally appear as "Inf"; and arithmetic operation involving the string "Inf" would cause "Inf" to be converted internally to Inf.
Thus "Inf" + 0 => Inf + 0 => Inf => "Inf".  This is nicely consistent with `"Inf" == "Inf"`.  We'd just have to document that in cases of potential ambiguity (such as `"Inf" + "Inf"`), the string interpretation has priority.

This approach would also invite us to rethink whether we really want `nan == nan` => false.

nicowilliams:
I don't agree with you, but as a compromise we can accelerate a feature
I've been wanting: the ability to jq-code JSON (and other) parsers and
encoders.  Minimally this means adding a builtin for output to stdout.  One
could then raise an error when a value deep in an array/object turns out to
be a NaN or infinity.  Or one could encode them as strings as you propose,
or in any other conventional manner, or even as some JSON extension, as
YAML, or anything else you might like.

pkoppstein:
@nicowilliams 
> I don't agree with you

Doesn't it trouble you that: 
```
$ jq -n '0/0 | (. == (tojson | fromjson))'
false
```
The combination of (a) the principle of invertibility of tojson/fromjson, and (b) your dictum about "only JSON", implies there's a problem, no?

nicowilliams:
@pkoppstein It bothers me that JSON doesn't have a representation of NaNs and infinities, but that can't be helped, full stop.  It doesn't bother me that numbers can't be expected to round-trip in JSON implementations.  It doesn't bother me that `0/0` produces a NaN.  It doesn't bother me that jq maps NaNs to `null` on output.  Let it rest: you'll get to write your own encoder and parser and handle NaNs as you wish.  And for now you get to check every math operation's result.

I'll make a note that anything that today produces NaNs or infinities might eventually produce errors instead (except `nan` and `infinite`, of course).  A better try/catch syntax might really help, something like `EXP ? EXP : EXP`, so that for math one could write `0/0?.:.value` to mean "divide zero by zero and let the NaN through if it raises an error.  This is a trivial addition to `parser.y`, so... maybe.  But we can revisit that later.

Summary: jq 1.4 already had the problems you're upset by, we won't be making anything worse in 1.5 as it stands in master, and if anything, we have improved things.

pkoppstein:
@nicowilliams wrote:
> ... except nan and infinite, of course

Yes, they're the fly in your ointment.  Would you consider retracting them at least for now?  

> Let it rest:

Peace.

nicowilliams:
No, those two are fine.  Suppose we make all math builtins raise instead of outputting NaNs, but then how would you produce a NaN as a value for your own internal purposes?

nicowilliams:
I.e., just because you have no use for NaNs doesn't mean others won't.

pkoppstein:
Currently, there is no need for them. They can always be added back later if needed.  Adding them now locks jq into a certain position prematurely.

nicowilliams:
@pkoppstein Let's go over this one last time.

IEEE754 has NaN and infinity values.  Other floating point implementations may or may not have these, but it stands to reason that even when they don't, a jq that used such an implementation could still represent NaNs and infinities.

JSON does not have a way to represent NaNs and infinities.  Extensions to JSON could, and other encodings (that jq does not yet support but in principle could) can as well.

NaN and infinity values may or may not be useful in a jq program.  When not useful the alternative is raising errors (that can be, and may or may not be caught).  When useful they can be checked for with the new `isnan` and related builtins (and indeed, this is how one might turn a NaN or infinity into an error).

Having an `isnan` and `nan` and related builtins does not paint us into a corner.  If we switch to a floating point implementation that lacks NaNs and/or infinities we can either raise errors (making `isnan` then pointless, and `nan` raise -perhaps-) or implement NaNs and infinities in the jv implementation separately from the remainder of the number implementation.

Given that we're currently not checking for floating point exceptions we are relying instead on NaNs and infinities to pass on (currently and admittedly less useful) error information.  We can always add exceptions later (provided that we document this possibility now, so that jq programmers can handle them, though we might not document what such exceptions might look like yet, since we've not researched how best to represent IEEE754/POSIX floating point exceptions).  We might also instead / additionally encode error information into signalling NaNs and provide builtins for extracting the signal.  There is much we can yet do, and it is far from clear at this point what is the best that we can do.

Since there are no new problems in 1.5 that aren't in 1.4, and since adding `isnan` and friends does not paint us into a corner (see above), there isn't really anything here that is a blocker for a 1.5 release.

nicowilliams:
And I certainly don't want to hold up 1.5 so we can think about how best to expose POSIX floating point exceptions to jq programs.

nicowilliams:
I should also add that NaNs and infinities, being C double/float values, and C doubles being what jq uses for its number type's implementation, NaNs and infinities are proper values for jq.  These are the only values in jq without a standard JSON representation, but they are values nonetheless, and so they will remain available as such for the time being.

PR #858, opened 2015-07-19
Resolve shift/reduce conflict of 'def' vs '|'

dtolnay:
This was an important conflict. In the following expression:

    def a: 0; . | a

Bison needs to decide between these two equally valid
parses:

    (def a: 0; .) | a
    def a: 0; (. | a)

For jq we want the second one, because the first results in
"a/0 is not defined". In the current parser the first parse
is a reduce and the second parse is a shift. Since Bison
prefers to shift in shift/reduce conflicts, we accidentally
got the correct behavior.

This commit adds a precedence level FUNCDEF which is lower
precedence than '|', meaning we explicitly choose the
correct parse.

Of course many unit tests already cover this case, but I
added one specifically for it.


dtolnay:
I will investigate the remaining conflicts as well. I suspect we can do better than "%expect".

dtolnay:
I pushed a second commit that fixes all remaining shift/reduce conflicts by lowering the precedence of rules that should never use the generic Exp '?' rule.

nicowilliams:
LGTM, thanks!

PR #865, opened 2015-07-24
fix references to --color-output

zackse:
The manpage and manual specify `colour`, but main.c uses `color`.

``` c
      if (isoption(argv[i], 'C', "color-output", &short_opts)) {
```


dtolnay:
Good catch. Thanks!

PR #868, opened 2015-07-25
Error on bytecodes longer than 0xFFFF (fix #269)

### LINKED ISSUE #269: Too many commas causes output to repeat
I'm trying to pick out many values (10,000) from a large object with many fields (100,000).  So as a small example, if I want to pick out 1 and 3:

    $ echo '{"1":1, "2":2, "3":3}' | jq '.["1"],.["3"]'
    1
    3

First, please let me know if there is a better way to do this.  My current approach is to generate a large jq expression with all the elements I want to pick out, and load it from a file with `-f`.  As a test, this should return the number of elements I requested (at a maximum).  This works fine with up to 3641 elements separated by commas:

    $ echo '{"1": 1}' | jq -f <(seq 3641 | python -c 'import sys; print(", ".join(""".["{}"]""".format(line.rstrip("\n")) for line in sys.stdin))') | wc -l
    3641

But if I ask for 3642, I get twice that:

    $ echo '{"1": 1}' | jq -f <(seq 3642 | python -c 'import sys; print(", ".join(""".["{}"]""".format(line.rstrip("\n")) for line in sys.stdin))') | wc -l
    7282

Looking at the actual output you'll see it repeats.  FWIW it doesn't seem to be a limit on the textual size of the expression, but the number of items.  I'm using jq 1.3 on Ubuntu 13.10 64-bit.

dtolnay:
This lets us use uint16_t for the program counter. JVM
languages have the same limit on method size so it is a
reasonable limit.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -1047,7 +1047,7 @@ static int expand_call_arglist(block* b) {
   return errors;
 }
 
-static int compile(struct bytecode* bc, block b) {
+static int compile(struct bytecode* bc, block b, struct locfile* lf) {
   int errors = 0;
   int pos = 0;
   int var_frame_idx = 0;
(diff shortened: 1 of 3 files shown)


nicowilliams:
The compiler could probably handle very large programs/functions by splitting the code into functions and join the code with calls to them.

nicowilliams:
LGTM!

dtolnay:
Rebased and merged 21309bc.

PR #879, opened 2015-07-29
Assignment operators should handle backtracking (fix #873)

### LINKED ISSUE #873: Unexpected "empty" behaviour
This could well be me misunderstanding the way `empty` works but this is unexpected:

```
borisov@deli:~ $ echo '{"key1": {"attr1": "foo"}, "key2": {"attr1": "foo", "attr2": "bar"}}' | jq '.[].attr2 |= if . then "bar" else empty end'                               
null
```

Expected something like this (but without .key1.attr2 set to "null"):

```
borisov@deli:~ $ echo '{"key1": {"attr1": "foo"}, "key2": {"attr1": "foo", "attr2": "bar"}}' | jq '.[].attr2 |= if . then "bar" else . end'
{
  "key1": {
    "attr1": "foo",
    "attr2": null
  },
  "key2": {
    "attr1": "foo",
    "attr2": "bar"
  }
}
```

jq version: jq-1.5rc2

nicowilliams:
This allows `.a=empty`, `.a|=empty`, and so on to work as one would
expect, namely leaving `.` alone.  The makes it easy to write things
like `.[].x|=(select(.=="foo")|"bar")` and have it work.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1369,8 +1369,8 @@ static const char* const jq_builtins[] = {
 #include "libm.h"
   "def add: reduce .[] as $x (null; . + $x);",
   "def del(f): delpaths([path(f)]);",
-  "def _assign(paths; value): value as $v | reduce path(paths) as $p (.; setpath($p; $v));",
-  "def _modify(paths; update): reduce path(paths) as $p (.; setpath($p; getpath($p) | update));",
+  "def _assign(paths; value): (value as $v | reduce path(paths) as $p (.; setpath($p; $v)))//.;",
+  "def _modify(paths; update): reduce path(paths) as $p (.; setpath($p; getpath($p) | update) // .);",
   "def map_values(f): .[] |= f;",
 
   // recurse


dtolnay:
Sorry I was away for most of the day so I missed the conversation on #873 but... I want to make sure we thoroughly consider whether this fix needs to be in 1.5 so late in the milestone. What is the likelihood that this introduces some behavior we haven't thought of that is even more confusing than the current behavior? I know that looking at the fix, I cannot say with certainty that this makes every use of `=` and `|=` behave intuitively.

The current behavior was in both 1.3 and 1.4 releases. The behavior questioned in #873 is not documented one way or the other. This leads me to prefer leaving this fix out of 1.5 so that we have a full milestone to test this fix and polish any remaining issues related to `=` and `|=`.

Also please add unit tests.

dtolnay:
As a concrete example, this probably was not intentional:

    $ ./jq-old -n '{} | . = false'
    false
    $ ./jq-new -n '{} | . = false'
    {}

nicowilliams:
Ah, good catch.  Yeah, needs more thought.

PR #887, opened 2015-08-07
Fix range(value;stream) (fix #886)

### LINKED ISSUE #886: range(VALUE, STREAM)
The Advanced Topics page (https://github.com/stedolan/jq/wiki/Advanced-Topics)
says: 

> range(0,1;3,4) is the same as range(0;3), range(0;3), range(1;3), range(1;4)

This was evidently intended to read:

> range(0,1;3,4) is the same as range(0;3), range(0;4), range(1;3), range(1;4)

but in any case, jq 1.4 and jq 1.5rc2 both treat range(0,1;3,4) simply as range(0,1;4).
Can treating the stream (3,4) as (4) be justified?

To sharpen the issue, compare:
```
jq -n 'range(0; (1,2))'
0
1

jq -n '(1,2) | range(0; . )'
0
0
1
```

dtolnay:


dtolnay:
I merged to master but not jq-1.5-branch.

nicowilliams:
Thanks.

PR #888, opened 2015-08-08
Script to update gh-pages from master

dtolnay:
Here is the script I have been using to do gh-pages updates. It seems useful enough to share.

### DIFF
--- a/scripts/update-website
+++ b/scripts/update-website
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+# This script builds the website from the docs directory of
+# the current branch and copies it over to the gh-pages
+# branch. WARNING: this script runs 'git clean -fd' after
+# checking out the gh-pages branch in order to remove files
+# that are gitignore'd on master but not gh-pages. If your
+# repo contains untracked files they will be lost.
+
+set -eu
+set -o xtrace
+
+# build website
+scriptdir=`dirname "$0"`
+cd "$scriptdir"/../docs
+rm -rf output
+rake build
+cd ..
+
+# copy to /tmp
+tmpdir=$(mktemp -d -t jq.website.XXXXXXXXXX)
+cp -r docs/output/* "$tmpdir"
+
+# copy to gh-pages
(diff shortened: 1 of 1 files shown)


nicowilliams:
Why not do a shallow clone in a temp dir to avoid having to git clean?
That's what i do... perhaps i ought to get used to never ever leaving any
files about uncommitted that i care about.

dtolnay:
I think there is a more general problem we can fix, which is that any time you checkout gh-pages from a branch on which you have run make, you end up with a ton of untracked files:

```
aclocal.m4       configure    jv_aux.o     jv_unicode.lo  main.o
autom4te.cache/  .deps/       jv_dtoa.lo   jv_unicode.o   Makefile
builtin.lo       execute.lo   jv_dtoa.o    lexer.lo       Makefile.in
builtin.o        execute.o    jv_file.lo   lexer.o        parser.lo
bytecode.lo      jq           jv_file.o    libjq.la       parser.o
bytecode.o       jq.1         jv.lo        .libs/         .remake-version-h
compile.lo       jq_test.lo   jv.o         libtool        tests/
compile.o        jq_test.o    jv_parse.lo  linker.lo      test-suite.log
config/          jv_alloc.lo  jv_parse.o   linker.o       util.lo
config.log       jv_alloc.o   jv_print.lo  locfile.lo     util.o
config.status    jv_aux.lo    jv_print.o   locfile.o      version.h
```

We could copy the .gitignore from master to gh-pages to make this nicer, then remove `git clean` from this script.

I don't know exactly what you have in mind for the shallow clone method but the reason I would not want this script to do a clone is that the clone would be missing any local git config in my existing clone. For example it would be missing `git config --local user.email`, which adds an extra step to committing and pushing to origin/gh-pages.

nicowilliams:
Good point, yeah.

nicowilliams:
Go ahead and push this.

dtolnay:
I added .gitignore to gh-pages, then rebased this without `git clean` as e404a1a.

PR #894, opened 2015-08-11
Fix #817

### LINKED ISSUE #817: incorrect handling of ./ in search path
Currently on master:

    jq$ ./jq -L tests/modules -n 'import "test_bind_order" as check; check::check'
    true
    jq$ ./jq -L ./tests/modules -n 'import "test_bind_order" as check; check::check'
    jq: error: module not found: test_bind_order0

    jq: 1 compile error

nicowilliams:
These two commits cleanup how we expand module search path elements.  Please review.

### DIFF
--- a/linker.c
+++ b/linker.c
@@ -25,6 +25,18 @@ static int load_library(jq_state *jq, jv lib_path, int is_data, int raw,
                         const char *as, block *out_block,
                         struct lib_loading_state *lib_state);
 
+static int path_is_relative(jv p) {
+  const char *s = jv_string_value(p);
+
+#ifdef WIN32
+  int res = PathIsRelativeA(s);
+#else
+  int res = *s != '/';
+#endif
+  jv_free(p);
+  return res;
+}
+
 
 // Given a lib_path to search first, creates a chain of search paths
 // in the following order:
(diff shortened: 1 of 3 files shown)


nicowilliams:
We could make assignment of PR to someone other than its author be a poke to another collaborator to do a code review.  Should we?

nicowilliams:
@joelpurra Can you be more specific?

nicowilliams:
I do need to add a test...

dtolnay:
Looks good to me. Seems to work correctly but I only tested on Linux. Please add a test for the case in the bug report.

dtolnay:
> We could make assignment of PR to someone other than its author be a poke to another collaborator to do a code review. Should we?

I don't think we need to assign PRs. The collaborators get emails about them so we know when one is waiting for a code review. After some reasonable amount of time (like 48 hours) a polite @mention should be enough.

nicowilliams:
@joelpurra ping.  Can you be more specific?

joelpurra:
@nicowilliams wrote:
> @joelpurra Can you be more specific?

@nicowilliams wrote:
> @joelpurra ping. Can you be more specific?

That was a comment on a code line; seems they're pulled in to the issue comment stream now; think they should be differentiated a bit more from ordinary comments.
https://github.com/nicowilliams/jq/commit/1d77ada1e6e5dbbe83bc29ba07c9440e57534eb6#diff-8950c7cecbfb6fbbee14e43bffd28221R199

Looking at the diff only, it seemed weird to use `err` when it's invalid, and not use it when it's valid.

dtolnay:
There is a type of jv called JV_KIND_INVALID which wraps an error message. The check for `!jv_is_valid(err)` is checking whether `err` is this type of jv.

I rebased these as 10df52e 5d08d54 b119853.

dtolnay:
Cherry-picked to jq-1.5-branch as 66f7ce5 0683002 b60c9eb.

joelpurra:
@dgryski: alright, thanks!

PR #895, opened 2015-08-11
walk/1

pkoppstein:
walk/1 has been stable for some time.  builtin.c has been tested, but manual.yml has not.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1640,7 +1640,18 @@ static const char* const jq_builtins[] = {
   "      else .[2]"
   "      end"
   "  end;",
+
+  // # Apply f to composite entities recursively, and to atoms
+  "def walk(f):"
+  "  . as $in"
+  "  | if type == \"object\" then"
+  "      reduce keys[] as $key"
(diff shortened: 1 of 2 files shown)


dtolnay:
The manual does not build. The tabs need to be replaced with 8 spaces and you are missing a closing single quote on line 1546.

dtolnay:
Rebased as 16d0e52 and dea7868. Thanks!

ulidtko:
The "run" button from docs https://stedolan.github.io/jq/manual/#walk(f) leads to a `walk/1 is not defined` error

pkoppstein:
jqplay currently uses jq 1.5, but `walk` was only added thereafter.

ulidtko:
How about not publishing docs for unreleased code then?

Not cool.

pkoppstein:
The "development" version of the documentation (https://stedolan.github.io/jq/manual/) not only states at the top "jq Manual (development version)" but also provides links to documentation corresponding to "official releases".

PR #902, opened 2015-08-12
Fix import-related assertion failures and leak

### LINKED ISSUE #900: jv_object_set: Assertion `jv_get_kind(object) == JV_KIND_OBJECT' failed.
Another bug from AFL:

```
jq -n 'import "a" as frt "b" ; def fooa: foas bar; def fooa: foo::a; [fooO, bar::a, bar:Nb, foo::a]'
```

gives:

```
jq: jv.c:1109: jv_object_set: Assertion `jv_get_kind(object) == JV_KIND_OBJECT' failed.
Aborted
```

when run with a version of jq built against 89897b4bcd7e2fbbe3fade3bd05fd39c712cc05f

dtolnay:


dtolnay:
I put two more commits in this PR because they both conflict with the original one, so separate PRs would be annoying.

compile.c
```diff
@@ -495,23 +495,26 @@ jv block_module_meta(block b) {
...
   meta = jv_object_set(meta, jv_string("is_data"), is_data ? jv_true() : jv_false());
   meta = jv_object_set(meta, jv_string("relpath"), jv_string(name));
   i->imm.constant = meta;
-  block_free(metadata);
   return inst_block(i);
 }
 
+block gen_import_meta(block import, block metadata) {
+  assert(block_is_single(import) && import.first->op == DEPS);
+  assert(block_is_const(metadata) && block_const_kind(metadata) == JV_KIND_OBJECT);
```
nicowilliams: Eh, we don't need to assert that the metadata is an object.  One could want to say `import "foo" 2.0;`!

parser.y
```diff
@@ -471,62 +482,58 @@ Term {
 }
 
 Import:
-"import" String "as" '$' IDENT ';' {
+ImportWhat ';' {
```
nicowilliams: I suppose the rules for import syntax could have used less violence, but it's fine.

parser.y
```diff
@@ -471,62 +482,58 @@ Term {
...
+  $$ = $1;
+} |
+ImportWhat Exp ';' {
+  if (!block_is_const($2)) {
+    FAIL(@$, "Module metadata must be constant");
+    $$ = gen_noop();
+    block_free($1);
+    block_free($2);
+  } else if (block_const_kind($2) != JV_KIND_OBJECT) {
+    FAIL(@$, "Module metadata must be an object");
```
nicowilliams: Ditto here: module and import metadata should be allowed to be things other than objects.

nicowilliams:
My substantive comments are that module and import metadata should probably be permitted to be things other than objects.  Of course, objects are most useful...

nicowilliams:
Otherwise LGTM.

compile.c
dtolnay: The old code requires an object because we put in `.as`, `.is_data`, and `.relpath`. The manual confirms:

> It should be an object with keys like “homepage” and so on.

So I will leave this commit as is, at least enforcing the requirement instead of failing an assertion. I filed #906 to support non-objects and fix the docs.

nicowilliams: M'ok.  Go ahead and push.

dtolnay:
Rebased as 5479d7f 7b5c6a9 58a053d.

PR #919, opened 2015-08-20
Use 'snprintf' instead of 'sprintf'

fcambus:


dtolnay:
Why? `buf` has room for 31 characters. The first case you changed always puts 6 characters. The second case always puts 12 characters.

fcambus:
The rationale behind this commit is getting rid of a linker warning when compiling on OpenBSD : 

./.libs/libjq.a(jv_print.o)(.text+0x2fc): In function `jvp_dump_string':
/root/jq-1.5/jv_print.c:118: warning: sprintf() is often misused, please use snprintf()

dtolnay:
Makes sense. I added your explanation to the commit message and merged as 5d23c24. Thanks!

nicowilliams:
There is a portability price to pay for this sort of change.  Depending on `snprintf()` affects portability much more than using `fileno()` and such.  Some open source projects, including one I also collaborate on (Heimdal) include portable implementations of such functions, and some go to great extents to be possible to build on rather old operating systems and with old build chains.  Personally I think that making jq build on lots of older systems and with older build tools is much too much work for a volunteer project, but I do object to this sort of change where we add a dependency on a newer function because some tool somewhere says that it's not safe to use the older function.  It would be like objecting to our use of git because it depends on an obsolete cryptographic hash function (SHA-1), which is true, and no, we're not going to abandon over SHA-1.  Security tools complain about the use of obsolete functions like SHA-1, of course, and as they should, but that does NOT mean that we should then fix every warning -- we must instead analyze warnings and decide whether to try to quieten them.

Incidentally, the SQLite3 dev community -quite rightly- refuses to fix warnings as reported by users (reports are quite frequent).  We should have a better policy: analyze the report and decide whether to fix/quieten it, but not guarantee that we will.

dtolnay:
Before merging I checked that we already depend on snprintf as of 63511a8 which you committed. We also depend on vsnprintf as of the much older a3e5090. Do you want me to revert this commit?

nicowilliams:
Ah, good.  No, no need to revert.  I just think we should have a policy for such reports.  You did the right thing.  Thanks!  (I should have checked myself.)

PR #924, opened 2015-08-23
Move source files to src/

dtolnay:
How do you feel about this?

This is the convention many other popular C projects on GitHub use. The code is easier to browse without having to dodge things like jq.spec or compile-ios.sh. See https://github.com/dtolnay/jq/tree/b377147c5b6aa787163263c11e59bd716fbd3e07/src.

This does not break git blame, neither in GitHub nor locally. It also does not cause problems rebasing old commits on top of this.

To tidy up further, I could make a separate directory called src/generated/ for the compiled lexer and parser and version.h.


nicowilliams:
This has come up before.  IIRC there are some change-following problems in git (git log?).  I personally am all "meh" about this, but I agree that for newcomers who want to browse the repo via the github web interface this would be nice.  What would be left at the root?

dtolnay:
git log --follow -- src/jv.h
    git log --follow -- jv.h

Both work perfectly. This is the easiest case for git because the files are *only* renamed - no lines changed.

Here is what is left at the root: https://github.com/dtolnay/jq/tree/b377147c5b6aa787163263c11e59bd716fbd3e07

nicowilliams:
I tried rebasing a branch of mine onto your rename-sources (`src`) branch and it worked fine.  I also checked that the similarity index reported for all renamed files is 100%.

I think this will be OK, and it LGTM.

dtolnay:
Rebased as 0c93eb3.

nicowilliams:
Thanks!

PR #926, opened 2015-08-24
Remove jq.1.default and related content

nicowilliams:
We no longer need the old default manual.  Remove it.


dtolnay:
Looks good

PR #937, opened 2015-09-05
Get oniguruma from github instead of geocities

dtolnay:


pkoppstein:
@dtolnay - Maybe wget doesn't mind but it might be better to remove the redundancy in the following line in Dockerfile:
```
wget https://github.com/kkos/oniguruma/archive/v5.9.6.tar.gz -O onig-5.9.6.tar.gz -O onig-5.9.6.tar.gz && \
```

dtolnay:
Good catch. I don't know how I got that in there twice. Fixed now.

PR #939, opened 2015-09-07
Segmentation fault on Windows with long paths

### LINKED ISSUE #811: File names with non ASCII characters

svnpenn:
Consider this script:

~~~sh
#!/bin/sh
set 6789012345678901234567{,8,89}.json
echo '{"delta": 2}' | tee $*
jq . $1
jq . $2
jq . $3
~~~

With JQ master, here is the result:

~~~
{
  "delta": 2
}

Segmentation fault
~~~

So path of 27 characters or less is good, 28 gives no output, and 29 or more is
segfault. Git bisect reveals 7811ef1e1746f0963deb960af2c9623fb53c24a5 to be the
culprit:

~~~
7811ef1e1746f0963deb960af2c9623fb53c24a5 is the first bad commit
commit 7811ef1e1746f0963deb960af2c9623fb53c24a5
Date:   Fri Jun 19 17:10:45 2015 -0500
Fix #811: use CommandLineToArgvW() and _wfopen()
~~~

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -47,7 +47,7 @@ void *alloca (size_t);
 
 #ifdef WIN32
 FILE *fopen(const char *fname, const char *mode) {
-  size_t sz = MultiByteToWideChar(CP_UTF8, 0, fname, -1, NULL, 0);
+  size_t sz = sizeof(wchar_t) * MultiByteToWideChar(CP_UTF8, 0, fname, -1, NULL, 0);
   wchar_t *wfname = alloca(sz);
   MultiByteToWideChar(CP_UTF8, 0, fname, -1, wfname, sz);
 


svnpenn:
@nicowilliams please review pull request, I use code from this example

http://lists.gnu.org/archive/html/bug-wget/2015-08/msg00146.html

nicowilliams:
Great, one of the msdn pages doesn't mention that it's wide characters that the function counts.

LGTM.  My current dev system is screwed up, so I won't merge this for a few days yet.

dtolnay:
LGTM too. I changed the commit message and merged as 63dd033.

nicowilliams:
Thanks.  We have a few bugs in 1.5 that are somewhat serious, like this one.  We might want to consider a 1.5.1 release.  What do you think @dtolnay?

dtolnay:
I made a 1.5.1 release milestone, but my preference is to just do 1.6 sooner instead. Were you thinking of doing 1.5.1 by cherry-picking only the critical fixes onto the 1.5 release, or by releasing from master? I guess the downside of doing 1.6 is there are no headline "features" yet, only bugfixes.

pkoppstein:
I just noticed there is no 1.5 tag, which is consequential e.g. when using SourceTree:
```
$ git tag
jq-1.0
jq-1.1
jq-1.2
jq-1.3
jq-1.4
jq-1.5rc1
jq-1.5rc2
$ 
```

dtolnay:
There is: https://github.com/stedolan/jq/releases/tag/jq-1.5. A `git fetch` should create it for you locally.

pkoppstein:
@dtolnay wrote: 
> There is

Hmmm. It doesn't show up in SourceTree at all (after pulling all commits); nor after a manual *git fetch*:
```
$ git fetch ; git log -1 ; git tag
commit 63dd033da7249d0afbbd52c5a1cb547b7653409f
Author: Steven Penny <svnpenn@gmail.com>
Date:   Mon Sep 7 11:29:32 2015 -0500

    Fix segmentation fault on Windows with long paths (fix #939)
jq-1.0
jq-1.1
jq-1.2
jq-1.3
jq-1.4
jq-1.5rc1
jq-1.5rc2
$

**Comment by svnpenn** [comment]
@dtolnay @pkoppstein is right. I had the same problem when I was troubleshooting this issue. You may need to `git push --tags` from your machine

**Comment by nicowilliams** [comment]
git fetch --tags

**Comment by svnpenn** [comment]
@dtolnay @pkoppstein is right. I had the same problem when I was troubleshooting this issue. You may need to `git push --tags` from your machine

**Comment by dtolnay** [comment]
```
$ git clone git@github.com:stedolan/jq.git && cd jq && git tag
Cloning into 'jq'...
remote: Counting objects: 6038, done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 6038 (delta 0), reused 0 (delta 0), pack-reused 6034
Receiving objects: 100% (6038/6038), 4.43 MiB | 696.00 KiB/s, done.
Resolving deltas: 100% (3604/3604), done.
Checking connectivity... done.
jq-1.0
jq-1.1
jq-1.2
jq-1.3
jq-1.4
jq-1.5
jq-1.5rc1
jq-1.5rc2
```

nicowilliams:
I'll double check after lunch.

svnpenn:
@dtolnay hm ok, I will check when I get home.

pkoppstein:
> git fetch --tags

That did it, but all the other tags always appeared without having to do that.  One of git's .... shall we say, oddities?

nicowilliams:
@pkoppstein @svnpenn Tags in git have a global namespace, unlike branches.  Therefore when fetching the remote's tags can conflict with the local clone.  To prevent accidents, git fetch does not fetch tags by default.  You have to ask for it.

From the git-fetch(1) manpage:

    The default behavior for a remote may be specified with the remote.<name>.tagopt setting. See git-config(1).

nicowilliams:
@dtolnay Either way would work for me.  I'd like to wait a bit longer though, see what else shakes out.

nicowilliams:
Of course, always fetching tags is almost always the right thing.  If you want something like local tags then use branches.  So, yes, this is one of git's oddities.

pkoppstein:
@nicowilliams wrote:
> git fetch does not fetch tags by default.

Ah, if only git were as simple as that.  From the git-fetch man page:

> By default, any tag that points into the histories being fetched is also fetched; the effect is to fetch tags that point at branches that you are interested in. 

That's presumably why in the past this has not been an issue.  

So the question now is: why has git "forgotten" that I'm still interested in the same branches as I was before?  I suppose the answer lurks somewhere in .git/config?

nicowilliams:
Ah, so what happened is that @dtolnay deleted the jq-1.5 branch when we were done with the 1.5 release.  It didn't seem necessary as we weren't going to do any further development on that branch.

And that explains that.  Thanks for that quote @pkoppstein.

dtolnay:
Glad we figured this out. It sounds like deleting the release branch after tagging was a bad move.

pkoppstein:
> None of us realized this would be a problem

Yes, a frequent refrain when it comes to git.  That would make a nice epitaph :-)

Without first deleting and then cloning a git repository, is there a recommended way to "reclone" a repository?   Is there a way to "reclone" a github fork?

nicowilliams:
@pkoppstein I don't recall having any serious problems with git ever, and I'd not consider this a serious problem.  Personally, I adore git.

There are things I think it's missing, well, *two* things: a) branch versioning (so that rebasing retains history at least locally for the developer's sanity), and b) grouping commits at push time and push history (so that one may see that so and so pushed some commits together, and who did it, which is different than who authored the commits).  Oh, I suppose there's one more thing: SHA-1 is long in the tooth, and it'd be nice if git could switch to a better hash function with a longer output.

nicowilliams:
I've restored the `jq-1.5-release` branch to avoid this in the future.

svnpenn:
Seems to be working as expected now, if it was ever broken. Thanks!

svnpenn:
https://github.com/stedolan/jq/issues/1155

cup:
https://github.com/stedolan/jq/issues/1066

PR #956, opened 2015-09-21
Delete negative indices in array (fix #954)

### LINKED ISSUE #954: Negative indices in path_expressions don't work?
Hi, negative indices don't appear to work as expected in path_expressions.

And by "as expected" I mean with the same semantics they have in value retrieval and assignment:

Retrieval:
`echo '[1,2,3]' | jq '.[-1]'`
Result:
`3`

Assignment:
`echo '[1,2,3]' | jq '.[-1]=5'`
Result:
```
[
  1,
  2,
  5
]
```

**But not:** removal with a path_expression:
`echo '[1,2,3]' | jq 'del(.[-1])'`
Result:
```
[
  1,
  2,
  3
]
```

Is this an oversight, or is there some reason this doesn't work?

dtolnay:


src/jv_aux.c
```diff
@@ -222,13 +222,17 @@ jv jv_dels(jv t, jv keys) {
     // no change
```
nicowilliams: Huh.  jv_dels() should be static or else it should have a prototype in jv.h.  I don't know that I care which (I see no reason that it wouldn't be stable, if we add the prototype).

dtolnay: I will make it static in a separate commit. `jv_delpaths` should be enough until someone specifically requests this in libjq for efficiency.

src/jv_aux.c
```diff
@@ -222,13 +222,17 @@ jv jv_dels(jv t, jv keys) {
...
     jv new_array = jv_array();
     jv starts = jv_array(), ends = jv_array();
-    jv_array_foreach(orig_keys, i, key) {
+    jv_array_foreach(keys, i, key) {
       if (jv_get_kind(key) == JV_KIND_NUMBER) {
-        keys = jv_array_append(keys, key);
+        if (jv_number_value(key) < 0) {
+          neg_keys = jv_array_append(neg_keys, key);
+        } else {
+          nonneg_keys = jv_array_append(nonneg_keys, key);
```
nicowilliams: Couldn't we normalize the negative index here and be done?

nicowilliams:
One of my comments needs an answer, but even so, LGTM, and if you want to push as is that's fine.

(If you make changes as a result of my comments then squash before pushing.)

src/jv_aux.c
dtolnay: No because the code below expects indices to be sorted. Say you start with an array of size 10 and indices [-2, -1, 1, 2]. After normalizing, it would be [8, 9, 1, 2]. Solution A would be to rotate the array after this loop to make it sorted again, which I am not happy with. Solution B would be to keep track of the first nonnegative index while normalizing, then continue to use two pointers in the code below, but they would be pointers into the same array. The code for solution B would look roughly the same as this solution, but I think having separate arrays is simpler to understand.

dtolnay:
I merged 58f082d which squashes in the following change:

    -jv jv_dels(jv t, jv keys) {
    +static jv jv_dels(jv t, jv keys) {

src/jv_aux.c
nicowilliams: Thanks.

PR #958, opened 2015-09-22
EOF after newline in string mishandled (fix #951)

### LINKED ISSUE #951: An unterminated string is not treated as an error
JSON input containing an unterminated string should be treated as an error. Instead, jq produces no output and exits with `0` status:

    $ echo '"foo' | jq
    $ echo $?
    0

That's a simple case, and maybe an uncommon one, but more complex JSON with the same issue behaves the same way:

    $ echo '{"foo: 1 }' | jq
    $ echo $?
    0

In these cases `jq` should print an error consistent with other kinds of invalid JSON (even if that error is a bad one right now, per issue #501), and exit with a non-zero status. In other words, it should match the behavior already present for other kinds of invalid JSON:

    $ echo 'foo"' | jq
    parse error: Invalid numeric literal at line 1, column 4
    $ echo $?
    4

nicowilliams:


tests/shtest
```diff
@@ -109,10 +109,25 @@ printf '1' | $JQ -ce --seq . > $d/out 2>&1
...
+fi
 cmp $d/out $d/expected
 
+# Regression test for #951
+printf '"a\n' > $d/input
+if $VALGRIND $Q $JQ -e . $d/input; then
+  printf 'Issue #951 is back?\n' 1>&2
+  exit 2
+else
+  true
```
dtolnay: Why is this required?

nicowilliams: Why is what required?  The `else true`?  That's me being sloppy (I'd had `else echo ...`).  I've pushed a commit to fix that.

nicowilliams:
The decrease in code coverage is... odd.  I'll have to spend some time on that.

dtolnay:
I squashed and merged as c4524da.

The decrease in code coverage is most likely because of the /dev/urandom test in shtest, not because of this. There is a setting in coveralls that controls how much coverage needs to drop for it to fail the build. I changed it to 10pp so this should not happen again.

nicowilliams:
Ah, OK, thanks.

Issue #435, opened 2014-06-21
Split up doc Rakefile to minimize dependencies


else:
Please consider splitting up the Rakefile for doc generation, because many of the dependencies are not needed for the generation of the manpage. Specifically, the following libraries are unused for the task `manpage`:

* bonsai
* liquid
* maruku

On Debian, bonsai is not available as a package, and as long as it's not required for the manpage generation, I'd prefer not to be required to package it.

Issue #434, opened 2014-06-21
jq build version generation: include version in tagged releases


elsbrock:
configure.ac looks up jq versioning information using the git command. In many packaging setups (such as Debian's) however, there is no access to the .git repository during package building. I propose to include versioning information in release tarballs, so that maintainers don't need to set that manually. One tool that could be used for this is `autorevision`, which can e.g. generate a header file.

nicowilliams:
Hmmm, I'm not sure what to do here.  The tarball includes a ./configure doesn't use git.  The configure.ac does, but why autoreconf if not in git?

nicowilliams:
I'll take a look at autorevision.  Perhaps it will help.

nicowilliams:
I guess the setup.sh script could run autorevision to make an m4 file for configure.ac, and a version.h so that we don't need anything else like scripts/version.  Although I'd rather avoid yet another external dependency for building from git.  And anyways, the trick is to make sure that this runs when building from git and that its output is dist'ed when doing `make dist`.

dtolnay:
@else can you point me to an autoconf project that does this correctly?

jsoref:
Please note that:
1. debian trixie shipped a `1.7.1` binary that claims to be `1.7`
2. ubuntu noble ships a similar version of that same package with that same problem
3. github's ubuntu-24.04 similarly claims to have a `1.7` which is also really a `1.7.1`

Saur2000:
OpenEmbedded always runs `autoreconf` for all Autotools-based code. When using the released tarballs of `jq` this results in `PACKAGE_VERSION` being set to the empty string, and, e.g., producing an invalid `libjq.pc` file with `version` also unset.

PR #984, opened 2015-10-12
docs: jq 1.5 is now available in Debian unstable

else:
I have updated the Debian packaging to the 1.5 upstream release. You can find a changelog of the packaging [here](http://packages.qa.debian.org/j/jq/news/20151012T071912Z.html).

The new version of jq in Debian does not build libjq separately but bakes it into the binary; however, libonig2 is linked dynamically.

You can check the list of [patches](https://anonscm.debian.org/cgit/users/else-guest/jq.git/tree/debian/patches) that are applied to the source in order to be compliant with the Debian policy and fix Lintian warnings.

Particularly, the following things could be improved:
* include version in tagged releases (without requiring .git): #434 
* split up documentation build so that manpage can be built with minimum set of dependencies: #435

This time jq had to be repackaged in order to remove a minified Bootstrap Javascript resource (DFSG requirement). Please consider splitting up jq-src and homepage source.

Thank you for your work!

### DIFF
--- a/docs/content/2.download/default.yml
+++ b/docs/content/2.download/default.yml
@@ -16,7 +16,7 @@ body:
 
       ### Linux
 
-       * jq 1.4 is in the official [Debian](https://packages.debian.org/jq) and
+       * jq 1.5 is in the official [Debian](https://packages.debian.org/jq) and
          [Ubuntu](http://packages.ubuntu.com/jq) repositories. Install using
          `sudo apt-get install jq`.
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks Simon. I will prioritize #434 and #435. Additionally, I filed #985 to get rid of the Bootstrap resource. Please continue to file issues for anything we could do to make packaging easier.

PR #987, opened 2015-10-12
Add KEYS file

nicowilliams:
I've no idea if there's a convention for this.

I saw this http://mail-archives.apache.org/mod_mbox/ambari-commits/201504.mbox/%3C197614c190e149099f7ad0f3d7ea2d6f@git.apache.org%3E which makes me think a file named KEYS constructed thusly will work:

    (printf '\n'; gpg --list-keys|grep -v ^/; gpg --export -a) >> KEYS

I like it well enough.


dtolnay:
I think we can just add the 160-bit fingerprints for the three maintainers to the AUTHORS file, as long as the keys are on a reasonably popular keyserver like pgp.mit.edu.

nicowilliams:
Sure.  I'll squash and push then.

nicowilliams:
Closed with 583ab58.

dtolnay:
Any reason to include the expired one that you don't use?

nicowilliams:
D'oh.

PR #989, opened 2015-10-14
Typo fix

netei:


dtolnay:
I think the original is correct. The wording means "does not output anything, not even an error." I don't think the new wording makes this clearer.

nicowilliams:
The original wording means "it outputs nothing (not even an error) when `.` is neither an object nor an array", and the proposed new wording means "it does not output an error when `.` is neither an object nor an array" but says nothing about any values that might be output in that case except implicitly by reference to "`like `.foo`".  So it's not really a change at all.

netei:
I think the current wording is still misleading. It should probably be : 

> Just like `.foo`, but does not output an error when `.` is not an array or an object.

Before it was : 

> Just like `.foo`, but does not output even an error when `.` is not an array or an object.

The "even" is unecessary here.

pkoppstein:
@netei - Evidently you are misunderstanding something. The "even" is not only correct, but essential to capture the intended meaning.

nicowilliams:
Just like .foo, but does not output an error when . is not an array or an object.

Does work.  It's true that the "even" is redundant when one knows how `.foo` works.  Still, the "even" isn't wrong.

PR #991, opened 2015-10-18
Support numbers and booleans in join (fix #930)

### LINKED ISSUE #930: Join different value types
It should be possible to join different types of values like in JS.
```
$ echo '["1",2]' | jq -r 'join(",")'
jq: error (at <stdin>:1): string (",") and number (2) cannot be added
```

dtolnay:


dtolnay:
I think it is pretty reasonable to expect join to work for scalar types. I don't know about objects and arrays.

nicowilliams:
And we don't want it to `tostring` `null` either, I guess?  Anyways, LGTM, but maybe the manual should say something about booleans and numbers being stringified?

dtolnay:
Rebased as e17ccf2.

PR #992, opened 2015-10-18
Move jq-coded builtins to non-C file (fix #424)

### LINKED ISSUE #424: jq-coded builtins should live in a separate, non-C file
It'd be nice if there was no need to escape double-quotes for C, and if the only thing to do to add jq-coded builtins was to append them to a file, then add docs and tests.

dtolnay:


Makefile.am
```diff
@@ -86,6 +86,10 @@ src/version.h: .remake-version-h
 	$(AM_V_GEN) $(generate_ver); echo "$$ver" > $@
 src/main.c: src/version.h
 
+src/builtin.inc: src/builtin.jq
+	$(AM_V_GEN) sed -e 's/\\/\\\\/g' -e 's/"/\\"/g' -e 's/^/"/' -e 's/$$/\\n"/' $^ > $@
```
nicowilliams: Why `s/$$/../`, why not `s/$/...`?

src/builtin.c
```diff
@@ -1698,10 +1413,8 @@ int builtins_bind(jq_state *jq, block* bb) {
     block_free(*bb);
     return nerrors;
   }
-  for (int i=(int)(sizeof(jq_builtins)/sizeof(jq_builtins[0]))-1; i>=0; i--) {
-    nerrors = builtins_bind_one(jq, bb, jq_builtins[i]);
-    assert(!nerrors);
-  }
+  nerrors = builtins_bind_one(jq, bb, jq_builtins);
```
nicowilliams: Excellent!

nicowilliams:
I only have one comment (see diffs).

I have a half-baked commit lying around that looks a lot like this, but with the addition of supporting multiple builtin modules (which can wait for some other day; the idea is that users could `include jq "1.4";`, which would allow us to start to make backwards-incompatible changes to the builtins (including `|=` and such syntax that depends on jq-coded helpers) comfortably.

nicowilliams:
Other than my comment about that `sed`, this LGTM.

Makefile.am
dtolnay: https://www.gnu.org/software/make/manual/html_node/Variables-in-Recipes.html

> if you want a dollar sign to appear in your recipe, you must double it (‘$$’)

dtolnay:
Rebased as 143e8dc.

Makefile.am
nicowilliams: Oh, duh, it's make, and I brainfarted.  Sorry.

PR #1004, opened 2015-10-28
Add address sanitizer (ASAN) support (fix #1002)

### LINKED ISSUE #1002: Option to use ASAN instead of valgrind

nicowilliams:


configure.ac
```diff
@@ -117,7 +122,7 @@ AS_IF([test "x$enable_docs" != "xno"],[
 *****************************************************************
 *  Ruby dependencies for building jq documentation not found.   *
 *  You can still build, install and hack on jq, but the manpage *
-*  will not be rebuilt and some of the tests won't run.         *
```
nicowilliams: Ah, I meant to commit this separately.  This fixes vim syntax mis-highlighting.

configure.ac
```diff
@@ -130,11 +135,16 @@ EOF
 ])
 
 AM_CONDITIONAL([ENABLE_VALGRIND], [test "x$enable_valgrind" != xno])
+AM_CONDITIONAL([ENABLE_ASAN], [test "x$enable_asan" = xyes])
 AM_CONDITIONAL([ENABLE_GCOV], [test "x$enable_gcov" = xyes])
 AM_CONDITIONAL([ENABLE_DOCS], [test "x$enable_docs" != xno])
 AM_CONDITIONAL([ENABLE_ERROR_INJECTION], [test "x$enable_error_injection" = xyes])
 AM_CONDITIONAL([ENABLE_ALL_STATIC], [test "x$enable_all_static" = xyes])
 
+if test x$enable_valgrind != xno -a x$enable_asan = xyes; then
```
nicowilliams: Ah, and I meant to remove this.

nicowilliams:
OK, I think it's ready.  ASAN is fast enough that one can run with it compiled-in in many cases.  I could see releasing executables with and without ASAN.

nicowilliams:
OTOH, it's not as good as valgrind...

configure.ac
```diff
@@ -89,6 +89,11 @@ dnl fix leaks.
 AC_ARG_ENABLE([valgrind],
    AC_HELP_STRING([--disable-valgrind], [do not run tests under Valgrind]))
 
+dnl Running tests with Valgrind is slow; address sanitizer (ASAN) is
+dnl faster.
+AC_ARG_ENABLE([asan],
+   AC_HELP_STRING([--enable-asn], [enable address sanitizer]))
```
dtolnay: `--enable-asan`

dtolnay:
Does asan catch anything valgrind does not? If so, we should run a fourth travis build with asan instead of valgrind.

nicowilliams:
From what I can tell, no, it catches a subset of the classes of errors that valgrind catches.  For example, it doesn't catch reads of uninitialized memory.  I'm thinking this is useful when there's no valgrind, or -because it's faster than valgrind- during development prior to sending a PR.

dtolnay:
Okay. Looks good once you fix the typo.

nicowilliams:
Thanks.

Issue #1013, opened 2015-11-07
'fold' option when pretty printing to print multiple numbers on a line


abetusk:
The problem:

```bash
$ jq . data.json
{
  "data": [
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
[...]

PR #1014, opened 2015-11-07
Resolves issue #1013

abetusk:
Added a '--fold' option to the command line.  As discussed in the issue, this only affects pretty printing and numbers in a list.

A new `process` was created along with `jv_print_extra_opt` to leave the original functions untouched.

There is an unused structure variable in the `jv_extra_opt` for indentation that could be used later.

Testing is minimal (I tested it on a few examples, some of which appeared in #1013).  I wasn't sure how to incorporate these changes in the tests that I saw.

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -18,6 +18,16 @@ typedef enum {
 
 struct jv_refcnt;
 
+
+struct jv_extra_opt_t {
+  uint32_t flag;
+  int array_fold;
+  int array_fold_indent;
+};
+
+typedef struct jv_extra_opt_t jv_extra_opt;
+
+
 /* All of the fields of this struct are private.
    Really. Do not play with them. */
 typedef struct {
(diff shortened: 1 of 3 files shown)


dtolnay:
Unfortunately I do not expect this to be broadly useful enough to warrant a command line flag.

If you are intending this to be used with bitmap data as in your first example, I think that is better served by this substantially simpler implementation which you could put in a library if you need it commonly:

    jq <data.json -r '
        def bmp($w):
            range(0; length/$w) as $row |
            .[$row*$w : ($row+1)*$w] | join("");
        .data | bmp(32)'

If you are intending this to make JSON output with unwieldy arrays more human-readable as in your second example, I think that is better served by JSON folding provided by [vim-json](https://github.com/elzr/vim-json) and similar for other editors.

Mentioning @nicowilliams in case he has a different opinion.

nicowilliams:
I like @dtolnay's solution.  Does it work for you @abetusk?

nicowilliams:
@dtolnay's approach isn't easy to apply recursively all the way down without jq-coding a JSON encoder.  A jq-coded JSON encoder is actually quite an interesting idea, as would be a jq-coded JSON parser.

abetusk:
OK, going to close, thanks for taking a look.

PR #1018, opened 2015-11-15
Fix manual signature for `split` (non-regex)

Loamhoof:
The documentation for `split` is inconsistent with the rest, as the function requires an argument which is not shown in the signature.


dtolnay:
Thanks!

PR #1019, opened 2015-11-16
Improve irc.lc link in the README

bmintz:


dtolnay:
Firefox and Chrome both block the https link because it contains an iframe that is not https, so I would prefer to keep the http link.

PR #1022, opened 2015-11-21
Support --without-oniguruma

### LINKED ISSUE #1021: No way to deterministically disable liboniguruma
You have --with-oniguruma, which specifies the oniguruma install path.
However, without --with-oniguruma configure will attempt to find it and either will or will not succeed.

--without-oniguruma breaks. You should allow to disable oniguruma even when it is installed for people who wouldn't like to link to it.

You should have --with-oniguruma, --without-oniguruma, and --oniguruma-path.
Currently --with-oniguruma is one for all, not very clear. --oniguruma-path will be used very rarely, only for really non-standard paths.

dtolnay:
fixes #1021


nicowilliams:
LGTM.  Thanks!

PR #1024, opened 2015-11-26
update docs to use dnf instead of yum

mmckinst:
`yum` has been replaced with `dnf` as of f22, using yum will give you a deprecation warning. see https://fedoraproject.org/wiki/Changes/ReplaceYumWithDNF for more info


dtolnay:
Thanks!

PR #1030, opened 2015-12-02
Assume . if either stdin/on isatty() (fix #1028)

### LINKED ISSUE #1028: Redirecting or piping output has different behaviour
This caught me by surprise -- omitting the (mandatory) jq filter causes the help to be printed instead, but only if the output is redirected to a file or piped: the `.` filter is implied if redirection does not occur.

```
> echo '"x"' | ./jq -r
x
> echo '"x"' | ./jq -r > foo
jq - commandline JSON processor [version 1.5]
Usage: ./jq [options] <jq filter> [file...]
```

This is particularly mystifying if jq is invoked within a shell script, since it means the shell scripts' behaviour radically changes.

The obvious workaround is to use the `.` filter, but it's not obvious you need it when it sometimes succeeds.

I don't have a strong preference for whether it should always fail or always assume `.` as a default filter.

nicowilliams:
And move the #define of isatty on Windows up.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -17,6 +17,15 @@
 #include <wtypes.h>
 #endif
 
+#if !defined(HAVE_ISATTY) && defined(HAVE__ISATTY)
+#undef isatty
+#define isatty _isatty
+#endif
+
+#if defined(HAVE_ISATTY) || defined(HAVE__ISATTY)
+#define USE_ISATTY
+#endif
+
 #include "compile.h"
 #include "jv.h"
 #include "jq.h"
@@ -423,7 +432,8 @@ int main(int argc, char* argv[]) {
     }
   }
 
-  if (isatty(fileno(stdout))) {
+#ifdef USE_ISATTY
+  if (isatty(STDOUT_FILENO)) {
     dumpopts |= JV_PRINT_ISATTY;
 #ifndef WIN32
   /* Disable color by default on Windows builds as Windows
@@ -431,6 +441,7 @@ int main(int argc, char* argv[]) {
     dumpopts |= JV_PRINT_COLOR;
 #endif
   }
+#endif
   if (options & SORTED_OUTPUT) dumpopts |= JV_PRINT_SORTED;
   if (options & ASCII_OUTPUT) dumpopts |= JV_PRINT_ASCII;
   if (options & COLOR_OUTPUT) dumpopts |= JV_PRINT_COLOR;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -15,6 +15,13 @@
 #include <shellapi.h>
 #include <wchar.h>
 #include <wtypes.h>
+
+#ifdef HAVE__ISATTY
+#ifdef isatty
```
dtolnay: C standard says about #undef:

> It is ignored if the specified identifier is not currently defined as a macro name.

So this ifdef is redundant.

src/main.c
```diff
@@ -15,6 +15,13 @@
 #include <shellapi.h>
 #include <wchar.h>
 #include <wtypes.h>
+
+#ifdef HAVE__ISATTY
```
dtolnay: This is not the same behavior as what you removed below. Below, if both are defined we replace isatty with _isatty even ifndef WIN32, but this is just for WIN32. Is that intentional?

src/main.c
```diff
@@ -457,16 +464,8 @@ int main(int argc, char* argv[]) {
...
 
-#if (!defined(WIN32) && defined(HAVE_ISATTY)) || defined(HAVE__ISATTY)
-
-#if defined(HAVE__ISATTY) && defined(isatty)
-#undef isatty
-#define isatty _isatty
-#endif
-
-  if (!program && isatty(STDOUT_FILENO) && !isatty(STDIN_FILENO))
+  if (!program && (isatty(STDOUT_FILENO) || !isatty(STDIN_FILENO)))
```
dtolnay: This needs an ifdef isatty around it.

dtolnay: Also this is not the right condition. We want to assume "." if either one is not a tty.

src/main.c
nicowilliams: I think so, yes :)  As in: I've never heard of `_isatty()` anywhere else.

src/main.c
nicowilliams: Oy, I thought I pushed the correction.  I noted it in the issue.

nicowilliams: We didn't guard the other use of `isatty()` with an `#ifdef`.  The code assumed there's always an `isatty()`, though it might be `_isatty()`, or it might not be a `#define`, so either way we shouldn't `#ifdef isatty`, but either leave it as is or `#define` some other macro for the `#ifdef`.  Comments?

dtolnay: No guard then. I didn't realize this was the only guarded one.

nicowilliams:
Oops, I squashed my fixup.

PR #1031, opened 2015-12-02
Fix JV_OBJECT() and go up to 18

nicowilliams:
The `JV_OBJECT()` macro was rather broken.  And this brings it up to 9 key/value pairs (exactly what I need to dump a `block`).


nicowilliams:
Thanks.

nicowilliams:
This kind of macro makes me wish for macro-defining macros.

wtlangford:
Is there a reason this was a macro instead of a vararg function? I suppose this is marginally faster, but is certainly unpleasant to behold.

nicowilliams:
Heh, yeah, maybe!

PR #1032, opened 2015-12-02
Dump block

nicowilliams:
This is a work in progress.  We'll need a new jq compile function to take the dump_block option.  Or we could overload the existing `jq_compile_args()` `args` argument to accept special arguments (ones whose names are not valid identifiers).  I vote for a new function.  Comments?

### DIFF
--- a/src/execute.c
+++ b/src/execute.c
@@ -1142,6 +1142,10 @@ jv jq_get_attr(jq_state *jq, jv attr) {
   return jv_object_get(jv_copy(jq->attrs), attr);
 }
 
+jv jq_get_option(jq_state *jq, jv option) {
+  return jv_getpath(jv_copy(jq->attrs), JV_ARRAY(jv_string("options"), option));
+}
+
 void jq_dump_disassembly(jq_state *jq, int indent) {
   dump_disassembly(indent, jq->bc);
 }
--- a/src/execute.c
+++ b/src/execute.c
@@ -1081,6 +1081,7 @@ int jq_compile_opts(jq_state *jq, const char* str, jv opts, jv args) {
     bytecode_free(jq->bc);
     jq->bc = 0;
   }
+  jq_set_attr(jq, jv_string("options"), jv_copy(opts));
   int nerrors = load_program(jq, locations, &program);
   if (nerrors == 0) {
     jv_array_foreach(args, i, arg) {
--- a/src/execute.c
+++ b/src/execute.c
@@ -1067,6 +1067,10 @@ static struct bytecode *optimize(struct bytecode *bc) {
 }
 
 int jq_compile_args(jq_state *jq, const char* str, jv args) {
+  return jq_compile_opts(jq, str, jv_object(), args);
+}
+
+int jq_compile_opts(jq_state *jq, const char* str, jv opts, jv args) {
   jv_nomem_handler(jq->nomem_handler, jq->nomem_handler_data);
   assert(jv_get_kind(args) == JV_KIND_ARRAY);
   struct locfile* locations;
@@ -1088,13 +1092,21 @@ int jq_compile_args(jq_state *jq, const char* str, jv args) {
 
     nerrors = builtins_bind(jq, &program);
     if (nerrors == 0) {
-      nerrors = block_compile(program, &jq->bc, locations);
+      program = block_drop_unreferenced(program);
+      /*
+       * We could dump_block(program) here, but block_compile() modifies
+       * the program it in important ways, so we leave it to
+       * block_compile() to dump it.
+       */
+      int dump = jv_equal(jv_true(), jv_object_get(jv_copy(opts), jv_string("dump_block")));
+      nerrors = block_compile(program, &jq->bc, locations, dump);
     }
   }
   if (nerrors)
     jq_report_error(jq, jv_string_fmt("jq: %d compile %s", nerrors, nerrors > 1 ? "errors" : "error"));
   if (jq->bc)
     jq->bc = optimize(jq->bc);
+  jv_free(opts);
   jv_free(args);
   locfile_free(locations);
   return jq->bc != NULL;
--- a/src/parser.y
+++ b/src/parser.y
@@ -135,6 +135,8 @@ struct lexer_param {
 void yyerror(YYLTYPE* loc, block* answer, int* errors,
              struct locfile* locations, struct lexer_param* lexer_param_ptr, const char *s){
   (*errors)++;
+  if (answer)
+    mark_error(*answer);
   if (strstr(s, "unexpected")) {
 #ifdef WIN32
       locfile_locate(locations, *loc, "jq: error: %s (Windows cmd shell quoting issues?)", s);
@@ -314,8 +316,8 @@ Module:
 "module" Exp ';' {
   if (!block_is_const($2)) {
     FAIL(@$, "Module metadata must be constant");
-    $$ = gen_noop();
-    block_free($2);
+    mark_error($2);
+    $$ = $2;
   } else {
     $$ = gen_module($2);
   }
@@ -363,6 +365,7 @@ Term "as" Pattern '|' Exp {
 } |
 "if" Exp "then" error {
   FAIL(@$, "Possibly unterminated 'if' statement");
+  mark_error($2);
   $$ = $2;
 } |
 
@@ -376,6 +379,7 @@ Term "as" Pattern '|' Exp {
 } |
 "try" Exp "catch" error {
   FAIL(@$, "Possibly unterminated 'try' statement");
+  mark_error($2);
   $$ = $2;
 } |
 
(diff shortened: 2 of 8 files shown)


dtolnay:
Looks good so far. I agree on a separate function.

wtlangford:
Definitely a separate function.

nicowilliams:
This works but has a small leak.

nicowilliams:
Try:

    $ ./jq --debug-dump-block '..?|select((.source?.fname?//empty)!="<unspec>")' < /dev/null | ./jq

then

    $ ./jq --debug-dump-block '..?|select((.source?.fname?//empty)!="<unspec>")' < /dev/null | ./jq --debug-dump-block '..?|select((.source?.fname?//empty)!="<unspec>")' < /dev/null

for example.

It would help to sprinkle more `gen_location()`s in `src/parser.y` so we can get more location information in the `--debug-dump-block` output.

nicowilliams:
I'm not sure how useful this will be.

We might also need an option to get errors as JSON or otherwise easy to parse, so that syntax highlighters could use that to get error location correctly.  Also, on error, it'd be nice to get the dumped block representation of the program as parsed up to wherever the error was reached.

nicowilliams:
Anyways, I think this is ready to be reviewed.

@vito-c Can you see if this PR helps your syntax highlighter work?  I hope so!  If not, well, an improved AST is not likely anytime soon...  Oh, and as to the paucity of source location information, see the above comments.

nicowilliams:
See #1032.

nicowilliams:
In any case, this will help us visualize how to detect tail-calls in the `block` representation of jq programs, and move TCO from a bytecode optimizer to the previous stage.

nicowilliams:
Hmmm, no, not quite ready.  The bodies of functions don't appear to be getting dumped.

nicowilliams:
Ah, I'm dumping the blocks after compilation, but compilation doesn't merely mutate to set each instruction's `bytecode_pos`: it also frees the subfunctions (which includes the body).

nicowilliams:
Better!

@dtolnay Please don't merge this.  I need to squash the last two commits.  I left them unsquashed because they seem worth reviewing separately and to not reset whatever review you were already doing.

Anyways, this makes clear how to find tail calls in `block` representation.

src/execute.c
```diff
@@ -1088,13 +1092,21 @@ int jq_compile_args(jq_state *jq, const char* str, jv args) {
...
     nerrors = builtins_bind(jq, &program);
     if (nerrors == 0) {
-      nerrors = block_compile(program, &jq->bc, locations);
+      program = block_drop_unreferenced(program);
+      /*
+       * We could dump_block(program) here, but block_compile() modifies
+       * the program it in important ways, so we leave it to
+       * block_compile() to dump it.
+       */
+      int dump = jv_equal(jv_true(), jv_object_get(jv_copy(opts), jv_string("dump_block")));
```
nicowilliams: It would probably be useful to gather into an object the pre- and post-`block_compile()` block representation of the jq program.  `block_compile()` does make some transformations on the input `block`, and it does do things like set `bytecode_pos`.  It is useful to be able to observe these transformations.

vito-c:
> @vito-c Can you see if this PR helps your syntax highlighter work?

@nicowilliams I haven't had time to check this out but I will this weekend. 

- Does this solve the issue where the column number was not showing up? 
- Will this give the ability to act as a "compile" step of a script as well? (so I can use it for linting/error checking) or is `jq -c -n ` better for that?

nicowilliams:
@vito-c This won't help in the case of errors.  But in the case of successful compilation it's a JSON view of the compilation, which includes line and column numbers (or, rather, byte offsets from the beginning of the file; I should add lineno/colno info too), which might help with syntax highlighting.  It won't be as helpful as it could be though because the compiler doesn't peper enough source location in the `block` intermediate representation, but if this could be helpful to you then we'd go fix that.

vito-c:
I was trying to compile from source and running into some issues. I have bison installed here `/usr/bin/bison` which is version `2.3` and I have a brew version of bison which is here `/usr/local/opt/bison/lib` which is version `3.0` I tried running `./configure --disable-maintainer-mode` but was failing in the make step

nicowilliams:
It does turn out that adding `gen_location()` calls in `parser.y` helps.  For example, this:

    diff --git a/src/parser.y b/src/parser.y
    index a37d217..9d13252 100644
    --- a/src/parser.y
    +++ b/src/parser.y
    @@ -359,7 +359,7 @@ Term "as" Pattern '|' Exp {
     } |

     "if" Exp "then" Exp ElseBody {
    -  $$ = gen_cond($2, $4, $5);
    +  $$ = gen_location(@$, locations, gen_cond($2, $4, $5));
     } |
     "if" Exp "then" error {
       FAIL(@$, "Possibly unterminated 'if' statement");

yields this in the dumped block representation of the first op in an example program:

      {
        "pos": 2,
        "op": {
          "name": "DUP", ...
        }, ...
        "source": {
          "fname": "<top-level>",
          "start": 23,
          "end": 53,
          "snippet": "if . > 0 then .-1|r else . end"
        },
        "imm": { ...
        }
      },

Nice?

We could add `gen_location()` in many more places in `parser.y`.

And! since now the blocks get printed even on compile/link error, this might be useful to show errors.

It might be useful if `struct inst` had a field to indicate whether there was an error when compiling it.  That would be great, because then errors could get highlighted trivially.  The only problem here is that this works for compilation errors, but not for parsing errors.  But `yyerror()` does get the `block` produced so far, so it could dump it, if only we had a way to pass the need to dump it...  ah, but I know how.  OK, so when I'm done we'll even get JSON `block` representation with error indications, even for parse errors.

nicowilliams:
OK, now with compile-time errors flagged.  Parse-time errors next.

nicowilliams:
Alright!  Now we're getting parse errors marked up in the block dumps.  There's an `"error"` key in each instruction in the dump that is set to `true` if that's where an error occurred.

Next up: more `gen_locations()` calls.

nicowilliams:
Feel free to kick the tires on this, let me know how it works for you.

nicowilliams:
Oy, I should `make check` before asking you to kick the tires...  Tomorrow...

nicowilliams:
@vito-c In your case you should be able to put the `/usr/local` bindir for bison in `$PATH` ahead of `/usr/bin` when you run `./configure`.  That should get `./configure` to find and use the bison 3.0 that you have installed.

nicowilliams:
Oy, so, some of the changes to `src/parser.y` to make it preserve information about where errors arose is having knock-on effects that will requires cascading error handling throughout it.  I think it will be worth it.  It will allow syntax highlighters to show where the error occurred, and all of what the parser could not parse as a result.

vito-c:
for future reference if someone is trying to compile by source and they try to install bison via homebrew they should do this:  `export PATH=/usr/local/opt/bison/bin:$PATH` then the homebrew version of bison will be used when compiling jq from source!

nicowilliams:
The error cascading thing in `src/parser.y` could get depressing.  I might need some macros to make it easier, something like a varargs macro that takes any (small) number of blocks and expands into an `if (any of the blocks is marked as an error) { $$ = <the first of those blocks>; } else`, then one would write `IF_BLOCKS_OK(...) { ... }` or so in just about every action.  This will then allow the erroneous block to be the one that's output as the parsed program.  The user of `--debug-dump-block` could then highlight the offending part of the program, and also highlight the correctly-parsed parts, and also highlight the parts not parsed as a result of the error.

nicowilliams:
Now tests pass, but I'm certain I've introduced more assertion tripping cases that are just not tested.  So the `src/parser.y` error cascading needs thought.

Also, I've not yet added `gen_location()` calls.  I think it should suffice to do it for errors.  But it might be interesting to do it for every rule.

src/parser.y
```diff
@@ -536,21 +544,23 @@ ImportWhat:
   jv_free(v);
 } |
 "include" ImportFrom {
-  jv v = block_const($2);
-  $$ = gen_import(jv_string_value(v), NULL, 0);
-  block_free($2);
-  jv_free(v);
+  if (block_is_error($2)) {
```
nicowilliams: @dtolnay, @wtlangford Note that this is a hack to prevent an assertion a few lines below in `block_const()`.  This is what I mean by cascading of error handling.  This becomes necessary because the `ImportFrom` rule below preserves the offending non-constant `block` so that `--debug-dump-block` can produce something useful.  Of course, without `gen_location()` that won't be so useful, but one of the next steps is to make all error cases also call gen_location().

nicowilliams: Any advice?

nicowilliams:
Another thing that might be useful to include is the `debuginfo` from the `bc` field of `struct inst`...

nicowilliams:
@dtolnay  @wtlangford  At this point I think the only thing that will change further is `src/parser.y`.  Feel free to review the rest.

vito-c:
hi @nicowilliams !

PR #1037, opened 2015-12-09
Avoiding conditional directives that break statements

RomeroMalaquias:


nicowilliams:
Cherry-picked as 6151f20.  Thanks!

PR #1045, opened 2015-12-15
Make it easier to use jq with shebangs (fix #1044)

### LINKED ISSUE #1044: Comment continuation needed on first comment following shebang
Historically shebangs allow only one argument in the shebang line.  This means that if the script author doesn't know the path to the script interpreter *and* they need to pass any additional arguments to said interpreter, then they can't write the script.

The typical thing to do is this:

    #!/bin/sh
    # this comment is here just for the continuation so foo interpret doesn't eval this next line \
    exec jq -f "$0" "$@"

But jq does _not_ have comment continuations now, so adding them now would break backwards compatibility.  But I think it'd be safe to scan for the shebang and continued comment on the first line in `main()` (as opposed to in `src/parser.y`).  Then the above would work and there chance of breakage would be astronomically small.

nicowilliams:
Allow a continuation on a comment immediately after a shebang to make
this traditional hack possible:

    #!/bin/sh
    # this next line is ignored by jq \
    exec jq -f "$0" "$@"
    # jq code follows

But continue only on the first line following a shebang, and only if
it's a comment.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -127,6 +127,21 @@ enum {
 };
 static int options = 0;
 
+static const char *skip_shebang(const char *p) {
+  if (strncmp(p, "#!", sizeof("#!") - 1) != 0)
+    return p;
+  const char *n = strchr(p, '\n');
+  if (n == NULL || n[1] != '#')
+    return p;
+  n = strchr(n + 1, '\n');
+  if (n == NULL || n[1] == '#' || n[1] == '\0' || n[-1] != '\\' || n[-2] == '\\')
+    return p;
+  n = strchr(n + 1, '\n');
+  if (n == NULL)
+    return p;
+  return n+1;
+}
+
 static int process(jq_state *jq, jv value, int flags, int dumpopts) {
   int ret = 14; // No valid results && -e -> exit(4)
   jq_start(jq, value, flags);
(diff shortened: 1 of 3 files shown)


dtolnay:
Looks good, cherry-picked as 8f6f28c.

nicowilliams:
Thanks.

svnpenn:
@nicowilliams does this mean we can do something like this?

~~~sh
#!/bin/jq
# jq code
~~~

nicowilliams:
@svnpenn You can do this:

    #!/bin/jq -f
    ...

today, with jq 1.4 and 1.5, and probably even earlier, but you have to know the path to jq for the shebang.  This commit PR was for cases where you don't know the path to jq, so you want the shell to evaluate that and then exec jq.

svnpenn:
@nicowilliams couldnt you just use the standard idiom?

~~~sh
#!/bin/env jq
# stuff
~~~

nicowilliams:
@svnpenn No.  The problem is that only one shebang line argument works portably, and we need one for the `-f` argument.

nicowilliams:
Aha, this hack is used with Tcl:
 - http://wiki.tcl.tk/812
 - http://docs.nscl.msu.edu/daq/bluebook/html/x3237.html

It was hard to find those links.

svnpenn:
@nicowilliams

http://wikipedia.org/wiki/Polyglot_%28computing%29

PR #1058, opened 2015-12-31
add debian/ directory

sigmonsays:


debian/copyright
```diff
@@ -0,0 +1,355 @@
...
+     License is not intended to restrict the license of any rights under
+     applicable law.
+
+Files: docs/public/bootstrap/*
+Copyright: 2012 Twitter, Inc
+License: Apache-2.0
+ On Debian systems the full text of the Apache-2.0 license can be found in the
+  `/usr/share/common-licenses/Apache-2.0' file.
+
+Files: jv_dtoa.c
```
nicowilliams: This is now `src/jv_dtoa.c`.

debian/control
```diff
@@ -0,0 +1,27 @@
...
+Architecture: any
+Multi-Arch: foreign
+Depends: ${shlibs:Depends}, ${misc:Depends}
+Description: lightweight and flexible command-line JSON processor
+ jq is like sed for JSON data – you can use it to slice
+ and filter and map and transform structured data with
+ the same ease that sed, awk, grep and friends let you
+ play with text.
+ .
+ It is written in portable C, and it has zero runtime
```
nicowilliams: Oh, we should probably update this.  With Oniguruma, jq now has 1 run-time dependency.

else:
Actually I think the patches in `debian/patches` should also be part of the repository then.

debian/gbp.conf
```diff
@@ -0,0 +1,3 @@
+[git-import-orig]
+filter = ['docs/public/bootstrap/js/bootstrap.min.js']
```
dtolnay: no longer exists, we now depend on BootstrapCDN

jeroenooms:
Can we please include `libjq.so` and the `jq.h` headers in the package as it is done for e.g. Fedora?

else:
I'd like to do this, but we'd need at least some kind of basic manpage describing the functions exposed by the lib. Last time I checked this did not exist, which is the reason I linked it statically.

else:
Actually I just checked debian-policy and it is not required to have a man page. I am going to package libjq-dev with the next release then.

emanuele6:
Outdated. Also, I don't think this stuff is necessary since `debian` distributes `jq`.

Closing...

nicowilliams:
Well, yes, there's the question of whether we ought to be responsible for all the packaging metadata for all the packaging systems.  At the very least they would have to not collide.

PR #1060, opened 2016-01-05
Mention select() in if-then-else docs

cweiske:


itchyny:
I don't think this one line explanation fits well to this section, and considering this PR is outdated, I'm closing.

PR #1061, opened 2016-01-05
Mention keywords "cast" and "integer" in "tonumber" docs

cweiske:
When searching for a way to cast a string to integer with `jq`, I did not find any results. Thus I'm adding the keywords "cast" and "integer" for future search engine users.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1111,6 +1111,8 @@ sections:
           will convert correctly-formatted strings to their numeric
           equivalent, leave numbers alone, and give an error on all other input.
 
+          It casts the string into an integer or a float, depending on its value.
+
         examples:
           - program: '.[] | tonumber'
             input: '[1, "1"]'


pkoppstein:
@cweiske - Although the intent of this PR as reflected in the title ("Mention keywords "cast" and "integer" in "tonumber" docs) is commendable, the proposed change is unfortunately inaccurate and misleading.  

First, jq follows JSON in not having a formal "integer" type, and indeed currently jq converts all JSON numbers to IEEE floats.  Sometimes these will print as though they were decimal integers, but the algorithm for doing so may well change.  Indeed, there are plans (not yet fully developed as far as I know) to overhaul the treatment of JSON numbers in general.

Second, the concept of "casting" (as understood in languages such as C, C++, Java, etc) is quite foreign to jq  -- it is no accident that the word does not appear even once in the jq manual.  However, a case can certainly be made for having the word in the manual to assist users who want to search for that term.

In summary, I'd suggest you reword the proposed changes.

Meanwhile, thanks for your other PR regarding select(). The addition of a reference to select() in connection with if-then-else-end has long been overdue.

nicowilliams:
I think "convert" is good enough.

PR #1073, opened 2016-01-15
Fix Windows build and crash; fix build in general

nicowilliams:
Not tested.  Submitting a PR to get Travis building this.


nicowilliams:
@dtolnay Note that had to `#include "src/..."` for built files, otherwise the cross-compilation system doesn't find them.  In principle we could do for all jq headers.

PR #1082, opened 2016-01-18
Make jv_sort stable regardless of qsort details.

stedolan:


nicowilliams:
Oh!  I should have looked closely.  I didn't know we had an index to look at.  Ha!  Good one!

Rebased so it's a fast-forward push and then I'll restore the `sort_by` tests to `tests/jq.test`.

nicowilliams:
Thanks @stedolan!

pkoppstein:
> I didn't know we had an index to look at
Adding the index was part of the update.

The FAQ now records this monumental event.

Thanks y'all.

nicowilliams:
@pkoppstein Let me rephrase.  I hadn't understood that we had `struct stort_entry` to which an index could be added.  I simply hadn't stopped to think about it, or even look.

PR #1086, opened 2016-01-18
Travis-CI build for OSX (fix #1083)

### LINKED ISSUE #1083: OS X and Travis-CI
Apparently Travis-CI supports OS X.  We should use this to make sure we don't break the jq on OS X.

nicowilliams:


nicowilliams:
Well, that sucks.  All the builds work except for the coveralls build, which now fails when compiling `src/parser.y`.  It took a lot of effort to figure out how to build with bison 3+ on OS X and Linux both, and now this fails.  It fails because of this:

    $ echo "$TRAVIS_OS_NAME"
    ["linux", "osx"]

Ah, so we *cannot* use `$TRAVIS_OS_NAME` to make decisions, contra the docs.

nicowilliams:
OK, so it all builds, but the coveralls build now fails to find Oniguruma, so it doesn't pass two of the tests.

OK, I'm out of time for this.  Maybe someone else can take a look.

nicowilliams:
Er, not the coveralls build, the coverage build.

nicowilliams:
Oh, so there's an interaction with the `os` and `matrix` stanzas such that the `addons` needed to build with don't get added in the gcov build.

nicowilliams:
Got it: had to set `os` in the `matrix` `include` for the gcov build.

Issue #47, opened 2012-11-17
base64 decoding function


Downchuck:
Add a function to base64 encode and decode a string result. Some serialization packages such as Jackson, a popular Java package, include base64 encoding for some data fields.

schroederc:
I find myself also wanting a base64 decoder.  Are there any updates on this feature?

nicowilliams:
base64 coding is certainly in master.

antonosmond:
+1 for base64 decoding

Overbryd:
As this is showing up on Google a lot, and good documentation on jq is sparse, here is to everybody who lands here:

Example:

```console
echo '{"foo": "Ym9iIGxpa2VzIGFsaWNlCg=="}' | jq '.foo | @base64d'
```

Or even use it when building new objects:

```console
echo '{"foo": "Ym9iIGxpa2VzIGFsaWNlCg=="}' | jq '{encoded: .foo, decoded: .foo | @base64d}'
```

rmetzler:
Thank you @Overbryd! After building jq in a Docker container (fixing the Dockerfile first), this works for me.

ameissnersofi:
@Overbryd , when I run your example of decoding, I get an error.

```sh
$ jq --version
jq-1.5
$ echo '{"foo": "Ym9iIGxpa2VzIGFsaWNlCg=="}' | jq '.foo | @base64d'
jq: error (at <stdin>:1): base64d is not a valid format
```
However, if I run `@base64` to encode things, that works just fine.  Am I doing something wrong?

Overbryd:
@ameissnersofi My best guess is that you have to have a recent version of jq. `@base64d` was only added recently.

ameissnersofi:
@Overbryd oh I'm sorry.  I thought you were actively developing jq.  My bad.  :/

@l8nite - can you verify my issue (assuming you are developing - the commit history seems to indicate so lol)  @stedolan ?  

Thanks!

ameissnersofi:
@l8nite awesome!  I see that 1.6 is in rc1 right now, which is probably why brew didn't pick it up for me.  Thanks so much!

pkoppstein:
@ameissnersofi - Have you tried `brew install --HEAD jq` ?

Downchuck:
Great job all, on the feature and documenting its access.

mterron:
@base64d seems to have issues with binary values, or at least does not match `base64 -d`:
```bash
[...]

PR #1089, opened 2016-01-24
Add @base64d for decoding base64 #47

l8nite:
This adds an `@base64d` to do base64 decoding (issue #47).

I needed this as part of querying [Consul's HTTP API](https://www.consul.io/docs/agent/http/kv.html), which returns k/v values base64 encoded in the result JSON.

Edit: The appveyor failures appear unrelated to my PR (something about maximum file size on the download of Oniguruma).

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1739,6 +1739,11 @@ sections:
 
             The input is converted to base64 as specified by RFC 4648.
 
+          * `@base64d`:
+
+            The inverse of `@base64`, input is decoded as specified by RFC 4648.
+            Note\: If the decoded string is not UTF-8, the results are undefined.
+
           This syntax can be combined with string interpolation in a
           useful way. You can follow a `@foo` token with a string
           literal. The contents of the string literal will *not* be
@@ -1767,6 +1772,14 @@ sections:
           - program: '@sh "echo \(.)"'
             input: "\"O'Hara's Ale\""
             output: ["\"echo 'O'\\\\''Hara'\\\\''s Ale'\""]
+
+          - program: '@base64'
+            input: '"This is a message"'
+            output: ['"VGhpcyBpcyBhIG1lc3NhZ2U="']
+
+          - program: '@base64d'
+            input: '"VGhpcyBpcyBhIG1lc3NhZ2U="'
+            output: ['"This is a message"']
 
       - title: "Dates"
         body: |
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -376,6 +376,24 @@ static jv f_utf8bytelength(jq_state *jq, jv input) {
 
 #define CHARS_ALPHANUM "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
 
+static const unsigned char BASE64_ENCODE_TABLE[64 + 1] = CHARS_ALPHANUM "+/";
+static const unsigned char BASE64_INVALID_ENTRY = 0xFF;
+static const unsigned char BASE64_DECODE_TABLE[255] = {
+  0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+  62, // +
+  0xFF, 0xFF, 0xFF,
+  63, // /
+  52, 53, 54, 55, 56, 57, 58, 59, 60, 61, // 0-9
+  0xFF, 0xFF, 0xFF,
+  99, // =
+  0xFF, 0xFF, 0xFF,
+  0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, // A-Z
+  0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+  26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51,  // a-z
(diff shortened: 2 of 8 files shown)


src/builtin.c
```diff
@@ -548,6 +564,38 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
...
+    input = f_tostring(jq, input);
+    const unsigned char* data = (const unsigned char*)jv_string_value(input);
+    int len = jv_string_length_bytes(jv_copy(input));
+    char result[(3 * len) / 4]; // decoded result buffer
+    int ri = 0;
+    int input_bytes_read=0;
+    uint32_t code = 0;
+    for (int i=0; i<len && data[i] != '='; i++) {
+      code <<= 6;
+      code |= BASE64_DECODE_TABLE[data[i]];
```
nicowilliams: If the input isn't base64, shouldn't we terminate early and return an error (`jv_invalid_with_msg()`)?  Particularly as we're reading the string a byte at a time rather than a Unicode codepoint at a time.

nicowilliams:
This is cool, thanks.  See review comments (so far only one).  Also, maybe add a test with invalid base64 data?  It should produce an errors (use `try ... catch ...` to check that you get an error).

nicowilliams:
The Appveyor build had what I hope is a transient failure due to it not being able to find Oniguruma.  I sure hope it's transient.  Anyways, you can ask Appveyor to rebuild: visit their site, click sign-in, click sign-in with github credentials, go to your projects page and find jq there, then click on the current build, click on rebuild.

src/builtin.c
```diff
@@ -548,6 +564,38 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     }
     jv_free(input);
     return line;
+  } else if (!strcmp(fmt_s, "base64d")) {
+    jv_free(fmt);
+    input = f_tostring(jq, input);
+    const unsigned char* data = (const unsigned char*)jv_string_value(input);
+    int len = jv_string_length_bytes(jv_copy(input));
+    char result[(3 * len) / 4]; // decoded result buffer
```
nicowilliams: VLAs are dangerous.  Suppose the incoming string is 10MB... and suppose the stack grows down...  you could easily start by writing to memory that's not on the stack but in some other memory mapping.  In the best case you'd just segfault. I would much prefer this be malloc()ed.

src/builtin.c
```diff
@@ -548,6 +564,38 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
...
+      }
+    }
+    if (input_bytes_read == 3) {
+      result[ri++] = (code >> 10) & 0xFF;
+      result[ri++] = (code >> 2) & 0xFF;
+    } else if (input_bytes_read == 2) {
+      result[ri++] = (code >> 4) & 0xFF;
+    }
+    jv line = jv_string("");
+    line = jv_string_append_buf(line, result, ri);
```
nicowilliams: Just use jv_string_sized().

Now, what if the decoded string is not UTF-8?  We won't get an error.  Instead we'll get a string with non-UTF-8 sequences replaced with `0xFFFD (REPLACEMENT CHARACTER)`.  I'd rather get an error but I know that will be harder, so we can leave that for some other day.

docs/content/3.manual/manual.yml
```diff
@@ -1739,6 +1739,10 @@ sections:
 
             The input is converted to base64 as specified by RFC 4648.
 
+          * `@base64d`:
+
+            The inverse of `@base64`, input is decoded as specified by RFC 4648.
+
```
nicowilliams: Please explain that if the decoded string is not UTF-8 then jq may or may not raise an error.

nicowilliams:
https://github.com/appveyor/ci/issues/613

nicowilliams:
Last comment for a while: long-term I want to have support for binary strings represented as arrays of small integers (0..255), but internally as plain `unsigned char` arrays, naturally.  So what I'd like to do with base64-encoded binary data then is to decode into this binary data "type" (it wouldn't be a proper type; `type` would report it as `"array"`).  We might need a new text format specifier for decoding base64-encoded data into this "type".

src/builtin.c
l8nite: Yea there are a couple places where we should bail out if something goes amiss, I will make the code more defensive.

l8nite:
@nicowilliams Thanks for the review. I've switched to alloca/memset for the result array, and I added error conditions for invalid base64 data and the case where there are not enough bytes in the input to successfully decode a character. 

I added a new base64 test suite independent of the others so it'd be a little faster to run, let me know if you want me to squash those back into `jq.test` instead.

nicowilliams:
@l8nite `alloca()` has exactly the same problem as VLAs (the two are morally the same thing: change the `sp` by the requested number of bytes and good luck to you).

You could do this (which I've done elsewhere):

    /*
     * Make alloca() safe by tripping any guard page if there is one and we
     * want to allocate enough bytes to blow the stack.
     */
    static void
    ensure_alloca(size_t bytes)
    {
        int strack_grows_down;
        size_t i;
        volatile char *mem = alloca(bytes);

        strack_grows_down = &mem[bytes - 1] < &mem[0];
        for (i = 0; i < bytes; i += 4096) {
            if (strack_grows_down)
                ((char *)mem)[bytes - i] = 0;
            else
                ((char *)mem)[i] = 0;
        }
    }
    
    #define SAFE_ALLOCA(bytes) (ensure_alloca(bytes), alloca(bytes))

which will merely segfault if you request too much memory, assuming there's a guard page anyways, as opposed to possibly corrupting memory.  (And this assumes pages are 4KB, which they might not be).  But still, that's a terrible way to fail.

Just malloc()/free() :)

l8nite:
@nicowilliams Thanks! You're exposing my seriously-underutilized C skills here ;)

nicowilliams:
Yes, you have to `jv_free(input);` before returning.

l8nite:
Turns out `jv_free` was already being called by `type_error` so all good there. I've squashed the commits down into the initial commit. Let me know if there's anything else I can fix.

nicowilliams:
@l8nite For now there's nothing to do.  I can't get AppVeyor to build.  I've tried several things, and I'll try more tomorrow.  Without clean builds, we won't merge.

nicowilliams:
I found the maintainer of the Oniguruma pkgs for MSYS.  I hope he responds.

nicowilliams:
If we don't hear from him I might just adapt https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-oniguruma to build in appveyor.yml.

l8nite:
I've tried a couple builds this week on my own appveyor account without success. Any word from the Onigurama maintainer?

nicowilliams:
No word yet.  I'll have to find time for making appveyor.yml build
oniguruma :(

shanielh:
Hi mate, I was trying to do the same, Instead I was using `base64` command (I'm using linux), I've created import/export script for the KV store of consul : 

```
map("curl \"http://NEW_ADDR:8500/kv/\(.Key)\" -XPUT -d \"$(echo \"\(.Value)\" | base64 --decode)\"")[]
```

Good luck with the pull-request.

royalaid:
Just curious as to the status of this

nicowilliams:
Sorry, i've been traveling.

What I'll do is disable the appveyor build in `appveyor.yml` and pull this.  Probably tonight.

dansteen:
+1 for this!  It will make working with consul so much easier.

so0k:
Any update on this? Please advise any help needed to get this merged?

wtlangford:
I'll take a look at this tonight, and if it's good, I'll rebase and merge
it.

wtlangford:
Rebased and merged manually with e6fa039

nicowilliams:
Thanks @wtlangford for reviewing and merging!  Thanks @l8nite for the contribution!

PR #1140, opened 2016-05-02
fix --exit-code with more than one object in input

### LINKED ISSUE #1139: jq -e returns 4 for non-empty output with select()
```
$ cat test.json
{"key": 1}
{"key": 2}
{"key": 3}

$ for i in 1 2 3 4
do
     echo "==== i=$i ===="
     jq --exit-status -c "select(.key == $i)" test.json
     echo ret=$?
done
==== i=1 ====
{"key":1}
ret=4
==== i=2 ====
{"key":2}
ret=4
==== i=3 ====
{"key":3}
ret=0
==== i=4 ====
ret=4
```

The man page says "Sets the exit status of jq to [...] 4 if no valid result was ever produced", but it seems the command returns 4 whenever last object is filtered out. (it should return '4' only for i=4 case above)

```
$ jq --exit-status -c 'select(.key == 1) | (.key!=1) ' test.json; echo ret=$?
false
ret=4
```

Similarly, I think this should return '1' instead of 4, because it's last output (which is not last one in the input, though) is 'false'.

ryo1kato:
Return the exit code 1 or 4 based on last output, not last input.
Fixes #1139

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -129,6 +129,17 @@ enum {
 };
 static int options = 0;
 
+enum {
+    JQ_OK              =  0,
+    JQ_OK_NULL_KIND    = -1, /* exit 0 if --exit-status is not set*/
+    JQ_ERROR_SYSTEM    =  2,
+    JQ_ERROR_COMPILE   =  3,
+    JQ_OK_NO_OUTPUT    = -4, /* exit 0 if --exit-status is not set*/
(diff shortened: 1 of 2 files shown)


ryo1kato:
Fix #1142 as well.

ryo1kato:
A new PR https://github.com/stedolan/jq/pull/1697 based on the latest master

PR #1133, opened 2016-04-14
Update download instructions for Arch

xiongchiamiov:
jq has been in the community repo for [just over a year now](https://projects.archlinux.org/svntogit/community.git/commit/trunk?h=packages/jq&id=8d5e06065bc92bf56ea651970c9bda07fb545554).

PR #1172, opened 2016-06-23
Update ArchLinux dowload url

arcan1s:
Actually jq package has been moved into official repositories some time ago. I believe users don't need to build the application from git repository directly and may just install compiled application from repositories


dequis:
This is a dupe of #1133 which was just merged, this can be closed.

Issue #1163, opened 2016-06-13
`paths(scalars)` bug?


alex-ozdemir:
Consider this JSON:

``` json
{
  "size": 2,
  "unsaf": false
}
```

The instructions `paths(scalars)` yield just

``` json
[
  [
    "size"
  ]
]
```

Whereas one might expect it should yield

``` json
[
  [
    "size"
  ],
  [
    "unsaf"
  ]
]
```

A bit of poking around makes me think the problem is when `false` is a value in a object.

alex-ozdemir:
Oh, hmm - paths is just getting confused by the output of scalars. Thats unfortunate, but perhaps not a bug.
It does make the `leaf_paths` alias somewhat unfortunate though.

slapresta:
This is almost surely a bug.

alex-ozdemir:
Okay, I submitted a PR (#1178) which fixes the issue with `false` leaves.
The same issue exists with `null` leaves, but it will be harder to fix (`paths` is what is messing us up there).

pkoppstein:
While we await the incorporation of a fix into "master", here is a stand-alone definition of paths/1 based on @alex-ozdemir's contribution:
```
def paths:
  def conditional_recurse(f):  def r: ., (select(.!=null) | f | r); r;
  path(conditional_recurse(.[]?)) | select(length > 0);
```

This definition is written so that anyone who wants to define conditional_recurse/1 as a top-level filter can easily do so. For reference:

* the builtin recurse(f) is defined in terms of:  def r: (f | select | r);
* conditional_recurse(f) is defined in terms of: def r: (select | f | r);

saxonww:
Came to report this same issue. I'm surprised it's been sitting for 18 months - is there a reason #1178 can't be accepted?

Another workaround may be e.g. `path(..|select(type!="array" and type!="object"))`

alex-ozdemir:
What a blast from the past.

I believe we're just waiting on someone with write access to the repo...

saxonww:
Hello @nicowilliams, I see you are the most frequent committer, can you recommend the best way to get a PR noticed?

nicowilliams:
I'll recite this next week.

ajasmin:
Until this is fixed a workaround is to check if `scalars` returns no values
```
paths([scalars] != [])

nicowilliams:
~This is quite mysterious:~

pkoppstein:
@nicowilliams wrote:
[...]

PR #1178, opened 2016-07-06
Added `is{scalar,value,..}`. Fixes leaf_paths bug.

alex-ozdemir:
Now paths to `false` leaves are correctly returned.
Added the new test functions to the docs.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,42 +8,43 @@ environment:
 
 install:
   # update mysy2
-  - C:\msys64\usr\bin\bash -lc "pacman --needed --noconfirm -Sy pacman-mirrors"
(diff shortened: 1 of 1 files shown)


alex-ozdemir:
This is a partial solution to #1163.

slapresta:
I proposed something very similar a long time ago. I like it! I find myself using `type == "array"` more often than `arrays`.

alex-ozdemir:
Okay, paths to `null` handled by fixing `paths`. The fix uses an alternative behavior for recurse. I chose not to include this new behavior in the stdlib, as having two floating around would probably just confuse people.

This PR now completely closes #1163.

nicowilliams:
Thanks!  I'll review.

alex-ozdemir:
Paging @nicowilliams 😄

PR #1185, opened 2016-07-16
Minor fixups and updates after reviewing `jv.c` and `main.c`

cipriancraciun:
I have started reviewing `jv.c` and `main.c` and made a (very) few fixups, optimizations, and assertions.

I have made a commit with each individual change (to allow cherry-picking) and prefixed it with the nature (change, fixup, or optimization), and where needed explained in the commit message the reasoning.

Thanks,
Ciprian.

P.S.:  I intend to continue my review and contribute back my findings.  Thus if the manner in which I submit the patches is not the correct one, please point me in the right direction.

### DIFF
--- a/src/jv.c
+++ b/src/jv.c
@@ -24,10 +24,12 @@ typedef struct jv_refcnt {
 static const jv_refcnt JV_REFCNT_INIT = {1};
 
 static void jvp_refcnt_inc(jv_refcnt* c) {
+  assert(c->count > 0);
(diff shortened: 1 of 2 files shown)


cipriancraciun:
Regarding the two Travis-CI failed checks, they seem to break on `autoreconf -i`, thus extremely unlikely to be from the patched code itself.

(The other three Travis-CI runs succeeded.)

PR #1201, opened 2016-08-09
Add snap packaging support

ltrager:
This adds [Snap packaging](http://snapcraft.io/) to jq. The snap can be built by running
`make snap`

The jq snap can then be published to the [Snapcraft store](https://uappexplorer.com/apps) by following [the Snap publishing document](http://snapcraft.io/docs/build-snaps/publish)

### DIFF
--- a/snapcraft.yaml.template
+++ b/snapcraft.yaml.template
@@ -1,0 +1,22 @@
+name: jq
+version: @JQ_VERSION@
+summary: jq is a lightweight and flexible command-line JSON processor.
+description: jq is like sed for JSON data - you can use it to slice and filter and map and transform structured data with the same ease that sed, awk, grep and friends let you play with text.
+confinement: strict
+
+apps:
+  jq:
+    command: bin/jq
+
+parts:
+  jq:
+    plugin: autotools
+    source: .
+    filesets:
+      binaries:
+       - bin/jq
+       - lib/libjq.so*
+    stage:
+      - $binaries
+    snap:
+      - $binaries
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,8 @@ tests/*.trs
 cscope.in.out
 cscope.out
 cscope.po.out
+
+snapcraft.yaml
+jq_*.snap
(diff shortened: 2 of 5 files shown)


snapcraft.yaml.template
```diff
@@ -0,0 +1,22 @@
...
+
+apps:
+  jq:
+    command: bin/jq
+
+parts:
+  jq:
+    plugin: autotools
+    source: .
+    filesets:
```
celso-providelo: I think it should be adjusted to match the suggested building setup.
{{{
  configflags:
    - --disable-maintainer-mode
}}}

snapcraft.yaml.template
```diff
@@ -0,0 +1,22 @@
...
+confinement: strict
+
+apps:
+  jq:
+    command: bin/jq
+
+parts:
+  jq:
+    plugin: autotools
+    source: .
```
celso-providelo: I think it should be adjusted to match the suggested building setup.
{{{
configflags:
- --disable-maintainer-mode
}}}

snapcraft.yaml.template
```diff
@@ -0,0 +1,22 @@
+name: jq
+version: @JQ_VERSION@
+summary: jq is a lightweight and flexible command-line JSON processor.
+description: jq is like sed for JSON data - you can use it to slice and filter and map and transform structured data with the same ease that sed, awk, grep and friends let you play with text.
```
celso-providelo: The long line can be avoided with YAML syntax:

    description: |
      jq is like sed for JSON data - you can use it to slice and filter and map
      and transform structured data with the same ease that sed, awk, grep and
      friends let you play with text.

celso-providelo:
@ltrager, I've completely missed your PR and proposed another, #1210, apologies.

I've suggested some changes here for your consideration so it supersedes mine.

Thanks for your work on this.
[]

rgrannell1:
Has there been any progress on this issue? I'm hoping to migrate away from .debs where possible

PR #1210, opened 2016-08-13
Tentative snap building.

cprov:
This is a proposal to include snap building support upstream.

It's temporarily pushing to the `jq-cprov` snap (installed with `snap install jq-cprov`), but once agreed upstream, we can easily setup the `jq` snap along with the most suitable auto-building infrastructure (Travis or Launchpad).


nicowilliams:
Hi.  What is "snap building"?

dequis:
Some fancy new packaging system that uses containers or something http://snapcraft.io/

This drops OSX building in travis, reindents travis.yml for no reason, and includes some random encrypted file which won't work if merged because travis encryption keys are tied to specific repos. If this were my project I'd say "no, thanks"

dequis:
Actually there's #1201 which looks significantly better, doesn't mess with travis or attempt to do uploads by itself, just adds build targets. And it looks like @cprov was interested in letting that other PR supersede this one.

nicowilliams:
OK, thanks.

PR #1214, opened 2016-08-19
Introduce limits on max parsing and printing depth

wmark:
Avoids _jq_ getting killed for running out-of stack.

Fixes: CVE-2016-4074 (and another)

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -10,6 +10,10 @@
 
 typedef const char* presult;
 
+#ifndef MAX_PARSING_DEPTH
+#define MAX_PARSING_DEPTH (256)
+#endif
+
 #define TRY(x) do {presult msg__ = (x); if (msg__) return msg__; } while(0)
 #ifdef __GNUC__
 #define pfunc __attribute__((warn_unused_result)) presult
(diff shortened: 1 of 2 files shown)


smithfarm:
Could the commits be squashed and a line like `Fixes: CVE-2016-4074` be added to the commit message?

ismail:
@stedolan Can you please comment on this patch?

wmark:
These and other patches are part of my Docker image, and can be installed with the package for Debian/Ubuntu from here:

https://hub.docker.com/r/blitznote/debootstrap-amd64/
`rkt image fetch blitznote.com/ubuntu:16.04`

https://s.blitznote.com/debs/ubuntu/amd64/all/jq_1.5+dfsg-17bn_amd64.deb

smithfarm:
@wmark It would be better for distros like ours (SUSE) if the patch were merged upstream.

smithfarm:
@wmark Can you fix the travis CI test failure?

wmark:
It did fail on *Darwin*, not *Linux*.

smithfarm:
@wmark OK, thanks for the clarification.

nicowilliams:
Maybe we should just not use a recursive print algorithm, instead keeping a logical stack rather than an implicit stack.  We don't, after all, have size limits on arrays, strings, objects...  Why should we have a limit on depth?  If you're not able to produce such a patch and this one is important to you, I could merge it and switch to a non-recursive implementation later.

wmark:
@nicowilliams Yes, to your offer. Please merge this PR for the time being. No, to if I want to rewrite parts of **jq**.

As to limits on depth: Any recursive functions needs either to be bound wrt. depth, or a proof that its callstack will always be finite, or that it always is transformed to an iterative equivalent.

nicowilliams:
Rebased and merged.  Thanks!

I'll eventually undo this and replace it with a non-recursive, explicit stack implementation.

wmark:
Thanks for looking into this!

tsibley:
> I'll eventually undo this and replace it with a non-recursive, explicit stack implementation.

It'd be nice to have the unlimited depth back. We sometimes use jq for inspecting and modifying deep trees of data and run into the 256 default limit fairly easily. In the meantime, I recently realized that we could explicitly use jq 1.5 instead of 1.6 when we need to avoid it.

Alternatively, if a limit is to remain, it'd be helpful if its value could be given by a command-line option or env var. e.g. CPython's got a default recursion limit, but you can change it at runtime.

PR #1217, opened 2016-08-23
Add summary_stats and variant builtins

erikbrinkman:
I worry this is getting in the realm of builtin bloat, but this adds a c function that computes mean, variance, min, and max in one pass. In addition to being more stable than doing it like `(add / length) as $mean | map((. - $mean) * (. - $mean)) | add / length` it is much faster than anything I could do in native jq.

The contributor guides indicated that keeping objects as jv is preferred. Here I explicitly do the arithmetic on doubles, but since the function mandates the inputs be numbers anyway, this seems like the correct way to go.

I didn't see a builtin way to do coverage, but every branch should be reasonably covered by the tests.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1231,6 +1231,28 @@ sections:
             input: '[{"foo":1, "bar":14}, {"foo":2, "bar":3}]'
             output: ['{"foo":2, "bar":3}']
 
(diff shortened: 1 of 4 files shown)


erikbrinkman:
Resubmitting to trigger a rebuild, hoping that the mac builds have been fixed.

Issue #1227, opened 2016-09-04
Make it possible to define new @foo formats in jq, not just C


dbohdan:
This is an offshoot of issue #685.

Is anyone working on this? If not, would you be interesting in a PR?

dossy:
Personally, I'd really like to see `@sql` implemented, either in C or jq.

Issue #685, opened 2015-01-30
A new output format: Tcl dictionaries 


dbohdan:
I've implemented a function that produces Tcl dictionary output from jq (see [man n tcl](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/mann/Tcl.ntcl.html) for a overview of the format). You will find it below.

If possible, I'd like to have it ship as part of jq (e.g., `@tcl`). I've looked and all the existing export formats (`@csv`, `@tsv`, etc.) seem to be implemented in C rather than jq itself, so before I start working on a pull request I would like to ask: would be acceptable if I implemented `@tcl` as a call from `f_format` to a function stored in `jq_builtins` or should avoid that and translate it to plain C?

```
def totcl:
    if type == "array" then
        # Convert array to object with keys 0, 1, 2... and process
        # it as object.
        [range(0;length) as $i
            | {key: $i | tostring, value: .[$i]}]
        | from_entries
        | totcl
    elif type == "object" then
        .
        | to_entries
        | map("{\(.key)} {\(.value | totcl)}")
        | join(" ")
    else
        tostring
        | split("{") | join ("\\{")
        | split("}") | join ("\\}")
    end;
. | totcl
```

nicowilliams:
Hey there, this is cool!  Yes, perhaps `format` should be jq-coded, calling out to the C-coded form only for the C-coded formats.  This should be easy to do, I think, and would be something like:

    def format(fmt): if fmt == "@tcl" then ... else _format(fmt) end;

nicowilliams:
We'll have to think about ways of making this something that can be delivered by modules.

dbohdan:
@nicowilliams The ability to define new formats this way would be great!

I've updated the code to use `gsub` rather than `split`/`join`, which doesn't work for strings like "{something}".

PR #1228, opened 2016-09-04
Define format in jq code

dbohdan:
The proposed change makes it easier to extend `format` with new formats (#1227). The reasoning is that it is often more straightforward to implement a serialization algorithm recursively in a high-level functional language than in C.

I've implemented serialization to Tcl dictionaries (#685) in pure jq code as an example. This version differs slightly from the one in #685 to work around forward slash (`\`) escaping issues in `src/builtin.jq`.

<s>Perhaps the function `totcl` should be renamed `_totcl`.</s> Edit: Done.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1295,7 +1295,7 @@ static const struct cfunction function_list[] = {
   {(cfunction_ptr)f_min_by_impl, "_min_by_impl", 2},
   {(cfunction_ptr)f_max_by_impl, "_max_by_impl", 2},
   {(cfunction_ptr)f_error, "error", 2},
-  {(cfunction_ptr)f_format, "format", 2},
+  {(cfunction_ptr)f_format, "_format", 2},
   {(cfunction_ptr)f_env, "env", 1},
   {(cfunction_ptr)f_get_search_list, "get_search_list", 1},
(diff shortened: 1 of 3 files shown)


nicowilliams:
Hi.  This is a wonderful contribution.  I may leave `totcl` that way instead of as `_totcl`; not sure.  I'll try to get to merging this this coming weekend.  Thanks!

dbohdan:
Great!

dbohdan:
I've realized there is a corner case that can cause `@tcl` to produce invalid Tcl data. I'll make a test for it and produce a patch shortly.

dbohdan:
Done. Now `@tcl` quotes data the way Tcl itself does. I've added tests for some corner cases that the earlier implementation failed to process correctly. I didn't rename `_totcl` back to `totcl` to leave that up to you.

I forgot to say that I appreciate you complimenting this change. I am a fan of the purely functional language that jq has grown.

dbohdan:
Ping. Any updates on this?

colindean:
I'd love to see this in jq. I could see this formatter plugin being used to output YAML, hocon, dhall, and some others that are effectively supersets of JSON.

Issue #369, opened 2014-05-22
Error with reading large integer?


bsbkeven:
Hello,

I am running jq (version 1.3), and find that it seems to be reading large integers (>= 17 digits) incorrectly. For example, if I run 
    `jq '.id'`
on the following input:

```
    {"id" : 125276004817190914}
    {"id" : 12527600481719091}
    {"id" : 1252760048171909}
    {"id" : 12527.6004817190914}
```

the output is:

```
    125276004817190910
    12527600481719092
    1252760048171909
    12527.600481719091
```

Results on the first and second line (18 and 17 digits) are incorrect, but the third line (16-digits integer) and the fourth (floating number) are correct.

Seems like this hasn't been reported before. So I just want to bring it up and wonder if anyone has had this issue as well? Thanks!

jrdriscoll:
I don't have a lot to add to this discussion other than to note that I was shocked that `echo 11111111111111111 | jq '.'` produced something other than its input.  That being said, I never expected infinite precision arithmetic.

dequis:
Submitted #1246, which solves this issue in a conservative way - no bigint libraries, only for ints up to 64 bits, and only if no operations are done over them, much like awk's behavior in previous comments:

    $ echo 111111111111111111 | jq -c '[., 1*.]'
    [111111111111111111,111111111111111100]

sp-james-mcmurray:
I had this issue today with jq-1.5.

Where for large integers, it is changing the last values to 0s, even converting to strings doesn't help (as it reads it as integer first):

> "3265374331746778600"
"3349146353896582000"
"3658445187091539500"
"381327942920288"
"4540495826739245000"
"4609284046671461000"

Should be:

> 3265374331746778747
3349146353896582298
3658445187091539618
4540495826739245237
4609284046671461009

nicowilliams:
@sp-james-mcmurray Internally jq uses IEEE754 doubles for number representation.  Any integers whose absolute values are larger than 2^52 - 1 will not be faithfully represented.

mitar:
I just use Python and its `json` module. Now that Python maintains the order in its dicts, it is easy to modify JSON and get output match the input, expect for the changes you want.

nicowilliams:
@mitar please don't leave this comment on every issue that deals with jq and IEEE754.  Thanks.

umairrafiq:
One workaround for this issue is to double quote integers as string before sending to jq. 
An easy way to do that in bash using perl :
[...]

PR #1246, opened 2016-10-03
jv: Add some support for 64 bit ints in a very conservative way

dequis:
This adds an extra int64_t field to jv, which is only written when
parsing number literals (and only if they are integers within range),
and only read when printing.

Any operations done over those numbers will downgrade them to a double
with 53 bit precision. This is intentional for the scope of this patch:

```
$ echo 111111111111111111 | jq -c '[., 1*.]'
[111111111111111111,111111111111111100]
```

For ints between 53 and 64 bits, this matches the behavior of awk,
as suggested in the bug tracker by tischwa:

```
$ echo 111111111111111111 | awk '{print $1, 1*$1}'
111111111111111111 111111111111111104
```

---

Fixes issue #369 (at least partially)

I tried to add tests but the stuff in jq.test seems to be parsed by jq itself, so they end up being comparisons between the double values, not the int64 ones.

This is an itch i've really needed to scratch for a long time. Using jq as a pretty printer for json, I almost started getting used to big numbers getting mangled. Not anymore!

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -27,7 +27,10 @@ typedef struct {
   int size;
   union {
     struct jv_refcnt* ptr;
-    double number;
+    struct {
+      double dbl;
+      int64_t int64;
+    } number;
   } u;
 } jv;
 
@@ -60,8 +63,10 @@ jv jv_true(void);
 jv jv_false(void);
 jv jv_bool(int);
 
+jv jv_number_full(double, int64_t);
 jv jv_number(double);
 double jv_number_value(jv);
+int64_t jv_number_value_int64(jv);
 int jv_is_integer(jv);
 
 jv jv_array(void);
--- a/src/jv.c
+++ b/src/jv.c
@@ -137,14 +137,27 @@ static void jvp_invalid_free(jv x) {
  * Numbers
  */
 
-jv jv_number(double x) {
-  jv j = {JV_KIND_NUMBER, 0, 0, 0, {.number = x}};
+jv jv_number_full(double x, int64_t y) {
+  jv j = {JV_KIND_NUMBER, 0, 0, 0, {.number.dbl = x, .number.int64 = y}};
   return j;
+}
+
+jv jv_number(double x) {
+  int64_t y = 0;
+  if(x == x && x < INT_MAX && x > INT_MIN && x == (int)x) {
+    y = x;
+  }
+  return jv_number_full(x, y);
 }
 
 double jv_number_value(jv j) {
   assert(jv_get_kind(j) == JV_KIND_NUMBER);
-  return j.u.number;
+  return j.u.number.dbl;
+}
+
+int64_t jv_number_value_int64(jv j) {
+  assert(jv_get_kind(j) == JV_KIND_NUMBER);
+  return j.u.number.int64;
 }
 
 int jv_is_integer(jv j){
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <assert.h>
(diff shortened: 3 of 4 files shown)


src/jv.h
```diff
@@ -27,7 +27,10 @@ typedef struct {
   int size;
   union {
     struct jv_refcnt* ptr;
-    double number;
+    struct {
+      double dbl;
+      int64_t int64;
```
nicowilliams: This is not desirable, unfortunately, as it increases the `sizeof(jv)`.

What I would recommend instead is a new `jv` kind flag that represents a signed 64-bit integer.  Then add `jv_int64()` (creates a `jv` from an `int64_t`), `jv_int64_value()` (returns an `int64_t` given a numeric `jv`).  For backwards compatibility I'd have `jv_get_kind()` return `JV_KIND_NUMBER`, and then add a `jv_get_number_kind()` that reports the numeric kind.  `jv_number_value()` and `jv_int64_value()` should both accept numbers of any kind, and do the appropriate conversion.

This would allow us to also add `uint64_t` as well.

All arithmetic and math functions would continue to operate on `double`s, though perhaps the arithmetic operations could eventually be made to operate on integers (falling back on `double`s in the event of overflow/underflow).

nicowilliams:
Hi.  Thanks for this contribution.  I left a comment on `jv.h`.  Let us know what you think.

dequis:
omg you're alive

nicowilliams:
heheh, yeah, I'm alive.  Sorry for the absence :(  I've been heads down in other things.

nicowilliams:
I'm working on an alternative version of this where a numeric `jv` can only be a `double`, and `int64_t`, or a `uint64_t`.  The parser will use `strtoumax()` or similar when a numeric value has no exponent and no decimal, and will produce an a 64-bit integer, signed or unsigned as necessary if the integer is small enough to fit in 64 bits.  The dumper will, of course, trivially dump 64-bit ints.  There will be new `jv` functions to go with all of this.

My main fear is that this will slow things down.  I may add macros for disabling this feature.

dequis:
Why is it not acceptable to increase sizeof(jv) anyway? Is there some sort of API/ABI guarantee for other projects?

nicowilliams:
There's no ABI constraint on the size of a `jv`.  But `jv`s are rather compact -- as compact as they can be, really, and this is useful because we pass them on the C and jq stacks a lot.  There are some issues with `jv`s though, mostly the size of the `size` and `offset` fields, which are way too small and have bad overflow semantics, but even for fixing those I'd be reluctant to change the size of a `jv`.

nicowilliams:
Another thing is that having multiple number representations at once in any one `jv` seems like asking for trouble?  Better just have one.

fadado:
> Another thing is that having multiple number representations at once in any one jv seems like asking for trouble? Better just have one.

I like a lot bitwise operations on integers, but I liked also XSLT  in the past, and XSLT has only one kind of "numbers": let jq do the same?

Speaking of XSLT, how about implementing [translate](http://zvon.org/xxl/XSLTreference/Output/function_translate.html)?

JJOR

nicowilliams:
BTW, your submission is awesome.  I'm running with it and modifying it to make the int value a part of the `u` union in the jv, but please don't feel bad about that.

nicowilliams:
@fadado jq will only have one kind of number.  The idea here is that when the input was an integer that fits in an `int64_t` then that should be the internal representation _and_ when printed it should be printed the way a `printf` function would do it (hmm, though we need to be careful to not have any internationalization, e.g., thousands separators!), which is to say: with no exponent, no decimal.

nicowilliams:
@dequis Checkout #1327.  What do you think?

PR #1256, opened 2016-10-20
Add summary_stats and variant builtins

erikbrinkman:
I worry this is getting in the realm of builtin bloat, but this adds a c function that computes mean, variance, min, and max in one pass. In addition to being more stable than doing it like `(add / length) as $mean | map((. - $mean) * (. - $mean)) | add / length` it is much faster than anything I could do in native jq.

The contributor guides indicated that keeping objects as jv is preferred. Here I explicitly do the arithmetic on doubles, but since the function mandates the inputs be numbers anyway, this seems like the correct way to go.

I didn't see a builtin way to do coverage, but every branch should be reasonably covered by the tests.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1231,6 +1231,28 @@ sections:
             input: '[{"foo":1, "bar":14}, {"foo":2, "bar":3}]'
             output: ['{"foo":2, "bar":3}']
 
+      - title: "`summary_stats`, `mean`, `variance`, `sample_variance`"
+        body: |
+
+          Compute summary statistics on an array of numbers.
+
+          `mean`, `variance`, and `sample_variance` compute mean, population
+          variance, and sample variance respectively. The output of
+          `summary_stats` is an object with five fields, "count", "mean",
+          "min", "max", and "sumsq". "sumsq" is the sum of squared differences
+          from the mean, everything else has the obvious definition.  If the
+          input list is empty, every field but count is `nan`. Computing the
+          mean through these filters is slightly faster and more stable than
+          `add / length`.
+
+        examples:
+          - program: 'summary_stats'
+            input: '[5,4,2,7]'
+            output: ['{"count":4,"mean":4.5,"sumsq":13,"min":2,"max":7}']
+          - program: 'sample_variance'
+            input: '[5,4,2,7,9]'
+            output: ['7.3']
+
       - title: "`unique`, `unique_by(path_exp)`"
         body: |
 
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -857,6 +857,43 @@ static jv f_max_by_impl(jq_state *jq, jv x, jv y) {
   return minmax_by(x, y, 0);
 }
 
+static jv f_summary_stats(jq_state *jq, jv values) {
+  // Takes an array of numbers and computes count, mean, min, max, and sum of
+  // squared mean differences
+  if (jv_get_kind(values) != JV_KIND_ARRAY)
+    return type_error(values, "cannot be iterated over");
+  int count = 0;
+  double mean = 0;
+  double sumsq = 0;
+  double min = INFINITY;
+  double max = -INFINITY;
+
+  for (int i=0; i<jv_array_length(jv_copy(values)); i++) {
+    jv item = jv_array_get(jv_copy(values), i);
+    jv_kind k = jv_get_kind(item);
+    if (k != JV_KIND_NUMBER) {
+      jv_free(values);
+      return type_error(item, "number required");
+    }
+    double val = jv_number_value(item);
+    double delta = val - mean;
+    jv_free(item);
+    mean += delta / ++count;
+    sumsq += delta * (val - mean);
+    max = fmax(max, val);
+    min = fmin(min, val);
+  }
+  jv_free(values);
+
+  jv ret = jv_object();
+  jv_object_set(ret, jv_string("count"), jv_number(count));
+  jv_object_set(ret, jv_string("mean"), jv_number(count ? mean: NAN));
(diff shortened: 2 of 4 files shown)


nicowilliams:
Thanks.  This is very nice.  I'll review and merge some time soon.

Question: why not jq-code this?

erikbrinkman:
I have done similar things in jq using a reduce, and what I remember is it being significantly slower. Since I'm going to go over the random pr and change the source, I'll put together a simple benchmark with an alternative jq only implementation and verify that that's the case. I generally agree that writing functions in jq is nicer than their alternative in c, it also might not be an important enough difference in time that a jq solution is still appropriate.

nicowilliams:
Fair enough.  However, we should have /1 versions of these that take operate on a stream, rather than an array.

erikbrinkman:
In my very quick profiling, I used `for _ in {1..1000000}; do echo "$RANDOM"; done` to generate a sequence of a million random numbers and piped this into the following three commands:

- `./jq -s 'summary_stats'`: 1.83s (the original c array implementation)
- `./jq -s 'summary_stats(.[])'`: 11.35s (the jq stream / reduce implementation)
- `./jq -n 'summary_stats(inputs)'`: 10.66s (above but using inputs)

My guess is that all of the jq object creation is costly, although maybe it could be optimized. Is there a way to operate on a stream from the c api? I couldn't find anything in the source that seems to allow or do this. Assuming there's not I'll just push what I have with a caveat that the stream implementation is much slower for large inputs.

erikbrinkman:
This supports streams now, but the appveyor build seems to be broken.

nicowilliams:
It's always something with the Windows/Appveyor build.  It's really a problem.  It's largely why I stopped working on jq for so long.

erikbrinkman:
That's reasonable, I remember this being an issue the first time I submitted this PR. Is there anything I can do other than wait and push an amended commit hoping that the build is fixed?

nicowilliams:
I'll push this regardless this weekend.  I'll see if I can do anything to get the appveyor build unstuck.  Another option might be to see if there's any way to cross-compile on Travis and screw Windows, though that would require cross-compiling Oniguruma too, which would mean having to include Oniguruma in-tree, which sucks.

erikbrinkman:
I should mention, that this passed the appveyor build a while ago, but a different PR was failing and so I did a blank amend to see if the problem was appveyor breaking or an issue with that PR. Unsurprisingly, it was appveyor.

pkoppstein:
Sorry for making the following two requests so late but I hope you will consider them:

1. "sumsq" here is potentially ambiguous between "sum of squares" and "sum of squared deviations". Since it is here intended to refer to the latter, I would request that either the name be changed (e.g. to "ssd") or that the (unbiased) sample variance be provided instead (e.g. under the name "sample_variance").

2. I'd also like to request summary_stats/2 for computing bivariate statistics (mainly to make it easy and efficient to compute the sample correlation-coefficient).  This could of course be a later addition, but it would probably be easier to take advantage of the existing momentum.

Thank you!

erikbrinkman:
@pkoppstein 

1. This seems totally reasonable. It is documented, but that's a reasonable request as `sumsq` is fairly ambiguous. I'm just going to change it to `ssd`. I didn't have it compute sample variance because it's a derivative. There is an alias that just computes it directly.

2. I'm not opposed to implementing a bivariate statistic, I think it's probably worth thinking about what the appropriate use case is before implementing it. Some thoughts include: are bivariate summary statistics an important return, or given that you already have them as lists, would it be more appropriate to just compute correlation coefficient since everything else can be achieved by summary_stats/1? I could also see a variant where the input is an array of arrays and does all of the pairwise comparisons. I also feel like this is already approaching builtin bloat, and so I'm hesitant to add too much functionality along the statistical analysis lines as there are better languages if that's your goal. Potentially @nicowilliams has thoughts?

pkoppstein:
@erikbrinkman - Re (1) - ssd is of course fine by me. Re (2) - The main goals in my view would be to make it very easy and efficient to compute correlation coefficients, while keeping the implementation very light-weight. I envision a very minor tweak to the existing underlying C code.

erikbrinkman:
@pkoppstein 

Do you just imagine the following with a few more checks?

    static jv f_correlation(jq_state *jq, jv xs, ys) {
      int count = 0;
      double meanx = 0;
      double ssdx = 0;
      double meany = 0;
      double ssdy = 0;
      double ssdxy = 0;

      for (int i=0; i<jv_array_length(jv_copy(xs)); i++) {
        jv xi = jv_array_get(jv_copy(xs), i);
        jv yi = jv_array_get(jv_copy(ys), i);
        double x = jv_number_value(xi);
        double y = jv_number_value(yi);

        count++;
        double dx = x - meanx;
        double dy = y - meany;
        meanx += dx / count;
        ssdx += dx * (x - meanx);
        ssdxy += dy * (x - meanx);
        meany += dy / count;
        ssdy += dy * (y - meany);

        jv_free(xi);
        jv_free(yi);
      }
      jv_free(xs);
      jv_free(ys)
  
      return jv_number(ssdxy / sqrt(ssdx * ssdy));
    }

pkoppstein:
@erikbrinkman - Presumably the univariate statistics (meanx, meany, etc) will come from two calls to  f_summary_stats?  

The big question, I think, is what exactly `summary_stats(x;y)` will return.  Presumably the main options are:
* a flat object?
* an object with an object each for the univariate statistics of x and y?
* an array of the univariate statistic objects, together with the correlation coefficient?

One appeal of the third option is that it circumvents naming issues, but I'm sure you'll have persuasive reasons for whatever you decide, so please don't feel the need to justify them to me :-)

erikbrinkman:
This branch should be ready to rebase again. I'm opting out of multivariate stats for now, but it could obviously be added later.

PR #1259, opened 2016-10-22
Fix cross-compile tmp dir removal

jay:
- cd out of the tmp dir before attempting to remove it.

---

Changes in 5d9d1b1 broke removal of the temporary directory.


nicowilliams:
Thanks!

PR #1270, opened 2016-11-04
Typo: s/more/mode/ in `--seq`

timmc:
I'm not actually sure I understand that sentence, but it does appear to be a typo either way.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -107,7 +107,7 @@ sections:
         output and an ASCII LF (line feed) is printed after every
         output.  Input JSON texts that fail to parse are ignored (but
         warned about), discarding all subsequent input until the next
-        RS.  This more also parses the output of jq without the `--seq`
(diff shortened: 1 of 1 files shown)


eric-brechemier:
@timmc If this sentence means that `jq` with the `--seq` option shall accept the output of `jq` used without the `--seq` option, it looks like a bug to me:

### jq to jq --seq

```
$ jq -n '1,2,3' | jq --seq '.' | xxd # produces no output
```

The other forms work as expected:

### jq --seq to jq --seq

```
$ jq --seq -n '1,2,3' | jq --seq '.' | xxd
0000000: 1e31 0a1e 320a 1e33 0a                   .1..2..3.
```

### jq to jq

```
$ jq -n '1,2,3' | jq '.' | xxd
0000000: 310a 320a 330a                           1.2.3.
```

even

### jq --seq to jq

```
$ jq --seq -n '1,2,3' | jq '.' | xxd
0000000: 310a 320a 330a                           1.2.3.
```

PR #1282, opened 2016-11-30
Mark startswith() and endswith() with "Since 1.4"

muffl0n:
It took me some time to figure out why the documented "startswith()" and "endswith()" [1] do not work. The reason: I'm using version 1.3. But the manual does not have an information since which version a feature exists. A simple "Since 1.4" would have helped a lot!

[1] https://stedolan.github.io/jq/manual/

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -1353,7 +1353,7 @@ sections:
       - title: "`startswith(str)`"
         body: |
 
-          Outputs `true` if . starts with the given string argument.
+          Outputs `true` if . starts with the given string argument. Since 1.4.
 
         examples:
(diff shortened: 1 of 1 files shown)


pkoppstein:
@muffl0n - The manual you refer to is clearly marked "development version" (at the top).  Notice also that at the top there are links to other versions ("For released versions, see jq 1.5, jq 1.4 or jq 1.3"). For better or worse, it was decided some time ago to version the manuals rather than litter each edition with "first introduced" details.

muffl0n:
You are right, I just missed the hint right at the top. I'm used to those "littering" hints as many other projects use them. But if that's the decision for "jq" then that's the way it is.

Thanks for your explanation!

Issue #1314, opened 2017-01-23
|= doesn't handle empty results from the update expression


nicowilliams:
The fix in `src/builtin.jq` is trivial:

    -def _modify(paths; update): reduce path(paths) as $p (.; setpath($p; getpath($p) | update));
    +def _modify(paths; update): reduce path(paths) as $p (.; setpath($p; getpath($p) | (.,update)));

nicowilliams:
This will probably be fixed by the fix to #1313.

nicowilliams:
IIRC we have another issue asking that empty in `|=` be the same as deleting the values at the given paths where the update expression outputs `empty`.  I'm not sure that's a good idea.

I can confirm that fixing #1313 fixes this.

pkoppstein:
Regarding the cautionary note ("I'm not sure that's a good idea"), I'd like to suggest that at this point, the sine qua non is that `map_values(select(f))` should behave in the way that most people expect and/or want.  This could, if need be, be accomplished in jq 1.5 quite simply by:

    def map_values(f):
      with_entries( [.value|f] as $v | select( $v|length == 1) | .value = $v[0] ) ;

nicowilliams:
@pkoppstein Well, but `map_values(f)` is defined in terms of `|=`, so `map_values(select(stuff))` never worked correctly.

We can argue about whether `stuff |= empty` means "delete stuff" or "leave stuff alone".  Here I think "delete stuff" is nice when you want it, but surprising when you don't.  We have a way to delete things when that's what we want, but if you want to do updates and deletions in one expression, then having `stuff |= empty` mean "delete stuff" sure is the better semantic to choose.  Implementing that will be tricky.

As for `reduce`, when the update expression produces `empty`, we currently have undefined results, and that's surprising and annoying.  I think the only choices as to `reduce` and empty updates are: "undefined behavior" or "leaves state unmodified".

nicowilliams:
Given a fix to #1313, this causes `|=` to delete paths when the update expression is `empty`:

```
def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
```

Example:

```
$ ./jq -cn '{a:{b:{c:1,d:2}}}|.a.b.c |= select(.!=1)'
{"a":{"b":{"d":2}}}
[...]

PR #1315, opened 2017-01-23
reduce: handle empty updates (fix #1313)

### LINKED ISSUE #1313: reduce does not handle empty updates well; breaks walk/1 and others
`reduce ... (...; update)` doesn't handle empty results from the `update` expression very well:

    $ ./jq -cn 'reduce range(3) as $x ({}; if $x == 2 then empty else .[$x|tostring] |= $x end)'
    null
    $ ./jq -cn 'reduce range(3) as $x ({}; .[$x|tostring] |= $x)'
    {"0":0,"1":1,"2":2}

Workaround: wrap the `update` expression like this: `(.,update)`:

    $ ./jq -cn 'reduce range(3) as $x ({}; (.,if $x == 2 then empty else .[$x|tostring] |= $x end))'
    {"0":0,"1":1}

`src/compile.c:gen_reduce()` needs to do this automatically.  This makes it clear what the change needs to be:

    $ diff -u <(./jq --debug-dump-disasm -cn 'reduce range(3) as $x ({}; if $x == 2 then empty else .[$x|tostring] |= $x end)'|cut -d\  -f2-) <(./jq --debug-dump-disasm -cn 'reduce range(3) as $x ({}; (.,if $x == 2 then empty else .[$x|tostring] |= $x end))'|cut -d\  -f2-)

which is: prepend to the update expression body: a `FORK` and a `JUMP`, with the fork bound to the address of the body, and the jump bound to the address of the `STOREV` that follows the body.

nicowilliams:
This fixes #1313 (and #1314).

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -2413,10 +2413,18 @@ sections:
                   (2 as $item | . + $item) |
                   (1 as $item | . + $item)
 
+          If the reduction update expression outputs `empty` (that is,
+          no values), then the reduction state is left as-is.  For
+          example, `reduce range(4) as $n ({}; if .==2 then empty else
+          .[$n|tostring] |= $n)` will produce `{"0":0,"1":1,"3":3}`.
+
         examples:
           - program: 'reduce .[] as $item (0; . + $item)'
             input: '[10,2,5,3]'
             output: ['20']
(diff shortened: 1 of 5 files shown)


nicowilliams:
I'll probably push this today.  We can argue about whether `|=` should take the first or last value of the RHS expression later.

nicowilliams:
Pushed.

PR #1327, opened 2017-01-29
jv: Add some support for 64 bit ints in a very conservative way (ALTERNATIVE)

nicowilliams:
This is a variant of #1246 that doesn't change the size of a `jv`.

@dequis What do you think?

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -363,6 +363,33 @@ static jv f_tonumber(jq_state *jq, jv input) {
   return type_error(input, "cannot be parsed as a number");
 }
 
+static jv f_toint(jq_state *jq, jv input) {
+  input = f_tonumber(jq, input);
+  if (!jv_is_valid(input))
+    return input;
+  if (jv_is_int64(input) || jv_is_uint64(input))
+    return input;
+  double d = jv_number_value(input);
+  if (d < 0 && d >= INT64_MIN) {
+    int64_t i = d;
+    if (i < 0)
+      return jv_int64(i);
+  } else if (d < UINT64_MAX) {
+    uint64_t u = d;
+    if (d == (double)u)
+      return jv_uint64(u);
+  }
+  return jv_number(nearbyint(d));
+}
+
+static jv f_isint(jq_state *jq, jv input) {
+  if (jv_get_kind(input) != JV_KIND_NUMBER)
+    return type_error(input, "only numbers can be integers");
+  if (jv_is_int64(input) || jv_is_uint64(input) || jv_is_integer(input))
+    return jv_true();
+  return jv_false();
+}
+
 static jv f_length(jq_state *jq, jv input) {
   if (jv_get_kind(input) == JV_KIND_ARRAY) {
     return jv_number(jv_array_length(input));
@@ -1285,6 +1312,8 @@ static const struct cfunction function_list[] = {
   {(cfunction_ptr)f_json_parse, "fromjson", 1},
   {(cfunction_ptr)f_tonumber, "tonumber", 1},
   {(cfunction_ptr)f_tostring, "tostring", 1},
+  {(cfunction_ptr)f_toint, "tointeger", 1},
+  {(cfunction_ptr)f_isint, "isinteger", 1},
   {(cfunction_ptr)f_keys, "keys", 1},
   {(cfunction_ptr)f_keys_unsorted, "keys_unsorted", 1},
   {(cfunction_ptr)f_startswith, "startswith", 2},
--- a/src/jv.c
+++ b/src/jv.c
@@ -191,6 +191,10 @@ int64_t jv_int64_value(jv j)
       return (int64_t)j.u.uint64;
     return INT64_MAX;
   }
+  if (j.u.number > 0 && (int64_t)j.u.number < 0)
+    return INT64_MAX;
+  if (j.u.number < 0 && (int64_t)j.u.number > 0)
+    return INT64_MIN;
 #endif
   return j.u.number;
 }
@@ -207,6 +211,10 @@ uint64_t jv_uint64_value(jv j)
       return j.u.int64;
     return 0;
   }
+  if (j.u.number < 0)
+    return 0;
+  if (j.u.number > (double)UINT64_MAX)
+    return UINT64_MAX;
 #endif
   return j.u.number;
 }
--- a/src/jv.c
+++ b/src/jv.c
@@ -47,6 +47,10 @@ jv_kind jv_get_kind(jv x) {
   return x.kind_flags & KIND_MASK;
 }
 
+jv_subkind jv_get_subkind(jv x) {
+  return x.subkind_flags & KIND_MASK;
+}
+
 const char* jv_kind_name(jv_kind k) {
   switch (k) {
   case JV_KIND_INVALID: return "<invalid>";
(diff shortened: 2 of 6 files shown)


nicowilliams:
Arithmetic (and math) continues to go through `jv_number()` and `jv_number_value()`.  Only parsing and printing check the integer members of the union, and the utility `jv` functions for C applications that want them.  Not changing the size of `jv`, and not having multiple internal representations of any given number at once, seems conservative to me :) and anyways, I kept the title of your PR.

nicowilliams:
As to the `JQ_OMIT_INTS`... I am copying the SQLite3 pattern.  It's fairly cheap, but the problem is that it probably won't get much testing.  It is not a premature optimization so much as... me not having measured the impact of this new feature.

pkoppstein:
@nicowilliams - No doubt I'm missing something, but using @dequis's test case, it seems this change entails a step backwards:
```
github/nicowilliams/jq$ ./jq --version
jq-1.2-889-gb28c886

github/nicowilliams/jq$ git log -1
commit b28c886edbb45c2128f7f5eb79da169190473cff
Date:   Sat Jan 28 18:56:55 2017 -0600

    jv int64 and uint64 support

github/nicowilliams/jq$ echo 111111111111111111 | ./jq -c '[., 1*.]'
echo 111111111111111111 | ./jq -c '[., 1*.]'
[111111111111111104,111111111111111100]

```
By contrast:
```
$ echo 111111111111111111 | jq1.5 -c '[., 1*.]'
[111111111111111100,111111111111111100]
```

nicowilliams:
@pkoppstein Yeah, and that's not the only thing that's wrong.  Try `echo 2 63^p|dc|jq .`.  We need more test cases, that's for sure.

nicowilliams:
Tests still needed.

nicowilliams:
@wtlangford Can you review this?  I think it's done.

Also, I'm thinking of taking this approach a bit further.  I'm thinking of adding `jv` sub-types for:

 - binary data (sub-type of array, appearing to be an array of numbers in the range of 0..256 -- i.e., bytes)
 - array slices when the offset/size does not fit in a `jv`

nicowilliams:
Ah, so, `jv_identical()` needs to be updated too.

src/jv.h
```diff
@@ -16,18 +16,26 @@ typedef enum {
...
+} jv_subkind;
+
 struct jv_refcnt;
 
 /* All of the fields of this struct are private.
    Really. Do not play with them. */
 typedef struct {
   unsigned char kind_flags;
-  unsigned char pad_;
+  unsigned char subkind_flags;
```
wtlangford: And now our padding is gone. 😢

wtlangford (changes requested):
Nothing here seems too concerning, other than the overflow checking.

We appear to continue to do double math with this change.  Do we want to consider being subkind-aware for adding and subtracting (and maybe multiplication?).  Division should, of course, use double behavior, since we don't want integer division semantics in jq.

I'm envisioning this problem (which is a floating-point precision error): 
`9223372036854774784 + 1` #9223372036854774784
Of course, jq still prints that as `9223372036854775000` because of a bug somewhere in `jv_dtoa`...

Is inconsistent integer semantics _worse_ than surprising double precision errors?

Also, we ought to use in our constant folding.
Also, tests.  Lots of them.

src/jv.h
nicowilliams: Well, yes, we lose the padding.  I suppose I could encode sub-kind into kind_flags.  We could always do that later if ever we want to use the padding for something else.

src/builtin.c
```diff
@@ -363,6 +363,33 @@ static jv f_tonumber(jq_state *jq, jv input) {
...
+static jv f_toint(jq_state *jq, jv input) {
+  input = f_tonumber(jq, input);
+  if (!jv_is_valid(input))
+    return input;
+  if (jv_is_int64(input) || jv_is_uint64(input))
+    return input;
+  double d = jv_number_value(input);
+  if (d < 0 && d >= INT64_MIN) {
+    int64_t i = d;
+    if (i < 0)
```
nicowilliams: No, it's just for deciding whether to return a signed 64-bit integer representation or unsigned 64-bit integer representation.

src/builtin.c
```diff
@@ -363,6 +363,33 @@ static jv f_tonumber(jq_state *jq, jv input) {
...
+      return jv_int64(i);
+  } else if (d < UINT64_MAX) {
+    uint64_t u = d;
+    if (d == (double)u)
```
nicowilliams: Ditto.  Here we know that `d` is positive, and if it's also less than `UINT64_MAX` then we can represent it as a `uint64_t`.

src/builtin.c
wtlangford: But if the check is false, the next code that runs is `return jv_number(nearbyint(d));`, which is a double representation.

src/jv.c
```diff
@@ -138,27 +142,139 @@ static void jvp_invalid_free(jv x) {
...
-  jv j = {JV_KIND_NUMBER, 0, 0, 0, {.number = x}};
+  jv j = {JV_KIND_NUMBER, JV_SUBKIND_NONE, 0, 0, {.number = x}};
+  return j;
+}
+
+jv jv_int64(int64_t x) {
+#ifndef JQ_OMIT_INTS
+  jv j = {JV_KIND_NUMBER, JV_SUBKIND_INT64, 0, 0, {.int64 = x}};
+#else
+  jv j = {JV_KIND_NUMBER, JV_SUBKIND_INT64, 0, 0, {.number = x}};
```
nicowilliams: Oops!  (It would have no effect, given that other code that would check it would be omitted.  But it's still a bug.)

src/jv.c
```diff
@@ -138,27 +142,139 @@ static void jvp_invalid_free(jv x) {
...
+#ifndef JQ_OMIT_INTS
+  char sk = jv_get_subkind(j);
+  if (sk == JV_SUBKIND_INT64)
+    return j.u.int64;
+  if (sk == JV_SUBKIND_UINT64) {
+    if (j.u.uint64 <= INT64_MAX)
+      return (int64_t)j.u.uint64;
+    return INT64_MAX;
+  }
+  if (j.u.number > 0 && (int64_t)j.u.number < 0)
```
nicowilliams: I like this!  Thanks!

src/jv.c
```diff
@@ -138,27 +142,139 @@ static void jvp_invalid_free(jv x) {
...
+  if (sk == JV_SUBKIND_UINT64)
+    return j.u.uint64;
+  if (sk == JV_SUBKIND_INT64) {
+    if (j.u.int64 >= 0)
+      return j.u.int64;
+    return 0;
+  }
+  if (j.u.number < 0)
+    return 0;
+  if (j.u.number > (double)UINT64_MAX)
```
nicowilliams: Yes!

src/jv.c
```diff
@@ -138,27 +142,139 @@ static void jvp_invalid_free(jv x) {
...
   }
+#ifndef JQ_OMIT_INTS
+  char sk = jv_get_subkind(j);
+  if (sk != JV_SUBKIND_NONE) {
+    assert(sk == JV_SUBKIND_UINT64 || sk == JV_SUBKIND_INT64);
+    return 1;
+  }
+#endif
   double x = jv_number_value(j);
+  /* XXX Check against actual double min/max integers */
```
nicowilliams: Fair enough.

src/builtin.c
wtlangford: These checks should use the `nextafter` stuff from below for `INT64_MIN` and `UINT64_MAX`

nicowilliams: Eh?  No, there's an `else if` at 377.

dequis:
>We appear to continue to do double math with this change

In my version of this, that was explicitly intended behavior. The idea was to fix the common case of using jq as a pretty-printer and/or filter, so as long as the numbers aren't touched they keep their old values.

src/builtin.c
wtlangford: Perhaps I misunderstand.
Line 373 is checking to see if `d is negative AND d fits into an int64_t`
Line 374 does the cast
What does line 375 check?

wtlangford:
Ah, then I have no complaints about the double-precision math.  Additionally, maintaining existing behavior is easier than changing it and then deciding to change it back later; so we can always revisit in the future.

src/builtin.c
nicowilliams: Oh yes, i'm the one who's confused here.  I need to s/else // here :)

nicowilliams:
Eventually we might add branches to basic arithmetic operators to avoid double-precision math when possible.  But I'm a bit loathe to do that right now.  The goal here really is to allow `jq .` to pass _integers_ through unmodified.  Of course, they'll still get parsed and formatted, but there won't be any scientific notation on output, which appears to be the most commonly desired thing here.

We could also fix the formatter to never use scientific notation for doubles with no non-zero decimal part.  But we do also get calls for larger integer ranges than -2^52..2^52, which I think is fair enough.

chrishmorris:
Thank you all for JQ.

It would be great if you could address this issue.

JalinWang:
Thank you all for JQ!
It would be great if you could address this issue.

image

PR #1329, opened 2017-02-05
Add `builtins` builtin

### LINKED ISSUE #1318: Add builtin that outputs list of builtins
```
jq -n builtins
```
should list all public builtins.

wtlangford:
Add a `builtins` builtin that lists all "public" builtins. (what a sentence)

"Public" builtins are builtins that do not start with an `_`.  This implementation delays dropping unreferenced defs and binding to a jq program until after all builtins have been bound (so that we can use the builtins block to generate a list of builtin functions).

The `builtins` function itself was bytecoded to avoid having to store the list of names in a static variable or in the `jq_state`.

This also fixes a memory leak in the libm cfunctions when the arguments were not numbers.

This also removes a level of unnecessary indirection for the libm math functions.
Previously, math functions were defined as such:
`def cos: _cos;`, where `_cos` is a cfunction.  Instead, we just define `cos` as a cfunction.

Closes #1318

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -101,9 +101,15 @@ static jv f_ ## name(jq_state *jq, jv input) { \
 
 #define LIBM_DDD(name) \
 static jv f_ ## name(jq_state *jq, jv input, jv a, jv b) { \
-  if (jv_get_kind(a) != JV_KIND_NUMBER || jv_get_kind(b) != JV_KIND_NUMBER) \
-    return type_error(input, "number required"); \
   jv_free(input); \
+  if (jv_get_kind(a) != JV_KIND_NUMBER) { \
+    jv_free(b); \
+    return type_error(a, "number required"); \
+  } \
+  if (jv_get_kind(b) != JV_KIND_NUMBER) { \
+    jv_free(a); \
+    return type_error(b, "number required"); \
+  } \
   jv ret = jv_number(name(jv_number_value(a), jv_number_value(b))); \
   jv_free(a); \
   jv_free(b); \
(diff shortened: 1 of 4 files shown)


src/builtin.c
```diff
@@ -1390,10 +1409,10 @@ static block bind_bytecoded_builtins(block b) {
                                             BLOCK(gen_param("start"), gen_param("end")),
                                             range));
   }
-
-  return block_bind_referenced(builtins, b, OP_IS_CALL_PSEUDO);
+  return block_bind(builtins, b, OP_IS_CALL_PSEUDO);
```
nicowilliams: So, this means we don't lose the bytecode for unreferenced builtins.  That seems... sad?  :)

Is there some other way?

src/builtin.c
```diff
@@ -1390,10 +1409,10 @@ static block bind_bytecoded_builtins(block b) {
...
 }
 
+
 #define LIBM_DD(name) "def " #name ": _" #name ";"
```
nicowilliams: These macros should be removed now, no?

src/builtin.c
```diff
@@ -1431,13 +1448,18 @@ static const char* const jq_builtins =
...
+}
+
 static int builtins_bind_one(jq_state *jq, block* bb, const char* code) {
   struct locfile* src;
   src = locfile_init(jq, "<builtin>", code, strlen(code));
   block funcs;
   int nerrors = jq_parse_library(src, &funcs);
   if (nerrors == 0) {
-    *bb = block_bind_referenced(funcs, *bb, OP_IS_CALL_PSEUDO);
+    *bb = block_bind(funcs, *bb, OP_IS_CALL_PSEUDO);
```
nicowilliams: See above comment.

src/builtin.c
```diff
@@ -1459,14 +1481,18 @@ static int slurp_lib(jq_state *jq, block* bb) {
...
-  nerrors = builtins_bind_one(jq, bb, jq_builtins);
+  nerrors = builtins_bind_one(jq, &builtins, jq_builtins);
   assert(!nerrors);
-  *bb = bind_bytecoded_builtins(*bb);
-  *bb = gen_cbinding(function_list, sizeof(function_list)/sizeof(function_list[0]), *bb);
+  builtins = bind_bytecoded_builtins(builtins);
+  builtins = gen_cbinding(function_list, sizeof(function_list)/sizeof(function_list[0]), builtins);
+  builtins = gen_builtin_list(builtins);
+  *bb = block_bind(builtins, *bb, OP_IS_CALL_PSEUDO);
+  *bb = block_drop_unreferenced(*bb);
```
nicowilliams: Oh, maybe here's the answer to my question.   Plus I can run your commits and see for myself.  In a few minutes I shall know.

src/builtin.c
nicowilliams: Right, never mind.  Looks good to me!

src/builtin.c
nicowilliams: They are harmless now, but 1416-1421 should now be removed.  The comment at 1427 should also be removed.

nicowilliams: 1428-1433 can also be removed.

src/builtin.c
nicowilliams: Never mind this comment.  LGTM.

src/builtin.c
nicowilliams: Indeed, my concern is gone.  Thanks!

PR #1334, opened 2017-02-07
fix truncate_stream to match docs

phanimahesh:
The current behaviour of `truncate_stream` is to always truncate 1 level from output, and to retain those nodes that would survive a truncation by `$n`. Most likely, this was a typo, since this behaviour doesn't make much sense to me.

This small change fixes it to match docs. Or, did I misunderstand docs completely?


nicowilliams:
Thanks!

Issue #1337, opened 2017-02-11
Documentations for .. operator


fadado:
If you follow the description for the operator `..` in the manual you are redirected to the definition of `recurse `, then to `recurse(.[]?)`, then to `recurse(f; . != null)`...

I propose the following equivalence:

### `..`

Short-hand for `.,  .[]?, (.[]?|.[]?), (.[]?|.[]?|.[]?), ...`.

nicowilliams:
Hmmm, how about:

          Recursively descends `.`.  Same as `recurse` without
          arguments.  This is intended to resemble the XPath `//`
          operator.  Note that `..a` does not work; use `..|.a` instead.
          In the example below we use `..|.a?` to find all the values of
          object keys "a" in any object found "below" `.`.

          This is particularly useful in conjunction with `path(EXP)`
          and the `?` operator.

fadado:
> Hmmm, how about:
> 
>       Recursively descends `.`

Very good.

I will reuse this tread for other documentation confusing aspects:

1)  Defining `error(message)` the manual says _Produces an error_. A will say _signals_ or _raises_, reserving _produces_ for filters. By the way, `error` accepts all kind of values as argument, not only strings: this must be clarified.

2) Defining array access the manual says _used with the slice operator_. Is slicing an _operator_? In coherence with the manual  it is a _syntax_ or _filter_, not an operator.

Thanks in advance for your attention.

nicowilliams:
To be absolutely pure, we have just these types of things in jq:

 - special operations that can only possibly be implemented with special syntax (e.g., `EXP as $name | ...`, `reduce`, ...)
 - special operations supported by the bytecode opcodes and syntax (e.g., `.[]` being implemented as the `EACH` opcode; the compiler maps `.[]` to `EACH`)
    - things like `+` also fall into this category -- the compiler converts them into calls to the `_plus` (etcetera) builtins
 - special operations supported by the bytecode opcodes and byte-coded builtins (e.g., `empty`, `path`)
 - special operations supported by C-coded builtins (e.g., math functions)
 - jq-coded builtins

Some of these things look like "operators": `+`, `-`, and so on.  While others look like something other than operators (e.g., `EXP as $name | ...`, `reduce`, ...).  The rest look like functions (e.g., `empty`).

So perhaps we can say we have three kinds of things:

 - syntax
 - operators
 - functions

Am I missing anything?

nicowilliams:
See #1340 for modest improvements related to this.

[...]

Issue #1326, opened 2017-01-27
Documentation/Feature request: Clarify expressions/scope


lylemoffitt:
## Exigent Question:
At any point in a **jq** script what does the filter `.` return? It may be easy for an experienced user, but it's not clear from the documentation. Put another way: what defines an expression? What delimits scope? The answers to these questions are **implied**, but not explicitly or clearly stated by the documentation. It's ironic that the dot filter is referred to the "least interesting filter", because it is the key to understanding the transformation of data through the script.
>  To clarify, a *script* is the string passed as the command-line argument `filter` or loaded with ` --from-file`. This is to disambiguate from the pragmatic units chained together within it, which are also called 'filters'.

## Problems:
The man page doesn't really say a whole lot about parenthesis. They pretty much only show up in function signatures and in examples. Yet, they have a fundamental relationship with the dot filter, and thus a critical role in the functioning the script. Their usage should be clarified. It would also be helpful to clarify  their relationship with the object constructors, `[]` and `{}`, as all three are used to create sub-expressions and return objects.

The easy thing to do here would be to just create a section where you define `()` as an *expression operator* or *scope operator*, and then stick all the missing explanation there. This might solve the immediate issue, but you could do a lot better. I'm trying to stick with one problem here, but in general the manual could be a lot clearer. I don't know if you're trying to intentionally hide that **jq** is a full-blown language, but it would certainly be a lot cleaner if you approached explaining the query language like it was the pure-function programming language it is. 

## Suggested Solutions:

1. Define the operator `()` as a *Value Constructor* and put it in the **Types and Values** section. It constructs a value from the output of the contained expression. The only thing that would be needed to be changed about its existing functionality in order to bring it in line with the other constructor operators is that it must also work when the expression is empty. Analogous to `[]` and `{}`, this should be implemented to construct a `null` value.

    #### Example:
    > For each serialized JSON input, the type construction operator should return the minimum viable value of same type when the contained expression is empty, or the result of composing the expression over the input otherwise.

    ```sh
    echo '<json>' | jq -c '[]'              #=> []
    echo '<json>' | jq -c '[<expr>]'        #=> [result of <expr> applied to <json>]
    echo '<json>' | jq -c '{}'              #=> {}
    echo '<json>' | jq -c '{<key>:<expr>}'  #=> {<key>: result of <expr> applied to <json>}
    echo '<json>' | jq -c '()'              #=> null
[...]

PR #1340, opened 2017-02-15
Improve manual

nicowilliams:
Improvements for the manual inspired by #1326 and #1337.

### DIFF
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -255,12 +255,12 @@ sections:
 
   - title: Basic filters
     entries:
-      - title: "`.`"
+      - title: "Identity: `.`"
         body: |
 
-          The absolute simplest (and least interesting) filter
-          is `.`. This is a filter that takes its input and
-          produces it unchanged as output.
+          The absolute simplest filter is `.` .  This is a filter that
+          takes its input and produces it unchanged as output.  That is,
+          this is the identity operator.
 
           Since jq by default pretty-prints all output, this trivial
           program can be a useful way of formatting JSON output from,
@@ -271,17 +271,24 @@ sections:
             input: '"Hello, world!"'
             output: ['"Hello, world!"']
 
-      - title: "`.foo`, `.foo.bar`"
+      - title: "Object Identifier-Index: `.foo`, `.foo.bar`"
         body: |
 
           The simplest *useful* filter is `.foo`. When given a
           JSON object (aka dictionary or hash) as input, it produces
           the value at the key "foo", or null if there's none present.
 
-          If the key contains special characters, you need to surround
-          it with double quotes like this: `."foo$"`.
-
           A filter of the form `.foo.bar` is equivalent to `.foo|.bar`.
+
+          This syntax only works for simple, identifier-like keys, that
+          is, keys that are all made of alphanumeric characters and
+          underscore, and which do not start with a digit.
+
+          If the key contains special characters, you need to surround
+          it with double quotes like this: `."foo$"`, or else `.["foo$"]`.
+
+          For example `.["foo::bar"]` and `.["foo.bar"]` work while
+          `.foo::bar` does not, and `.foo.bar` means `.["foo"].["bar"]`.
 
         examples:
           - program: '.foo'
@@ -294,7 +301,7 @@ sections:
             input: '{"foo": 42}'
             output: [42]
 
-      - title: "`.foo?`"
+      - title: "Optional Object Identifier-Index: `.foo?`"
         body: |
 
           Just like `.foo`, but does not output even an error when `.`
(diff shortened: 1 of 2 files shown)


docs/content/3.manual/manual.yml
```diff
@@ -271,17 +271,24 @@ sections:
             input: '"Hello, world!"'
             output: ['"Hello, world!"']
 
-      - title: "`.foo`, `.foo.bar`"
+      - title: "Object Identifier-Index: `.foo`, `.foo.bar`"
```
lylemoffitt: I think `Member Selector-Index:` might be clearer, since `.foo` doesn't return an object, what it identifies is a *key*, and it uses syntax that most people will recognize as "dot member" notation.

docs/content/3.manual/manual.yml
```diff
@@ -314,37 +321,22 @@ sections:
...
-      - title: "`.[<string>]`, `.[2]`, `.[10:15]`"
+      - title: "Generic Object Index: `.[<string>]`"
         body: |
 
           You can also look up fields of an object using syntax like
-          `.["foo"]` (.foo above is a shorthand version of this). This
-          one works for arrays as well, if the key is an
-          integer. Arrays are zero-based (like javascript), so `.[2]`
-          returns the third element of the array.
+          `.["foo"]` (.foo above is a shorthand version of this, but
```
lylemoffitt: Need a proper example for `.[<string>]`, since `.["foo"]` won't actually be verified/tested during the manual build.

docs/content/3.manual/manual.yml
```diff
@@ -294,7 +301,7 @@ sections:
             input: '{"foo": 42}'
             output: [42]
 
-      - title: "`.foo?`"
+      - title: "Optional Object Identifier-Index: `.foo?`"
```
lylemoffitt: Would prefer `Optional Member Selector-Index`. Same reason as above.

docs/content/3.manual/manual.yml
lylemoffitt: Alternatively `Object Member-Index` would work. This would be more in keeping with the Object/Array distinction you're drawing.

docs/content/3.manual/manual.yml
```diff
@@ -408,15 +412,16 @@ sections:
           Like `.[]`, but no errors will be output if . is not an array
           or object.
 
-      - title: "`,`"
+      - title: "Comma, Output Stream Join: `,`"
```
lylemoffitt: `Output Stream:` is simpler. Can remove `comma` b/c it's redundant. Would avoid referring to it as a "join", since it's actually a FORK opcode and will will thus come across very much as a "split". The steaming aspect is important. Would focus on that instead.

docs/content/3.manual/manual.yml
```diff
@@ -408,15 +412,16 @@ sections:
...
-          right. For instance, filter `.foo, .bar`, produces
-          both the "foo" fields and "bar" fields as separate outputs.
+          same input will be fed into both and the two filters' output
+          value streams will be joined: first, all of the outputs
+          produced by the left expression, and then all of the outputs
+          produced by the right. For instance, filter `.foo, .bar`,
+          produces both the "foo" fields and "bar" fields as separate
+          outputs.
 
         examples:
```
lylemoffitt: Should include an example that makes it clear that the next step in the pipeline is called separately (and sequentially) with each value. Ex: `'["a","b","c","d"]' >>> '.[1],.[3] | [.]' >>> ["b"],["d"]`

docs/content/3.manual/manual.yml
```diff
@@ -1631,12 +1639,16 @@ sections:
       - title: "`..`"
```
lylemoffitt: This is now the only operator without a name. How about `Recurse operator` or `Expansion operator`?

docs/content/3.manual/manual.yml
nicowilliams: JSON speaks of `objects` with `name`/`value` pairs.  jq calls them `objects` with `key`/`value` pairs.  There will be no talk of fields or members.

docs/content/3.manual/manual.yml
nicowilliams: There are tests of this elsewhere.  I may or may not add a test here.

docs/content/3.manual/manual.yml
nicowilliams: Calling it the "comma" operator is... concise, but not precise.  I can't think of a one-word precise name for it.  "Output Stream" is not correct since `.[]` too (and others) creates an output stream, for example.

The opcodes don't always provide good names for things.  `.[]` is clearly a value iterator, but internally it's called `EACH`.  I would rather call `.[]` an iterator than an "each".

The actual semantics of `,` is that the whole thing outputs the outputs of the expression on the left, then those of the expression on the right.

Perhaps "Expression Concat"?  Something like that.

I must say I don't like "Output Stream Join" either, though mostly because it will seem to a new user like an odd, off-putting name for what should obviously be a trivial thing.  It's too soon here to make it clear that jq is all about generators, and that generators are pervasive.

nicowilliams: (Avoiding surprising new users with generators this early is partly why I'd like to call this just `Comma`.  There's a whole section later about generators.)

docs/content/3.manual/manual.yml
nicowilliams: I don't follow.

docs/content/3.manual/manual.yml
nicowilliams: Good point!

docs/content/3.manual/manual.yml
lylemoffitt: Good point.

docs/content/3.manual/manual.yml
nicowilliams: I'll also move this up in the manual.

docs/content/3.manual/manual.yml
nicowilliams: Perhaps `.["string"]` should come before `.string` in the manual.  After all, "how do I get the value of <non-identifier-like string>?" is almost certainly the most FAQ...

lylemoffitt: I would say `.string` is most common and most intuitive, and should therefore go first. Maybe add a small note in the desc that `.["string"]` is an alternative notation?

nicowilliams: It's already that way.

lylemoffitt: Then no change is needed. My only point of criticism here was on the name `Object Identifier-Index:`

docs/content/3.manual/manual.yml
lylemoffitt: A one word, precise name is not really necessary. Your point about opcodes and names is valid, but would like to point out that "each" is just another way of expressing iteration. I actually think "Fork" wouldn't be a bad candidate for a one-word, precise name. Your suggestion of "Expression Concat" is another good choice, but less desirable, imo.

docs/content/3.manual/manual.yml
nicowilliams: Hmm, well, I want users to understand why `.foo` doesn't work for keys like "foo$bar.baz"...  I want the reason to jump at them.  Maybe there's no hope!  :)

lylemoffitt: I think the only thing you could do is move the `.[***]` operators up to immediately follow `.foo` and be before `.foo?`. This isn't really needed, though (imo)

PR #1343, opened 2017-02-17
Alternation destructuring operator

wtlangford:
Adds support for a `?//` operator to enable chained/alternation destructuring.

Also comes with a fix for a tripped assertion in `reduce . as $n (.; .)` (a notably useless expression, but tripped assertions are bad).

See https://github.com/stedolan/jq/issues/856 for more rationale, etc on the alternation syntax.

## Syntax
```
./jq '. as {$a, $b} ?// [$a,$b,$c] | [$a,$b,$c]'
{"a":1, "b":2}
[1,2,null]
[1,2,3]
[1,2,3]
```

The syntax should work with any previously valid destructuring syntax, but allows chaining alternatives.  After this expression, all variables named in the expression are bound, even if the syntax that bound them wasn't used, in which case, they are bound to `null`.

This syntax is also supported in reduce and foreach:
```
reduce .[] as [$k,$v] ?// {key:$k,value:$v} ( {}; .[$k] |= $v )
foreach .[] as [$k,$v] ?// {key:$k,value:$v} ( null; $k,$v )
foreach .[] as [$k,$v] ?// {key:$k,value:$v} ( null; .; $k,$v )
```

## Implementation
The bytecode for this is a bit heavy-handed, relying on `FORK_N` for flow-control.
Essentially, we lay out the bytecode like this:

```
. as MATCHER1 ?// MATCHER2 ?// MATCHER3 | remainderOfProgram
```
becomes:
```
10 FORK_N 30 // Jump to the next matcher on error
20 MATCHER1 (Matches $a and $b)
25 jq code: `| {$a,$b}` // convert the variables back into an object
29 JUMP 70 // successful match, jump to the remainder
30 FORK_N 50
40 MATCHER2 (Matches $a and $c)
45 jq code: `| {$a,$c}`
49 JUMP 70
50 MATCHER3 (Matches $d and $e)
60 jq code: `| {$d,$e}`
70 jq code: `. as {$a, $b, $c, $d, $e}` // convert the output object back into variables
80 remainderOfProgram
```

This has the unfortunate side-effect of destructuring twice, even in the shortest path, but there's no easy way to avoid this, since later LOADV instructions need to bind to a specific STOREV instruction.

The implementation comes with two new opcodes, though one is only used in the compiler.
`PUSH_UNDER` is a new opcode to replace the 
```
SUBEXP_BEGIN
LOADK $imm
SUBEXP_END
```
paradigm, which has the effect of placing `$imm` just below the top item on the stack.  Instead, we do this in one op.  The speed boost here is not likely to be significant (though it _is_ two fewer opcodes each time we do this), but it could lead to easier processing/optimizations in the future.

`DESTRUCTURE_ALT` is used to ease the compiler's job of detecting destructuring alternatives.  Each alternative (except the last!) is placed as the `subfn` of a `DESTRUCTURE_ALT` instruction.

Anyways, bang on this please.  I'm not convinced it's not broken in some subtle way.
Certainly don't merge it.  It still needs proper tests and documentation, but I wanted to get opinions before I bothered writing that.


nicowilliams:
This is fantastic.  I will need a solid block of time to review this, and it will probably take me a few days in total.

nicowilliams:
Ah, it's mostly `src/compile.c` that needs review.  The rest is obvious enough.

nicowilliams:
Have you run this with `valgrind --leak-check=full`?

nicowilliams:
That's all my comments so far.  I need to do a second pass to look for leaks, but it's sure nice to let valgrind do that for me, provided you have good enough tests.

My only substantive comment is that at first glance it may be better to generate first a sequence of `STOREV`s of `null`s to all the vars, then the matchers with the `FORK_OPT` prefixes, then the body.  I suspect this would a) cause less garbage (no short-lived object to store the matched variables) and b) fewer instructions (no need to generate the final "destructuring" of that short-lived object).  Maybe I'm missing something as to why that wouldn't work or would require more work, and if so, please let me know what.

I also had some nits that you are free to disregard -- we're not sticklers for C style here.

src/execute.c
```diff
@@ -417,6 +417,15 @@ jv jq_next(jq_state *jq) {
       break;
     }
 
+    case PUSH_UNDER: {
+      jv v = jv_array_get(jv_copy(frame_current(jq)->bc->constants), *pc++);
```
wtlangford: I like it.  Will do.

src/compile.c
```diff
@@ -553,6 +564,11 @@ block gen_call(const char* name, block args) {
 
 
 block gen_subexp(block a) {
+  if (block_is_const(a)) {
```
wtlangford: You're right.  I'll tweak this to actually check for a `LOADK` instead.  It'll look a little less clean, but it explicitly avoids the case where we accidentally drop a swap.  I don't want this optimization to break things surprisingly in that case.

src/compile.c
```diff
@@ -570,27 +586,35 @@ block gen_const_object(block expr) {
   jv k = jv_null();
   jv v = jv_null();
   for (inst *i = expr.first; i; i = i->next) {
-    if (i->op != SUBEXP_BEGIN ||
+    if (i->op == PUSH_UNDER) {
+      k = jv_copy(i->imm.constant);
+      i = i->next;
+    } else if (i->op != SUBEXP_BEGIN ||
         i->next == NULL ||
```
wtlangford: I don't _think_ so, but I didn't want to rule out anything that was hand-bytecoded or that I'd otherwise missed.

src/compile.c
```diff
@@ -702,13 +726,139 @@ static block bind_matcher(block matcher, block body) {
...
+ * END will destructure the object returned from a wrapper into all of the
+ * variables contained in the entire destructuring expression and is what the
+ * program's body binds to.
+ */
+static block alternation_wrapper(block submatcher, block *final_matcher, int is_first, int is_last, jv vars_list) {
+    // We need to generate an output object of what succeeded here so we can
+    // bind everyone to those values.
+    block end = gen_subexp(gen_const(jv_object()));
+
+    // Manually restructure the values we extracted into a new object.
```
wtlangford: This is primarily because I couldn't find a way to `STOREV` into existing variables, though maybe that's just a lack of understanding of the bytecode format.  This is certainly still less garbage/bytecode than any existing way to implement this.

I'll take a look into making two `STOREV` instructions store to the same memory location, but I'm not sure that `inst` has a way to do that.

src/compile.c
```diff
@@ -702,13 +726,139 @@ static block bind_matcher(block matcher, block body) {
...
+
+  // Pass through the matchers to find all destructured names.
+  int num_matchers = 0;
+  while (matchers.first && matchers.first->op == DESTRUCTURE_ALT) {
+    block_append(&altmatchers, inst_block(block_take(&matchers)));
+    num_matchers++;
+  }
+
+  // We don't have any alternations here, so we can use the simplest case.
+  if (num_matchers == 0) {
```
wtlangford: Hah.  I've learned to love explicit braces from my time spent writing Go.  I find it easier to read despite the extra lines now.  I'll drop them if you'd prefer, though.  Both are still sufficiently readable in my eyes.

src/compile.c
nicowilliams: `STOREV` doesn't create a variable though, and `reduce` and `foreach` definitely `STOREV` multiple times.

Ultimately it's `gen_op_var_fresh()` and this code in `src/compile.c:compile()` that creates/allocates a variable:

```
    if ((opcode_describe(curr->op)->flags & OP_HAS_VARIABLE) &&
        curr->bound_by == curr) {
      curr->imm.intval = var_frame_idx++;
      localnames = jv_array_append(localnames, jv_string(curr->symbol));
    }
```

In `gen_reduce()` we see this:

```
        block gen_reduce(block source, block matcher, block init, block body) {
------>block res_var = gen_op_var_fresh(STOREV, "reduce");
------>block update_var = gen_op_bound(STOREV, res_var);
        ...
                                                gen_op_target(FORK, jmp),
                                                jmp,
                                                body,
var update ------------------->update_var)),
        ...
          return BLOCK(gen_op_simple(DUP),
                       init,
var create --->res_var,
                       gen_op_target(FORK, loop),
                       loop,
                       gen_op_bound(LOADVN, res_var));
        }

```

And this is what a reduction looks like after compilation:

```
$ jq -n --debug-dump-disasm 'reduce . as $i (null; $i)'
0000 TOP
0001 DUP
0002 LOADK null
0004 STOREV $reduce:0
0007 FORK 0023
0009 DUPN
0010 STOREV $i:1
0013 LOADVN $reduce:0
0016 LOADV $i:1
0019 STOREV $reduce:0
0022 BACKTRACK
0023 LOADVN $reduce:0
0026 RET

null
$ 
```

An you can see that multiple `STOREV`s to the same variable can work.  The key is that the `STOREV` in the patterns can't be "fresh", otherwise they *will* create new variables and my plan would fail.

src/compile.c
```diff
@@ -702,13 +726,139 @@ static block bind_matcher(block matcher, block body) {
...
+    if (!is_last) {
+      submatcher = BLOCK(gen_op_target(FORK_OPT, submatcher), submatcher);
+    }
+    return submatcher;
+}
+
+// Extract destructuring var names from the block
+// *all_vars should be a jv_object (for set semantics)
+// *matcher_vars should be a jv_array
+static void alternation_vars(block b, jv *all_vars, jv *matcher_vars) {
```
nicowilliams: Though if you can make the change above re `STOREV`s, then you don't need the `matcher_vars` output here, just the `all_vars` output, which then can be an input and then this function can just return the modified object.  Then this function would indeed be generic.

src/compile.c
nicowilliams: Ahhh, so you'll need to have a prologue of `gen_op_var_fresh()` or `gen_op_unbound()` `STOREV`s, but the corresponding `STOREV`s in the `Pattern`s will need to be bound to those in the prologue in order to NOT create new vars.

I think that's the trick.  Of course, that will make the non-alternation case, e.g., `. as $foo |`, generate two `STOREV`s per-variable where one would suffice, though that could be optimized with a bit more effort.

nicowilliams:
BTW, this is fun, and works the same with and without your patches:

```
$ jq -n '[1,[2]]|. as [$a,[$a]] | $a'
2
```

Causing such a pattern to raise a compile time error isn't worth the effort, so I'd leave it as-is.  But it illustrates the point that currently the `STOREV`s generated for each pattern are fresh, but with the scheme I'm proposing you'd have to make sure they're bound to the corresponding `STOREV`s in the prologue.

nicowilliams:
Back to the pathological example, `. as [$a, [$a]] |`, I gave earlier...

With your scheme the `body` gets bound to fresh vars stored into at the end by destructuring the special-use object that gathers the matching pattern's bindings.

With mine the `body` gets bound to the vars from the prologue, so what happens to the second `$a` in this pathological case?  All the vars in the matchers in this case have to bind to the ones in the prologue, so what would happen is that both `$a`s in this example would match to the same `$a`, and both `.[0]` and `.[1][0]` get stored, in that order, into `$a`... getting the same result as in your patch (and in master): `$a` binds to `.[1][0]`.

nicowilliams:
As to the non-alternation case producing an unnecessary `STOREV`, I could live with that, though I think we can also later optimize it away by simply fresh-binding the `STOREV`s in the one matcher and removing/not generating the prologue.  Or we could create a bunch of very similar grammar rules to distinguish the no-alternation case from the alternation case -- a bit silly seeming, but whatever.

wtlangford:
Detecting patterns such as `[1,[2]] |. as [$a,[$a]] | $a` at compile time shouldn't be _too_ hard, given that we are already walking through the variable list anyways.

nicowilliams:
To recap, for the no-garbage approach you'd:

 - change `alternation_vars()` so it only extracts/updates a list of unbound vars from the given block
 - change `bind_alternation_matchers()` so that:
    - if there's only one matcher, it uses it as is and binds the body to the matcher
    - if there's more than one matcher then it extracts the hash of vars from all, generates the prologue of fresh vars (`STOREV`s), and binds the matchers and the body to the prologue
 - change `alternation_wrapper()` so all it does is add a `FORK_OPT` prefix to each matcher but the last; you'd no longer need the `is_first` input argument, I don't think

This actually should avoid the two-`STOREV` issue in the non-alternation case.

nicowilliams:
I think there's no need to do anything about the `. as [$a, [$a]] |` case if we don't have to.  We'd just document it as having undefined behavior, thus giving us the ability to do something interesting with it in the future.

src/compile.c
nicowilliams: Keep them if you like them.  I just won't write them if I don't need them.

wtlangford:
An interesting note on the `. as [$a, [$a]] |` case:

The old matching code actually generates two separate fresh vars for this, and the body is bound to the first var generated.  Since we build the matchers in right-to-left order, this has the neat side effect of giving the rightmost occurrence of a var name precedence.
```
jq --debug-dump-disasm '. as [$a,[$a]] | $a'
0000 TOP
0001 DUP
0002 DUP
0003 SUBEXP_BEGIN
0004 LOADK 1
0006 SUBEXP_END
0007 INDEX
0008 DUP
0009 SUBEXP_BEGIN
0010 LOADK 0
0012 SUBEXP_END
0013 INDEX
0014 STOREV $a:0
0017 POP
0018 DUP
0019 SUBEXP_BEGIN
0020 LOADK 0
0022 SUBEXP_END
0023 INDEX
0024 STOREV $a:1
0027 POP
0028 LOADV $a:0
0031 RET
```

The new code (using your recommendations above, which work great, btw) does the same thing (because we use the old behavior when possible)... until we add alternation to the mix:
```
./jq --debug-dump-disasm '. as [$a,[$a]] ?// $alt | $a'
0000 TOP
0001 DUP # I feel like so many jq programs start with `TOP, DUP`.  Not all of them, though.
# binding prologue
0002 DUP
0003 LOADK null
0005 STOREV $a:0
0008 DUP
0009 LOADK null
0011 STOREV $alt:1
# [$a, [$a]] matcher
0014 FORK_OPT 0038
0016 DUP
0017 PUSHK_UNDER 1
0019 INDEX
0020 DUP
0021 PUSHK_UNDER 0
0023 INDEX
0024 STOREV $a:0
0027 POP
0028 DUP
0029 PUSHK_UNDER 0
0031 INDEX
0032 STOREV $a:0
0035 POP
0036 JUMP 0043
# $alt matcher
0038 POP
0039 DUP
0040 STOREV $alt:1
# body
0043 LOADV $a:0
0046 RET
```

This gives us back the leftmost(!) occurrence of the varname.  I don't mind the change, as this is clearly undefined behavior.

However.

This gave me an idea.
```
./jq -n '[{},"a"] as [[$b],$a] ?// $fallback | {$a, $b, $fallback}'
{
  "a": "a",
  "b": null,
  "fallback": null
}
```
Well that's concerning... Let's check something...
```
./jq -n '"Uh-Oh" | [{},"a"] as [[$b],$a] ?// [$first,$second] | {$a, $b, $first, $second}'
{
  "a": "a",
  "b": null,
  "fallback": "Uh-Oh"
}
```
... Oh dear.

There are two issues here.
One's a lack of sufficient `DUP`ing, which I can resolve (Which is why `"Uh-Oh"` showed up instead of `[{},"a"]`).
The other is a bit more insidious... Since all of the matchers share the prologue, if we fail inside of one matcher, the variables retain that partial state for the remaining matchers.

Thoughts on the second issue?  I don't think leaving previous matcher state laying around is acceptable.

nicowilliams:
(I too had noticed that the old code generated two vars for varnames that occurred more than one in each matcher.  I also agree that which of the first or last -lexically- binding wins is something we're free to pick.)

nicowilliams:
I think, ultimately, that we can't really be greedy in making bindings for partial matches, and so the results will seem rather inconsistent to any users who aren't familiar with how this works inside, and who aren't thinking about that when they attempt to take advantage of this.

Now, we do have the option to document partial matching as undefined behavior, so maybe we can say that after you add the missing `DUP` you're done.  However, I do feel queasy about it.  How hard would it be to generate additional `STOREV` prefixes for the non-first matchers?

wtlangford:
Should be fairly simple. I'll check into it. I'd rather the extra cost for
less surprising behavior.

On Sun, Feb 19, 2017, 22:08 Nico Williams <notifications@github.com> wrote:

I think, ultimately, that we can't really be greedy in making bindings for
partial matches, and so the results will seem rather inconsistent to any
users who aren't familiar with how this works inside, and who aren't
thinking about that when they attempt to take advantage of this.

Now, we do have the option to document partial matching as undefined
behavior, so maybe we can say that after you add the missing DUP you're
done. However, I do feel queasy about it. How hard would it be to generate
additional STOREV prefixes for the non-first matchers?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/stedolan/jq/pull/1343#issuecomment-280977706>, or mute
the thread

nicowilliams:
I agree.  It's interesting that the code that yesterday I thought was less optimal (and probably still is) got this right naturally.

There... may be another way though, one that might be even be elegant.  We could have a variant of `STOREV` that either restores the previous value, or, more easily, resets to `null`, on `BACKTRACK` or maybe only on error.  Then if a pattern only partially matches it will reset the match vars to `null` on its way back to the `FORK_OPT`!

wtlangford:
Hm, I actually quite like this implementation.  I threw it together pretty quickly (the `null` variant, at least) and got pleasant results out of it, with no increase in bytecode. 😄 

I'm not sure what the semantic difference between doing it on backtrack vs rewind would be, but either way, what would you name such an op?

wtlangford:
Pushed a version that has the opcode you suggested.  I haven't run it through valgrind yet (or written proper tests), but it looks correct.  Play with it when you get a chance, and see how you like it.
I named that op `STOREV_BT_N`, but I think that's a terrible name.  Feel free to (read: please do) come up with a better one.

wtlangford:
Also, have a debug flag I've been playing with locally that enabled a more detailed debug trace output.

nicowilliams:
I'll look tomorrow.  Maybe `STOREVR` or `STOREVN`?  ('R' for reversible, though it wouldn't quite be, or 'N' for "null".)

nicowilliams:
It loos very good to me.  Still reviewing a bit, but so far no comments.

tests/jq.test
```diff
@@ -701,6 +701,11 @@ reduce range(5) as $n ([]; select($n%2 == 1) | . + [$n])
 null
 [1,3]
 
+# This, while useless, should still compile.
```
nicowilliams: Let's add a few more tests.

nicowilliams:
This part of the `ObjPat rule in `master` is broken:

```
825 IDENT ':' Pattern {
826   $$ = gen_object_matcher(gen_const($1), $3);
827 } |
```

It causes jq to think that `$1` is a function.

nicowilliams:
I've added tests.

nicowilliams:
@wtlangford Are you waiting on me?  I guess I need to finish looking at this.  It LGTM so far though, and I drove it a bit.

wtlangford:
I rebased this off of `master`, and squashed two changes in while I was there.
1. Rename the new opcode to `STOREVN` instead of `STOREV_BT_N`.
2. I only committed half of the work for the new debug flag (I forgot the part in main.c).  That's been added now.

I didn't re-run the tests locally, so I'm waiting on travis to confirm that the rebase didn't break them.

Otherwise, if you've banged on this enough, it's probably alright.

nicowilliams:
OK, so I'll see if I still have energy today, if not tomorrow.  Thanks for finishing the detailed trace option.

nicowilliams:
Can you also include this change, which I hope will make Appveyor happy:

```
diff --git a/appveyor.yml b/appveyor.yml
index 519c945..d068dbf 100644
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,7 @@ environment:

 clone_script:
   - bash -lc "git clone -q --branch=$APPVEYOR_REPO_BRANCH https://github.com/${APPVEYOR_REPO_NAME}.git $APPVEYOR_BUILD_FOLDER"
-  - bash -lc "cd $APPVEYOR_BUILD_FOLDER && git checkout -qf $APPVEYOR_REPO_COMMIT"
+  - bash -lc "cd $APPVEYOR_BUILD_FOLDER && git checkout -qf $APPVEYOR_REPO_BRANCH"
   - bash -lc "cd $APPVEYOR_BUILD_FOLDER && git submodule update --init --recursive"

 install:
```
?

nicowilliams:
Running builds, make check.  Will do a light re-review in a bit.

nicowilliams:
I'll have to finish tomorrow.

nicowilliams:
Rebased and squashed one commit as b142d48..7935d30.

PR #1346, opened 2017-02-20
Add $ARGS, --args, --jsonargs (fix #1345)

### LINKED ISSUE #1345: ER: programmatic way to test whether a $variable has been set
Currently there does not appear to be a way to test programmatically whether a "dollar-variable" has been set, e.g.

    $ jq -n 'try $a catch null'
    jq: error: a/0 is not defined at <top-level>, line 1:
    try $a catch null    
    jq: 1 compile error

Similarly for `$a?`

This is particularly problematic if one wants to allow, but not require, a variable to be specified on the command line (e.g. using the `--arg <var> <value>` option).

nicowilliams:
This PR adds:

 - a builtin `$ARGS` binding -- an object with two keys:
    - `named` -> object of arguments specified with `--arg` and such
    - `positional` -> array of positional arguments (when `--args` or `--jsonargs` are used)
 - `--args` CLI argument: remaining arguments are positional string arguments, not files
 - `--jsonargs` CLI argument: remaining arguments are positional JSON arguments, not files

```
./jq --arg foo bar -n --jsonargs '$foo, $ARGS' -- '{}' '[]' 1
"bar"
{
  "positional": [
    {},
    [],
    1
  ],
  "named": {
    "foo": "bar"
  }
}
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -41,7 +41,7 @@ static const char* progname;
  * For a longer help message we could use a better option parsing
  * strategy, one that lets stack options.
  */
-static void usage(int code) {
+static void usage(int code, int keep_it_short) {
   FILE *f = stderr;
 
   if (code == 0)
@@ -49,7 +49,9 @@ static void usage(int code) {
 
   int ret = fprintf(f,
     "jq - commandline JSON processor [version %s]\n"
-    "\nUsage:\n\n\t%s [options] <jq filter> [file...]\n\n"
+    "\nUsage:\t%s [options] <jq filter> [file...]\n"
+    "\t%s [options] --args <jq filter> [strings...]\n"
+    "\t%s [options] --jsonargs <jq filter> [JSON_TEXTS...]\n\n"
     "jq is a tool for processing JSON inputs, applying the given filter to\n"
     "its JSON text inputs and producing the filter's results as JSON on\n"
     "standard output.\n\n"
@@ -59,22 +61,35 @@ static void usage(int code) {
     "For more advanced filters see the jq(1) manpage (\"man jq\")\n"
     "and/or https://stedolan.github.io/jq\n\n"
     "Example:\n\n\t$ echo '{\"foo\": 0}' | jq .\n"
-    "\t{\n\t\t\"foo\": 0\n\t}\n\n"
-    "Some of the options include:\n"
-    "  -c               compact instead of pretty-printed output;\n"
-    "  -n               use `null` as the single input value;\n"
-    "  -e               set the exit status code based on the output;\n"
-    "  -s               read (slurp) all inputs into an array; apply filter to it;\n"
-    "  -r               output raw strings, not JSON texts;\n"
-    "  -R               read raw strings, not JSON texts;\n"
-    "  -C               colorize JSON;\n"
-    "  -M               monochrome (don't colorize JSON);\n"
-    "  -S               sort keys of objects on output;\n"
-    "  --tab            use tabs for indentation;\n"
-    "  --arg a v        set variable $a to value <v>;\n"
-    "  --argjson a v    set variable $a to JSON value <v>;\n"
-    "  --slurpfile a f  set variable $a to an array of JSON texts read from <f>;\n"
-    "\nSee the manpage for more options.\n", JQ_VERSION, progname);
+    "\t{\n\t\t\"foo\": 0\n\t}\n\n",
+    JQ_VERSION, progname, progname, progname);
+  if (keep_it_short) {
+    fprintf(f,
+      "For a listing of options, use %s --help.\n",
+      progname);
+  } else {
+    (void) fprintf(f,
+      "Some of the options include:\n"
+      "  -c               compact instead of pretty-printed output;\n"
+      "  -n               use `null` as the single input value;\n"
+      "  -e               set the exit status code based on the output;\n"
+      "  -s               read (slurp) all inputs into an array; apply filter to it;\n"
+      "  -r               output raw strings, not JSON texts;\n"
+      "  -R               read raw strings, not JSON texts;\n"
+      "  -C               colorize JSON;\n"
+      "  -M               monochrome (don't colorize JSON);\n"
+      "  -S               sort keys of objects on output;\n"
+      "  --tab            use tabs for indentation;\n"
+      "  --arg a v        set variable $a to value <v>;\n"
+      "  --argjson a v    set variable $a to JSON value <v>;\n"
+      "  --slurpfile a f  set variable $a to an array of JSON texts read from <f>;\n"
+      "  --args           remaining arguments are string arguments, not files;\n"
+      "  --jsonargs       remaining arguments are JSON arguments, not files;\n"
+      "  --               terminates argument processing;\n\n"
+      "Named arguments are also available as $ARGS.named[], while\n"
+      "positional arguments are available as $ARGS.positional[].\n"
+      "\nSee the manpage for more options.\n");
+  }
   exit((ret < 0 && code == 0) ? 2 : code);
 }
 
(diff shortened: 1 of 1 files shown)


nicowilliams:
Currently `tests/shtest` fails with a strange failure.

nicowilliams:
Ahh, `tests/shtest`'s check for constant folding needs to be updated because now we get two more instructions every time:

```
0002 LOADK {"positional":[],"named":{}}
0004 STOREV $ARGS:0
```

nicowilliams:
`tests/shtest` fixed.

src/main.c
```diff
@@ -394,36 +429,34 @@ int main(int argc, char* argv[]) {
...
+          jv data = jv_load_file(argv[i+2], 0);
+          if (!jv_is_valid(data)) {
+            data = jv_invalid_get_msg(data);
+            fprintf(stderr, "%s: Bad JSON in --%s %s %s: %s\n", progname, which,
+                    argv[i+1], argv[i+2], jv_string_value(data));
+            jv_free(data);
+            ret = 2;
+            goto out;
+          }
+          if (isoption(argv[i], 0, "argfile", &short_opts) &&
```
wtlangford: Can't we just check against `which` here?

src/main.c
```diff
@@ -508,11 +541,25 @@ int main(int argc, char* argv[]) {
       goto out;
     }
     jq_set_attr(jq, jv_string("PROGRAM_ORIGIN"), jq_realpath(jv_string(dirname(program_origin))));
+    if (jv_object_has(jv_copy(program_arguments), jv_string("ARGS"))) {
+      jv_free(ARGS);
```
wtlangford: So we don't override their `$ARGS` with ours?

src/main.c
```diff
@@ -508,11 +541,25 @@ int main(int argc, char* argv[]) {
...
+    } else {
+      ARGS = JV_OBJECT(jv_string("positional"), ARGS,
+                       jv_string("named"), jv_copy(program_arguments));
+      program_arguments = jv_object_set(program_arguments, jv_string("ARGS"), jv_copy(ARGS));
```
wtlangford: I think you just put `program_arguments` inside itself as `program_arguments.ARGS.named`...
I'm not sure why, either?

src/main.c
```diff
@@ -335,53 +365,58 @@ int main(int argc, char* argv[]) {
...
-        if (!short_opts) continue;
+        continue;
       }
       if (isoption(argv[i], 'e', "exit-status", &short_opts)) {
         options |= EXIT_STATUS;
         if (!short_opts) continue;
       }
       // FIXME: For --arg* we should check that the varname is acceptable
+      if (isoption(argv[i], 0, "args", &short_opts)) {
+        further_args_are_strings = 1;
```
nicowilliams: Here we should reset `urther_args_are_json`.

src/main.c
```diff
@@ -335,53 +365,58 @@ int main(int argc, char* argv[]) {
...
+        continue;
+      }
+      if (isoption(argv[i], 0, "jsonargs", &short_opts)) {
+        further_args_are_json = 1;
```
nicowilliams: And here we should reset `further_args_are_strings`.  The idea is that one could intermingle `--args` and `--jsonargs`.

src/main.c
nicowilliams: Yes, but I only changed indentation here.

src/main.c
nicowilliams: I checked what happens when you repeat an arg: in `master` the first one wins, so I decided I should switch to using an object and not set when a key already existed.  As to `ARGS`, I figured "if the user really wants a different `ARGS`, why not?", but I'm actually more inclined to not let them.

src/main.c
nicowilliams: I want to make the named arguments available as an object, so the jq program can check whether a named argument was set (see #1345), but also I wanted to still provide bindings for named arguments.  So I do this.  Note that because `jv` types are COW, there's no circular structure here ;)

src/execute.c
```diff
@@ -1079,11 +1079,31 @@ int jq_compile_args(jq_state *jq, const char* str, jv args) {
...
+      jv_array_foreach(args, i, arg) {
+        jv name = jv_object_get(jv_copy(arg), jv_string("name"));
+        jv value = jv_object_get(arg, jv_string("value"));
+        program = gen_var_binding(gen_const(value), jv_string_value(name), program);
+        jv_free(name);
+      }
+    } else if (jv_get_kind(args) == JV_KIND_OBJECT) {
+      jv_object_foreach(args, name, value) {
+        program = gen_var_binding(gen_const(value), jv_string_value(name), program);
+        jv_free(name);
```
nicowilliams: `jv_object_foreach()` does not free the key nor the value.  Ditto `jv_array_foreach()`.  You're expected to consume them.  `gen_const()` takes ownership, which is why I don't `jv_free(value)`, but `jv_string_value()` does not, which is why I do `jv_free(name)`.  FWIW, I did use valgrind.

src/main.c
```diff
@@ -59,22 +61,35 @@ static void usage(int code) {
...
-    "  --slurpfile a f  set variable $a to an array of JSON texts read from <f>;\n"
-    "\nSee the manpage for more options.\n", JQ_VERSION, progname);
+    "\t{\n\t\t\"foo\": 0\n\t}\n\n",
+    JQ_VERSION, progname, progname, progname);
+  if (keep_it_short) {
+    fprintf(f,
+      "For a listing of options, use %s --help.\n",
+      progname);
+  } else {
+    (void) fprintf(f,
```
nicowilliams: Oh, that's just habit from my days working on Solaris, where the C-style guidelines required casting ignored return values to (void).  IIRC some lints actually do the right thing with them  (Whereas GCC will not let you quiet warnings about ignoring important return values this way.  Always a reason to spit curses its way.)

src/execute.c
wtlangford: Huh.  I thought it did.  Sure enough, you're right. Don't mind me, then; I definitely didn't check before I said that.

src/main.c
wtlangford: Ahh, I see it now.  Clever.

src/main.c
wtlangford: Fair enough, but I saw you add the `which`, so I thought we could improve while we were here. 😄

src/main.c
wtlangford: And here's my weird arcane C lesson for the month.

src/main.c
```diff
@@ -394,36 +431,34 @@ int main(int argc, char* argv[]) {
...
+          jv data = jv_load_file(argv[i+2], 0);
+          if (!jv_is_valid(data)) {
+            data = jv_invalid_get_msg(data);
+            fprintf(stderr, "%s: Bad JSON in --%s %s %s: %s\n", progname, which,
+                    argv[i+1], argv[i+2], jv_string_value(data));
+            jv_free(data);
+            ret = 2;
+            goto out;
+          }
+          if (strcmp(which, "argfile") == 0 &&
```
nicowilliams: Fixed the `which` thing.

tests/shtest
```diff
@@ -29,50 +29,50 @@ $VALGRIND $Q $JQ -Rne '[inputs] == ["a\u0000b", "c\u0000d", "e"]' $d/input
 ## move all of these into tests/all.test
 
 # String constant folding (addition only)
-n=`$VALGRIND $Q $JQ -n --debug-dump-disasm '"foo"' | wc -l`
-if [ $n -ne 5 ]; then
+nref=`$VALGRIND $Q $JQ -n --debug-dump-disasm '"foo"' | wc -l`
+if [ $nref -gt 8 ]; then
```
nicowilliams: It's not necessary to compare `$nref` to any particular constant.  It will be more robust to only compare the subsequent counts to `$nref`.

PR #1353, opened 2017-02-26
Add $ENV builtin variable for access to environment; also fix error msg for unbound vars

nicowilliams:
Turns out it's pretty easy to add this.

Also, if we passed in the `$ARGS` object to this function (`compile.c:expand_call_arglist()`) we could greatly speed up the binding of many `--arg*`s to the program: instead of `O(N*M)` we'd have `O(N log M)`, where `N` is the size of the program and `M` is the number of variables...  We should probably do it.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -1007,6 +1007,8 @@ static int expand_call_arglist(block* b) {
       if (!curr->bound_by) {
         if (curr->symbol[0] == '*' && curr->symbol[1] >= '1' && curr->symbol[1] <= '3' && curr->symbol[2] == '\0')
           locfile_locate(curr->locfile, curr->source, "jq: error: break used outside labeled control structure");
+        else if (curr->op == LOADV)
+          locfile_locate(curr->locfile, curr->source, "jq: error: $%s is not defined", curr->symbol);
         else
           locfile_locate(curr->locfile, curr->source, "jq: error: %s/%d is not defined", curr->symbol, block_count_actuals(curr->arglist));
         errors++;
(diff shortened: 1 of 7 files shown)


src/compile.c
```diff
@@ -997,14 +998,48 @@ static int count_cfunctions(block b) {
...
+}
 
 // Expands call instructions into a calling sequence
 static int expand_call_arglist(block* b) {
   int errors = 0;
   block ret = gen_noop();
   for (inst* curr; (curr = block_take(b));) {
     if (opcode_describe(curr->op)->flags & OP_HAS_BINDING) {
-      if (!curr->bound_by) {
+      if (!curr->bound_by && curr->op == LOADV && strcmp(curr->symbol, "ENV") == 0) {
```
nicowilliams: Here we could bind command-line argument bindings!

nicowilliams:
Well now, that's an interesting Travis build failure!  Tomorrow.

src/compile.c
```diff
@@ -997,16 +998,52 @@ static int count_cfunctions(block b) {
...
+    return r;
+  for (size_t i = 0; environ[i] != NULL; i++) {
+    const char *eq;
+
+    if ((eq = strchr(environ[i], '=')) == NULL)
+      r = jv_object_delete(r, jv_string(environ[i]));
+    else
+      r = jv_object_set(r, jv_string_sized(environ[i], eq - environ[i]), jv_string(eq + 1));
+  }
+  atexit(free_env);
```
nicowilliams: Actually, I'm unhappy with the use of atexit() for this.  I think I'd rather pass `$ARGS` all the way down here and output `$ENV` so it can be freed after compilation.

nicowilliams:
*Ugh*  I accidentally pushed this trying to push just the appveyor change.

PR #1354, opened 2017-02-26
def isempty(g)  # Testing 'isempty(empty)' at line number 1364

pkoppstein:
Documentation has been updated (but please check).

Test passes successfully.


nicowilliams:
Cherry-picked, conflicts corrected, added a test to make sure that `isempty/1` only evals its argument once.  Pushed as `4b4cf78`.  Thanks!

PR #1385, opened 2017-04-08
correct spelling of "post" in manual under Variables

tomryanx:


pkoppstein:
@tomryanx - The misspelling was intentional.  That's why "first" is also misspelled!

tomryanx:
Whoops! I'll get my coat....

:)

PR #1403, opened 2017-05-03
Use unsorted keys in walk

Argoday:
Preserve key sorting order when executing in walk, if sorted keys is desired `--sort-keys` should be used to explicitly obtain sorted keys.

This caused unexpected behavior when executing jq programs like `jq 'walk(if (type == "object") then with_entries( if .key == "target_key" then .value |= sort else . end ) else . end'`

Before fix:
```json
#Input
{
    "target_key": [
        "def",
        "abc"
    ],
    "another_key": 1
}

#Output
{
    "another_key": 1,
    "target_key": [
        "abc",
        "def"
    ]
}

#Note how key order for "another_key" changed despite it being unrelated to the walk(f) execution
```

After fix:
```json
#Input
{
    "target_key": [
        "def",
        "abc"
    ],
    "another_key": 1
}

#Output
{
    "target_key": [
        "abc",
        "def"
    ],
    "another_key": 1
}

#Note how key order for "another_key" is preserved as expected
```

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -284,7 +284,7 @@ def bsearch(target):
 def walk(f):
   . as $in
   | if type == "object" then
-      reduce keys[] as $key
+      reduce keys_unsorted[] as $key
         ( {}; . + { ($key):  ($in[$key] | walk(f)) } ) | f
   elif type == "array" then map( walk(f) ) | f
   else f


nicowilliams:
I'm all for this.

In fact, I think I'm all for making `keys` be `keys_unsorted`.  If you want them sorted, use `keys|sort`, no?

Argoday:
I'm not familiar with jq's backwards compatibility policy, but that would be a potentially breaking change if anyone is depending on the sorting behavior of base `keys` ... otherwise I agree that `keys`'s natural ordering should be the input order unless explicitly requested to be sorted via a `keys_sorted` call.

wtlangford:
I'm not in love with the break in backwards compatibility, especially since the manual does state that `keys` codepoint-sorts the result.  If it were an implementation detail, I'd be fine with changing it...

A couple notes:
- `--sort-keys` is only relevant for output purposes (its a flag to the jq executable, not the jq language itself), so it wouldn't have any actual effect on the mid-program behavior of `keys`.
- I do like the change to not sort keys when `walk`ing from an optimization standpoint, but I feel I should ask why it was important to your use case to preserve key order?

carecki:
Any plans to merge it in near future?

As to the usecases, having this option would open possibility to modify single values in existing files, without rewriting whole files, i.e. changing every line in them. Especially useful when such files are under version control.

Argoday:
@carecki , exactly - my use case was to rewrite single keys in files under git version control and I didn't want to rewrite the whole file.  It also helps keep key order that makes sense to humans (instead of alphabetical) while managing inner keys.

wtlangford:
@carecki, @Argoday - those are... quite good reasons, actually.  You've convinced me.

That said, I'm not sure I want to break backwards compatibility, especially since `keys` is well-documented as being sorted.

I'm considering a `walk_unsorted` or similar, that would provide the behavior you want without breaking any previous code.
```
def _walk(keyfn; f):
  . as $in
  | if type == "object" then
      reduce keyfn[] as $key
        ( {}; . + { ($key):  ($in[$key] | _walk(keyfn; f)) } ) | f
  elif type == "array" then map( _walk(keyfn; f) ) | f
  else f
  end;
def walk(f): _walk(keys; f);
def walk_unsorted(f): _walk(keys_unsorted; f);
```
`_walk` is just to prevent having duplicate code for the actual implementation.

My only dislike of this approach is that it locks us into the "`keys/0` is sorted" path.  But I don't see a good way to break that compatibility right now.

@nicowilliams thoughts on this?

wtlangford:
Actually...
We don't need my suggestion.  `walk/1` makes no promises about its walk order, and we can resolve the `keys/0` issue separately.

Merging.

Issue #1409, opened 2017-05-13
Parsing of milliseconds in dates


juliangamble:
Here the author writes:
> Would it also parse ISO timestamps like `2015-03-05T19:40:53.324Z`? Eg. what you get from javascript `new Date().toISOString()`.

https://github.com/stedolan/jq/issues/364#issuecomment-77442886

What we need is 

    strptime("%Y-%m-%dT%H:%M:%S.%fZ")

similar to how python does it.

c0b:
is progress in https://github.com/stedolan/jq/pull/1413#issuecomment-351315539 kind of abandoned?

Alanscut:
hi @juliangamble 
jq-1.6 had supports strptime usage.you can try it
```
root@oss-001:test_jq#  echo '"2015-03-05T23:51:47Z"' | jq 'strptime("%Y-%m-%dT%H:%M:%SZ")'
[
  2015,
  2,
  5,
  23,
  51,
  47,
  4,
  63
]

```
You can see detailed usage in the manual or jq-1.6 manual.yml.
https://github.com/stedolan/jq/blob/master/docs/content/manual/v1.6/manual.yml

itchyny:
I think this is an issue about `%f` support which jq-1.6 does not support.
```
 % echo '"2015-03-05T23:51:47.487Z"' | jq 'strptime("%Y-%m-%dT%H:%M:%S.%fZ")'
jq: error (at <stdin>:1): date "2015-03-05T23:51:47.487Z" does not match format "%Y-%m-%dT%H:%M:%S.%fZ"
 % python -c 'from datetime import datetime; print(datetime.strptime("2015-03-05T23:51:47.487Z", "%Y-%m-%dT%H:%M:%S.%fZ"))'
2015-03-05 23:51:47.487000
```

ecojan:
Struggled today with this, problem is that the unix "strptime" doesn't support milliseconds. As such, I switched to using a regex replace
```
.last_updated |= sub("(?<time>.*)\\..*Z"; "\(.time)Z")
```
As such, will transform this "2015-03-05T23:51:47.487Z" to "2015-03-05T23:51:47Z"

janosmiko:
When you want to format a timestamp which contains offset:

```
sub("(?<time>.*)\\.[\\d]{3}(?<tz>.*)"; "\(.time)\(.tz)")
```

This will transform the following:
`2015-03-05T23:51:47.487Z` to `2015-03-05T23:51:47Z`

[...]

PR #1413, opened 2017-05-19
Add support for milliseconds formatted as %f in strptime(fmt).

reidwagner:
Addresses #1409.

The milliseconds should be added at the end of the fmt string. They're chopped off of the original fmt string which is passed to strptime(3). As before, if the format does not match, e.g. milliseconds in date but not fmt string, a jv error is returned. 

I also moved some of the cleanup in case of an error to a goto label to reduce redundancy. 

Please let me know if I can change or improve anything!


src/builtin.c
```diff
@@ -1219,24 +1219,49 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
-  if (end == NULL || (*end != '\0' && !isspace(*end))) {
-    jv e = jv_invalid_with_msg(jv_string_fmt("date \"%s\" does not match format \"%s\"", input, fmt));
-    jv_free(a);
-    jv_free(b);
-    return e;
+  char *inputdup = strdup(input);
+  char *fmtdup = strdup(fmt);
+
+  int ms = -1;
+  char *pf = strstr(fmtdup, "%f");
```
nicowilliams: Watch out! `%` can be escaped with a preceding (unescaped) `%`.  You have to step over the format string character (well, byte) by character (byte).

nicowilliams:
I wonder if we may not have to implement `strptime()` from scratch, rather than depend on the OS `strptime()`.  That may be advisable as to all system date functions, since they tend to vary somewhat by OS.  That would be painful though, especially as it means having to deal with timezones, which we should absolutely not want to do!  So I guess we'll hack it somewhat like your PR.

src/builtin.c
```diff
@@ -1219,24 +1219,49 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
+  if (pf) {
+    char *msfmtloc = pf - 1;
```
nicowilliams: Buffer underrun if the format string starts with `"%f"`.

src/builtin.c
```diff
@@ -1219,24 +1219,49 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
+    char delim = *msfmtloc;
+    *msfmtloc = '\0';
+    char *msloc = strrchr(inputdup, delim);
```
nicowilliams: The assumption here is that `delim` will not be used for anything else after the `%f`...  I'm not very comfortable with this.

nicowilliams:
Note that `strptime()` returns a pointer to the first character not consumed by `strptime()`.  So I think what we want to do is something like this:

 - implement a function that supports all the format string features that `strptime()` supports and which half-parses the input string, only skipping each field as it goes
 - if and when it reaches `%f` this function should
    - parse the field
    - construct new format and input strings with the `%f` and milliseconds field removed, respectively
 - call the real `strptime()` with the possibly-new format and input strings to do the real parsing

That's a lot of work.  In the mean time users can use regular expressions to parse and fixup timestamps.

Lastly, `%f`... is that used by any implementation anywhere?  We don't want to conflict with C library implementations nor POSIX...

nicowilliams:
We're in luck as to `%f`: it's already in use in various places to mean "milliseconds" (e.g., Python 2.6 and up).

nicowilliams:
(Hit wrong button; didn't mean to close this!)

nicowilliams:
Bah, Ruby uses `%L` and `%N`.  R uses `%OSn` (which conflicts with glibc!).

What a mess.

rwagnerr:
Thanks for the fast response. I'm going to try working out the plan you proposed and address the code comments.

c0b:
have progress since May?

reidwagner:
I haven't gone any further.

nicowilliams:
I'd definitely like to complete this.  I can't promise that I will soon though :(

spier:
I just also ran into this issue with parsing milliseconds in jq.
Is there anything that I could do to help with this PR?

reidwagner:
I'm taking a fresh look at this using the plan put forth by @nicowilliams.

reidwagner:
I was hoping to find a clever way to simply parse out `%f` from the format string without paying attention to the values of each field. I.e. determining the delimiters and matching fields until `%f`. And you can get pretty close making some assumptions. But there's no two ways around something like `"Februaryr%Br%Y"`. So yeah, it would have to support every feature as you said.

Because I probably don't have the bandwidth for that I'm going to close this to encourage anyone else who wants to try.

PR #1429, opened 2017-06-15
Dockerfile build fix: oniguruma doesn't have a distclean target

pjeby:
I was trying to build a docker image from `master` (because I needed `@base64d`), and the build failed saying there was no "distclean" target for oniguruma.  Removing the corresponding command from the Dockerfile fixed this; presumably the need to delete oniguruma build files is sufficiently met later in the Dockerfile, by the already-existing ` rm -rf oniguruma-5.9.6` command.


pjeby:
Closing in favor of #1631

Issue #1438, opened 2017-07-01
abysmal IN(s)


pkoppstein:
I believe the following establishes that the current implementation of IN(s) is not only needlessly complex but also both needlessly and abysmally slow.  

Unless I'm missing something, here is an alternative (and much simpler) implementation:

    def IN0(s): . as $in | (first(s == $in) | true) // false;

And here is the test I ran (for various values of N):

     [range(0;N) | IN( range(0;N) )] | length

Results:

```
# N             u+s

#   1000   IN0 0.01
#   1000   IN  0.43

#  10000   IN0 0.136s
#  10000   IN  32.9s

# 100000   IN0 0.59s
# 100000   IN  did not complete (?)

```

Wolf-at-SO:
What part of code are you talking about?

pkoppstein:
For the record, the current definitions of `IN/1` and `IN/2` in builtin.jq are:

    def IN(s): reduce (first(select(. == s)) | true) as $v (false; if . or $v then true else false end);
    def IN(src; s): reduce (src|IN(s)) as $v (false; if . or $v then true else false end);


I would also propose simplifying the current definition of IN/2 to:

    def IN(src; s): any( src|IN(s); .);

fadado:
> I believe the following establishes that the current implementation of IN(s) is not only needlessly complex but also both needlessly and abysmally slow.
> 
> Unless I'm missing something, here is an alternative (and much simpler) implementation:
> 
> def IN0(s): . as $in | (first(s == $in) | true) // false;

You are right to consider the implementation of IN unpleasant, but  I think your `IN0` filter is wrong. Simple counter example:

```
def IN0(s): . as $in | (first(s == $in) | true) // false;
7777777 | IN0(range(88))
```

*This outputs `true`*!

My solution in the next post...

fadado:
> Unless I'm missing something, here is an alternative (and much simpler) implementation:
> 
> def IN0(s): . as $in | (first(s == $in) | true) // false;

I have my own versions for some  jq builtins, more elegant but only a bit faster (around a 10% in this case). For example, using my version for `any` (I add `all` for completeness)  I can define a very short `IN/0`:

```
def all(stream; predicate):
    isempty(stream | predicate and empty)
;

def any(stream; predicate):
    isempty(stream | predicate or empty) | not
;

def IN(s): . as $in | any(s; . == $in);
```

I like your `IN/2`. I will integrate in my library a bit changed:

```
def some(stream):
[...]

PR #1440, opened 2017-07-03
Revisions to any/2, all/2, IN/1, IN/2 as per #1438

pkoppstein:
Revisions to any/2, all/2, IN/1, IN/2 as per #1438; 
./jqtest 306 of 306 tests passed (0 malformed); no new valgrind issues

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -303,3 +303,6 @@ def JOIN($idx; stream; idx_expr; join_expr):
   stream | [., $idx[idx_expr]] | join_expr;
 def IN(s): . as $in | first( if (s == $in) then true else empty end ) // false;
 def IN(src; s): any( src|IN(s); .);
+
+# f should always evaluate to a string:
+def GROUPS_BY(stream; f): reduce stream as $x ({}; .[$x|f] += [$x] ) | .[];
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -32,18 +32,12 @@ def index($i):   indices($i) | .[0];       # TODO: optimize
 def rindex($i):  indices($i) | .[-1:][0];  # TODO: optimize
 def paths: path(recurse(if (type|. == "array" or . == "object") then .[] else empty end))|select(length > 0);
 def paths(node_filter): . as $dot|paths|select(. as $p|$dot|getpath($p)|node_filter);
-def any(generator; condition):
-        [label $out | foreach generator as $i
-                 (false;
-                  if . then break $out elif $i | condition then true else . end;
-                  if . then . else empty end)] | length == 1;
+
+def isempty(g): 0 == ((label $go | g | (1, break $go)) // 0);
+def any(stream; predicate): isempty(stream | predicate or empty) | not;
 def any(condition): any(.[]; condition);
 def any: any(.);
-def all(generator; condition):
-        [label $out | foreach generator as $i
-                 (true;
-                  if .|not then break $out elif $i | condition then . else false end;
-                  if .|not then . else empty end)] | length == 0;
+def all(stream; predicate):isempty(stream | predicate and empty);
 def all(condition): all(.[]; condition);
 def all: all(.);
 def isfinite: type == "number" and (isinfinite | not);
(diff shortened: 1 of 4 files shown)


src/builtin.jq
```diff
@@ -304,5 +301,8 @@ def JOIN($idx; stream; idx_expr):
   stream | [., $idx[idx_expr]];
 def JOIN($idx; stream; idx_expr; join_expr):
   stream | [., $idx[idx_expr]] | join_expr;
-def IN(s): reduce (first(select(. == s)) | true) as $v (false; if . or $v then true else false end);
-def IN(src; s): reduce (src|IN(s)) as $v (false; if . or $v then true else false end);
+def IN(s): . as $in | first( if (s == $in) then true else empty end ) // false;
```
nicowilliams: Nice.  This can be further simplified as:

```
def IN(s): first(select(.==s), false);
def IN(src; s): first((src|IN(s)),false);
```

src/builtin.jq
```diff
@@ -32,18 +32,12 @@ def index($i):   indices($i) | .[0];       # TODO: optimize
 def rindex($i):  indices($i) | .[-1:][0];  # TODO: optimize
 def paths: path(recurse(if (type|. == "array" or . == "object") then .[] else empty end))|select(length > 0);
 def paths(node_filter): . as $dot|paths|select(. as $p|$dot|getpath($p)|node_filter);
-def any(generator; condition):
-        [label $out | foreach generator as $i
-                 (false;
-                  if . then break $out elif $i | condition then true else . end;
-                  if . then . else empty end)] | length == 1;
+
+def isempty(g): 0 == ((label $go | g | (1, break $go)) // 0);
```
nicowilliams: `isempty/1` is already in master.  Can you rebase?

src/builtin.jq
```diff
@@ -32,18 +32,12 @@ def index($i):   indices($i) | .[0];       # TODO: optimize
...
+def any(stream; predicate): isempty(stream | predicate or empty) | not;
```
nicowilliams: Nice!  This is much simpler than the `any/2` in master.

src/builtin.jq
```diff
@@ -32,18 +32,12 @@ def index($i):   indices($i) | .[0];       # TODO: optimize
...
 def any(condition): any(.[]; condition);
 def any: any(.);
-def all(generator; condition):
-        [label $out | foreach generator as $i
-                 (true;
-                  if .|not then break $out elif $i | condition then . else false end;
-                  if .|not then . else empty end)] | length == 0;
+def all(stream; predicate):isempty(stream | predicate and empty);
```
nicowilliams: Nice simplification!

src/builtin.jq
nicowilliams: Oh, never mind, you just moved it.

src/builtin.jq
```diff
@@ -304,5 +301,8 @@ def JOIN($idx; stream; idx_expr):
   stream | [., $idx[idx_expr]];
 def JOIN($idx; stream; idx_expr; join_expr):
   stream | [., $idx[idx_expr]] | join_expr;
-def IN(s): reduce (first(select(. == s)) | true) as $v (false; if . or $v then true else false end);
-def IN(src; s): reduce (src|IN(s)) as $v (false; if . or $v then true else false end);
+def IN(s): . as $in | first( if (s == $in) then true else empty end ) // false;
+def IN(src; s): any( src|IN(s); .);
+
+# f should always evaluate to a string:
+def GROUPS_BY(stream; f): reduce stream as $x ({}; .[$x|f] += [$x] ) | .[];
```
nicowilliams: This should be all lower-case.  Though it's SQLish, it's also jq-ish.

pkoppstein: @nicowilliams - As I mentioned elsewhere, git and github have me all tied up in knots at the moment, so please make any changes you see fit.  

I used uppercase for GROUPS_BY to emphasize the distinction with the existing `group_by`.  Incidentally, is it too late to re-implement `group_by` so that it does not incur the costs of sorting?  (For example, `group_by` could be implemented using `GROUPS_BY`.)

pkoppstein:
Here is what I believe now to be a much better GROUPS_BY:

```
# f need not be string-valued
def GROUPS_BY(stream; f): 
   def unwind:
     to_entries[] | .value | to_entries[] | .value ;

   reduce stream as $x ({};
     ($x|f) as $s
     | ($s|type) as $t
     | (if $t == "string" then $s else ($s|tojson) end) as $y
     | .[$t][$y] += [$x] )
   | unwind
   ;

```
It is better because it is efficient and versatile, while not being subject to all the caveats that come with the simpler GROUPS_BY.

PR #1454, opened 2017-07-21
Example of object creation attribute shortcut

sloanlance:
A suggestion for the tutorial:

Use an attribute name without a value in the object creation to give an example of the shortcut syntax.

### DIFF
--- a/docs/content/1.tutorial/default.yml
+++ b/docs/content/1.tutorial/default.yml
@@ -164,11 +164,12 @@ body:
       There's a lot of info we don't care about there, so we'll restrict it down
       to the most interesting fields.
 
-  - command: "jq '.[0] | {message: .commit.message, name: .commit.committer.name}'"
+  - command: "jq '.[0] | {message: .commit.message, name: .commit.committer.name, sha}'"
     result: |
       {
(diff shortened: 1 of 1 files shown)


sloanlance:
Tell me if I need to do anything about the Travis CI failure.

docs/content/1.tutorial/default.yml
```diff
@@ -164,19 +164,22 @@ body:
...
       to the most interesting fields.
 
-  - command: "jq '.[0] | {message: .commit.message, name: .commit.committer.name}'"
+  - command: "jq '.[0] | {message: .commit.message, name: .commit.committer.name, sha}'"
     result: |
       {
         "message": "Merge pull request #162 from stedolan/utf8-fixes\n\nUtf8 fixes. Closes #161",
-        "name": "Stephen Dolan"
+        "name": "Stephen Dolan",
+        "sha": d25341478381063d1c76e81b3a52e0592a7c997f
```
itchyny, line 172: This is not a valid result JSON.

itchyny (changes requested):

itchyny:
The tutorial explains very basic concepts for beginners, and adding a shortcut syntax to this page is not appropriate in my opinion. Closing the PR, but thank you.

PR #1458, opened 2017-08-03
fix portability problems, found in NetBSD-7.1/amd64

n-soda:
There are 4 portability problems in jq,

For example, it dumps core as follows on NetBSD-7.1/amd64, as described in https://gnats.netbsd.org/52460 :
```
$ echo 334 | jq 'strftime("%Y-%m-%dT%H:%M:%SZ")'
Memory fault (core dumped)
```
This is caused by missing function prototype of gmtime_r() due to inconsistent use of *_SOURCE macros, and that's fixed in df9a0963 in this pull request.

There are 3 other problems, and all of them are fixed in this pull request.

NOTE: this problem was originally found by Takehiko NOZAKI, and basic fixes are made by him and Kimihiro Nonaka.  But if there is any problem in these fixes, it's my fault since I made final modification.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,7 @@ if MAINTAINER_MODE
 BUILT_SOURCES = src/lexer.h src/lexer.c src/parser.h src/parser.c 	\
 		src/builtin.inc src/version.h
 src/lexer.c: src/lexer.l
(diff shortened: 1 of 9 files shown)


emanuele6:
This PR is now outdated; it addresses problems that were already fixed when jq 1.6 was released in 2018.

Closing.

PR #1460, opened 2017-08-08
Change tutorial to highlight filter part of query versus output part of query

stagrlee:
Intent of this change is to let the reader who is totally unfamiliar with the JQ syntax see how the input JSON is filtered during a query by the right side of the of the syntax and the output is specified by the left side.  In the current tutorial, the search filter is

`.commit.message`

and the output is 

`message:`

The word message is used in both the search side and the output side and might confuse a new user to try to search JSON input with the left side.  (It has happened to me a few times after leaving JQ coding for a few months).  With this patch, search remains the same but "message" turns into "comment" and "committer" turns into "contributor" showing the reader that the left side is clearly output and the right side is the search.

The configure.ac change is to keep autoconf syntax highlighters from getting confused.


stagrlee:
Travis-ci error is the same error seen on master so I believe the problem is upstream.

itchyny:
Closing not just because this looks stale, but wrong usage of terms. Both comments and contributors are different concepts from the extracted fields in the examples.

PR #1512, opened 2017-10-22
remove unused conditional branch

yoichi:
It has no effect after the change c4524da.

Bug reported by the clang static analyzer.

Description: Value stored to 'value' is never read
jq/src/util.c
Line: 439


nicowilliams:
Thanks!

PR #1517, opened 2017-10-27
Fix build with linux on travis-ci

yoichi:
I've fixed build error with linux on travis-ci

* use json 1.8.5 to fix error with ruby 2.4
* recover the 1st test of tests/optional.test (error exits of jq cause valgrind error)

Test with macos is fails since it is broken as I commented in 
https://github.com/stedolan/jq/issues/1415


itchyny:
We no longer use TravisCI and switched to GitHub Actions, closing this issue.

PR #1519, opened 2017-10-28
Added rawfile

davidfetter:
In passing, clean remnants of argfile from slurpfile docs.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -83,6 +83,7 @@ static void usage(int code, int keep_it_short) {
       "  --arg a v        set variable $a to value <v>;\n"
       "  --argjson a v    set variable $a to JSON value <v>;\n"
       "  --slurpfile a f  set variable $a to an array of JSON texts read from <f>;\n"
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -443,18 +444,22 @@ int main(int argc, char* argv[]) {
         continue;
       }
       if (isoption(argv[i], 0, "argfile", &short_opts) ||
+          isoption(argv[i], 0, "rawfile", &short_opts) ||
           isoption(argv[i], 0, "slurpfile", &short_opts)) {
+        int raw = isoption(argv[i], 0, "rawfile", &short_opts);
         const char *which;
         if (isoption(argv[i], 0, "argfile", &short_opts))
           which = "argfile";
+        else if(isoption(argv[i], 0, "rawfile", &short_opts))
```
nicowilliams: You don't have to repeat this `isoption()` call here.  You can just `if (raw) ...`.  I'll fix this and push.

davidfetter: Thanks!

nicowilliams:
Fixed with b4742c1.

PR #1522, opened 2017-10-31
Add ruby config dotfiles for docs

AndrewSpeed:
Add
 - `.ruby-gemset` to isolate jq docs dependencies from other ruby dependencies installed for other projects, as recommended by the [rvm best practices guide](https://rvm.io/rvm/best-practices)
 - `.ruby-version`  to automatically have rvm switch to the desired version of ruby automatically when `cd`ing into the directory.


itchyny:
We removed the dependency on Ruby at a4756fde3fb059d93259442fd2e469a1831a32c0 on Feb 27, 2019, so closing.

PR #1523, opened 2017-10-31
Update maruku and rake dependencies to latest versions

AndrewSpeed:
Update the `maruku` and `rake` dependencies to their latest versions.

I've tested that the site builds and examples appear as expected, but if there are additional things you'd like me to check before this is merged, please let me know :slightly_smiling_face:


itchyny:
We removed the dependency on Ruby at a4756fde3fb059d93259442fd2e469a1831a32c0 on Feb 27, 2019, so closing.

PR #1537, opened 2017-11-23
Update manual.yml

yonatan-py:
I think there was a small mistake :)

`echo "[{\"x\": 1},{\"x\":2}]" | jq "[.[] | x]"` don't work
but `echo "[{\"x\": 1},{\"x\":2}]" | jq "[.[] | .x]"` does

### DIFF
--- a/docs/content/3.manual/v1.5/manual.yml
+++ b/docs/content/3.manual/v1.5/manual.yml
@@ -872,7 +872,7 @@ sections:
           Similarly, `map_values(x)` will run that filter for each element,
           but it will return an object when an object is passed.
 
-          `map(x)` is equivalent to `[.[] | x]`. In fact, this is how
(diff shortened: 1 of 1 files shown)


pkoppstein (changes requested):
The update is partly wrong. The .x in the update should be x, as in the original.

yonatan-py:
@pkoppstein just tried my example again, they seem to be valid.
:dango:

pkoppstein:
I am referring to the sentence: map(x)` is equivalent to `[.[] | .x]`

It doesn't matter what you tried, the above is absolutely incorrect.  As should be self evident, it should read:

     `map(x)` is equivalent to `[.[] | x]`

EugenGez:
`map(x) is defined as [.[] | x]`

The `x` is a placeholder...

itchyny:
Yes, as commented above, the manual is correct.

PR #1542, opened 2017-11-29
Keep object keys in parsing order in `tostream` output

### LINKED ISSUE #1541: Order differs in output of --stream vs tostream
Using latest version from `master`:

```
$ ./jq --version
jq-1.5rc2-250-g239278f
```

The `--stream` parameter preserves the order of properties in the output:

```
$ echo '{"a":"one","B":"two"}' | ./jq --stream '.'
[
  [
    "a"
  ],
  "one"
]
[
  [
    "B"
  ],
  "two"
]
[
  [
    "B"
  ]
]
```

while the `tostream` builtin doesn't:

```
$ echo '{"a":"one","B":"two"}' | ./jq '. | tostream'
[
  [
    "B"
  ],
  "two"
]
[
  [
    "a"
  ],
  "one"
]
[
  [
    "a"
  ]
]
```

In the context of streaming, the parsing order should be reflected in the output.

eric-brechemier:
As noted by @nicowilliams, `tostream` used `keys`, which sorts the keys in alphabetical order, instead of `keys_unsorted`, which preserves the parsing order.

After the change, the behavior of `tostream` matches the behavior of `--stream`:

```
$ echo '{"a":"one","B":"two"}' | ./jq '. | tostream'
[
  [
    "a"
  ],
  "one"
]
[
  [
    "B"
  ],
  "two"
]
[
  [
    "B"
  ]
]
```

matches

```
 $ echo '{"a":"one","B":"two"}' | ./jq --stream '.'
[
  [
    "a"
  ],
  "one"
]
[
  [
    "B"
  ],
  "two"
]
[
  [
    "B"
  ]
]
```

which fixes #1541.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -24,6 +24,7 @@ David R. MacIver     <david@drmaciver.com>            - bug fixes
 David Tolnay         <dtolnay@gmail.com>              - destructuring, build improvements
 Doug Luce            <doug@github.con.com>            - build
 Eiichi Sato          <sato.eiichi@gmail.com>
+Eric Bréchemier      <eric@egull.co>                  - bug fix
 Filippo Giunchedi    <fgiunchedi@gmail.com>           - bug fixes
 Filippo Valsorda     <filippo.valsorda@gmail.com>     - recursive object merge (`*`)
 Hanfei Shen          <qqshfox@gmail.com>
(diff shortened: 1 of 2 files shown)


nicowilliams:
Thanks!

PR #1549, opened 2017-12-07
Use an unreachable commit ID instead of a tag

pjw91:
For git v1.7.1 (eg. on CentOS 6), a shallow clone achieved by `git clone --depth 1` clones not only the latest commit on master branch but also all branches and all tags, which makes the tag `jq-1.0` available.

As a result, the shallow-clone-detecting command passes and the `git describe` is incorrectly called.

This patch replaces the tag jq-1.0 with eca89ac, which is the initial commit.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 m4_define([jq_version],
-          m4_esyscmd_s([(git rev-parse --verify -q jq-1.0 > /dev/null &&
+          m4_esyscmd_s([(git rev-parse --verify -q eca89ac > /dev/null &&
                         (git describe --tags --dirty --match 'jq-*'|sed 's/^jq-//')) ||
                         echo `git rev-parse --abbrev-ref HEAD`-`git describe --always --dirty`])))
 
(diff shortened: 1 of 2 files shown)


itchyny:
As of git v1.7.10 (April 2012), using `--depth=1` implies `--single-branch` flag so tags are not fetched (ref: [RelNotes/1.7.10.txt#L80-L82](https://github.com/git/git/blob/v1.7.10/Documentation/RelNotes/1.7.10.txt#L80-L82), https://lkml.org/lkml/2012/3/28/418). So the version shows `master-[commit sha]` right after a shallow clone of the repository. But doing `git fetch --tags` after the shallow clone, `autoreconf` fails with `fatal: No tags can describe` error. In this situation, `git rev-parse --verify -q eca89ac` resolves the full commit sha (not sure why) so the proposed change does not fix the error. We can use `git describe --tags --match 'jq-*'` (or just `git describe --tags` for simplicity) for this check IMO. Here are the related commits: f7bf9a8e96801d1aaef6148929098c1c92024378, 687812320224e4e5de0080fdb708204891ddec84.

PR #1558, opened 2017-12-15
Fix a try-catch example

zogwarg:
Fixing a bad syntax in a try-catch example

It was probably based on the inputs builtin definition,
The if statement was messed up and the semi-colon actually belongs to the function definition.


itchyny:
This resolves #1777, but make sure to edit the YAML file in the `master` branch, not the HTML file in `gh-pages` branch directly.

zogwarg:
Thanks!

PR #1574, opened 2018-01-04
fix FreeBSD install instructions

chvostek:
Current instructions for FreeBSD are incorrect. This patch includes the two standard methods (source and binary) for installation in FreeBSD, along with links to documentation for each method.


nicowilliams:
Thanks!

PR #1596, opened 2018-02-08
Build and deploy to GH statically linked binaries for various arches and test against musl libc

jirutka:
1. Refactored `.travis.yml` to use Travis [Build Stages](https://docs.travis-ci.com/user/build-stages/) (this allows to define jobs with completely different configuration).
2. Added job that builds and tests jq with gcc 6.4 on Alpine Linux v3.7 (inside chroot), i.e. verifies that it works with musl libc.
3. Added jobs that build fully statically linked (with musl libc) jq on Alpine Linux v3.7 aarch64, armhf, and ppc64le architectures emulated using QEMU user-mode emulator and binfmt. Emulation is quite slow, so it takes more time to build.
4. **Configured automatic deployment of statically linked jq built for x86_64, x86, aarch64, armhf, and ppc64le to GitHub Releases.**

You just need to replace `secret:  # TODO…` with your **encrypted** [personal GitHub token](https://github.com/settings/tokens) (use gem [travis](https://rubygems.org/gems/travis) to encrypt it: `travis encrypt YOUR_TOKEN`). I recommend to create a separate GitHub account for this, because personal access tokens have too wide scope. Unfortunately Travis does not support any better authentication method for deployment to GitHub Releases.

Once you set up it and tag new release, Travis should automatically build and deploy a tar.gz archive with statically linked binaries for each architecture into Releases (it looks like [this](https://github.com/alpinelinux/apk-tools/releases)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,6 +111,34 @@ jobs:
         -   <<: *test-osx
             compiler: clang
 
+
+        # Build with gcc and run tests on Alpine Linux v3.7 (inside chroot).
+        # Note: Alpine uses musl libc.
+        -   &test-alpine
+            stage: test
+            os: linux
+            language: minimal
+            compiler: gcc
+            sudo: true
+
+            before_install:
+                - "wget 'https://raw.githubusercontent.com/alpinelinux/alpine-chroot-install/v0.7.0/alpine-chroot-install' \
+                    && echo '090d323d887ef3a2fd4e752428553f22a52b87bb  alpine-chroot-install' | sha1sum -c || travis_terminate 1"
+                - alpine() { /alpine/enter-chroot -u "$USER" "$@"; }
+
+            install:
+                - sudo sh alpine-chroot-install -b v3.7 -a "$ARCH"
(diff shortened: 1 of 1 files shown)


jirutka:
@stedolan, this is now ready for review&merge.

nicowilliams:
Merged.  Thanks!  This is going to be very useful!

jirutka:
You’re welcome! But you [forgot to set the deploy key](https://github.com/stedolan/jq/blob/2ce1e323df8291e4d2fdb32e4bb85e525309c987/.travis.yml#L179) (see the PR comment), so it will not deploy binaries into GitHub Releases.

nicowilliams:
Ah, yes, sorry.  That actually requires @stedolan to get involved.

SerialVelocity:
@nicowilliams @stedolan Was the deploy key set up? jq 1.6 seems to be missing the statically built binaries for aarch64

jirutka:
Hmm, Travis CI seemed like a good idea four years ago… Now I should probably rewrite it for GitHub Actions.

Issue #1587, opened 2018-01-26
Can't use both the --raw-output and --ascii-output switches


eelcohn:
If both the --raw-output and the --ascii-output command line switches are used, the --ascii-output switch is ignored:
```
$ echo "{\"key1\":\"\\ud83d\\ude00\"}" | jq --raw-output '.key1'
😀
$ echo "{\"key1\":\"\\ud83d\\ude00\"}" | jq --ascii-output '.key1'
"\ud83d\ude00"
$ echo "{\"key1\":\"\\ud83d\\ude00\"}" | jq --raw-output --ascii-output '.key1'
😀
```
What I'd expect is that in the last example the output would be:
```
$ echo "{\"key1\":\"\\ud83d\\ude00\"}" | jq --raw-output --ascii-output '.key1'
\ud83d\ude00
```

rain-1:
https://github.com/stedolan/jq/pull/1611

nicowilliams:
Closed with bf88c73c20ff1642aa2566ae67df18ca5b4b63ab via #1587 by @rain-1.  Thanks!

PR #1611, opened 2018-03-01
src/main.c: fix issue #1587 by using jv_dumpf in the raw-output and a…

rain-1:
…scii-output case. retain fwrite in the raw-output case.


nicowilliams:
Thanks!

PR #1631, opened 2018-03-19
Fix Dockerfile

rmetzler:
target `distclean` did not exist in Makefile
target `dist-clean-local` works.

This should override #1429


Alanscut (approved):

nicowilliams:
Could you rebase and drop the merge commit?

PR #1643, opened 2018-04-04
Add mul builtin

andremarianiello:
Includes more cases where `*` operator succeeds.


andremarianiello:
I built and tested this using the Dockerfile, but the `make distclean` command failed with no target. I don't know for sure if the change I made is correct.

Additionally, I don't know if this is something that you actually want merged in, so I didn't update the documentation, but I can update it if necessary.

PR #1647, opened 2018-04-12
pick init file from multiple candidates and filter search path to dirs

benjaminran:
This adds a JQ_INIT_FILE_PATH jq attr which is set as the first of ~/.jqrc, ~/.jq, or ~/.jq/jqrc that exists and is a file. This gets sourced into the program as before. It also filters JQ_LIBRARY_PATH to only directories. Let me know how it looks.


aberezin:
Is there a reason this PR was not taken?

PR #1671, opened 2018-05-28
"writing output failed" with non ASCII input

ghost:
Using this file:

~~~json
$ cat example.json
{
  "alfa": "Alt 173",
  "bravo": "hello¡"
}
~~~

Gives this result:

~~~
$ jq -r .bravo example.json
hello¡
Error: writing output failed: No such file or directory
~~~

Note if you remove `-r`, it works as expected:

~~~json
$ jq .bravo example.json
"hello¡"
~~~

Looks like the problem is an `fwrite` call that was not replaced with the
portable version.


cup:
test

cup:
test

nicowilliams:
Thanks for the report!

PR #1672, opened 2018-05-30
Add scoop installation instructions

rasa:


itchyny:
This PR modifies a generated HTML file, and we have a newer PR #2188 which correctly modifies the source file, so closing. Thank you.

PR #1673, opened 2018-05-31
Add support for CRLF line endings in filters

### LINKED ISSUE #1219: Support \r as whitespace in filters
```
jq -n "$(echo -e "1\n+\n2")"
3    
jq -n "$(echo -e "1\r\n+\r\n2")"
jq: error: syntax error, unexpected INVALID_CHARACTER, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:

jq: 1 compile error
```

Is there a particular reason for this? I spent a while banging my head on this trying to call jq from a powershell script.

Edit: looking at `lexer.l`, there may also be an issue with carriage returns in comments. I would argue a comment should only end on a line feed character, and not on a solitary `\r`.

kvonbredow:
In files created in windows, the CRLF line endings in filters cause jq to crash with an "unexpected INVALID_CHARACTER" syntax error.  Since this should just be treated as any other whitespace character to aid in Windows support, I modified the lexer to ignore \r whitespace in filters.

resolves #1219


itchyny:
This issue has been resolved by #2942.

emanuele6:
No, #2942 did not add `\r` as a whitespace character, it just made it so `\r` in a comment does not end the comment, and trigger an INVALID_CHARACTER error for CR.
If you have a `\r` character in your script, the code will still not compile because `\r` is invalid; but now `\r` is at least allowed in comments.

emanuele6:
Anyhow, this patch probably has conflicts, and extra code would be needed to support `# \<cr><lf>` comment continuations. So we can keep it closed, I guess.

itchyny:
Oops, thanks for comment. That sounds a halfway support, so I'd like to see full support CRLF newlines.

PR #1697, opened 2018-07-29
fix --exit-code issues #1142 and #1139

### LINKED ISSUE #1139: jq -e returns 4 for non-empty output with select()

### LINKED ISSUE #1142: jq -e always returns 0 for empty input
`jq -e` always exits with status 0 when passed empty input. For example:

``` shell
$ echo | jq --exit-status 'false'; echo $?
0
```

Consider a use case like this one:

``` shell
call-some-json-api |
jq --exit-status '.field == "some-value"' &&
take-action
```

If the API call fails and output is only written to stderr, `take-action` will still run as though the call succeeded, and "some-value" had been set.

This may be related to issue #667, but that one has been fixed on master, and this bug is still present.

ryo1kato:
* Set default error code to -4 in main(), Fixes #1142
* fix --exit-code with more than one object in input, Fixes #1139
    - Return code 1 or 4 based on last output, not last input.

This is a rebase of PR #1140 to the latest master.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -146,6 +146,17 @@ enum {
 };
 static int options = 0;
 
+enum {
+    JQ_OK              =  0,
+    JQ_OK_NULL_KIND    = -1, /* exit 0 if --exit-status is not set*/
+    JQ_ERROR_SYSTEM    =  2,
+    JQ_ERROR_COMPILE   =  3,
+    JQ_OK_NO_OUTPUT    = -4, /* exit 0 if --exit-status is not set*/
+    JQ_ERROR_UNKNOWN   =  5,
+};
+#define jq_exit_with_status(r)  exit(abs(r))
+#define jq_exit(r)              exit( r > 0 ? r : 0 )
+
 static const char *skip_shebang(const char *p) {
   if (strncmp(p, "#!", sizeof("#!") - 1) != 0)
     return p;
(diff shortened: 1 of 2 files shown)


dongcarl:
Are there any updates on this? Would love for it to be merged as it's very much needed when using `jq` in shell scripts.

nicowilliams:
Thanks!

DrStrangepork:
This merge did not fix the problem in #1142 
```
# echo | jq -e .; echo $?
0
```

xiruizhao:
> This merge did not fix the problem in #1142
> 
> ```
> # echo | jq -e .; echo $?
> 0
> ```

because this commit is merged after the jq 1.6 release. Hopefully the maintainer will release a new version soon.

Issue #1659, opened 2018-04-27
Make fails on Ubuntu 18.04


gharris999:
`./.libs/libjq.a(builtin.o): In function `f_pow10':
/home/daadmin/StdUtils/jq/src/libm.h:253: undefined reference to `pow10'
collect2: error: ld returned 1 exit status
Makefile:984: recipe for target 'jq' failed
make[2]: *** [jq] Error 1
make[2]: Leaving directory '/home/daadmin/StdUtils/jq'
Makefile:1148: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/daadmin/StdUtils/jq'
Makefile:825: recipe for target 'all' failed
make: *** [all] Error 2
`
Is this the same issue as #1565 ??

gharris999:
Output of..
./configure --with-oniguruma=builtin --disable-maintainer-mode
..at
https://pastebin.com/irSA4qVS

doronbehar:
I'm experiencing a very similar problem here on Arch Linux,

Here is my trace:
```
make[2]: Entering directory '/var/cache/aursync/jq-git/src/jq'
  CCLD     jq
./.libs/libjq.a(util.o): In function `get_home':
util.c:(.text+0x3f2): warning: Using 'getpwuid' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
./.libs/libjq.a(builtin.o): In function `f_pow10':
builtin.c:(.text+0x50c6): undefined reference to `pow10'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:984: jq] Error 1
make[2]: Leaving directory '/var/cache/aursync/jq-git/src/jq'
make[1]: *** [Makefile:1147: all-recursive] Error 1
make[1]: Leaving directory '/var/cache/aursync/jq-git/src/jq'
make: *** [Makefile:824: all] Error 2
```

Here is the [output of `./configure --prefix=/usr --with-oniguruma=builtin --disable-maintainer-mode`](https://gist.github.com/doronbehar/bda71f564f16b506c10897c10baad13b)

marcelpaulo:
Also happening with a fresh install of Xubuntu 18.04. 

The output of `./configure --with-oniguruma=builtin` has:

```
checking for pow10... yes
checking for pow... yes
```

But `make` fails:

```
./.libs/libjq.a(builtin.o): In function `f_pow10':
[...]

PR #1702, opened 2018-08-02
Bugfix: Math function checking

alex-ozdemir:
tl;dr:
Some of our configuration machinery was subtly broken, and it broke compilation for versions of glibc >= 2.27. Now it's fixed (and simplified!).

Long Version:

We had config machinery that determined which math functions are
available in libc. If a c math function was missing on the host system,
then the corresponding jq function would be removed from the source,
enabling the build to proceed anyway. The detection machinery was broken
in a subtle way, as was shown after glibc updated to 2.27, dropping the
`pow10` function. This caused compilation to fail.

The essential problem was that we detected whether a math function was
available by compiling and linking a small program evaluating that
function on constants. However, since gcc's optimization machinery has
special knowledge of some math functions (e.g. `pow10`), it can
optimize them away, even if they don't exist in the library and are not
linkable. That is, the following example compiles and links against
glibc 2.27, even though `pow10` has been removed:
```
int main () {
  printf("%f", pow10(0.5));
  return 0;
}
```

What?!
On the other hand, this program does not link:
```

int main () {
  double f;
  printf("%f", &f);
  printf("%f", pow10(f));
  return 0;
}
```

In the first program the call to `pow10` can be optimized away as a
constant expression. This requires GCC to know about `pow10` (which it
does!), but it does not require `pow10` to be in the library (and
actually linkable).

The solution is to use autoconf's machinery for detecting function
presence, instead of our own (buggy) machinery. This has the added
benefit of simplifying the code.

The bug was reported in issue #1659


yarikoptic:
FWIW 
- would've been nice to see it merged since prevents newbee build-from-source users (e.g. to get `@base64d` not available in the release)  fail to build from source otherwise
- travis failures seems to be unrelated 
```
0.11s$ bundle install --gemfile=docs/Gemfile
Ignoring executable-hooks-1.5.0 because its extensions are not built.  Try: gem pristine executable-hooks --version 1.5.0
Ignoring gem-wrappers-1.3.2 because its extensions are not built.  Try: gem pristine gem-wrappers --version 1.3.2
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems.rb:241:in `bin_path': can't find gem bundler (>= 0.a) (Gem::GemNotFoundException)
	from /usr/local/bin/bundle:22:in `<main>'
The command "bundle install --gemfile=docs/Gemfile" failed and exited with 1 during .
Your build has been stopped.
```
so may be just a matter of restarting a build?

andrewmiller1:
newbee build-from-source user here. is this the same issue?

```text
make[2]: Entering directory '/home/andrew/Downloads/jq'
  CCLD     jq
/usr/bin/ld: ./.libs/libjq.a(builtin.o): in function `f_pow10':
/home/andrew/Downloads/jq/src/libm.h:253: undefined reference to `pow10'
collect2: error: ld returned 1 exit status
```

alexozdemir-coursera:
Yeah, it is.

You could pull my fork and build that.

You can also mess around with the config files to specifically remove `pow10`.

jpnarkinsky:
Thanks @alex-ozdemir for the branch.  That solved my problem on Ubuntu 18.04 without me having to spend a bunch of time digging in to the code.  (The 90's called, they want their skills back!)

On another subject, what would it take to get a new release of jq?  It's literally been years, and for me the base64 support is absolutely essential.  It would save a lot of unnecessary traffic and effort on the issues board to get it released so distros would just include it

Issue #1285, opened 2016-12-03
jq should detect when stdout is closed (if it's a pipe or socket) and exit


ghost:
Running:

~~~
jq | head
~~~

hangs on master. 5fe053671a768d28bc9773719cc4a9a35e72f30d is the culprit. Here
is Git bisect:

~~~
5fe053671a768d28bc9773719cc4a9a35e72f30d is the first bad commit
commit 5fe053671a768d28bc9773719cc4a9a35e72f30d
Author: Nicolas Williams <nico@cryptonector.com>
Date:   Tue Dec 1 22:16:03 2015 -0600

    Assume . if either stdin/on isatty() (fix #1028)

    And move the #define of isatty on Windows up.
~~~

svnpenn:
https://github.com/stedolan/jq/issues/1028

svnpenn:
Looks like workaround is:

~~~
jq -h | head
~~~

but this is not intuitive

nicowilliams:
*sigh*

svnpenn:
@nicowilliams development seems to have stalled? Would you be interested to add me as collaborator?

nicowilliams:
@svnpenn We certainly need more developers, yes.  What I'd like to do is review PRs and then upgrade authors to collaborator status as they need less and less hand-holding.  Send me a PR!

svnpenn:
Hm, no thanks. I have made commits to this repo already, so I had hoped I had already proven myself.

nicowilliams:
@svnpenn I can't hand out committer access.  I can only recommend new prospective committers to @stedolan.  There are two small commits from you in the repo (and maybe neglected PRs? let me know) -- nothing wrong with that, but maybe @stedolan would like to see more?  I wouldn't mind, and he's hard to get a hold of, so a bit more from you would help.  Besides, it's just access to push, not access to author.  You can author any time you like and get me (or @wtlangford) to pull your commits.  If we're unresponsive to issues, there's always e-mail.

itchyny:
This does not look like a bug. The command just waits for standard input. Waiting for comments from other maintainers.

nicowilliams:
@itchyny I agree it's not a bug.  I suppose we might could do something about this though trying to detect error conditions on `STDOUT_FILENO` when it's a pipe or a socket.

nicowilliams:
https://unix.stackexchange.com/a/736795
https://unix.stackexchange.com/a/736865

Basically, this is a common problem.  We could:

[...]

PR #1703, opened 2018-08-03
wip: feat: terminate jq immediately after the outgoing pipe closed

### LINKED ISSUE #1691: funny behaviour with pipes
I want to use jq to filter a stream/pipe of json inputs and then terminate the whole pipeline as soon as a criterion has been met.
I would like to use jq to express that criterion.

So I have the following bash code (as an example):
```bash
for ((i=0; ;i=$(((i+1))))); do echo "{\"value\":$i}"; done | jq --unbuffered -c "(select(.value==2))" | head -1
```
This does return the correct result:
```
{"value":2}
```
However, the pipeline does not terminate.
If I change the code so that the criterion is met multiple times
```bash
for ((i=0; ;i=$(((i+1)%3)))); do echo "{\"value\":$i}"; done | jq --unbuffered -c "(select(.value==2))" | head -1
```
I get the correct result and the pipeline terminates.

But then, the following snippet also creates multiple hits, returns one of them and does not terminate:
```bash
for ((i=0; ;i=$(((i+1))))); do echo "{\"value\":$i}"; done | jq --unbuffered -c "(select(.value<4))" | head -1
```

When I use cat instead of jq:
```bash
for ((i=0; ;i=$(((i+1))))); do echo "{\"value\":$i}"; done | cat | head -1
```
the pipeline terminates after printing one line.

So my suspicion is that the SIGPIPE from head somehow gets lost in jq.

Is this an issue with my invocation or is there something wrong with jq?

I have built jq today from master
```bash
git log -1
commit 90bc29c1b544c0436ec44246e180fdbb6d6384df (HEAD -> master, origin/master, origin/HEAD)
```

I have done my best to search through the open issues to find something related.
My apologies if this has already been reported.

Thanks for your help.

mumoshu:
This should work. In reality, this does compile, but doesn't function when I actually try to emit SIGPIPE by doing:

```
cat | ./jq . | bash -c 'sleep 1; foo'
bash: foo: command not found
```

My understanding is that the missing command `foo` should immediately result in SIGPIPE on the jq command, so that
this feature terminates jq immediately.

I'd greatly appreciate it if anyone could share your insight on how I could make it actually work :pray:

Resolves #1691

### DIFF
--- a/src/signal.c
+++ b/src/signal.c
@@ -1,0 +1,23 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <signal.h>
+#include "signal.h"
+
+void jq_handle_sigpipe(int signum)
+{
+#ifndef WIN32
+  signal(SIGPIPE, SIG_IGN);
+  fprintf(stderr, "jq: error: received SIGPIPE\n");
+  exit(128 + SIGPIPE);
+#endif
+}
+
+void jq_register_signal_handler()
+{
+#ifndef WIN32
+  if (signal(SIGPIPE, jq_handle_sigpipe) == SIG_ERR ) {
+    fprintf(stderr, "jq: failed to register signal handler");
+    exit(1);
+  }
+#endif
+}
--- a/Makefile.am
+++ b/Makefile.am
@@ -105,7 +105,7 @@ src/builtin.inc: src/builtin.jq
 src/builtin.o: src/builtin.inc
 
 bin_PROGRAMS = jq
-jq_SOURCES = src/main.c src/version.h
+jq_SOURCES = src/main.c src/version.h src/signal.c
 jq_LDFLAGS = -static-libtool-libs
 jq_LDADD = libjq.la -lm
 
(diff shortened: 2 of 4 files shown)


src/signal.c
```diff
@@ -0,0 +1,23 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <signal.h>
+#include "signal.h"
+
+void jq_handle_sigpipe(int signum)
+{
+#ifndef WIN32
+  signal(SIGPIPE, SIG_IGN);
+  fprintf(stderr, "jq: error: received SIGPIPE\n");
```
nicowilliams, line 10: `fprintf` is not async-signal-safe, so it can't be used from a signal handler.

src/signal.c
```diff
@@ -0,0 +1,23 @@
...
+  exit(128 + SIGPIPE);
```
nicowilliams, line 11: `exit()` is not async-signal-safe, so it can't be used from a signal handler.  Use `_exit()` instead.

nicowilliams: If and when we add an FFI this (exiting in a signal handler) will probably not be a safe thing to do.

nicowilliams:
See also commentary in #1285.

PR #1718, opened 2018-09-10
Add fromstring builtin

lorenz:
This adds the `fromstring` builtin (analogous to `tostring`) which parses JSON from a string. I found myself needing this quite a lot mainly for Kubernetes work.

Also kudos for the very well-structured codebase, this took less than 20mins including documentation!


wtlangford:
This exists, and is called `fromjson`.

lorenz:
D'oh! I searched Google and the manual for some time and didn't find it, so I assumed it must not exist. Thnaks!

PR #1726, opened 2018-09-26
Rename gen_* functions to jq_gen_*

jgarvin:
This prevents multiple definition errors when static linking against
libpcap.

All I did was: find . -name '*.c' -o -name '*.h' -o -name '*.y' | xargs sed -i -e 's/gen_/jq_gen_/g' 

Seems to build, but I haven't been able to run the tests because my automake is too old for oniguruma.


nicowilliams:
Why shouldn't it be libpcap that gets modified?

Why should we support static linking of anything other than the jq executable with libjq?  I'm inclined to support only dynamic linking.

Static linking has very bad semantics.  That is not a result of the _static_ part, but rather of the historical lack of metadata in static link archives, thus the lack of link-edit-time metadata for the link-editor.  Dynamic linking fixed this by adding that metadata -- the fact that it's also dynamic linking is not relevant.  It's sad that we can't have static linking with ELF's semantics.

The relevant ELF semantics that static linking doesn't have (but should) are:

- recording only direct dependencies at every library (this helps the linker resolve symbol name conflicts)
- recording where to search for those direct dependencies
- listing only direct dependencies at executable link-edit time
- versioned symbols (Linux) / direct binding (Illumos)
- symbol visibility control

Every time demands to use static linking come up I wonder if I could find a way to dedicate a year of my time to fixing all the common link-editors to implement the above semantics.  I must not be the only one who has ever had that thought.  It's been decades since ELF happened and no one has bothered to go back and implement these semantics for static linking (libtool tries, but libtool is garbage), which means that no one ever finds these problems troublesome enough to go fix static linking.  That's because dynamic linking works very well.  So let's just use dynamic linking.

nicowilliams:
I won't close this yet.  I want to hear from @jgarvin.  Maybe @jgarvin really has a problem that can't be fixed in any other way.  Maybe @jgarvin can just keep this patch as a local patch, but maybe that's not good enough and maybe they can convince us to accept this.

jgarvin:
> Why shouldn't it be libpcap that gets modified?

It probably should be as well, but jq has at least made the effort to prefix most of its functions to work around the lack of namespacing in C so this seems consistent.

> Static linking has very bad semantics.

Static linking has better performance and some compiler features require it (whole program optimization). This is an environment where I control the entire stack so everything that is going to run on the machine using jq gets rebuilt if jq is upgraded.

PR #1728, opened 2018-09-27
Enhanced builtins all, any and range/3

fadado:
Changed definitions in builtin.jq for `all`, `any` and `range/3` for faster and clearer versions.

Moved `isempty` above in the file.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -32,20 +32,13 @@ def index($i):   indices($i) | .[0];       # TODO: optimize
 def rindex($i):  indices($i) | .[-1:][0];  # TODO: optimize
 def paths: path(recurse(if (type|. == "array" or . == "object") then .[] else empty end))|select(length > 0);
 def paths(node_filter): . as $dot|paths|select(. as $p|$dot|getpath($p)|node_filter);
-def any(generator; condition):
-        [label $out | foreach generator as $i
-                 (false;
-                  if . then break $out elif $i | condition then true else . end;
-                  if . then . else empty end)] | length == 1;
-def any(condition): any(.[]; condition);
-def any: any(.);
-def all(generator; condition):
-        [label $out | foreach generator as $i
-                 (true;
-                  if .|not then break $out elif $i | condition then . else false end;
-                  if .|not then . else empty end)] | length == 0;
-def all(condition): all(.[]; condition);
-def all: all(.);
+def isempty(g): 0 == ((label $go | g | (1, break $go)) // 0);
+def all(generator; condition): isempty(generator | condition and empty);
+def all: isempty(.[] and empty);
+def all(condition): isempty(.[] | condition and empty);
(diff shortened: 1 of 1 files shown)


pkoppstein:
@fadado - Very nice!

A more efficient version of range/3 is given below (it also includes a check against $by==0), though maybe it should be a C function.

```
def range($init; $upto; $by): #:: (number;number;number) => *number
  if $by == 0 then error("increment in range cannot be 0")
  else label $out
  | $init
  | if $by>0
    then recurse(.+$by) | if . < $upto then . else break$out end
    else recurse(.+$by) | if . > $upto then . else break$out end
    end
  end ; 

```

fadado:
Instead of an error `empty` should produced like `range(0;0)` does.

pkoppstein:
@fadado - For the sake of simplicity, let's consider the two essential cases when $by == 0:

* range(0;0;0) # if allowed, it would be reasonable to regard this as identical to `empty` as you suggest;

* range(0;1;0) # if allowed, this should be equivalent to the infinite stream of 0s, i.e. `repeat(0)`.

These behaviors would follow from the analogy between jq's `range(x;y;by)` and C's `for(i=x;i<y; i+=by)`.

It seems to me, however, than since jq already has `repeat/1`, it is more useful for `range(_;_;0)` to produce an error consistently as I cannot think of any compelling circumstance where 0 would intentionally be the third argument.

fadado:
In *JQ* you can design partial functions using three different techniques:

1. Union type: return `null` for not defined results in addition to "normal" values.
2. Produce `empty` for no defined results.
3. Raise an error if the function cannot fulfil his contract because  there are failed preconditions.

The technique 1 is used by `add/0` and `last/1`, technique 2 is used by `range/2` and technique 3 by `nth/2` and your `range/3`. I always prefer  technique 2  in *JQ* to promote an style of design in the __Icon__ language tradition, but all techniques are OK  if not mixed at random.

pkoppstein:
Since the current implementation emits `empty` when $by==0,
I'll amend my proposed enhancement (with regrets) accordingly:

```
def range($init; $upto; $by): #:: (number;number;number) => *number
  if $by == 0 then empty
  else label $out
  | $init
  | if $by>0
    then recurse(.+$by) | if . < $upto then . else break$out end
    else recurse(.+$by) | if . > $upto then . else break$out end
    end
  end ; 
```

Even better would be an implementation in C.  It if agrees with
the semantics of C's for(int i=x; i<y; i+=by), so much the better.

fadado:
I close this pull request, and will open a new one on top of 1.6 with several builtins modified with better definitions.

leonid-s-usov:
The range function is implemented in `builtin.c` as a generated bytecode. Maybe it would be better to extend that one with the third parameter and have the /2 variant defined in `builtin.jq`?

PR #1736, opened 2018-10-08
Fix truncate_stream example

twosider:
The input is `1`, then the script also gives an input of `1`.


itchyny (changes requested):
Looks a reasonable change fixing the example of `truncate_stream` to use the `input` as the filter input. Needs rebasing and modifying all the version docs though.

PR #1738, opened 2018-10-10
Add new builtin `fetch`

bbbco:
I added a new builtin called `fetch`, which allows you to specify a key and returns the value or nothing (or another default you define). This is similar to the Ruby [`#fetch`](https://ruby-doc.org/core-2.5.0/Hash.html#method-i-fetch) method on Hash-like objects.

````
           jq ´fetch(.foo)´
              {"foo": "bar"}
           => "bar"

           jq ´fetch(.fu)´
              {"foo": "bar"}
           =>

           jq ´fetch(.fu; "baz")´
              {"foo": "bar"}
           => "baz"
````

I've added tests and docs for this new builtin, and running `make check` passes.

### CODE REVIEW

**Comment by pkoppstein** [comment]
Most people, I think, would agree that 

    {"a": false} | fetch(.a)

should emit `false` rather than the empty stream.  

Perhaps the following achieves the semantics you have in mind:

```
    def fetch(key):
      del(key) as $o
      | if $o == . then empty else key end;

```
However this is evidently very inefficient, so perhaps a C implementation is warranted.

**Comment by bbbco** [comment]
Ah, yes. Great edge cases! I think if we're following the pattern of the Ruby implemention, we should pass falsey and null values through if they are being set on the key value (as we can see happening in Ruby here http://rubyfiddle.com/riddles/71c57/2 ).

Also, to answer your other question about deeper nesting of the key that is being fetched, maybe we should take cue from Ruby and require a string of just a single key and use `has()` instead of `?` to see if it's available and pass it through.

**Comment by pkoppstein** [comment]
@bbbco -  For `fetch(_)`, it would make sense to allow _ to be a jq path path expression (like .a.b), or an array path expression (as in `getpath(_)`), or maybe either.  Here is one possibility that allows _ to be a jq path:

```
def fetch(key; default):
  ({} | path(key)) as $k
  | if any(path(..); . == $k) then key else default end;

def fetch(key): fetch(key; empty);
```
Example: 

    {a:0, b: {c:1, d:2}} | fetch(.b.c)

**Comment by bbbco** [comment]
Thanks for the suggestion @pkoppstein ! I've incorporated your code, slightly tweaked so that the default can also refer to a path expression as well. I've also added a few more tests to the tests and docs.

**Comment by pkoppstein** [comment]
`.|default` is the same as `default`.  Remove the `.|`.

**Comment by bbbco** [comment]
Huh. You're right. I could have sworn it wasn't working right without that. 

Anyways, I've updated and squashed the commits into one tidy one.

**Comment by pkoppstein** [comment]
The def using `path(..)` was only meant to illustrate one possible semantics.  As an implementation, it's ridiculously inefficient.

Here's one way to have our cake and eat it:

```
# $key must be an array-path, as with getpath/1
def fetchpath($key; default):
  ($key|length) as $len
  | if $len > 0
    then $key[0] as $k
    | if has($k)? // false
      then .[$k] | (if $len == 1 then . else fetchpath($key[1:]; default) end)
      else default
      end
    else default
    end;

def fetch(key; default):
  fetchpath({} | path(key); default);

def fetchpath(key): fetchpath(key; empty);

def fetch(key): fetch(key; empty);
```

**Comment by bbbco** [comment]
The problem with this approach can be seen in the following examples:

````
╰─ ./jq '{"foo": {"bar": 2}, "fu": "baz"} | fetch(.foo.bar; .fu)' -n
2                                                                                                           # This is good
╰─ ./jq '{"foo": {"bar": 2}, "fu": "baz"} | fetch(.foo.bar2; .fu)' -n
null                                                                                                       # This is not
╰─ ./jq '{"foo": "foo", "fu": "baz"} | fetch(.foo.bar; .fu)' -n                       # Neither is this (returns empty)
````
and
````
╰─ ./jq '{"foo": {"bar": 2}, "fu": "baz"} | fetch(.fu2; .foo)' -n                        # This is good
{
  "bar": 2
}
╰─ ./jq '{"foo": {"bar": 2}, "fu": "baz"} | fetch(.fu2; .foo.bar)' -n                  # So is this
2
╰─ ./jq '{"foo": {"bar": 2}, "fu": "baz"} | fetch(.foo.bar2; .foo.bar)' -n          # This is not
null
╰─ ./jq '{"foo": {"bar": 2}, "fu": "baz"} | fetch(.foo.bar2; .bar)' -n                # Neither is this
2
````

**Comment by pkoppstein** [comment]
Revised to meet your expectations.

**Comment by pkoppstein** [comment]
The following version of `fetchpath/2` should be much faster.  Maybe you'd like to test and benchmark it, though as I originally said,  this kind of low-level functionality probably should be implemented in C.

```
# $key must be an array-path, as with getpath/1
def fetchpath($key; default):
  getpath($key) as $value
  | if $value != null then $value
    elif ($key|length) == 1 then default
    else getpath($key[:-1]) as $x
    | if $x == null then default
      elif $x|has($key[-1]) then null
      else default
      end
   end;
  
```

bbbco:
This version does not pass through null values:

````
╰─ ./jq '{"foo": null} | fetch(.foo)' -n
````
(returns empty)

Here's the performance benchmarks:
````
╰─ time cat citylots.json | ./jq -cM '.features[].properties.LOT_NUM' >/dev/null                                       
cat citylots.json  0.00s user 0.29s system 5% cpu 5.469 total
./jq -cM '.features[].properties.LOT_NUM' > /dev/null  5.32s user 0.55s system 96% cpu 6.082 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties.LOT_NUM)' >/dev/null
cat citylots.json  0.00s user 0.32s system 5% cpu 5.628 total
./jq -cM '.features[] | fetch(.properties.LOT_NUM)' > /dev/null  5.84s user 0.58s system 96% cpu 6.633 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties.LOT_NUM2)' >/dev/null
cat citylots.json  0.05s user 0.24s system 5% cpu 5.477 total
./jq -cM '.features[] | fetch(.properties.LOT_NUM2)' > /dev/null  6.21s user 0.52s system 97% cpu 6.930 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM)' >/dev/null
cat citylots.json  0.03s user 0.27s system 5% cpu 5.434 total
./jq -cM '.features[] | fetch(.properties2.LOT_NUM)' > /dev/null  5.97s user 0.58s system 97% cpu 6.742 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry.type)' >/dev/null
cat citylots.json  0.00s user 0.29s system 5% cpu 5.400 total
./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry.type)' >   6.30s user 0.51s system 97% cpu 6.964 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry2.type)' >/dev/null
cat citylots.json  0.02s user 0.28s system 5% cpu 5.412 total
./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry2.type)' >   6.12s user 0.53s system 97% cpu 6.841 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM; fetch(.geometry2.type; .geometry.type))' >/dev/null
cat citylots.json  0.01s user 0.17s system 3% cpu 5.686 total
./jq -cM  > /dev/null  7.11s user 0.64s system 95% cpu 8.100 total
````

Hmm... so seems like this implementation is slightly better when run `fetch/1`, but slightly worse with `fetch/2`.

I'm not as familiar with working with lower level C. Do you have suggestions for a lower level C implementation? Or are you going to be ok with accepting one of these implementations for the time being?

bbbco:
Thanks for fixing the bug. Looks good as far as tests passing now.

Here's the updated benchmarks:

````
╰─ time cat citylots.json | ./jq -cM '.features[].properties.LOT_NUM' >/dev/null
cat citylots.json  0.00s user 0.33s system 5% cpu 5.516 total
./jq -cM '.features[].properties.LOT_NUM' > /dev/null  5.24s user 0.63s system 95% cpu 6.126 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties.LOT_NUM)' >/dev/null
cat citylots.json  0.00s user 0.27s system 4% cpu 5.474 total
./jq -cM '.features[] | fetch(.properties.LOT_NUM)' > /dev/null  5.78s user 0.53s system 96% cpu 6.548 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties.LOT_NUM2)' >/dev/null
cat citylots.json  0.00s user 0.26s system 4% cpu 5.518 total
./jq -cM '.features[] | fetch(.properties.LOT_NUM2)' > /dev/null  6.17s user 0.54s system 95% cpu 6.999 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM)' >/dev/null
cat citylots.json  0.03s user 0.23s system 4% cpu 5.562 total
./jq -cM '.features[] | fetch(.properties2.LOT_NUM)' > /dev/null  6.02s user 0.56s system 95% cpu 6.866 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry.type)' >/dev/null
cat citylots.json  0.00s user 0.26s system 4% cpu 5.599 total
./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry.type)' >   6.16s user 0.60s system 95% cpu 7.109 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry2.type)' >/dev/null 
cat citylots.json  0.00s user 0.26s system 4% cpu 5.629 total
./jq -cM '.features[] | fetch(.properties2.LOT_NUM; .geometry2.type)' >   6.20s user 0.53s system 95% cpu 7.063 total

╰─ time cat citylots.json | ./jq -cM '.features[] | fetch(.properties2.LOT_NUM; fetch(.geometry2.type; .geometry.type))' >/dev/null 
cat citylots.json  0.00s user 0.27s system 4% cpu 5.505 total
./jq -cM  > /dev/null  7.04s user 0.63s system 96% cpu 7.966 total
````

bbbco:
Closing request. Seems that @pkoppstein decided to implement as `pick()` in 1.7 on their own. I'm glad that in some way I helped inspire this type of change. My implementation that I worked on with @pkoppstein five years ago was probably too heavy handed anyways.

PR #1748, opened 2018-10-25
Make reverse more DWIM-y

davidfetter:
Strings and numbers go backward.
Booleans flip.
Objects switch keys and values.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -23,7 +23,7 @@ def recurse_down: recurse;
 def to_entries: [keys_unsorted[] as $k | {key: $k, value: .[$k]}];
 def from_entries: map({(.key // .Key // .name // .Name): (if has("value") then .value else .Value end)}) | add | .//={};
 def with_entries(f): to_entries | map(f) | from_entries;
-def reverse: [.[length - 1 - range(0;length)]];
+def reverse: def r: [.[length - 1 - range(0;length)]]; if type == "string" then explode | r | implode else r end;
 def indices($i): if type == "array" and ($i|type) == "array" then .[$i]
   elif type == "array" then .[[$i]]
   elif type == "string" and ($i|type) == "string" then _strindices($i)


leonid-s-usov:
Reversing strings totally makes sense!

Boolean - it's weird. I'd not include it though, you have `not` to reverse bools in a predictable and typesefe way.

Numbers just don't make sense. If that's really what you want, I'd rather that you make it explicit in your code `.number | tostring | reverse`. The issue is that reversing an arbitrary valid number's string representation will give you an error. Consider `-1`, `3e+3`.

pkoppstein:
@davidfetter - Some observations:

a) there is no reason to define $in in the body of the def;

b) for strings, `explode | reverse | implode` is much faster (approx 2 times according to my measurements) and also significantly more economical with memory than using split/join;

c) In my opinion, the proposed def is inappropriate for objects as the proposed operation is a kind of inversion or transposition.   A more appropriate def (perhaps even *the* DWIM definition) would reverse the order of keys.  In addition, this def makes it very easy for users to ignore all the issues that key/value inversion can present. If someone wants to invert keys and values in some fashion,  they should be aware of all the issues, and be responsible for their resolution.

d) I would go even further than @leonid-s-usov and argue that the domain of `reverse` should not be extended beyond arrays and strings, mainly to avoid bloat and duplication of function. The simplicity of `explode|reverse|implode` could even be taken as an argument against the wisdom of an extension to strings :-)

leonid-s-usov:
Yeah.. I’d say that anything “explodable” should be also “reversible”

If one defines explode for objects to be to_entries then yeah, reverse would change the order of keys. 

Regards,
Leonid S. Usov

davidfetter:
I've trimmed the extension back just to strings and removed fat from the code.

The behavior for other types is compatible with the current behavior.

pkoppstein:
@davidfetter - The trimmed down def still assumes $in has been defined and is thus wrong.

Also, to avoid the second type check, I'd suggest:

    def reverse:
      def r: [.[length - 1 - range(0;length)]];
      if type == "string" then explode | r | implode else r end;

davidfetter:
I really appreciate your patient help here.

nicowilliams:
Sure, why not.

itchyny:
Related issue: #412.

PR #1752, opened 2018-10-27
Decimal literal number

leonid-s-usov:
Hi all

@pkoppstein @wtlangford @nicowilliams, let me present you the descendant of #1743, and a proud fixer of #1741 and such, the decimal number literal implementation.

Based on the other PR it features full PASS under valgrind. 

Unlike my suggestion at #1741 I haven't implemented the simple math (+ and -) to be run by the decimal number logic, although it's trivial to achieve currently.
Comparisons, however, are fully decimal, so please feel free to try
```
./jq -n '100000000000000000000000000000002 < 100000000000000000000000000000003'
```
and vice versa


leonid-s-usov:
With the latest minor change the significance of the literal is preserved.

leonid-s-usov:
@wtlangford @nicowilliams I wonder what this silence means. Does this pull request have a future?

wtlangford:
> I wonder what this silence means. Does this pull request have a future?

Just that I've been busy this last week. I just need to scrounge up the time to take a proper look at it.

leonid-s-usov:
:confused:

wtlangford:
Hey, sorry, I've been super busy. I'll have time in the next day or two to give this a good, proper read-through, due to the holiday.

leonid-s-usov:
Hi. 
Any news?

wtlangford:
Yep, I'm reading my way through it. It's large, so I'm making sure to give
it all the attention it needs. Should be done soon!

src/jv_type_invalid.c
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  jv_free(inv);
+  return x;
+}
+
+int jv_invalid_has_msg(jv inv) {
+  assert(JVP_HAS_KIND(inv, JV_KIND_INVALID));
+  int r = JVP_HAS_FLAGS(inv, JVP_FLAGS_INVALID_MSG);
+  jv_free(inv);
```
wtlangford: This free changes the semantics of this function. Previously, it would not free `inv`, and was only freeing the extra copy on the `msg` performed by `jv_invalid_get_msg` (see https://github.com/stedolan/jq/pull/1752/files/bf7f1f2b8176597a04bf4a8c00243567d3c7c83d#diff-5c46af0444da42be404b9ff2acf24da2L124)

src/jv_type_private.h
```diff
@@ -0,0 +1,89 @@
...
+} payload_flags;
+
+
+#define JVP_MAKE_PFLAGS(ptype, allocated) ((((ptype) << 4) & PTYPE_MASK) | ((allocated) ? JVP_PAYLOAD_ALLOCATED : 0))
+#define JVP_MAKE_FLAGS(kind, pflags) ((kind & KIND_MASK) | (pflags & PFLAGS_MASK))
+
+#define JVP_FLAGS(j)  ((j).kind_flags)
+#define JVP_KIND(j)   (JVP_FLAGS(j) & KIND_MASK)
+
+#define JVP_HAS_FLAGS(j, flags) (JVP_FLAGS(j) == flags)
```
wtlangford: Let's put parens around `flags` and `kind` here.

Also, I'd like the name to be a bit clearer. `JVP_HAS_FLAGS` would imply to me that I can ask about any particular bits on the flags, but this is actually saying "does it have only and exactly these specific flags?". Can we find a better name?

src/jv_type_private.h
```diff
@@ -0,0 +1,89 @@
...
+#define JVP_HAS_KIND(j, kind)   (JVP_KIND(j) == kind)
```
wtlangford: Would this be better called `JVP_IS_KIND`?

src/jv_type_array.c
```diff
@@ -0,0 +1,282 @@
...
+  return jvp_array_slice(a, start, end);
+}
+
+jv jv_array_indexes(jv a, jv b) {
+  jv res = jv_array();
+  int idx = -1;
+  jv_array_foreach(a, ai, aelem) {
+    jv_free(aelem);
+    jv_array_foreach(b, bi, belem) {
+      if (!jv_equal(jv_array_get(jv_copy(a), ai + bi), jv_copy(belem)))
```
wtlangford: remove the `jv_copy` from `belem` here, and you can skip the `jv_free(belem)` below.

src/jv_type_number.c
```diff
@@ -0,0 +1,309 @@
...
+  DEC_ROUND_HALF_UP,
+  0, /*no errors*/
+  0, /*status*/
+  0, /*no clamping*/
+};
+
+static struct {
+  uint64_t refcnt;
+  struct dtoa_context C;
+} shared_dtoa_context;
```
wtlangford: I have concerns about these shared contexts and thread-safety. While the jq vm itself is single-threaded (and the main `jq` executable as well), we do expose a C API that other, multi-threaded programs could use. Two programs attempting to use the same contexts here could encounter issues.

I understand the purpose was to reduce allocations/improve performance, but we have to be careful with globals and shared state.

src/jv_print.c
```diff
@@ -229,8 +230,12 @@ static void jv_dump_term(struct dtoa_context* C, jv x, int flags, int indent, FI
     put_str("true", F, S, flags & JV_PRINT_ISATTY);
     break;
   case JV_KIND_NUMBER: {
-    const char * str = jv_number_as_string(x, buf, JVP_DTOA_FMT_MAX_LEN);
-    put_str(str, F, S, flags & JV_PRINT_ISATTY);
+    if (jvp_number_is_nan(x)) {
+      jv_dump_term(C, jv_null(), flags, indent, F, S);
```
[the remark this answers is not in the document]
leonid-s-usov: I have actually fixed this annoying thing that when we fall back to null here we used output it as plain text null and it wouldn't get highlighted as real `null` values in the output. Just try it in console, the actual `null` and a `null` of a NAN are shown differently today.

src/jv_type_invalid.c
leonid-s-usov: Well, it's a change of implementation to follow the actual semantics of every single place this function was called, which was resulting a memory leak. 
Please check that in all places where it was called the parameter has been `jv_copy`ied.

This is exactly one of the things I have tried addressing in my API name refactoring. Today there is a big mess because one can't tell by the naming convention whether the function will be consuming the value or not.

At the same time, there is a general tendency to consume parameters in the calls, so instead of fixing all places where this function was called I have decided to change the implementation. And actually optimised it as well - previously it would construct a message just to see that it wasn't nil

src/jv_type_private.h
leonid-s-usov: Maybe. Or maybe not. 

This way it's HAS_FLAGS and HAS_KIND. But i'm not too married to this.

leonid-s-usov: Agreed about parens.

Naming - i'm open to suggestions

src/jv_type_private.h
leonid-s-usov: Maybe. Or maybe not.

This way it's HAS_FLAGS and HAS_KIND. But i'm not too married to this.

src/jv_type_array.c
leonid-s-usov: 👍

src/jv_type_number.c
leonid-s-usov: OK. Indeed, this shared data will not be usable even with external syncrhonization techniques. 
I can change it to thread local storage. The only constraint here would be that the jv vm would need to be started and used and destroyed on the same thread.

Or, we could think of some "vm context" to be used anywhere inside the implementation its needed, but it looks like a much more impacting refactor, it would require passing the context everywhere around. But it would mean that a given vm instance could be used across multiple threads given some external sync to prevent simultaneous execution.

leonid-s-usov:
🎉 let's rock and roll!

leonid-s-usov:
Re. performance.

I have linked in my other posts to the benchmarks of this library. Just to clarify, we aren't using any of the math except for comparison, and only in case two literals are being compared.

Also, if you have noticed, I have delayed the `dtoa` call until it's actually required to convert a decimal fraction into a binary fraction, and then i only do that once per literal. 
Parsing a decimal literal from decimal string representation is cheaper so if no conversion is needed we actually should gain some performance here.
However, if all parsed numbers are also converted to double for some operations, it will be slightly less performant since it will first generate a decimal structure and then run the `dtoa` algo for every number. 

I haven't run any checks cause I didn't have any infra for that at the time. I remember you mentioned trying things out so I thought that maybe you had some handy setup available, so left it out of my work scope.

pkoppstein:
> maybe you had some handy setup available

Please see
https://github.com/stedolan/jq/wiki/X---Experimental-Benchmarks

src/jv_type_invalid.c
wtlangford: > Well, it's a change of implementation to follow the actual semantics of every single place this function was called, which was resulting a memory leak.
> Please check that in all places where it was called the parameter has been `jv_copy`ied

Ah, I hadn't realized (yes, more evidence of your point. 😃). This is fine, then.

src/jv_type_number.c
wtlangford: I'd like to avoid the thread local storage so we don't have to pin a vm to its original thread. That seems like a pretty heavy cost on our users.

The vm context probably costs a bit more to solve than we'd like (certainly in implementation complexity).

That said, why do these need to be shared at all? Why not just make a `dtoa_context` or a `dec_context` on the stack when we need one? These don't seem to be particularly large (though the `dtoa_context` could be sneaky if it has to allocate a lot during processing). I'd like to investigate the performance cost of just making these on the stack every time.

leonid-s-usov: Actually, I have just realised that I can come up with a thread local implementation immune to thread move.

The shared objects will be managed per thread, with refcount; once a new jv_number requires a shared context it received an initialized pointer. Then, every time it needs the contex it should pass the pointer and the system will compare it to the current thread's shared pointer. If it's different then it will reduce the refcount for the other thread's context and increment the refcount for this thread's context, and return this thread's context, so that the jv_number already has a different pointer to work with currently.
this way, the refcount will properly flow between threads and the contexts will be freed once the last object is deallocated  or moved from that thread

wtlangford:
Sorry for the delay! The holidays got the better of me.

After further reflection, I feel that thread-locals are the right way to go, so I pushed up a copy of your branch to my fork with an extra commit on it (https://github.com/wtlangford/jq/commit/b2cf0f977de6b2c132d0420ce6dcfdef4db48417) that switches the shared contexts over to using pthread's TLD (and makes pthread no longer optional).

There's no major speed penalty to it compared to your branch, and we go back to being thread-safe. (Interestingly, most libc implementations I've looked at use similar mechanisms for their strtod context allocations).

Relating to documentation:
I'm happy to document as guarantees the semantics we intend to guarantee. For semantics we don't intend to guarantee (details about the specific implementation, etc), I'm fine with documenting them, but we need to be very explicit that we are not making guarantees about them.

I don't want to make any _guarantees_ that we'll maintain precision on your number (or even that we'll be keeping bignum support- maybe this causes lots of issues we aren't aware of yet and want to revert back to IEEE754 doubles). However, I'm happy with documenting our current behavior as "behavior", but not a "promise":
- Numbers you don't do math on will retain their precision

The last thing I might want is to put some `#ifdef USE_DEC_NUMBER` guards around all of the decnumber code and a configure flag to control it. I'd like us to be able to build without this support for testing/measurement purposes. (We'd likely also want to use that configure flag to control whether or not the decNumber sources get built into the binary, too).

leonid-s-usov:
🎉

Regards,
Leonid S. Usov

wtlangford:
Sorry again for the delays! We're at the end of them, though; at this point, all that's left is a review from @nicowilliams!

I've pushed a branch (https://github.com/stedolan/jq/compare/dec_literal_number) which contains all of this awesome work (less the `jv.c` split, as we've discussed elsewhere), rebased against the current master. I squashed most of the commits together to ease the process, and added a commit providing a configure flag (off by default!) to disable the literal numbers (meant for benchmarking, etc. I don't want it to stick around particularly long, though.).

@leonid-s-usov: The commits on my branch still have you as an author. Would you prefer to reset your branch to those or do you want me to close this PR and open one under the branch I pushed up? Either is fine with me.

nicowilliams:
I'll review this weekend.  Thanks @wtlangford!

leonid-s-usov:
Thanks much, I’ll pull from your repo and reset

Regards,
Leonid S. Usov

src/parser.y
```diff
@@ -172,7 +172,7 @@ static jv check_object_key(block k) {
     char errbuf[15];
     return jv_string_fmt("Cannot use %s (%s) as object key",
         jv_kind_name(block_const_kind(k)),
-        jv_dump_string_trunc(jv_copy(block_const(k)), errbuf, sizeof(errbuf)));
+        jv_dump_string_trunc(block_const(k), errbuf, sizeof(errbuf)));
```
nicowilliams: Was this a leak?

tests.out
```diff
@@ -0,0 +1,101 @@
+Skipped 9 tests
+Testing '.[0]' at line number 37
+Hit the number of tests limit (1), breaking
+1 of 1 tests passed (0 malformed)
+==9681== 17 bytes in 1 blocks are definitely lost in loss record 4 of 45
```
nicowilliams: Are these expected leaks??

src/parser.y
leonid-s-usov: There were plenty of leaks I've traced down and killed.
Apparently, this was one of them. To tell the truth I don't recall already but I can review. There were plenty of unobvious places which lead to my API renaming suggestion which hasn't caught on with the folks.

src/jv_print.c
```diff
@@ -229,16 +231,29 @@ static void jv_dump_term(struct dtoa_context* C, jv x, int flags, int indent, FI
...
-      put_str("null", F, S, flags & JV_PRINT_ISATTY);
+    if (jvp_number_is_nan(x)) {
+      jv_dump_term(C, jv_null(), flags, indent, F, S);
     } else {
-      // Normalise infinities to something we can print in valid JSON
-      if (d > DBL_MAX) d = DBL_MAX;
-      if (d < -DBL_MAX) d = -DBL_MAX;
-      put_str(jvp_dtoa_fmt(C, buf, d), F, S, flags & JV_PRINT_ISATTY);
+#ifdef USE_DECNUM
+      const char * literal_data = jv_number_get_literal(x);
```
nicowilliams: Do you need to check that `jv_number_has_literal(x)` before calling `jv_number_get_literal(x)`?

tests.out
leonid-s-usov: On my version of the branch there were no leaks. Need to check.

nicowilliams:
I don't suppose there was a decnum github repo we could have used as a submodule.

nicowilliams:
This weekend I'll do the review, and I may rework the TSD dtoa bits a bit to make sure this works on Windows, and there's a reasonable fallback on `__thread` on Unix if there's no pthread TSD (accepting leakage in that case).

src/jv_print.c
leonid-s-usov: nope, it will return NULL if no literal is available on the number.

src/parser.y
nicowilliams: No, it's ok, thanks.

tests.out
nicowilliams: I think you don't use this file, `tests.out`, right (`git grep tests.out` produces nothing)?  So let's remove this file.

leonid-s-usov:
> I don't suppose there was a decnum github repo we could have used as a submodule.
I've done some research and couldn't find anything better than the vanilla version of that particular library, distributed as source from their dedicated website.

tests.out
leonid-s-usov: AH right this file was accidentally committed. It was an intermediate support file I've generated during my hunt for leaks.

BTW it shows the usage of a side feature I've implemented which allows to -skip and -take a slice out of a test file. I've used it to bisect and find the leaking tests

nicowilliams: Yeah, I saw.  I liked it.

BTW, with the new `eval` stuff in #1843 we could replace the `src/jq_test.c` with jq-coded test-running utilities, and then `--run-tests` could even be removed.

nicowilliams:
@leonid-s-usov I'm not terribly interested in reviewing decNumber :) but we should check that it's roughly that from ICU.  Thanks.

leonid-s-usov:
> but we should check that it's roughly that from ICU. Thanks.

I'm not sure I'm following... 
Do you mean we should compare the source code with the most recent version on the ICU site? http://download.icu-project.org/files/decNumber/  
image

leonid-s-usov:
Just one thing, I believe that in this version there is still no usage of the decNumber library for addition (and subtraction of course). Doing a +1 will convert it to float.
Question if you think it is worth preserving the decimal representation for addition (and subtraction) operations.
I'd vote for it.

src/jv_print.c
nicowilliams: Ok.  I had taken a brief look and gotten the impression it would assert that the number has a literal.  I'll figure that out for sure this weekend.

nicowilliams:
> Just one thing, I believe that in this version there is still no usage of the decNumber library for addition (and subtraction of course). Doing a +1 will convert it to float.
> Question if you think it is worth preserving the decimal representation for addition (and subtraction) operations.
> I'd vote for it.

We can use more things from decNumber later.  It would be nice to be able to do arithmetic on large numbers without having to convert to `double` (I assume you meant `double`, not `float`).  But obviously not all math will be doable without converting to `double`.  So this can wait.

src/jv_print.c
leonid-s-usov: It asserts that it is a _number_, but gracefully checks for literal value presence.

nicowilliams:
FYI, I think we should adopt Heimdal's [`DllMain()`-based TSD implementation](https://github.com/heimdal/heimdal/blob/master/lib/base/dll.c).

nicowilliams:
Here's an issue to think about...

Until this contribution, if we merely added support for atomic reference count increment/decrement, then that's all we'd need to support `jv` value sharing among threads thread-safely.

Starting with this contribution we do mutate `jvp` values even when the reference count is larger than 1.  The values we mutate into place are deterministic, so in the event of racing we only risk leaking.

To avoid this we'd need to change the `jv` API to require that a `jv` with partial `jvp` state be "normalized" or "evaluated" using a function that takes a `jv` and returns an equivalent `jv` (possibly the same `jv` if it was already normalized).

Comments?

leonid-s-usov:
Can you give an example from the current API which doesn't already have such semantic? 
Most of the functions are consuming, meaning that the value passed is not usable after the call. This means that if the state is thread-sensitive the internals could perform the synchronization as required without affecting the API user who would only expect the result and would not be sending the same value elsewhere.
`jv_copy` itself could actually invoke some internal methods in a critical section manner, apart from the default count increment.


Actually, today jv_number is not having any thread-sensitive mutable state. All operations are idempotent; the only thing that is thread dependent is the dtoa context, which is thread dependent only for the sake of memory and speed optimization. 
The internally cached double representation of the number is, as well, merely an optimization. In a case of a thread race, the worst thing to happen is that the string to double conversion will be run multiple times. And that is easily fixable with additional internal synchronization without affecting the API users.

To conclude, I totally dig the memory management model used in jq and I don't think that this branch anyhow violates its principles. It does add internal complexity, but that's totally internal implementation details.

nicowilliams:
Currently `jvp` parts of `jv` values are never mutated except for their `refcnt` field.

leonid-s-usov:
ok.. but how does that fact change the assumption that in all cases, jv_number included, a jv value is not mutable in a way visible via the API? There is a reason why the pointer is called jvp, i.e. private.

nicowilliams:
@leonid-s-usov right, it changes nothing, I agree, about the API.  But if we wanted to have a thread-safe build of libjq that supports sharing a `jv` across multiple threads, then we'd need: a) atomic increment/decrement for reference counts (obviously), but also b) atomic compare-and-swap to mutate `jvp`s when fully-evaluating numbers.

I'm just pointing this out.

nicowilliams:
So, never mind, I think we'd just use an atomic CAS to mutate `jvp` when the reference count is not 1.

leonid-s-usov:
That assumes one would need to create a copy of the whole jvp object to mutate.
I am not sure this is needed in a generic enforcing fashion. Why not let every type decide whether or not that is needed? In a number of cases making a jvp copy would be an overhead, as sometimes a CAS on some internal var could be sufficient.

However, on the other hand, I would agree that some pre-coded routines to work with thread-local cache per type (like the one we've implemented here) and generic thread safe jvp management could be good tooling for jv types to use as needed.

nicowilliams:
@leonid-s-usov Let's backup.

Until now all `jv` values, and their `jvp` parts were immutable as long as `refcnt` was larger than 1.  Now when calling `jv_number_value()` on a `number` whose `jvp`'s `refcnt` is larger than 1 can mutate the `jvp`.  The way it mutates the `jvp` is deterministic, so the only danger is leaking.

Using an atomic CAS would prevent the leaking.

Of course, at this time we don't use any atomic operations, and we'd have to start using atomic increment/decrement for the `refcnt` in order to make any of this thread-safe, so this is all a bit academic.

Anyways, please disregard as I am convinced we can make this thread-safe.

nicowilliams:
@leonid-s-usov Let's backup.

Until now all `jv` values, and their `jvp` parts were immutable as long as `refcnt` was larger than 1.  Now when calling `jv_number_value()` on a `number` whose `jvp`'s `refcnt` is larger than 1 can mutate the `jvp`.  The way it mutates the `jvp` is deterministic, so the only danger is leaking.

Using an atomic CAS would prevent the leaking.

Of course, at this time we don't use any atomic operations, and we'd have to start using atomic increment/decrement for the `refcnt` in order to make any of this thread-safe, so this is all a bit academic.

Anyways, please disregard as I am convinced we can make this thread-safe.

leonid-s-usov:
That's cool, I understand it, and I agree.

What I'm adding on top is that I believe in most cases it will be sufficient to make an atomic CAS to _a member_ of the private structure pointed to by the jvp, under the implementing types' responsibility. This may be much more efficient than a common requirement to make a swap of the full jvp structure.

nicowilliams:
Yes, that's what I had in mind.

nicowilliams:
Also, no need to CAS a `double` into place in the case of numbers...  since the computation will be deterministic, having two or more threads racing to write the same `double` into the same address is fine.

leonid-s-usov:
For this branch it's not about the double, apparently, but about this one, for example
```
... 
  if (plit->literal_data == NULL) {
    int len = jvp_dec_number_ptr(n)->digits + 14;
    plit->literal_data = jv_mem_alloc(len);

    // Preserve the actual precision as we have parsed it
    // don't do decNumberTrim(pdec);

    decNumberToString(pdec, plit->literal_data);
  }```

wtlangford:
I've chatted with @nicowilliams: I'm going to go ahead and merge this and we'll just watch for any issues that may pop up as time goes on. The thread-safety issues we discussed above can be addressed whenever we're ready to cross that bridge.

Sorry to let this languish here for so long!

leonid-s-usov:
Wow!

Amazing. Thank you!

Regards,
Leonid S. Usov

mmotwani:
This was merged back in 2019, but no new builds since then.  It would be nice not to have to create custom builds with this fix. @stedolan @leonid-s-usov

PR #1759, opened 2018-11-05
Fix for missing directory bdir/src

mail6543210:
This occurs given:

* out-of-source-build: (eg. building in the directory `build`) since `${build-dir}/src` is not present
* `--disable-dependency-tracking`: the ${build-dir}/src is not created since it's unneeded

As a result, the `${build-dir}/src` is absent.
Then, the shell output redirection fails.

This patch `mkdir -p` in advance of the redirection output.


nicowilliams:
Thanks!

PR #1772, opened 2018-11-21
fix comparison of integers of different signs warning

itchyny:
Building on macOS emits the following warning so I fixed by casting to int.
```
src/lexer.c:1500:47: warning: comparison of integers of different signs: 'int' and 'yy_size_t' (aka 'unsigned long') [-Wsign-compare]
        if ((int) (yyg->yy_n_chars + number_to_move) > YY_CURRENT_BUFFER_LVALUE->yy_buf_size) {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -1497,7 +1497,7 @@ static int yy_get_next_buffer (yyscan_t yyscanner)
 	else
 		ret_val = EOB_ACT_CONTINUE_SCAN;
 
-	if ((int) (yyg->yy_n_chars + number_to_move) > YY_CURRENT_BUFFER_LVALUE->yy_buf_size) {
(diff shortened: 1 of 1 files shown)


leonid-s-usov:
You shouldn't be editing `lexer.c` directly. It's generated automatically from `lexer.l`. Your fixes must be done on the `.l` file (or in other supporting files) and then you should commit the regenerated `.c`

Specifically this issue seems to be rooted deeper in the `flex` generator, as this piece of code is entirely autogenerated. I'm not good at this tool. The only idea I am having is to ask it not generate the typedef or `yy_size_t` and define it manually as a synonym of `int`. That fixes the warning but might have side effects. Anyhow, a LEX expert comments are needed

itchyny:
Ah ok, thanks.

PR #1782, opened 2018-12-26
Two fixes for MinGW64, and one for non-root builds

wiiaboo:
Great software, use it a lot. Keep it up 👍

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -34,7 +34,7 @@ void *alloca (size_t);
 #endif
 #include <string.h>
 #include <time.h>
-#if defined(WIN32) && !defined(HAVE_SETENV)
+#ifdef WIN32
 #include <windows.h>
 #endif
 #include "builtin.h"
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -34,6 +34,9 @@ void *alloca (size_t);
 #endif
 #include <string.h>
 #include <time.h>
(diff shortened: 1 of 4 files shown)


src/builtin.c
```diff
@@ -34,6 +34,9 @@ void *alloca (size_t);
 #endif
 #include <string.h>
 #include <time.h>
+#if defined(WIN32) && !defined(HAVE_SETENV)
```
nicowilliams: Make that just `#ifdef WIN32`.  Whether we have `setenv()` is of no relevance here, right?

src/lexer.l
```diff
@@ -1,7 +1,7 @@
 %{
 #include <assert.h>
-#include "jv_alloc.h"
-#include "compile.h"
+#include "src/jv_alloc.h"
```
nicowilliams: Oh?  `-I${srcdir}` is not enough?

wiiaboo: I ran make with `V=1` and sure enough, there were just `-I. -I..`. The `..` is the `-I${srcdir}`, not `root/src`.

nicowilliams: I see the same.  Why does jq build for me on Unix when building out of tree?

nicowilliams:
@wiiaboo Thanks for your contribution!  I've rebased it, made the above change, and pushed it.

PR #1789, opened 2019-01-14
Provide a jvp_dump_raw_string function to implement the combination o…

rain-1:
…f flags -r and -a (raw and ascii)

### DIFF
--- a/src/jv_print.c
+++ b/src/jv_print.c
@@ -114,13 +114,18 @@ static void put_indent(int n, int flags, FILE* fout, jv* strout, int T) {
 }
 
 static void jvp_dump_string(jv str, int ascii_only, FILE* F, jv* S, int T) {
+  put_char('"', F, S, T);
+  jvp_dump_string_raw(str, ascii_only, F, S, T);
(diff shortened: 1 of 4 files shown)


src/jv_print.c
```diff
@@ -114,13 +114,18 @@ static void put_indent(int n, int flags, FILE* fout, jv* strout, int T) {
 }
 
 static void jvp_dump_string(jv str, int ascii_only, FILE* F, jv* S, int T) {
+  put_char('"', F, S, T);
+  jvp_dump_string_raw(str, ascii_only, F, S, T);
```
nicowilliams: What happens if the string has double-quote characters in it?

nicowilliams: Oh, I guess this isn't new...

rain-1: That's actually a good question.. can jvp_dump_string be tricked into producing bad output?

nicowilliams:
Hey, thanks for this.

Alanscut (approved):

PR #1794, opened 2019-01-22
Add tests for built-in function `contains`

haguenau:
Add tests for built-in function `contains` on trings, as per comments on #1732.

I believe this covers everything that would detect a bug wherein embedded null characters would be interpreting as string terminators, by making sure that characters after a null both in the "haystack" and in the "needle" are considered.

### DIFF
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1092,9 +1092,26 @@ null
 [true, true, false]
 
 # containment operator (embedded NULs!)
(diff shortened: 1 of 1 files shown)


nicowilliams:
Thanks @haguenau!

nicowilliams:
(The build failures come from `master`.  I'll take a look at some point.)

PR #1798, opened 2019-01-27
Add fromstream_with_dups() builtin (fix #1795)

### LINKED ISSUE #1795: Unexpected behavior with duplicate attributes
**Actual behavior:**
```
$ echo '{"a" : "b", "a" : "c" }' | jq '.a'
"c"
```
Duplicate keys are valid, but discouraged, in JSON. https://dzone.com/articles/duplicate-keys-in-json-objects

**Expected behavior:**
Both `b` and `c` are output.

**Environment:**
 - OS and Version: macOS 10.14.2
 - jq version: 1.5

nicowilliams:
This commit adds a pair of built-ins for dealing with streamed JSON
texts that have duplicate keys.

Eventually we may need a `tostream_with_dups` as well that implements a
convention that array values in objects are to be presented as duplicate
keys with array elements as values.  There are ambiguity issues, like
how should one stream `{"a":[]}`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ jobs:
                         - valgrind
                         - bison
                         - automake
+                        - git
 
             before_install:
                 - uname -s
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,8 +105,8 @@ jobs:
                 - uname -s
                 - brew update
                 - brew install flex bison
-                - rvm install ruby-1.9.3-p551
-                - rvm use 1.9.3
+                - rvm install ruby-2.3.0
+                - rvm use 2.3.0
                 - gem install bundler
                 - rm src/{lexer,parser}.{c,h}
                 - sed -i.bak '/^AM_INIT_AUTOMAKE(\[-Wno-portability 1\.14\])$/s/14/11/' modules/oniguruma/configure.ac
--- a/docs/content/3.manual/manual.yml
+++ b/docs/content/3.manual/manual.yml
@@ -2924,6 +2924,9 @@ sections:
       single JSON text that is 1GB in size, streaming it will allow you
       to process it much more quickly.
 
+      Streaming not only allows for online consumption of large JSON
+      texts, but also allows for processing of duplicate object keys.
+
       However, streaming isn't easy to deal with as the jq program will
       have `[<path>, <leaf-value>]` (and a few other forms) as inputs.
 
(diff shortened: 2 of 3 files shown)


emanuele6:
I just found this PR; `fromstream_with_dups` does not seem to work very well:

```
$ ./jq -c --stream <<< '{"a":0,"b":2,"a":[10]}'
[["a"],0]
[["b"],2]
[["a",0],10]
[["a",0]]
[["a"]]
```

If I run `fromstream_with_dups([["a"],0],[["b"],2],[["a",0],10],[["a",0]],[["a"]])`, I get `{"a":[10],"b":[2]}` instead of `{"a":[0,[10]],"b":[2]}`

nicowilliams:
> I just found this PR; `fromstream_with_dups` does not seem to work very well:

Ay, blast from the past -- stuff I've forgotten about.

> ```
> $ ./jq -c --stream <<< '{"a":0,"b":2,"a":[10]}'
> [["a"],0]
> [["b"],2]
> [["a",0],10]
> [["a",0]]
> [["a"]]
> ```
> 
> If I run `fromstream_with_dups([["a"],0],[["b"],2],[["a",0],10],[["a",0]],[["a"]])`, I get `{"a":[10],"b":[2]}` instead of `{"a":[0,[10]],"b":[2]}`

Why would you expect the above to yield `{"a":[0,[10]],"b":[2]}`?  I definitely only expect it to yield `{"a":[10],"b":[2]}`.

emanuele6:
> Why would you expect the above to yield `{"a":[0,[10]],"b":[2]}`? I definitely only expect it to yield `{"a":[10],"b":[2]}`.

Because the input object had two "a" keys, one was `0`, and the other one was `[10]`. If i had two "a" keys of which one was `0`, and the other one was `1`, this would have yielded `{"a":[0,1]}` so why would it not yield `{"a":[0,[10]]}` if one of the values is an array?
Is this function not supposed to return a value in which all the objects have array values with one or more elements that are all the values in the input for a given key?

nicowilliams:
> Because the input object had two "a" keys, one was `0`, and the other one was `[10]`. If i had two "a" keys of which one was `0`, and the other one was `1`, this would have yielded `{"a":[0,1]}` so why would it not yield `{"a":[0,[10]]}` if one of the values is an array?

We have first `[["a"],0]` which says that the top-level is an object and it has an `"a"` with value `0`.

Then we have `[["a"],<something-else>]`, so that _replaces_ the `0`.

If `.a` is to be an array then we expect `[["a",<index>],<value>]`, and then if we see multiple different array indices, then we build an array at `.a` with those values at those indices.

In this example though we're not doing that the first time we see a `.a` (`[["a"],0]`), and later when we see `[["a",0],10]` and until then `.a` was not an array, well then we have to replace `.a` with an array with `10` as the first value, so `{"a":0, ...}` becomes `{"a":[10], ...}`.

Now if we see `[["a",0],10]` then `[["a",1],20]`, then by the time we see `[["a",1],20]` we have `{"a":[10], ...}` and `["a",1]` is not a duplicate, so we can set the value at index `1` of `.a` to `20` and end up with `{"a":[10,20], ...}`.

If the stream isn't exactly well-formed and we see `[["a",0],10]` then `[["a",3],20]`, we'll end up with `{"a":[10,null,null,20], ...}`, so it's a bit forgiving, just like it's forgiving of dups in the same way that the jq JSON parser is in that the last value of a duplicate wins -- we don't gather all the dups into arrays:

```
: ; echo '{"a":0,"a":1}'|./jq -c .
{"a":1}
```

and the streaming parser is even more forgiving because it's stateless, so it preserves dups that the non-streaming parser doesn't:

```
: ; echo '{"a":0,"a":1}'|./jq -c --stream .
[["a"],0]
[["a"],1]
[["a"]]
```

> Is this function not supposed to return a value in which all the objects have array values with one or more elements that are all the values in the input for a given key?

No, because the values of any object keys don't have to be arrays -- they can be any other JSON type.

emanuele6:
> If the stream isn't exactly well-formed and we see `[["a",0],10]` then `[["a",3],20]`, we'll end up with `{"a":[10,null,null,20], ...}`, so it's a bit forgiving, just like it's forgiving of dups in the same way that the jq JSON parser is in that the last value of a duplicate wins -- we don't gather all the dups into arrays:
> ```
> : ; echo '{"a":0,"a":1}'|./jq -c .
> {"a":1}
> ```

I know how `--stream` works, but this `fromstream_with_dups` function does try to gather all the dups (and unique key element too) into arrays (note that `b` is also wrapped in an array):
```
$ ./jq --stream -n 'fromstream_with_dups(inputs)' <<< '{"a":0,"b":2,"a":[],"a":3}'
{
  "a": [
    0,
    [],
    3
  ],
  "b": [
    2
  ]
}
```
And it only succeeds in doing that if the duplicate values are scalars or `[]`/`{}`. If you have an object duplicate, it errors, and if you have an array duplicate, it overwrites what has been collected so far with that array.

```
{"a":1,"a":[10,3],"b":"foo","a":4}
=> {"a":[10,3,4],"b":"foo"}
{"a":1,"a":2,"b":{"x":3,"y":4}}
=> {"a":[1,2],"b":{"x":[3],"y":[4]}
```

If that is your intended behaviour, it does not seem very useful.

nicowilliams:
Oh, hah, I hadn't read this code :/

I say we close this.

Issue #1713, opened 2018-08-30
fromdate output may be off by an hour


muhmuhten:
**Describe the bug**
When DST applies to the input time, the return value of `fromtime` is an hour later than expected.

**To Reproduce**
`TZ=America/New_York jq -n '"2018-08-31T00:00:00Z"|fromdate|todate'`

**Expected behavior**
The above should output `"2018-08-31T00:00:00Z"`, not `"2018-08-31T01:00:00Z"`.
`fromdate|todate` and `todate|fromdate` should probably be idempotent.

**Environment (please complete the following information):**
 - macOS 10.11
 - jq 1.6rc2
 - bug not present in jq 1.5

**Additional context**
Swapping out `my_mktime` for `my_timegm` seems to fix the problem, but I'm not sure about the implications.

```diff
diff --git a/src/builtin.c b/src/builtin.c
index cc8a36e..db0f636 100644
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1415,7 +1415,7 @@ static jv f_mktime(jq_state *jq, jv a) {
   struct tm tm;
   if (!jv2tm(a, &tm))
     return jv_invalid_with_msg(jv_string("mktime requires parsed datetime inputs"));
-  time_t t = my_mktime(&tm);
+  time_t t = my_timegm(&tm);
   if (t == (time_t)-1)
     return jv_invalid_with_msg(jv_string("invalid gmtime representation"));
   if (t == (time_t)-2)
```

PR #1800, opened 2019-01-30
Unify timegm fallbacks into my_mktime

muhmuhten:
my_timegm was introduced in 1900c7 to add a mktime fallback used in
4a6241, the subsequently removed in c53823. As a result of this code
shuffling, the jq mktime function (f_mktime) wound up using a fallback
even on systems where timegm is available, with incorrect results in
some time zones with DST, e.g.

% TZ=America/New_York jq -n '"2018-08-31T00:00:00Z"|fromdate|todate'
"2018-08-31T01:00:00Z"

This undoes 1900c7 and moves the workaround it added into the #else of
my_mktime.

The original changes were made by @nicowilliams, so it would be nice if you could look over these to make sure they make sense.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1225,10 +1225,21 @@ static int setenv(const char *var, const char *val, int ovr)
  *
  * Returns (time_t)-2 if mktime()'s side-effects cannot be corrected.
  */
-static time_t my_timegm(struct tm *tm) {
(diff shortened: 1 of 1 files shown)


muhmuhten:
This is intended to fix #1713.

nicowilliams:
Thanks!!

Issue #1804, opened 2019-02-02
Crash in jq_next calling jvp_utf8_decode_length on master 3c5b141


pallas:
At 3c5b141 compiled with CC=afl-gcc and otherwise default configure, the following test case causes a SEGFAULT. Because it contains strange characters, it is xxd encoded.
````
00000000: 5b6d 6174 6368 2822 2820 292a 223b 2022  [match("( )*"; "
00000010: 6722 295d 0a22 6180 6322 0a5b 7b22 6f66  g")]."a.c".[{"of
00000020: 6673 6574 223a 302c 2022 6c65 6e67 7468  fset":0, "length
00000030: 223a 302c 2022 3a30 2c04 2222 6565 6565  ":0, ":0,.""eeee
00000040: 6565 6565 6565 6565 6565 6570 7475 7265  eeeeeeeeeeepture
00000050: 7322 3a5b 5d7d 2c7b 226f 6666 7365 7422  s":[]},{"offset"
00000060: 3a31 2c20 226c 656e 6774 6822 3a30 2c20  :1, "length":0,
00000070: 2273 7440 696e 6722 3a22 222c 2022 6361  "st@ing":"", "ca
00000080: 7074 7572 6573 223a 5b5d 7d2c 7b22 6f66  ptures":[]},{"of
00000090: 6673 6574 223a 402c 2322 6c65 6e67 7468  fset":@,#"length
000000a0: 223a 302c 2022 7374 7269 6e67 223a 2222  ":0, "string":""
000000b0: 2c20 2263 6170 7475 7265 7322 3a5b 5d7d  , "captures":[]}
000000c0: 5d0a                                     ].
````

The backtrace is as follows.
````
Testing '[match("( )*"; "g")]' at line number 1
*** Expected result is invalid on line 3: [{"offset":0, "length":0, ":0,""eeeeeeeeeeeeeeeptures":[]},{"offset":1, "length":0, "st@ing":"", "captures":[]},{"offset":@,#"length":0, "string":"", "captures":[]}]

Program received signal SIGSEGV, Segmentation fault.
f_match (jq=<optimized out>, input=..., regex=..., modifiers=..., testmode=...) at src/builtin.c:869
869               fr += jvp_utf8_decode_length(*fr);
#0  f_match (jq=<optimized out>, input=..., regex=..., modifiers=..., testmode=...) at src/builtin.c:869
#1  0x000055555556babd in jq_next (jq=0x555555693730) at src/execute.c:855
[...]

PR #1805, opened 2019-02-03
Fix UTF8 decoding overrun

pallas:


nicowilliams:
To recap, the problem is that `jvp_utf8_decode_length()` could not figure out the length of the UTF-8 encoding of `U+FFFD` `REPLACEMENT CHARACTER`, which is inserted by `jvp_string_copy_replace_bad()` in lieu of the offending `0x80` in the input.

nicowilliams:
I single-stepped through it in the debugger, and `jvp_utf8_decode_length()` correctly returned `1` for `'a'`, `3` for `U+FFFD`, `1` for `"c", and `1` for the `NUL`, but it shouldn't have been called for the `NUL`.  The principal bug is in `f_match()`.

src/builtin.c
```diff
@@ -935,7 +935,7 @@ static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
       break;
     } else { /* Error */
       UChar ebuf[ONIG_MAX_ERROR_MESSAGE_LEN];
-      onig_error_code_to_str(ebuf, onigret, einfo);
+      onig_error_code_to_str(ebuf, onigret, &einfo);
```
nicowilliams: Because it's varargs, this wasn't caught.  Ugh.  Thanks.

src/jv_unicode.c
```diff
@@ -87,7 +87,8 @@ int jvp_utf8_decode_length(char startchar) {
 	if ((startchar & 0x80) == 0) return 1;         // 0___ ____
 	else if ((startchar & 0xE0) == 0xC0) return 2; // 110_ ____
 	else if ((startchar & 0xF0) == 0xE0) return 3; // 1110 ____
-	else return 4;                                 // 1111 ____
+	else if ((startchar & 0xF0) == 0xF0) return 4; // 1111 ____
```
nicowilliams: Note that this change isn't needed in order to fix #1804.  This function correctly decoded the length of the UTF-8-encoded `U+FFFD` as `3`.

nicowilliams: Actually, this commit, I think is not needed at all.  At 88 we know at least the high bit is set, and if it's actually two are set, we return 2.  At 89 we know at least the two high bits are set, and if it's actually three, then we return three.  At 90 it can only be that the four high bits are set, so the new condition at 90 is always true.

nicowilliams:
@pallas I think the middle commit is not needed at all, and that `jvp_utf8_decode_length()` is correct.  Can you confirm my reasoning?

src/jv_unicode.c
pallas: Doesn't it skip the 10__ ____ case if a broken continuation byte sneaks in? But assuming that the UTF-8 is well-formed, yes, this change is unnecessary.

src/builtin.c
pallas: n/p

pallas:
> @pallas I think the middle commit is not needed at all, and that `jvp_utf8_decode_length()` is correct. Can you confirm my reasoning?

Yes, I think the reasoning is correct assuming that the bytes are valid UTF-8.

pallas:
> I single-stepped through it in the debugger, and `jvp_utf8_decode_length()` correctly returned `1` for `'a'`, `3` for `U+FFFD`, `1` for `"c", and `1`for the`NUL`, but it shouldn't have been called for the `NUL`. The principal bug is in `f_match()`.

Ah, makes sense.

src/jv_unicode.c
nicowilliams: I take your point.  Perhaps `f_match()` should use `jvp_utf8_next()` and re-encode it.  But as long as only `f_match()` calls this, and `f_match()` checks that there are as many bytes left in the put as this function returns, then it's safe.  If standard C had local functions...

nicowilliams:
Dropped the middle commit and pushed.  Thanks!

nicowilliams:
Yeah, it was important to notice that this code was never seeing `0x80` because it was getting replaced with `U+FFFD`.

PR #1806, opened 2019-02-06
Export input/0 directly from C, removing _input/0

muhmuhten:
The distinction between input/0 and _input/0 has been unnecessary since
a9c613..7dc34b, which moved end-of-input handling into C and replaced
the jq definition of input with `def input: _input;`


nicowilliams:
Thanks!

Issue #1380, opened 2017-04-02
Linking is O(M_binders * N_body_instructions), could be O(N' log M) or even O(N'), where N' is just unbound instructions


nicowilliams:
Linking is O(M_binders * N_body_instructions), could be O(N' log M) or even O(N'), where N' is just unbound instructions.

As we add builtins, linking will slow.

We've already optimized binding of global `$var`s defined with `--arg*`.

We could now optimize the binding of builtins and binders exported by modules.

There are two independent things to do about this:

 - Build a list of unbound instructions as we do any binding so that once we have that list we need only traverse it (this is where  `N'` comes from).

 - Sort the globals to bind by name and ariness, then traverse the body (or list of unbound insts in it) to find unbound insts and then binary search the table of binders (this is where `log M` comes from).

   Alternatively build a hash table of the globals to get down to `O(N')` (assuming we really get `O(1)` from the hash table).

   We could build the sorted / hash tables as we compile a module.

   The cost of sorting or building a hash table has to be taken into account.  For large programs it will be negligible, as it is a one-time cost, but most jq programs are small...  A hash table will probably perform best in this sense, since insertion should be fast too.

   The C-coded builtins can be kept sorted manually.  No need to sort it at run-time, nor hash it.

   `src/builtin.jq` can't be sorted because some of those builtins depend on others.

nicowilliams:
We can add `inst *next_unbound` and `inst *prev_unbound` fields to `inst`, but where `inst` values are generated we don't have the context to link them up.  So such a list would have to be setup the first time we bind.

Another thought is that we may eventually want to close the set of builtins for the "jq" module (as it were) or otherwise compile only the ones we need.

nicowilliams:
I think I'll do the following:

 - phase 1: use binary search of binder arrays instead of lists
    - keep `block_bind_subblock()` as-is
    - add a `block_bind_subblock2()` that takes an array (not a block) of binder `inst`s
    - make `gen_cbinding()`, `bind_bytecoded_builtins()`, and friends, build a sorted array of binders and call `block_bind_subblock2()` to bind them using binary search of the binders for each unbound `inst`

 - phase 2: use a list of unbound `inst`s for binding global symbols
    - add a `block_gather_unbound()` function to make a list of unbound `inst`s
    - add a `block_bind_unbounds()` function that works like `block_bind_subblock2()` but takes the unbound `inst` list, and make `gen_cbinding()`, `bind_bytecoded_builtins()`, and friends use it

[...]

PR #1807, opened 2019-02-06
Speed linking

nicowilliams:
This is an attempt to begin ameliorating #1380...  Let's see if this passes tests!


muhmuhten:
From running it alone, somewhat mixed results―the degenerate libraries with thousands of unused symbols load much faster, but it actually takes longer to load builtins and run the empty program.

nicowilliams:
Closing for now.

PR #1808, opened 2019-02-07
Define error/1 in terms of error/0

muhmuhten:
The parser generates a call to error/0 on break. Having that exported
from C directly removes a language dependency on builtin.jq.

In any case, error/0 seems to be the more fundamental operation, seeing
as the old C impementation of error/1 did nothing useful with its input.


nicowilliams:
Thanks!

PR #1809, opened 2019-02-07
 Work around fgets()'s lack of length reporting

### LINKED ISSUE #1504: Raw input inconsistently loses NULs at end of file
```
% printf 'a\0b\0' | jq -R .
"a\u0000b"
% printf 'a\0\nb\0\n' | jq -R . # this one's okay
"a\u0000"
"b\u0000"
% printf 'a\n\0b\0' | jq -Rs . # loses the NUL
"a\n\u0000b"
% printf 'a\n\0b\n\0' | jq -Rs . #doesn't lose the NUL
"a\n\u0000b\n\u0000"
% printf 'a\n\0b\n\0\0' | jq -Rs . # loses one NUL but not the other
"a\n\u0000b\n\u0000"
% seq -f 'a' 2048 | tr 'a\n' '\0' | jq -R length # correct
4096
% seq -f 'a' 2047 | tr 'a\n' '\0' | jq -R length # should be 4094
1
```

The problem is basically that because we use `fgets`, we have no idea how many bytes were actually read, and the current implementation (see util.c:337) is a hack that guesses based on strlen or the buffer size depending on the return value. This sort of works if the input doesn't contain NULs, but the result when there are NULs fails in incomprehensible ways.

muhmuhten:
Stop losing \0s from the end of files because we don't know how much
fgets() read into our buffer. If we prefill the buffer with anything
other than zeroes, we can actually tell the difference between the
terminator and unused space in the buffer.

This is possibly more expensive, but at least it should be correct.

Fixes #1504

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -10,7 +10,7 @@ def max_by(f): _max_by_impl(map([f]));
 def min_by(f): _min_by_impl(map([f]));
 def add: reduce .[] as $x (null; . + $x);
 def del(f): delpaths([path(f)]);
-def _assign(paths; value): value as $v | reduce path(paths) as $p (.; setpath($p; $v));
+def _assign(paths; $value): reduce path(paths) as $p (.; setpath($p; $value));
 def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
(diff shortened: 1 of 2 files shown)


nicowilliams:
I split your commit in two, FYI.

nicowilliams:
Clever!  Thanks!

PR #1812, opened 2019-02-08
Bind data as $var when already resolved too

muhmuhten:
b5560d84 lets the data imported by `import "data" as $data` be accessed as either $data or $data::data, except that when "data" has already been imported before, it isn't bound to $data because that code isn't in the other if branch.

### DIFF
--- a/src/linker.c
+++ b/src/linker.c
@@ -263,26 +263,38 @@ static int process_dependencies(jq_state *jq, jv jq_origin, jv lib_origin, block
       jv_free(lib_origin);
       return 1;
     }
-    uint64_t state_idx = 0;
-    for (; state_idx < lib_state->ct; ++state_idx) {
-      if (strcmp(lib_state->names[state_idx],jv_string_value(resolved)) == 0)
-        break;
-    }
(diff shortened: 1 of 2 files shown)


nicowilliams:
Do we need a test?

muhmuhten:
Definitely--my patch doesn't actually fix the problem, and in fact doesn't do anything useful...

muhmuhten:
Unifying the logic looked too messy, so I tried splitting it up as cleanly as possible instead.

Apparently 'import "data" as $d; import "data" as $e;" used to bind $d::d and $e::d (!) by reusing the resolved library, which is both not as documented *and* a violation of common sense.

It does $d::d, $d, $e::e, and $e now.

nicowilliams:
Oof.  So sorry.

muhmuhten:
Wrote a test, and _actually_ fixed the behaviour, I think...

nicowilliams:
We need to change `.travis.yml` to cat all `tests/*.log` corresponding to `tests/*.trs` files that indicate test failures...  I did something like that for Heimdal recently.

nicowilliams:
Thanks!

PR #1814, opened 2019-02-08
Attempt to improve linking time (#1380)

muhmuhten:
Here's my stab at the linking problem. This is definitely a band-aid that doesn't solve the underlying issues, and the implementation is iffy.

The idea here is to mark whole definitions when all the symbols inside are bound, then avoid descending into them on later iterations.

It's still pretty far from ideal, but at least it manages to cut close to half the test suite runtime.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -49,9 +49,10 @@ struct inst {
   // Unbound instructions (references to other things that may or may not
   // exist) are created by "gen_foo_unbound", and bindings are created by
   // block_bind(definition, body), which binds all instructions in
-  // body which are unboudn and refer to "definition" by name.
+  // body which are unbound and refer to "definition" by name.
   struct inst* bound_by;
   char* symbol;
+  int bound_flags;
 
   int nformals;
   int nactuals;
@@ -73,6 +74,7 @@ static inst* inst_new(opcode op) {
   i->bytecode_pos = -1;
   i->bound_by = 0;
   i->symbol = 0;
+  i->bound_flags = 0;
   i->nformals = -1;
   i->nactuals = -1;
   i->subfn = gen_noop();
@@ -324,7 +326,7 @@ static int block_count_refs(block binder, block body) {
   return nrefs;
 }
 
-static int block_bind_subblock(block binder, block body, int bindflags, int break_distance) {
+static int block_bind_subblock_inner(block binder, block body, int* bound_flags, int bindflags, int break_distance) {
   assert(block_is_single(binder));
   assert((opcode_describe(binder.first->op)->flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD));
   assert(binder.first->symbol);
@@ -336,6 +338,9 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
     binder.first->nformals = block_count_formals(binder);
   int nrefs = 0;
   for (inst* i = body.first; i; i = i->next) {
+    if ((i->bound_flags & bindflags) == bindflags)
+      continue;
+
     int flags = opcode_describe(i->op)->flags;
     if ((flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD) && i->bound_by == 0 &&
         (!strcmp(i->symbol, binder.first->symbol) ||
@@ -357,12 +362,24 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       // a break whenever we come across a STOREV of *anonlabel...
       break_distance++;
     }
+
+    if ((flags & bindflags) != (bindflags & ~OP_BIND_WILDCARD) || i->bound_by != 0) {
+      i->bound_flags |= bindflags;
+    }
+
     // binding recurses into closures
-    nrefs += block_bind_subblock(binder, i->subfn, bindflags, break_distance);
+    nrefs += block_bind_subblock_inner(binder, i->subfn, &i->bound_flags, bindflags, break_distance);
     // binding recurses into argument list
-    nrefs += block_bind_subblock(binder, i->arglist, bindflags, break_distance);
+    nrefs += block_bind_subblock_inner(binder, i->arglist, &i->bound_flags, bindflags, break_distance);
+
+    *bound_flags &= i->bound_flags;
   }
   return nrefs;
+}
+
+static int block_bind_subblock(block binder, block body, int bindflags, int break_distance) {
+  int bound_flags;
+  return block_bind_subblock_inner(binder, body, &bound_flags, bindflags, break_distance);
 }
 
 static int block_bind_each(block binder, block body, int bindflags) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ jobs:
                         - valgrind
                         - bison
                         - automake
+                        - git
 
             before_install:
                 - uname -s
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,8 +104,8 @@ jobs:
                 - uname -s
                 - brew update
                 - brew install flex bison
-                - rvm install ruby-1.9.3-p551
-                - rvm use 1.9.3
+                - rvm install ruby-2.3.0
+                - rvm use 2.3.0
                 - gem install bundler
                 - rm src/{lexer,parser}.{c,h}
                 - sed -i.bak '/^AM_INIT_AUTOMAKE(\[-Wno-portability 1\.14\])$/s/14/11/' modules/oniguruma/configure.ac


src/compile.c
```diff
@@ -336,6 +338,9 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
     binder.first->nformals = block_count_formals(binder);
   int nrefs = 0;
   for (inst* i = body.first; i; i = i->next) {
+    if ((i->bound_flags & bindflags) == bindflags)
```
nicowilliams: And this is the thing that makes it go faster?  I'll review when I'm fresh.  Right now I don't get it.

nicowilliams:
What think ye of:

```
index bd0cb6d..9ddf3ea 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -27,6 +27,7 @@
 struct inst {
   struct inst* next;
   struct inst* prev;
+  struct inst* next_unbound;

   opcode op;

@@ -68,7 +69,7 @@ struct inst {

 static inst* inst_new(opcode op) {
   inst* i = jv_mem_alloc(sizeof(inst));
-  i->next = i->prev = 0;
+  i->next = i->prev = i->next_unbound = 0;
   i->op = op;
   i->bytecode_pos = -1;
   i->bound_by = 0;
@@ -103,6 +104,23 @@ int block_is_single(block b) {
   return b.first && b.first == b.last;
 }

+static void block_unbounds(block b) {
+  if (!b.first || b.first->next_unbound)
+    return;
+  inst *last = 0;
+  for (inst *i = b.last; i; i = i->prev) {
+    if ((!i->symbol || i->bound_by) &&
+        block_is_noop(i->subfn) &&
+        block_is_noop(i->arglist))
+      continue;
+    block_unbounds(i->subfn);
+    block_unbounds(i->arglist);
+
+    /* i->next_unbound == i indicates we've done this operation */
+    i->next_unbound = last ? last : i;
+  }
+}
+
 static inst* block_take(block* b) {
   if (b->first == 0) return 0;
   inst* i = b->first;
@@ -238,6 +256,7 @@ static void inst_join(inst* a, inst* b) {
   assert(!b->prev);
   a->next = b;
   b->prev = a;
+  a->next_unbound = b;
 }

 void block_append(block* b, block b2) {
@@ -331,11 +350,15 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
   assert(binder.first->bound_by == 0 || binder.first->bound_by == binder.first);
   assert(break_distance >= 0);

+
   binder.first->bound_by = binder.first;
   if (binder.first->nformals == -1)
     binder.first->nformals = block_count_formals(binder);
+
+  block_unbounds(body);
+
   int nrefs = 0;
-  for (inst* i = body.first; i; i = i->next) {
+  for (inst* i = body.first; i; i = i->next_unbound) {
     int flags = opcode_describe(i->op)->flags;
     if ((flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD) && i->bound_by == 0 &&
         (!strcmp(i->symbol, binder.first->symbol) ||
@@ -361,6 +384,8 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
     nrefs += block_bind_subblock(binder, i->subfn, bindflags, break_distance);
     // binding recurses into argument list
     nrefs += block_bind_subblock(binder, i->arglist, bindflags, break_distance);
+    if (i->next_unbound == i)
+      break;
   }
   return nrefs;
 }
```
?

Passes tests w/o valgrind.  Now checking with valgrind.  It's only a start, but...

What does this do?  It builds a shadow AST linking just instructions with unbound symbol references or with sub-trees.  Then it traverses this for binding sub-blocks.

Now, there's no `prev_unbound` in this case, and I'm abusing the new `next_unbound` link to indicate whether the unbound-instruction scan has been done...  With a `prev_unbound` we could remove from the list all bound instructions.  Also, we could have some special values of `next_unbound` to indicate that the entire `subfn` sub-tree is bound.  If we add these enhancements then the amount of work done by each call to `block_bind_subblock()` will decrease with each symbol resolution, until there's nothing left to be done.

Note that I didn't add new `first_unbound`/`last_unbound` fields to `struct block`.  That's because `block`s are generally passed by value, and changing that would be too large a change.

muhmuhten:
I'm not sure block_unbounds is doing what you wanted? It looks like `b.first->next_unbound` is going to be set for any block with more than one instruction (by inst_join), and otherwise `last` is never set, which I think is wrong if the block does contain more than one instruction.

I'll note that when I tried the approach from https://github.com/stedolan/jq/issues/1380#issuecomment-461201718, it didn't seem to work as well as I hoped, though I probably didn't get the details right.

As far as I can tell, figuring out when the whole subfn/arglist tree is bound is actually the most important piece, as without that we have to assume that any subfn may contain unbound symbols and can never skip them, so the only thing we can jump over are leaf symbols and constants and the like.

In fact, I think we need a prev_unbound pointer and modify prev_unbound->next_unbound once we know that the whole tree is bound to skip anything other than constants.

muhmuhten:
Hm, if it's safe to assume that any instruction with a binding has a non-NULL `symbol`, though, a cleaner implementation is possible.

nicowilliams:
So... setting `next_unbound` on an `inst` that's not part of the unbound sub-tree is harmless -- that's why that's there.  And this patch does pass tests, ...

...but it doesn't handle removal of `inst`s from the unbound sub-tree, which is critical to improving performance.  However, as you point out, just marking sub-trees as all-bound would have most of the same effect as removing `inst`s from the sub-tree as they get bound -- that's a very nice idea that will allow me to stop trying to doubly-link the unbound `inst` sub-trees! and it will make it much easier to reason about the whole thing.

nicowilliams:
Ah, and if it's just the `subfn` and `arglist` sub-trees that we want to know are all bound... maybe we don't need a `next_unbound` list at all.  Just have a bit in the first `inst` in each sub-tree indicate whether it's all bound, and then in `block_bind_subblock()` we can skip all-bound sub-trees and mark all-bound sub-trees as such we find them.  Let me try that approach -- it sounds fantastically simpler.

nicowilliams:
So, here's something that's very close -- it fails only `shtest`and passes the others (w/o valgrind though).  The failure in `shtest` looks like this:

```
(gdb) run
Starting program: /home/nico/ws/jq/jq -c .\ as\ \$d\|path\(..\)\ as\ \$p\|\$d\|getpath\(\$p\)\|scalars_or_empty\|\[\$p,.\] tests/torture/input0.json

Program received signal SIGSEGV, Segmentation fault.
0x000055555559819a in locfile_locate (l=0x0, loc=..., fmt=0x5555555de700 "jq: error: %s/%d is not defined") at src/locfile.c:80
80          jq_report_error(l->jq, jv_string_fmt("jq: error: %s\n<unknown location>", jv_string_value(m1)));
(gdb) p jv_string_value(m1)
$1 = 0x555555875020 "jq: error: recurse/0 is not defined"
(gdb)
```

so it's not ready, but it's very very simple and on the right track (I think):

```
diff --git a/modules/oniguruma b/modules/oniguruma
--- a/modules/oniguruma
+++ b/modules/oniguruma
@@ -1 +1 @@
-Subproject commit 6fa38f4084b448592888ed9ee43c6e90a46b5f5c
+Subproject commit 6fa38f4084b448592888ed9ee43c6e90a46b5f5c-dirty
index bd0cb6d..978f9b2 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -64,6 +64,7 @@ struct inst {
   struct bytecode* compiled;

   int bytecode_pos; // position just after this insn
+  int all_bound;
 };

 static inst* inst_new(opcode op) {
@@ -77,6 +78,7 @@ static inst* inst_new(opcode op) {
   i->nactuals = -1;
   i->subfn = gen_noop();
   i->arglist = gen_noop();
+  i->all_bound = 0;
   i->source = UNKNOWN_LOCATION;
   i->locfile = 0;
   return i;
@@ -334,9 +336,15 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
   binder.first->bound_by = binder.first;
   if (binder.first->nformals == -1)
     binder.first->nformals = block_count_formals(binder);
+
+  if (!body.first || body.first->all_bound)
+    return 0;
+
   int nrefs = 0;
+  int all_bound = 1;
   for (inst* i = body.first; i; i = i->next) {
     int flags = opcode_describe(i->op)->flags;
+
     if ((flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD) && i->bound_by == 0 &&
         (!strcmp(i->symbol, binder.first->symbol) ||
          // Check for break/break2/break3; see parser.y
@@ -356,12 +364,20 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       // Increment the break distance required for this binder to match
       break_distance++;
+      all_bound = 0;
+    } else {
+      all_bound = 0;
     }
     // binding recurses into closures
     nrefs += block_bind_subblock(binder, i->subfn, bindflags, break_distance);
+    if (i->subfn.first && !i->subfn.first->all_bound)
+      all_bound = 0;
     // binding recurses into argument list
     nrefs += block_bind_subblock(binder, i->arglist, bindflags, break_distance);
+    if (i->arglist.first && !i->arglist.first->all_bound)
+      all_bound = 0;
   }
+  body.first->all_bound = all_bound;
   return nrefs;
 }
```

muhmuhten:
https://github.com/muhmuhten/jq/commit/38f766ace8623665449c474023896f83eb3e941d is a cleaner (no bitfield mangling) rework on this one.

nicowilliams:
@muhmuhten this is exciting.  Let's see who arrives at a fix first :)  More seriously, I don't think it's a counter that you want.  Just have `block_bind_subblock()` mark all-bound sub-trees as such by setting `body.first->all_bound = 1` whenever all of a tree and all its subtrees are bound.

nicowilliams:
Ahhh, when appending blocks to blocks we need to reset the `all_bound` boolean on the left side...

nicowilliams:
I.e., I think I need this hunk too:

```
@@ -244,6 +246,7 @@ void block_append(block* b, block b2) {
   if (b2.first) {
     if (b->last) {
       inst_join(b->last, b2.first);
+      b->first->all_bound = b->first->all_bound && b2.first->all_bound;
     } else {
       b->first = b2.first;
     }
```

muhmuhten:
Yes, turns out that having a counter is pretty superfluous when all I'm doing is checking whether it's zero or not.

Possibly keeping the all-bound flag in the first inst of a block can avoid the nasty pointer-passing shenanigans I'm pulling, but I think that still requires going into subfn and arglist each time. It might not be that big an impact, though.

nicowilliams:
@muhmuhten try this patch:

```
index bd0cb6d..e047579 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -64,6 +64,7 @@ struct inst {
   struct bytecode* compiled;

   int bytecode_pos; // position just after this insn
+  int all_bound;
 };

 static inst* inst_new(opcode op) {
@@ -77,6 +78,7 @@ static inst* inst_new(opcode op) {
   i->nactuals = -1;
   i->subfn = gen_noop();
   i->arglist = gen_noop();
+  i->all_bound = 0;
   i->source = UNKNOWN_LOCATION;
   i->locfile = 0;
   return i;
@@ -109,6 +111,8 @@ static inst* block_take(block* b) {
   if (i->next) {
     i->next->prev = 0;
     b->first = i->next;
+    if (b->first)
+      b->first->all_bound = 0;
     i->next = 0;
   } else {
     b->first = 0;
@@ -244,6 +248,7 @@ void block_append(block* b, block b2) {
   if (b2.first) {
     if (b->last) {
       inst_join(b->last, b2.first);
+      b->first->all_bound = b2.first->all_bound = 0;
     } else {
       b->first = b2.first;
     }
@@ -334,9 +339,15 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
   binder.first->bound_by = binder.first;
   if (binder.first->nformals == -1)
     binder.first->nformals = block_count_formals(binder);
+
+  if (!body.first || body.first->all_bound)
+    return 0;
+
   int nrefs = 0;
+  int all_bound = 1;
   for (inst* i = body.first; i; i = i->next) {
     int flags = opcode_describe(i->op)->flags;
+
     if ((flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD) && i->bound_by == 0 &&
         (!strcmp(i->symbol, binder.first->symbol) ||
          // Check for break/break2/break3; see parser.y
@@ -349,6 +360,8 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       if (i->nactuals == -1 || i->nactuals == binder.first->nformals) {
         i->bound_by = binder.first;
         nrefs++;
+      } else {
+        all_bound = 0;
       }
     } else if ((flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD) && i->bound_by != 0 &&
                !strncmp(binder.first->symbol, "*anonlabel", sizeof("*anonlabel") - 1) &&
@@ -356,12 +369,19 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       // Increment the break distance required for this binder to match
       break_distance++;
+    } else {
+      all_bound = 0;
     }
     // binding recurses into closures
     nrefs += block_bind_subblock(binder, i->subfn, bindflags, break_distance);
+    if (i->subfn.first && !i->subfn.first->all_bound)
+      all_bound = 0;
     // binding recurses into argument list
     nrefs += block_bind_subblock(binder, i->arglist, bindflags, break_distance);
+    if (i->arglist.first && !i->arglist.first->all_bound)
+      all_bound = 0;
   }
+  body.first->all_bound = all_bound;
   return nrefs;
 }

```

nicowilliams:
That patch has a slight pessimisation.  I think the `block_append()` hunk should be:

```
@@ -244,6 +248,7 @@ void block_append(block* b, block b2) {
   if (b2.first) {
     if (b->last) {
       inst_join(b->last, b2.first);
+      b->first->all_bound = b->first->all_bound && b2.first->all_bound;
     } else {
       b->first = b2.first;
     }
```

nicowilliams:
And no changes should be needed in `block_take()`.  Testing w/ valgrind.

nicowilliams:
I do think this patch works then -- valgrinding tests will take forever though.  Next we should measure how much performance we gain from it.

nicowilliams:
> Possibly keeping the all-bound flag in the first inst of a block can avoid the nasty pointer-passing shenanigans I'm pulling, but I think that still requires going into subfn and arglist each time. It might not be that big an impact, though.

Well, in my patch we have this right before the loop in `block_bind_subblock()`:

```
+  if (!body.first || body.first->all_bound)
+    return 0;
```

That's the don't-bother-with-all-bound-`subfn`s/`arglist`s optimization.  I could pull it up a bit to avoid a call that returns quickly (and so avoids wasting time on all those assertions):

```
 373     // binding recurses into closures
 374     if (i->subfn.first && !i->subfn.first->all_bound) {
 375       nrefs += block_bind_subblock(binder, i->subfn, bindflags, break_distance);
 376       if (i->subfn.first && !i->subfn.first->all_bound)
 377         all_bound = 0;
 378     }
 379     // binding recurses into argument list
 380     if (i->arglist.first && !i->arglist.first->all_bound) {
 381       nrefs += block_bind_subblock(binder, i->arglist, bindflags, break_distance);
 382       if (i->arglist.first && !i->arglist.first->all_bound)
 383         all_bound = 0;
 384     }
```

That's almost certainly worth doing.

nicowilliams:
I'm seeing a noticeable difference in performance, but it's on the order of 50%, but I'm not measuring rigorously.

nicowilliams:
Timing `jq --run-tests` on a subset of `tests/jq.test` I'm getting a 39% improvement.  It's... not as good as I'd been hoping, but it's a start, and this is a patch that looks like it will pass tests w/ valgrind -- that's not nothing!

Armed with this patch we could investigate going back to the `next_unbound` thing and having `block_bind_subblock()` unlink the `inst`s it binds.  I think that now will look simpler.  Then next up we should try your patch to stop dropping unreferenced blocks.  Altogether that might yield the gains we're hoping for.

muhmuhten:
Didn't look like it was doing much on either "how long does jqtest take to run" or the `repeat 100 jq -n empty` microbenchmark, but much better after.

Order of 50% sounds about right to me; 4caa3aa was around 40%, while 38f766a is at about 70% after dropping block_drop_unreferenced.

muhmuhten:
>  It's... not as good as I'd been hoping,

Ah, how much we've probably both been feeling this particular sentiment.

nicowilliams:
> > It's... not as good as I'd been hoping,

> Ah, how much we've probably both been feeling this particular sentiment.

Yeah...

We can then start doing things like:

 - build an object of unbound symbols in the program to which we're binding builtins, then skip bindings of bodies that don't need some binder
 - add your no-builtins patch with some command-line option (`--no-builtins`, say)
 - split up `src/builtin.jq` into several modules and allow `import jq::some::builtins::subset ...`

then users can start using jq this way to speed up startup.

nicowilliams:
FYI, my patch passes tests w/ valgrind, so I'm inclined to push it.  I'll open a PR.  Can you review it?  If not I'll get @wtlangford to do it.

nicowilliams:
I've opened #1816.

muhmuhten:
fwiw the business of reaching into subfn.first/arglist.first is what I was trying to avoid by passing a pointer down.

I think https://github.com/muhmuhten/jq/commit/25ec73637af5158ed43d96688fd330736d90fd78 (currently the head of my linking_3 branch), which is my latest attempt *without* stubbing block_drop_unreferenced for fair comparison, still performs a fair better than my reassembly of your last patchset (https://github.com/muhmuhten/jq/commit/bd75c449eedf82ce4abbbc841bf1ab574ce58f12), despite skipping subfn/arglist together at greater granularity. Mind taking that out for a run?

muhmuhten:
Guess this PR is superseded now.

nicowilliams:
I'll take a look, but i think I'm about to turn into a pumpkin and won't have time again till Sunday.

nicowilliams:
I wouldn't say this PR is superseded.  I opened mine to get Travis going, not to supersede this one.

nicowilliams:
I wonder if https://github.com/muhmuhten/jq/commit/25ec73637af5158ed43d96688fd330736d90fd78 is faster because it adds fewer branches.  I think that's very likely.

nicowilliams:
You https://github.com/muhmuhten/jq/commit/25ec73637af5158ed43d96688fd330736d90fd78 gets me 55% faster than master -- that's better than my 39%.  And with the stubbing out of `block_drop_unreferenced()` it's 77% faster.  So I'll take https://github.com/muhmuhten/jq/commit/25ec73637af5158ed43d96688fd330736d90fd78  if you make a PR for it alone, then we'll see about `block_drop_unreferenced()` (I still want it for the `--debug-dump-disasm` case).

PR #1816, opened 2019-02-09
Speed up linking by 39% (partial fix to #1380)

nicowilliams:
This patch improves linking performance by marking parse sub-trees as
"all-bound" when all instructions in them are bound, and then this is
used to avoid doing pointless work when asked to bind an all-bound
sub-tree.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -64,6 +64,7 @@ struct inst {
   struct bytecode* compiled;
 
   int bytecode_pos; // position just after this insn
+  int all_bound;
 };
 
 static inst* inst_new(opcode op) {
@@ -77,6 +78,7 @@ static inst* inst_new(opcode op) {
   i->nactuals = -1;
   i->subfn = gen_noop();
   i->arglist = gen_noop();
+  i->all_bound = 0;
   i->source = UNKNOWN_LOCATION;
   i->locfile = 0;
   return i;
@@ -244,6 +246,7 @@ void block_append(block* b, block b2) {
   if (b2.first) {
     if (b->last) {
       inst_join(b->last, b2.first);
+      b->first->all_bound = b->first->all_bound && b2.first->all_bound;
     } else {
       b->first = b2.first;
     }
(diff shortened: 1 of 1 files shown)


muhmuhten:
This is faster for reasons that are mysterious to me, and is the same logic as 25ec736.

Passes tests with valgrind.

```diff
diff --git a/src/compile.c b/src/compile.c
index c22ace3..2280cb6 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -358,8 +358,6 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       if (i->nactuals == -1 || i->nactuals == binder.first->nformals) {
         i->bound_by = binder.first;
         nrefs++;
-      } else {
-        all_bound = 0;
       }
     } else if ((flags & bindflags) == (bindflags & ~OP_BIND_WILDCARD) && i->bound_by != 0 &&
                !strncmp(binder.first->symbol, "*anonlabel", sizeof("*anonlabel") - 1) &&
@@ -367,9 +365,11 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       // Increment the break distance required for this binder to match
       // a break whenever we come across a STOREV of *anonlabel...
       break_distance++;
-    } else {
-      all_bound = 0;
     }
+
+    if (i->symbol && !i->bound_by)
+      all_bound = 0;
+
     // binding recurses into closures
     if (i->subfn.first && !i->subfn.first->all_bound) {
       nrefs += block_bind_subblock(binder, i->subfn, bindflags, break_distance);
```

nicowilliams:
> This closes pretty much the entire gap.

Meaning?  It doesn't seem any faster for me :(

muhmuhten:
Hm, interesting. It's not as much of a difference, or it might only be true on my system...

Basically I'm seeing ~20% speedup on `jq -n empty` and jqtest for 4bddb4228835bd063f9b8f71eb7f47d742326f41 vs 0b49cc80ecd9ce6d09cdd5f417461ddf9c880aed and

nicowilliams:
@muhmuhten has a better version of this!

PR #1817, opened 2019-02-09
Improve linking time by marking subtrees with unbound symbols

muhmuhten:
As far as we know, this is faster than #1816 and better than #1814.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -49,9 +49,10 @@ struct inst {
   // Unbound instructions (references to other things that may or may not
   // exist) are created by "gen_foo_unbound", and bindings are created by
   // block_bind(definition, body), which binds all instructions in
-  // body which are unboudn and refer to "definition" by name.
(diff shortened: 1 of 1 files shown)


src/compile.c
```diff
@@ -357,14 +365,25 @@ static int block_bind_subblock(block binder, block body, int bindflags, int brea
       // a break whenever we come across a STOREV of *anonlabel...
       break_distance++;
     }
+
+    i->any_unbound = (i->symbol && !i->bound_by);
```
nicowilliams: Do we not need to check the state of the `subfn` and `arglist` sub-trees too here?

muhmuhten: We do, which is why we pass a pointer to it down into the block_bind_subblock_inner calls on subfn and arglist. It gets set to 1 at L377 in those calls.

nicowilliams:
Thank you so much @muhmuhten, this was long needed!

PR #1818, opened 2019-02-09
Make sure to free in jv_string_indexes

muhmuhten:
2660b04a731568c54eb4b91fe811d81cbbf3470b makes jqtest fail valgrind since it skips the frees when idxlen == 0.

This was kind of an interesting experience when making other changes that could plausibly have leaked memory.


nicowilliams:
Thanks!

Issue #1813, opened 2019-02-08
Is block_drop_unreferenced useful?


muhmuhten:
`block_drop_unreferenced()` is called after loading the program and again after binding builtins and I think is intended to drop unused code to save work during the compile step.This behaviour is enshrined in the regression test for #479 (cb2737a4), which checks that only referenced functions make their way into the disassembly.

Unfortunately, `block_drop_unreferenced` is pretty expensive in its own right, and spends quite a lot of time in looping over `block_count_refs`

I think it'd make sense to stub out `block_drop_unreferenced` until it can be implemented in a way that isn't horribly inefficient:

```diff
diff --git a/src/compile.c b/src/compile.c
index cea36a6..d86fb28 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -464,28 +464,7 @@ block block_bind_referenced(block binder, block body, int bindflags) {
 }

 block block_drop_unreferenced(block body) {
-  inst* curr;
-  block refd = gen_noop();
-  block unrefd = gen_noop();
-  int drop;
-  do {
-    drop = 0;
-    while ((curr = block_take(&body)) && curr->op != TOP) {
-      block b = inst_block(curr);
-      if (block_count_refs(b,refd) + block_count_refs(b,body) == 0) {
-        unrefd = BLOCK(unrefd, b);
-        drop++;
-      } else {
-        refd = BLOCK(refd, b);
-      }
-    }
-    if (curr && curr->op == TOP) {
-      body = BLOCK(inst_block(curr),body);
-    }
-    body = BLOCK(refd, body);
-    refd = gen_noop();
-  } while (drop != 0);
-  block_free(unrefd);
+  /* XXX needs a non-#accidentallyquadratic implementation */
   return body;
 }

diff --git a/tests/shtest b/tests/shtest
index fc2ef15..4d2d130 100755
--- a/tests/shtest
+++ b/tests/shtest
@@ -223,11 +223,6 @@ if [ "$(HOME="$mods" $VALGRIND $Q $JQ -nr fg)" != foobar ]; then
     exit 1
 fi

-if [ $(HOME="$mods" $VALGRIND $Q $JQ --debug-dump-disasm -n fg | grep '^[a-z]' | wc -l) -gt 3 ]; then
-    echo "Binding too many defs into program" 1>&2
-    exit 1
-fi
-
 cd "$JQBASEDIR" # so that relative library paths are guaranteed correct
[...]

PR #1819, opened 2019-02-09
Reimplement block_drop_unreferenced in linear time

muhmuhten:
See also #1813.

This lets us have our cake and eat it, and is probably actually slightly faster than not dropping unreferenced blocks to boot.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -53,6 +53,7 @@ struct inst {
   struct inst* bound_by;
   char* symbol;
   int any_unbound;
+  int referenced;
 
   int nformals;
   int nactuals;
@@ -75,6 +76,7 @@ static inst* inst_new(opcode op) {
   i->bound_by = 0;
   i->symbol = 0;
   i->any_unbound = 0;
+  i->referenced = 0;
   i->nformals = -1;
   i->nactuals = -1;
   i->subfn = gen_noop();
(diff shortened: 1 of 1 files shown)


muhmuhten:
The implementation of `block_drop_unreferenced` here is actually pretty dumb (one backward pass to mark referenced insts, then a forward pass to construct a block with only referenced insts) since we can tell that an inst is referenced as soon as we see it, but I think removing it at that point requires mucking with block/inst guts.

muhmuhten:
It can be done in one pass, but I think it's less elegant for no particular benefit.

```diff
diff --git a/src/compile.c b/src/compile.c
index d6346dc..a59aa25 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -53,6 +53,7 @@ struct inst {
   struct inst* bound_by;
   char* symbol;
   int any_unbound;
+  int referenced;

   int nformals;
   int nactuals;
@@ -75,6 +76,7 @@ static inst* inst_new(opcode op) {
   i->bound_by = 0;
   i->symbol = 0;
   i->any_unbound = 0;
+  i->referenced = 0;
   i->nformals = -1;
   i->nactuals = -1;
   i->subfn = gen_noop();
@@ -466,28 +468,36 @@ block block_bind_referenced(block binder, block body, int bindflags) {
 }

 block block_drop_unreferenced(block body) {
-  inst* curr;
-  block refd = gen_noop();
-  block unrefd = gen_noop();
-  int drop;
-  do {
-    drop = 0;
-    while ((curr = block_take(&body)) && curr->op != TOP) {
-      block b = inst_block(curr);
-      if (block_count_refs(b,refd) + block_count_refs(b,body) == 0) {
-        unrefd = BLOCK(unrefd, b);
-        drop++;
+  int saw_top = 0;
+
+  inst* prev = 0;
+  for (inst* i = body.last; i; i = prev) {
+    if (saw_top && i->bound_by == i && !i->referenced) {
+      /* unreferenced */
+      assert(i != body.last);
+      if (i == body.first) {
+        assert(i->next && !i->prev);
+        body.first = i->next;
       } else {
-        refd = BLOCK(refd, b);
+        assert(i->next && i->prev);
+        i->prev->next = i->next;
       }
+      i->next->prev = prev = i->prev;
+      inst_free(i);
+    } else {
+      if (i->op == TOP) {
+        saw_top = 1;
+      }
+      if (i->bound_by) {
+        i->bound_by->referenced = 1;
+      }
+
+      i->arglist = block_drop_unreferenced(i->arglist);
+      i->subfn = block_drop_unreferenced(i->subfn);
+      prev = i->prev;
     }
-    if (curr && curr->op == TOP) {
-      body = BLOCK(inst_block(curr),body);
-    }
-    body = BLOCK(refd, body);
-    refd = gen_noop();
-  } while (drop != 0);
-  block_free(unrefd);
+  }
+
   return body;
 }

```

nicowilliams:
We could mark binders as referenced in `block_bind_subblock_inner()`...

muhmuhten:
Not easily--if we mark things referenced in `block_bind_subblock_inner()` we get extraneous references because the referencing code can itself be unreferenced. e.g. if we have

```jq
def select(f): if f then . else empty end;
def arrays: select(type == "array");
def recurse: ., (.[]|recurse);
```

`select` and `recurse` won't get removed.

nicowilliams:
But an approximation might be good enough, no?

muhmuhten:
Not good enough for the test at https://github.com/stedolan/jq/blob/master/tests/shtest#L226-L229 :-)

In any case, if we just set `binder.first->referenced = 1` in `block_bind_subblock_inner`, we get the following functions compiled in regardless of the program: IN INDEX walk tostream repeat combinations last first limit until sub splits _nwise _nwise capture test match match todateiso8601 fromdateiso8601 range _flatten scalars objects arrays isfinite all all any any paths paths index indices from_entries to_entries recurse recurse _modify _assign add group_by select map error range path not empty null

That pulls in pretty substantial chunk of builtin.jq, which makes `--debug-dump-disasm` almost as useless as if when block_drop_unreferenced is stubbed out entirely.

I think just doing a real mark + sweep is more elegant simpler to understand than trying to paper over the shortcomings of that approach. It's pretty cheap at just two passes through the tree, which is pretty negligible compared to all the run-arounds we do during binding, and we almost certainly make up for it by being able to skip compiling the extra functions.

nicowilliams:
Fair enough.  I'll be able to review this tonight or tomorrow.

nicowilliams:
Pushed with minor de-indentation change.

Thanks!

PR #1821, opened 2019-02-12
Allow globbing on Windows (#1644)

### LINKED ISSUE #1644: Wildcards on Windows
I'm using the 64-bit Windows version (jq-1.6rc1-10-g7fd9e86) from AppVeyor (1.0.285). Passing wildcards on the command line aborts:

```
J:\>jq *.json
Assertion failed!

Program: J:\jq.exe
File: src/main.c, Line 256

Expression: wargc == argc
...
```

adoxa:
The mingw-w64 runtime expands wildcards, which causes a discrepancy
between `main`'s `argv` (multiple file names) and Windows' `wargv` (a
single wildcard).  Use `wmain` to retrieve the wide character args.
Resolves #1644.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,6 +18,10 @@ LIBJQ_SRC = src/builtin.c src/bytecode.c src/compile.c src/execute.c    \
 AM_CFLAGS = -Wextra -Wall -Wno-missing-field-initializers               \
         -Wno-unused-parameter -Wno-unused-function
 
+if WIN32
+AM_CFLAGS += -municode
+endif
+
 ACLOCAL_AMFLAGS = -I config/m4
 
 ### Generating the lexer and parser
--- a/src/main.c
+++ b/src/main.c
@@ -247,7 +247,26 @@ static void debug_cb(void *data, jv input) {
   fprintf(stderr, "\n");
 }
 
+#ifdef WIN32
+int umain(int argc, char* argv[]);
+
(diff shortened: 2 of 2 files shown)


nicowilliams:
Should we do this only if building with mingw?

adoxa:
I think other compilers will automatically use `wmain` (VC certainly will), which is probably better than going through the shell's arguments. VC can also expand wildcards by linking with `wsetargv.obj` (but it's broken, as quoted arguments are also expanded, although that's highly unlikely to be an issue with `jq`).

nicowilliams:
I meant this:

```
   wchar_t **wargv = CommandLineToArgvW(GetCommandLineW(), &wargc);
```

should we lose that line in non-mingw Windows builds?

adoxa:
What non-mingw builds are there?  AFAICT you don't support Visual Studio (which anyway supports `wmain`, so no need for `CommandLineToArgvW`). However, it might still be needed if you support the original MinGW (i.e. not mingw-w64). But then you should turn off globbing (better to say `*.json` not found than to assert the argument counts are different). Although that leads to an alternative approach, disabling mingw's automatic globbing and doing it manually.

nicowilliams:
Thanks!

(We don't have non-mingw Windows builds, it's true, but that doesn't mean that we shouldn't keep that possibility in mind.  In any case, we'll leave that for some other time.)

PR #1823, opened 2019-02-14
Add ISO 8601-specific strptime

adoxa:
Implement `strptime` for systems lacking it, but only supporting the
ISO 8601 format string.


nicowilliams:
Thanks!

PR #1825, opened 2019-02-15
Support "if" without "else" clause

### LINKED ISSUE #1824: Feature request: making "else" optional
I think it would really help shorten expressions if the `else` clause inside an `if` could be omitted, returning `.` by default.

For example the following two programs would be considered equivalent:

```
if .foo then
    .foo = "bar"
else
    .
end
```

```
if .foo then
    .foo = "bar"
end

```

chancez:
This implements #1824. I decided to take a stab at it to see how difficult it
was, and it wasn't so difficult. Currently this PR only contains changes to the
yacc file because my yacc is quite newer and I'm seeing a lot of extraneous
changes to the `parser.c` and `parser.h` files.

Defaults the else clause to the identity filter "." if an else clause is
missing.

The following two jq programs are considered equivalent:

```
if .foo then
    .foo = "bar"
else
    .
end
```

```
if .foo then
    .foo = "bar"
end
```

Closes #1824


chancez:
I'm having troubles compiling bison 3.0.4 on my Mac, but I did manage to run the build in a fedora container with bison 3.0.5.

I basically added all the changes and removed the updates to `YYBISON_VERSION` and the comment headers at the top updating the copyright dates. I can't be 100% sure the other changes are the same as 3.0.5, but someone else with bison 3.0.4 could give it a try.

chancez:
Looks like failures are related to bundler missing in some environments, and one failed `tests/shtest`, but I'm not sure if that's related.

src/parser.y
```diff
@@ -357,9 +357,8 @@ Term "as" Patterns '|' Exp {
 "if" Exp "then" Exp ElseBody {
   $$ = gen_cond($2, $4, $5);
 } |
-"if" Exp "then" error {
```
nicowilliams: I think we still want to retain an error case though, for when the `end` token is missing.

chancez: Good point.

chancez: Fixed in latest commit. (Can squash when everything looks good)

nicowilliams:
@chancez LGTM.  Please squash.  Do also please update the docs (and squash that too), then I'll merge.

nicowilliams:
@chancez LGTM.  Please squash.  Do also please update the docs (and squash that too), then I'll merge.

chancez:
Ok, I'll try to get to that later in the week when I have some more time.

nicowilliams:
@chancez I've squash/rebased, added a commit for docs, and pushed to your branch.  I'll merge shortly.

nicowilliams:
Thanks!

chancez:
Awesome thanks!

PR #1827, opened 2019-02-18
Explore keeping an unbounds sub-tree

nicowilliams:
DO NOT PUSH THIS.

This commit explores the idea of keeping a sub-tree of unbound instructions, unlinking instructions from the sub-tree as they get bound.

It turns out that this doesn't really help because most sub-trees in the ASTs for the tests have very short runs of unbound instructions, so removing bound instructions doesn't help.

Using the first 154 tests in `tests/jq.test` for measuring performance, this patch makes no real difference.  There are only about 304 cases where removing an unbound instruction from the sub-tree when it is bound actually helps, again, because the lengths of runs of unbound instructions from `first` to `last` in any `block` is just short.

It's possible that I've made an error though, so I'll keep this around for now.

Things to do would include keeping some stats in the `first` inst of every body.

Note that one reason this is not helping so much is that the `first` instruction of every `block` is always on the unbound sub-tree.  The only way to prevent this would be to add a field to `struct block` and then arrange for `block_append()`, `block_join()`, `block_take()`, and `BLOCK()` to maintain this new field.  That might actually help.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -27,6 +27,7 @@
 struct inst {
   struct inst* next;
   struct inst* prev;
+  struct inst* next_unbound;
 
   opcode op;
 
@@ -70,7 +71,7 @@ struct inst {
 
 static inst* inst_new(opcode op) {
   inst* i = jv_mem_alloc(sizeof(inst));
-  i->next = i->prev = 0;
+  i->next_unbound = i->next = i->prev = 0;
   i->op = op;
   i->bytecode_pos = -1;
   i->bound_by = 0;
@@ -342,7 +343,9 @@ static int block_bind_subblock_inner(int* any_unbound, block binder, block body,
   if (binder.first->nformals == -1)
     binder.first->nformals = block_count_formals(binder);
   int nrefs = 0;
-  for (inst* i = body.first; i; i = i->next) {
+  int fast_path = body.first && body.first->next_unbound != 0;
+  inst *prev = fast_path ? body.first : 0;
+  for (inst* i = fast_path ? body.first : body.last; i; i = fast_path ? i->next_unbound : i->prev) {
     if (i->any_unbound == 0)
       continue;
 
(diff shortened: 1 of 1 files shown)


nicowilliams:
@muhmuhten @wtlangford FYI.

src/compile.c
```diff
@@ -375,8 +378,32 @@ static int block_bind_subblock_inner(int* any_unbound, block binder, block body,
...
+       * we make its next_unbound point into the unbounds sub-tree anyways.
+       */
+      i->next_unbound = prev; /* ditto */
+    } else if (fast_path && i != body.first) {
+      /*
+       * In this case we must have finally fully-bound the sub-tree at this
+       * inst, so we remove this inst from the sub-tree of unbounds.
+       */
+      prev->next_unbound = i->next_unbound;
+    }
```
nicowilliams: If we add `else if (fast_path) printf("\n");` here we see 5528 cases in the first 154 tests in `src/jq.test` where we don't benefit because `i == body.first` and we can't remove `body.first` from the unbound sub-tree.

src/compile.c
```diff
@@ -375,8 +378,32 @@ static int block_bind_subblock_inner(int* any_unbound, block binder, block body,
...
+       * This inst is bound, so it need not go into the unbounds sub-tree, but
+       * we make its next_unbound point into the unbounds sub-tree anyways.
+       */
+      i->next_unbound = prev; /* ditto */
+    } else if (fast_path && i != body.first) {
+      /*
+       * In this case we must have finally fully-bound the sub-tree at this
+       * inst, so we remove this inst from the sub-tree of unbounds.
+       */
+      prev->next_unbound = i->next_unbound;
```
nicowilliams: If we add a `printf("\n");` here we see 8977 cases where this happens, but if we add `if (i->next_unbound) printf("\n");` instead, then we see just 763 cases.

muhmuhten:
Yeah, it's interesting to see shortening how little impact we see from skipping instructions *within* a block. Part of it seems to just be that not only are runs of unbound instructions short―blocks are pretty short in general, as we saw from the difference it made to skip out on the innermost call in https://github.com/stedolan/jq/pull/1814#issuecomment-461984546.

Actually, commenting out the asserts at the top of `block_bind_subblock_inner` altogether gives me a small but consistent 5~10% under no-valgrind conditions...

nicowilliams:
If we add a `first_unbound` field to `block` -- and find a cheap way to do so -- then perhaps things might be better.

I've tried before to make a single linked list, rather than a sub-tree, of unbounds, but that fails the moment you have any `block` tree mutations.  That would make linking faster, if we could make it work at all.

nicowilliams:
This is decidedly not needed now thanks to @muhmuhten's several contributions, especially #1834.

Issue #1828, opened 2019-02-18
Linking order confusion for libraries


muhmuhten:
`block_bind_library` (and `block_bind_each`, `block_bind_referenced`, and maybe others) traverse the binder block from first by next, so the first function declared with a given name gets used, even if the library contains another function of the same name declared later.

The parser, on the other hand, binds backwards, and later definitions "override" (really, get locked in before) those earlier in the file. This leads to an inconsistency:

```
% jq -n 'def a: 1; def a: 2; a'
2
% echo 'def a: 1; def a: 2;' > a.jq
% jq -n 'include "a"; a'
1
```

This is a pretty trifling point, but I'm leaving it here for consideration because from cursory examination, the change I suggested in https://github.com/stedolan/jq/issues/1826#issuecomment-464597334 would require doing `bind_block_referenced`s _backward_ in order to get the linkage within builtin.jq right, which seems like a good opportunity to do the same thing when linking libraries, which would (from my perspective) fix this inconsistency.

There's a bit of trickiness when a single file includes multiple libraries; the libraries need to be bound in backwards, too, I believe.

Thoughts?

nicowilliams:
Oof.  This would be a breaking change, but we must make it.

nicowilliams:
Fixed by @muhmuhten in #1834.

muhmuhten:
Not fixed yet.

Another wrinkle: when we include multiple libraries, e.g.

```
# e.jq
def e: 1;
def e: 2;
```
```
# f.jq
def e: 3;
```
```
# main program
include "e";
include "f";
e
```

We probably want 3, but we actually get 1 (2 after above patch)

muhmuhten:
The quick way to fix this is flipping the imports in the parser but that's a wtf.

We could do it in block_take_imports, but that actually does extra work...

muhmuhten:
Actually, this behaviour is already tested since c4cc62a053f632f5daf6c40837f5dcfa01ddf590, so I guess the other way around is intentional?

Issue #1826, opened 2019-02-17
jq-1.6 **extremely** slow compared to 1.5


waldner:
Sample timing test:

```
# time for i in {1..1000}; do echo '{"foo":"bar"}' | ./jq-1.5-linux64 '.foo' > /dev/null; done

real	0m4.852s
user	0m1.352s
sys	0m0.564s
# time for i in {1..1000}; do echo '{"foo":"bar"}' | ./jq-1.6-linux64 '.foo' > /dev/null; done

real	0m46.192s
user	0m42.448s
sys	0m0.996s
```

Scripts that used to finish very fast with jq 1.5 are now timing out with 1.6. This is under Debian stretch 9.6, with the official binaries downloaded from jq site.

nicowilliams:
This is quadratic linking of built-ins, which is now fixed in `master` thanks to a kind contribution by @muhmuhten.

waldner:
Thanks. Is a stable release with the fix in sight anytime soon?

nicowilliams:
I think this warrants a new release soon, but we'll probably wait for a few more things..net

nicowilliams:
Ay, a few more things yet.

pkoppstein:
The good news is that the "master" version is significantly better than 1.6.
The not-so-good news is that it's still slower than 1.5 and significantly slower than 1.4.
On a Mac:

```
u+s # version

 3.81 jq-1.4  
 5.85 jq-1.5  
30.6  jq-1.6 
10.7  jq-1.6-35-g80b064f 

```

pkoppstein:
> Scripts that used to finish very fast with jq 1.5 are now timing out with 1.6.

I would venture to guess that jq is not being used optimally :-)

You might consider asking a usage question at stackoverflow.com with the jq tag.

muhmuhten:
AFAICT the difference between jq 1.5 and jq 1.6 startup time comes down to a change in the binding strategy: 1.5 binds each builtin individually (in reverse order) only if used, while 1.6 parses a big block of builtins as a library, resolving all the symbols internally (the quadratic part).

The change seems to been made to facilitate make the `builtins` easy to generate, but in the process forces jq to link all the builtins every time.

Meanwhile, linking was always (and is still) quadratic.

nicowilliams:
[...]

PR #1830, opened 2019-02-19
Load ~/.jq as a library

muhmuhten:
There are no performance improvements in this patchset, unfortunately.

It's just an attempt at cleaning up `builtins_bind` to make a better springboard for working out the details of #1826 without having ~/.jq guts in the same function, when it can better be loaded as a library that just happens to have the same name and location.

There is a small incompatibility in the resolution order of shadowed names, but that's already buggy IMO (#1828) so I don't think it matters.

(https://github.com/muhmuhten/jq/commit/070299079ae68c8c56f50eabf42d7c1b4aae21da is a slightly more truthful history of this changeset.)

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1746,9 +1746,7 @@ static block bind_bytecoded_builtins(block b) {
   return block_bind(builtins, b, OP_IS_CALL_PSEUDO);
 }
 
-
-
-static const char* const jq_builtins =
+static const char jq_builtins[] =
 /* Include jq-coded builtins */
 #include "src/builtin.inc"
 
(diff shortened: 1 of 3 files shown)


src/linker.c
```diff
@@ -140,6 +135,8 @@ static jv jv_basename(jv name) {
 static jv find_lib(jq_state *jq, jv rel_path, jv search, const char *suffix, jv jq_origin, jv lib_origin) {
   if (jv_get_kind(search) != JV_KIND_ARRAY)
     return jv_invalid_with_msg(jv_string_fmt("Module search path must be an array"));
+  if (!jv_is_valid(rel_path))
+    return rel_path;
```
nicowilliams: Here, at 137, and 141, we probably need to `jv_free(search)` and maybe the other `jv` arguments...

src/linker.c
```diff
@@ -375,6 +377,16 @@ int load_program(jq_state *jq, struct locfile* src, block *out_block) {
   if (nerrors)
     return nerrors;
 
+  char* home = getenv("HOME");
+  if (home) {    // silently ignore no $HOME
+    /* Import ~/.jq as a library named "" found in $HOME */
+    block import = gen_import_meta(gen_import("", NULL, 0),
```
nicowilliams: Nice!

src/builtin.c
```diff
@@ -1784,45 +1782,17 @@ static block gen_builtin_list(block builtins) {
   return BLOCK(builtins, gen_function("builtins", gen_noop(), gen_const(list)));
 }
 
-static int builtins_bind_one(jq_state *jq, block* bb, const char* code) {
```
nicowilliams: We might end up needing such a function again in the future, FYI.  I'm thinking we should have a convention that two consecutive newlines in `src/builtin.jq` should separate logical units, and then we can compile each of those separately, and use them as binders only if a program needs them.

src/linker.c
nicowilliams: Granted, we already had a leak, so let's fix that separately.

nicowilliams:
Tests fail, FYI.

src/linker.c
muhmuhten: Good point. I wasn't thinking about it too hard since the existing early exits are pretty sloppy, but generally `find_lib` is going to want to free `rel_path` and `search` (only?).

muhmuhten:
Well okay, looks like I introduced a(?) memory leak, but it's a works-on-my-machine so this space might get a bit noisy with pushes as I try to suss that out with the CI.

muhmuhten:
Tests are passing now, or at least the ones that normally pass anyway.

I squashed bugfixes into their corresponding commits; sorry for the extra noise, though.

nicowilliams:
Yeah, I've been meaning to figure out what's up with the tests that fail all the time...

nicowilliams:
Thanks!!!

PR #1832, opened 2019-02-19
Bytecode _assign, _modify, and recurse

muhmuhten:
Unchecked calls to these functions are generated by the parser, so it is
unsafe to skip their definitions in builtin.jq.

This is arguably a terrible idea since the bytecoded versions are giant opaque blobs, but they must always be defined even if we implement --no-builtins because otherwise user programs using `..`, `=`, `|=`, `+=` and variants can segfault or worse.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1702,12 +1702,52 @@ static const struct cfunction function_list[] = {
 
 struct bytecoded_builtin { const char* name; block code; };
 static block bind_bytecoded_builtins(block b) {
+  {
+    /* _assign and _modify depend on path/1, so they need to be bound separately */
+    block builtins = BLOCK(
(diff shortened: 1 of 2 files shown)


muhmuhten:
It occurs to me that this was a dumb idea, and it'd be easier to just have a "critical" builtins string to load instead of actually expanding out bytecode.

Although making `recurse` a bytecoded builtin might still be fine idea, seeing how short it is.

nicowilliams:
I agree that the no-builtins mode must still have a few builtins, including:

 - not
 - empty
 - input/inputs
 - _assign and _modify

Maybe we could not have all C-coded builtins, but it would take effort to do that, so might as well provide all of them :)

PR #1834, opened 2019-02-20
Binding only referenced builtins

muhmuhten:
aka "make startup time as fast as (probably a bit faster than) 1.5"
aka "omg omg I think I actually managed to pull it off in a kind of elegant way and it's like 3x faster now"

By having the parser spit out libraries without bindings, we can get a list of builtins and bind them backwards onto the main program, dropping the unreferenced ones:

- without "two newlines"-style contortions―we can get the information directly from the parse tree
- we can keep `builtins/0`, since we can read off symbols before we drop unreferenced defs
- **we don't need to search unreferenced blocks for symbols**

Now, in theory this should also be possible for libraries, but `block_bind_library` is kind of black magic and the caching looks like it would play poorly with the "blocks coming out of the parser *aren't* bound to each other" assumption (which gives us the property that a binder is referenced *iff it binds something* that gets counted in `block_bind_subblock_inner`).
So for now we just link libraries to themselves as soon as possible and do the same thing we used to. I don't think this is as big a problem as the builtins anyway.

Builds on top of and supersedes #1830, which lays the groundwork by redefining block_bind_referenced to a form simple enough to understand in context here, and ripping out ~/.jq's special-casing which would've been nasty to add support for in `builtins_bind`.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1743,7 +1743,7 @@ static block bind_bytecoded_builtins(block b) {
                                             BLOCK(gen_param("start"), gen_param("end")),
                                             range));
   }
-  return block_bind(builtins, b, OP_IS_CALL_PSEUDO);
+  return BLOCK(builtins, b);
 }
 
 static const char jq_builtins[] =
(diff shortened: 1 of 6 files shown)


nicowilliams:
Wow!  jq startup is now, and tests now run in under 5s for me w/o valgrind, and w/ valgrind in 4.25m!

Thanks!

PR #1835, opened 2019-02-21
Fix regression on ~/.jq being a directory

muhmuhten:
I did an oops and optional libraries shouldn't break if they're directories.

This actually only matters for ~/.jq because the default library path is also ~/.jq, but it's too late to correct that mistake.


nicowilliams:
I believe this got fixed by #1836.

PR #1836, opened 2019-02-21
Reverse library linking order

### LINKED ISSUE #1828: Linking order confusion for libraries

muhmuhten:
Actually fixes #1828.

A breaking change, as noted, and changes the test_bind_order test, but on the bright side it only affects situations where exported function names are shadowed, which is a bad situation to be in in the first place.


nicowilliams:
Is this still relevant?

muhmuhten:
Yes, we're still binding the first definition in a library instead of the last one.

nicowilliams:
@muhmuhten can you rebase and merge the conflicts?

muhmuhten:
@nicowilliams got it.

nicowilliams:
I had started, but got interrupted before i could finish it, otherwise I'd have saved you the bother.

src/linker.c
```diff
@@ -239,7 +240,11 @@ static int process_dependencies(jq_state *jq, jv jq_origin, jv lib_origin, block
   block bk = *src_block;
   int nerrors = 0;
 
-  jv_array_foreach(deps, i, dep) {
+  // XXX This is a backward jv_array_foreach because bindings go in reverse
+  for (int i = jv_array_length(jv_copy(deps)); i > 0; ) {
```
nicowilliams: a) Maybe we ought to have a macro for this too,

b) I've long wanted to do something a wee bit safer w.r.t. lengths.  A signed length works well for a backwards foreach, but for just about everything else I'd rather use `size_t` due to it being unsigned (which has well-defined overflow semantics in C).

I never convinced @stedolan of (b) and I'm not going to wait for (a).

tests/jq.test
```diff
@@ -604,6 +604,10 @@ def f(a;b;c;d;e;f): [a+1,b,c,d,e,f]; f(.[0];.[1];.[0];.[0];.[0];.[0])
 [1,2]
 [2,2,1,1,1,1]
 
+def f: 1; def g: f, def f: 2; def g: 3; f, def f: g; f, g; def f: 4; [f, def f: g; def g: 5; f, g]+[f,g]
```
nicowilliams: Oof, this one is hard to parse.  For anyone following along:

```
def f: 1;
def g: f,
       def f: 2;
       def g: 3;
       f,
       def f: g;
       f, g;
def f: 4;
[f, def f: g; def g: 5; f, g]+[f,g]
```
The first `f` in `[f, ...]` will be `4`, the second will be `g`, which will b `1, 2, 3, 3`, then `g` will be `5`.  The we add `[f,g]` to that, which will be the first `f` and `g`, so, again `[1,2,3,3]`.

nicowilliams:
Thanks!

PR #1838, opened 2019-02-23
No more ruby

wtlangford:
Ruby has, historically, been a bit of an issue in our CI builds. Builds would suddenly start failing because we wouldn't be able to get/install the particular ruby version.

This PR removes the ruby dependency and replaces it with python. Most systems already have a python, and they're generally easier to install than a ruby on systems where they don't already.

This uses python3, because it's 2019.

The docs are now built via python and the website is functionally identical to the ruby version (minor differences include that we're no longer minifying the js/css output, but that's _fine_, they aren't big anyway).
The manpage is slightly different than it was before, though I feel it looks a bit better this way.
The mantests work identically to before.

I'm still considering removing the need for pipenv in favor of an explicit `requirements.txt` (and maybe a virtualenv. We don't really need that for CI, but they're nice for local dev)

There's also a change to `contains/1` that fixes the test behavior on macOS platforms. The behavior of `memmem` with a zero-length needle is undefined, so we need to be consistent about the behavior ourselves. The tests imply that we think a zero-length needle should be truthy, so I've enforced that behavior directly in `contains/1`.


fadado:
> The docs are now built via python and the website is functionally identical to the ruby version (minor differences include that we're no longer minifying the js/css output, but that's _fine_, they aren't big anyway).

If `sed` is available you can minify CSS with this code:

```
sed '
# trim spaces
s/^[ \t]\+//
s/[ \t]\+$//
# delete empty lines
/^$/d
# remove more spaces
s/[\t ]*\([{}:,;>~]\)[\t ]*/\1/g
' | sed '
# join rules in one big line
/[^@]\+{$/,/}$/ {
	:loop
	N;
	/}$/! b loop
	s/\n//g
}
```

nicowilliams:
So, on Ubuntu I had a bit of pain getting `jq.1` to build.  The experience resembled the Ruby experience :(  On the plus side, my dislike of Ruby says that pain was worth it.  Most of it was my fault, starting with me not noticing the difference between `pip` and `pip3`...

Things I had to do:

 - teach myself to add a `3` everywhere
    - though `make jq.1` doesn't add a `3`... -- I guess `pipenv run python ... ` knows it's version 3 I want??
 - blow away an old `pip` in `/usr/local` -- oof, that caused me a lot of pain
 - `sudo aptitude install python3-dev python3-pip`
 - `pip3 install pipenv`
 - `pipenv --version 3.6.7` (where 3.6.7 is the version from `python3 --version`) -- not sure if this was needed
 - `pipenv run pip3 install markdown lxml pyyaml` -- these are dependencies needed to run `build_manpage.py`

There's three places in your commits where you refer to `pip` instead of `pip3`.  Are these bugs?  I think so.

I still can't figure out how to install pyenv, or if it's needed...

That said, it works!  Yay!

src/jv.c
```diff
@@ -1344,8 +1344,13 @@ int jv_contains(jv a, jv b) {
   } else if (jv_get_kind(a) == JV_KIND_ARRAY) {
     r = jv_array_contains(jv_copy(a), jv_copy(b));
   } else if (jv_get_kind(a) == JV_KIND_STRING) {
-    r = _jq_memmem(jv_string_value(a), jv_string_length_bytes(jv_copy(a)),
```
nicowilliams: Is it also the case that a haystack length of zero has undefined behavior anywhere?

And, what should `""|contains("")` return?

nicowilliams:
I left a comment on `jv_contains()`.

There's also the question above as to whether the three places where you use `pip` should be `pip3`.

And you need to add docs at least about having to `pipenv run pip3 install markdown lxml pyyaml`, and I think that's it and it's ready to go!

nicowilliams:
The `alpine` build (I didn't know or remember we had one) fails.  It fails the `shtest` test.  I don't think it's your fault, so just ignore it.

wtlangford:
> `make jq.1` doesn't add a `3`... -- I guess `pipenv run python ... ` knows it's version 3 I want??

It should say python3 for consistency/sanity. I'll fix that.

> blow away an old `pip` in `/usr/local` -- oof, that caused me a lot of pain

Really? Interesting.

> `pip3 install pipenv`

I'm considering getting rid of pipenv to simplify all this. It's apparently not as great as I thought it was, and the main benefit I was seeking (strict version pinning of requirements) can be done in a `requirements.txt` with just plain `pip3`.
We don't need to use a virtualenv in CI, and it's easy enough to do for local work.

> `pipenv run pip3 install markdown lxml pyyaml` -- these are dependencies needed to run `build_manpage.py`

This should be spelled `pipenv sync` and was a documentation failure on my end.

> There's three places in your commits where you refer to `pip` instead of `pip3`. Are these bugs? I think so.

Not _strictly_, but I should be better about it. I'll track those down.

> I still can't figure out how to install pyenv, or if it's needed...

I don't think I referenced pyenv anywhere?

src/jv.c
wtlangford: It might be, actually.
My mac says this:
```If big_len is smaller than little_len, if little_len is 0, if big_len is 0 or if little
     occurs nowhere in big, NULL is returned; otherwise a pointer to the first character of
     the first occurrence of little is returned.```

Other systems say this:
```
The memmem() function returns a pointer to the beginning of the
       substring, or NULL if the substring is not found.

BUGS
In glibc 2.0, if needle is empty, memmem() returns a pointer to the
       last byte of haystack.  This is fixed in glibc 2.1.
```

So, it's not consistent anywhere. What behavior do we want a haystack of length zero to have? And what do we want `""|contains("")` to return?

nicowilliams:
`pyenv` is mentioned in `.travis.yml`.  I basically started by replicating what the Travis build does...

nicowilliams:
The previous `pip` in `/usr/local` probably only confused me by being all about Python 2...

nicowilliams:
Also, why didn't the `docs/Pipfile` suffice to make `pip` figure out that it needed to install `markdown`, `lxml`, and `yaml`?

src/jv.c
nicowilliams: I think `"" | contains("")` can be true, false, null, an error -- undefined behavior, if you wish -- but the user-friendly thing to do is to say that it outputs `true`, IMO, and, for that matter, `"foobar" | contains("")` should be `true` as well, no?, after all, any string contains the empty string in a sense.

wtlangford:
> The `alpine` build (I didn't know or remember we had one) fails. It fails the `shtest` test. I don't think it's your fault, so just ignore it.

That's been failling prior to my changes. I was going to look into why, because I'd like the tests to be green by the end of this, but I might just let that ride for later.

> Also, why didn't the `docs/Pipfile` suffice to make `pip` figure out that it needed to install `markdown`, `lxml`, and `yaml`?

`Pipfile`, in a _marvel_ of naming, is actually part of pipenv, and not pip itself. `pip` supports a `requirements.txt` file (you'd run `pip install -r requirements.txt`), but `Pipfile` is something of `pipenv`'s creation.

> `pyenv` is mentioned in `.travis.yml`. I basically started by replicating what the Travis build does...

Ah, that's just for selecting the right python on travis. It's not something you'd need locally, unless you knew you were using `pyenv`. It lets you have multiple `python`s that don't interfere with each other at all.

src/jv.c
nicowilliams: Let's deal with the zero-length haystack question separately.

nicowilliams:
Thanks @wtlangford, this is fantastic!

Now we'll have all Travis builds except alpine's `shtest` failure.

nicowilliams:
I've made `shtest` verbose to help diagnose its failures.

PR #1844, opened 2019-02-26
Remove block_drop_unreferenced after binding builtins

muhmuhten:
Since the new builtin binding procedure implicitly drops unreferenced builtins, dropping unreferenced blocks *again* after should never do anything useful.

(It happens that it undid a block-flipping bug, though.)

Combines block_bind_referenced and block_bind_incremental in the process since they do the same thing and the latter is kind of a nonsense name.


nicowilliams:
Thanks!

I think we can probably soon also stop using `block_drop_unreferenced()` in the linker.

Issue #1840, opened 2019-02-23
scalars_or_empty is undocumented


muhmuhten:
Actually, can we get rid of it? It's kind of misnamed (empty means something else...) and it's only used for one test (for the streaming parser 5bfb9781f7e5ba2ad99b71150bccc559d53b928f), which isn't even testing it specifically.

nicowilliams:
Sure, let's get rid of it.

nicowilliams:
Removed in #1845.

Issue #1724, opened 2018-09-24
walk function not applied to child object of match


cwingrav:
**Describe the bug**
The walk function says it, "applies f recursively to every component of the input entity." However, it does not seem to apply to an inner object, after a match has been made. i.e. a child object will not be matched if its parent matches. 

**To Reproduce**
In my example, I am pulling data from a child object, into the parent object. This works just fine for the 'a.MATCH.c' case below, but fails for 'a.inner.MATCH.b'. Possibly a documentation issue (in that case, how do I recursively match?), a bug or is my code pulling into an object that gets removed?

CODE:
```
jq '. * walk( if type == "object" and has("MATCH") then .|=.MATCH else . end) ' a.json
```

INPUT:
```
{
  "a" : {
    "inner" : {
      "MATCH" : {
        "b" : 1
      } 
    },
    "MATCH" : {
      "c" : 2
    } 
  }
}
```

OUTPUT:

The a.MATCH.c is brought out to a.c but a.inner.MATCH.b is not brought out into a.inner.b.
```
{
  "a": {
    "inner": {
      "MATCH": {
        "b": 1
      }
    },
    "MATCH": {
      "c": 2
    },
    "c": 2
  }
}
```

**Expected behavior**

EXPECTED:
I would expect a.inner.b to exist.
```
{
  "a" : {
    "inner" : {
      "MATCH" : {
        "b" : 1
      },
      "b" : 1
    },
    "MATCH" : {
      "c" : 2
    },
    "c" : 2
  }
}
```

**Environment (please complete the following information):**
 - macOS - 10.13.6
 - jq version -  jq-HEAD-8eff744

**Additional context**

pkoppstein:
I believe the problem is a mismatch between your code and your expectations.

The result you've indicate you expect can be achieved using the following invocation of `walk`:

    walk( if type == "object" and has("MATCH") then .+.MATCH else . end) 

By the way, if you look at the code for `def walk(_):`, I think you'll agree the description is fair, but it could perhaps be improved because of what some might say is an ambiguity, as seen in the following variant of `walk`:


```
def postwalk(f):
  . as $in
  | if type == "object" then
      reduce keys_unsorted[] as $key
        ( {}; . + { ($key):  ($in[$key] | f) | postwalk(f) } ) 
  elif type == "array" then map( f ) | postwalk(f)
  else f
  end;

```

PR #1845, opened 2019-02-26
Big pile of builtin changes

muhmuhten:
Rationale:

- Remove scalars_or_empty per #1840

- The isempty/any/all are coming out of the discussion in #1839. Some benchmarking indicated that an extra `not` in the loop really does make `any` significantly slower.

- I was trying to get fromstream going faster, but it doesn't want to go any faster. This implementation uses an explicit emit/reset flag to handle when it outputs, instead of implicitly clearing the object being built by shuffling it between .[0] and .[1] on array/object closure, and thereby avoids duplicating what-are-we-doing-with-the-path logic between the update and extract steps. I think this this makes the control flow easier to follow so I rolled it in, but it's certainly arguable.

- It is pretty hard to explain this tostream. That last reduce in particular does not look like it should be optimal but everything I've tried doing with it makes it slower. It feels like there should be a better way to keep the last key than going through the array/object a second time, but keeping extra state to do that makes it slower. Maybe someone else might have a better idea. The key insights are (a) .[]? produces zero elements for precisely the leaf nodes of a stream; (b) `paths` on that provides the streaming keys we apparently wanted; and (c) oh, so reduce is how you implement last/1.

- bsearch was doing a really weird thing if its argument generated multiple arguments because it'd get contradictory-looking comparisons by using different values each go-around. That's, uh, bad. Added a test, too.

- ~walk/1: another one based on the "paths of backward (post-order) recurse" primitive, which, I'm starting to feel is more use than the pre-order one. It is actually *really cool* that this one Just Works when you think about it with jq goggles on. See also #1724.~ Retracted because |= does *not* do the right thing for multiple/zero outputs on arrays and uh, actually does some really weird stuff on arrays.

- As #1654 observed, `if type != "string" then tojson else . end` winds up doing precisely `tostring`, except the latter does the type check in C: https://github.com/stedolan/jq/blob/9a0d5be790f0c89c02ffde11144157f067b19fd1/src/builtin.c#L452-L458 And note that $row is always going to be one value that doesn't depend on its input, so we can assign rather than modify here.

- turns out IN was just a shouty way of writing any, except with less short-circuiting and a lot more `if ... then true else false end`. This became a lot more obvious after rewriting any and all!

These should be reasonably cherry-pickable if some of them are disagreeable, too.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -215,22 +215,9 @@ def fromstream(i): {x: null, e: false} as $init |
     else setpath(["e"]; $i[0]|length==1) end
   ; if .e then .x else empty end);
 def tostream:
-  {string:true,number:true,boolean:true,null:true} as $leaf_types |
-  . as $dot |
-  if $leaf_types[$dot|type] or length==0 then [[],$dot]
-  else
-    # We really need a _streaming_ form of `keys`.
-    # We can use `range` for arrays, but not for objects.
-    keys_unsorted as $keys |
-    $keys[-1] as $last|
-    ((# for each key
-      $keys[] | . as $key |
-      $dot[$key] | . as $dot |
-      # recurse on each key/value
-      tostream|.[0]|=[$key]+.),
-     # then add the closing marker
-     [[$last]])
-  end;
+  path(def r: (.[]?|r), .; r) as $p |
+  getpath($p) |
+  reduce path(.[]?) as $q ([$p, .]; [$p+$q]);
 
 
 # Assuming the input array is sorted, bsearch/1 returns
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -138,11 +138,6 @@ def gsub($re; s; flags): sub($re; s; flags + "g");
 def gsub($re; s): sub($re; s; "g");
 
 ########################################################################
-# range/3, with a `by` expression argument
-def range($init; $upto; $by):
-    def _range:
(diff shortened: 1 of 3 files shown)


src/builtin.jq
```diff
@@ -215,22 +215,9 @@ def fromstream(i): {x: null, e: false} as $init |
...
-    $keys[-1] as $last|
-    ((# for each key
-      $keys[] | . as $key |
-      $dot[$key] | . as $dot |
-      # recurse on each key/value
-      tostream|.[0]|=[$key]+.),
-     # then add the closing marker
-     [[$last]])
-  end;
+  path(def r: (.[]?|r), .; r) as $p |
```
nicowilliams: I think you can make this just `path(..?) as $p |`.

nicowilliams: Recall that `..` is syntactic sugar for `recurse`, which is `def recurse: recurse(.[]?);`, and `recurse/1` is `def recurse(f): def r: ., (f | r); r;`.  However, inlining as here is best.

nicowilliams:
Thanks!!!

src/builtin.jq
muhmuhten: .. (=recurse) is pre-order and we need post-order here to close nonempty objects/arrays.

nicowilliams: Ah, right.  Thanks for the clarification (and, I know, I missed this nugget of info for not reading the description together with the changes).

Issue #1168, opened 2016-06-21
Failure to parse .x.["y.z"] notation


mcint:
### mcint@work MINGW64 /c/git

The document

```
$ less test.json | jq -c '.'
{"a":"A","b":{"c":{"d.e.f":"DEF"}}}
```

The complex key

```
$ less test.json | jq -c '.b.c'
{"d.e.f":"DEF"}
```

One naive approach

```
$ less test.json | jq -c '.b.c.d.e.f'
null
```

Solution proposed by the docs

```
$ less test.json | jq -c '.b.c.["d.e.f"]'
jq: error: syntax error, unexpected '[', expecting FORMAT or QQSTRING_START (Windows cmd shell quoting issues?) at <top-level>, line 1:
.b.c.["d.e.f"]
jq: 1 compile error
```

Work-around

```
$ less test.json | jq -c '.b.c | .["d.e.f"]'
"DEF"
```

Boundaries of quoted key solution

```
$ less test.json | jq -c '.b | .c.["d.e.f"]'
jq: error: syntax error, unexpected '[', expecting FORMAT or QQSTRING_START (Windows cmd shell quoting issues?) at <top-level>, line 1:
.b | .c.["d.e.f"]
jq: 1 compile error
```

mcint:
Trials with 2-part ("z.y") versus 3-part ("z.y.x") key
```
$ less test2.json | jq -c '.b.c'
{"d.e": "DEF"}
```
```
$ less test2.json | jq -'c .b.c.d.e'
null
```
```
$ less test2.json | jq -c '.b.c.["d.e"]'
jq: error: syntax error, unexpected '[', expecting FORMAT or QQSTRING_START (Windows cmd shell quoting issues?) at <top-level>, line 1:
.b.c.["d.e"]
jq: 1 compile error
```

slapresta:
The syntax is either `.foo` (for keys without extraneous characters) or `["foo.bar"]` (note the absence of a trailing dot). That is:

`jq '.b | .c["d.e.f"]'`
`jq '.b.c["d.e"]'`

So, you may be wondering: how come `jq '.b.c | .["d.e.f"]'` does work? The answer lies in that the output of the previous filter is represented by `.`. Therefore, `.["d.e.f"]` performs `["d.e.f"]` over `.`. The dot can be elided when using `.foo` as the first filter: that's why it's `.b.c` and not `..b.c`.

mcint:
[...]

PR #1849, opened 2019-02-26
add nested array indexing example

kevpl:
This PR adds an example to the docs to help people who've hit the snag
discussed in #1168.

Having an example allows people who hit this snag to relatively quickly
see what they've done wrong, rather than having to probe further into a
cryptic error message:
```
jq: error: syntax error, unexpected $end, expecting FORMAT or QQSTRING_START (Unix shell quoting issues?)
```

Having it just as an example allows people to be informed of the behavior
without the burden of needing to explain it with additional text.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -378,6 +378,10 @@ sections:
             input: '[1,2,3]'
             output: ['2']
 
+          - program: '.letters[1]'
+            input: '{"letters":["a","b","c","d"]}'
+            output: '"b"'
+
       - title: "Array/String Slice: `.[10:15]`"
         body: |
 


nicowilliams:
I mean, the manual is littered with examples like this, but I guess the point is that just the section on array indexing can confuse people, so I'll take it.

Alanscut (approved):

docs/content/manual/manual.yml
```diff
@@ -378,6 +378,10 @@ sections:
             input: '[1,2,3]'
             output: ['2']
 
+          - program: '.letters[1]'
+            input: '{"letters":["a","b","c","d"]}'
+            output: '"b"'
```
nicowilliams: This needs to be `output: ['"b"']`.

itchyny (changes requested):
The example output needs to be corrected as commented before.

PR #1850, opened 2019-02-26
Fix alpine build(?)

muhmuhten:
`echo` with escapes isn't portable, so use `printf` instead.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -1,4 +1,4 @@
-#!/bin/sh -vx
+#!/bin/sh
 
 . "${0%/*}/setup" "$@"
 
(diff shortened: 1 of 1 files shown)


muhmuhten:
In retrospect, the first X at the end of the checks marks the spot, I guess
https://github.com/stedolan/jq/commits/master?after=b436156f5bdcc9ce3f3f93dd30a7bcbdb87910f7+0

nicowilliams:
Well, i feel dumb today!  :P

muhmuhten:
All green now :-)

nicowilliams:
Thanks!

PR #1851, opened 2019-02-27
Restore cfunction arity in builtins/0

muhmuhten:
Count arguments up-front at definition/invocation instead of doing it at
bind time, which comes after generating builtins/0 since e843a4f.

(mea culpa)

(and also add some tests for builtins/0)

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -290,18 +290,6 @@ int block_has_only_binders(block binders, int bindflags) {
   return 1;
 }
 
-// Count a binder's (function) formal params
-static int block_count_formals(block b) {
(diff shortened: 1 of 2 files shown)


muhmuhten:
Specifically, the bug this fixes is that builtins/0 would claim all C functions had 0 arguments (`["input_line_number/-1","input_filename/-1","now/-1",` ...) because they never went through block_bind_subblock_inner, which actually sets nformals if it's at the default -1. (jq functions weren't affected because they always go through block_bind_referenced at parse time).

nicowilliams:
Thanks!

PR #1862, opened 2019-03-14
Added jv_is_integer_large() to the library.

mfeit-internet2:
This change adds a new function to the library called `jv_is_integer_large()`, which returns true for any `JV_KIND_NUMBER` that fits into a `double` and has a fractional part smaller than the system's epsilon.

It was written to let [PyJQ](https://github.com/doloopwhile/pyjq) make decisions about using the arbitrarily-large integer types available in Python.  (A companion pull request for that project is forthcoming.)

Gory details about why:  https://github.com/perfsonar/pscheduler/issues/717

### DIFF
--- a/src/jv.c
+++ b/src/jv.c
@@ -7,6 +7,7 @@
 #include <stdarg.h>
 #include <limits.h>
 #include <math.h>
+#include <float.h>
 
 #include "jv_alloc.h"
 #include "jv.h"
@@ -157,6 +158,18 @@ int jv_is_integer(jv j){
   }
 
   return x == (int)x;
+}
+
+int jv_is_integer_large(jv j){
+  if(jv_get_kind(j) != JV_KIND_NUMBER){
+    return 0;
+  }
+  double x = jv_number_value(j);
+
+  double ipart;
+  double fpart = modf(x, &ipart);
+
+  return fabs(fpart) < DBL_EPSILON;
 }
 
 /*
--- a/src/jv.c
+++ b/src/jv.c
@@ -153,18 +153,6 @@ int jv_is_integer(jv j){
     return 0;
   }
   double x = jv_number_value(j);
-  if(x != x || x > INT_MAX || x < INT_MIN){
-    return 0;
-  }
-
-  return x == (int)x;
-}
-
-int jv_is_integer_large(jv j){
-  if(jv_get_kind(j) != JV_KIND_NUMBER){
-    return 0;
-  }
-  double x = jv_number_value(j);
 
   double ipart;
   double fpart = modf(x, &ipart);
--- a/src/jv.h
+++ b/src/jv.h
@@ -63,6 +63,7 @@ jv jv_bool(int);
 jv jv_number(double);
 double jv_number_value(jv);
 int jv_is_integer(jv);
+int jv_is_integer_large(jv);
 
 jv jv_array(void);
 jv jv_array_sized(int);
--- a/src/jv.h
+++ b/src/jv.h
@@ -63,7 +63,6 @@ jv jv_bool(int);
 jv jv_number(double);
 double jv_number_value(jv);
 int jv_is_integer(jv);
-int jv_is_integer_large(jv);
 
 jv jv_array(void);
 jv jv_array_sized(int);


nicowilliams:
Do we need a jq function for this?  It can be coded using the builtin `modf` and `fabs` functions...

mfeit-internet2:
Probably not for this change, although I'd imagine there could be some uses for it in scripts.

Jq _mostly_ leaves numbers as they came in on the way out, so there's no need for scripts to make corrections unless they're trying do explicitly convert ints to floats.  
There might be a different discussion to be had about how jq deals with real numbers, because it does chop non-fractional floats into ints.  That means someone specifying JSON numbers with _n_ significant figures will have them mangled into numbers with none (e.g., `123.000` on the way in becomes `123` on the way out).

I haven't looked at the sources for any of the other language bindings, but I assumed that `jv_is_integer()` was put there as a utility for them to use.  It is used in [one spot internally](https://github.com/stedolan/jq/blob/3ea0199e031e98e92670a25e4323bd711005b5db/src/jv_aux.c#L63).

nicowilliams:
Not every `jv` function has a jq function binding.  Some might only be used internally, some might be used by "internal libjq applications" (`src/main.c` and `src/jq_test.c`), some might be used by external libjq applications.  Who would use this one?

As to jq throwing away zero-valued fractional parts... well, it's just about how it formats numbers.  On input jq immediately parses JSON numeric values into IEEE754 `double`s.  On output it formats `double`s back into JSON numeric values.  As a result the original form of a numeric value is lost.  Now, if in C you have `double r = 1.0;` that's the same as `double n = 1;` in that the two values will be bit-by-bit equal (i.e., `memcmp((void*)&r, (void*)n, sizeof(double)) == 0`)

leonid-s-usov:
Just my two cents, the decimal number branch #1752 has the feature of keeping number precision intact. 

1 == 1.00 == 1.0

but 

[1, 1.00, 1.0] => jq map(. + 1) => [2, 2.00, 2.0]

Regards,
Leonid S. Usov

nicowilliams:
@leonid-s-usov Oh, I see, that is nice.  @wtlangford is rebasing your PR and undoing the big file split.

leonid-s-usov:
yeah, I have received his message. I loved that split BTW, I think it brings some order.

nicowilliams:
I understand.  The idea is not to not have that split, but to first merge the contribution, and then do the split  in a commit that only does the split.

mfeit-internet2:
Side discussions about bindings and how integers will be handled in the future, does this look like something that will be in the next release?

nicowilliams:
@mfeit-internet2 sure, a few questions:

 - how would this work once we integrate #1752?
 - do you have libjq-using code that needs this?
 - should we expose this as a jq function so jq programs can use it?

mfeit-internet2:
**1752:**   If the plan is to merge that change into the next release and there's still going to be a `jv_is_integer()` in the library without the same larger-than-31-bits bug as the existing version, the `_large` version can go away.  In retrospect, I should probably have proposed this change as a wholesale replacement for `jv_is_integer()` instead of adding a new function but didn't want to risk breaking existing code that might depend on the 31-bit behavior.   I wasn't aware this effort was underway, but as long as whatever function ends up in the new version works correctly, I'm good with it.

**Code needing it:**  I have a patched version of jq with the new function and a patched PyJQ that uses it.  Both are privately built and privately distributed, so that code is totally contained and under my control.  I'm not the only PyJQ user with this problem, and my plan is to submit a patch to PyJQ patch once jq either pulls this, fixes `jv_is_integer()` so it works or does something else that still lets me reason about what I'm about to be handed.

**Exposure as a function:**  No need.  Scripts shouldn't care since jq takes care of all of this internally and seems to work fine.  (See the perfSONAR ticket I referred to in the first comment.)  Nothing internal uses `jv_is_integer()`, so it's a library-only problem.

nicowilliams:
@mfeit-internet2 thanks.

OK, so, I think this should be folded into `jv_is_integer()`, and we should have an `isinteger` function for jq.  I'm sure that when #1752 gets merged we can make `jv_is_integer()` do the right thing (though I suppose we can then have a fun debate about whether `1.0` is an integer or not).

If you can force push a version that fixes `jv_is_integer()`, that'd be great, and I'll merge it.  If you can't before this weekend, i'll probably do it (while retaining you as author).

nicowilliams:
Thanks!  I'll squash and rebase as soon as I get a chance.

PR #1865, opened 2019-03-20
Add const modifiers to pointer function arguments that don't get modified

petdance:
jq uses `const` pointers all over, but there are a few instances that I found where `const` can get added to function arguments.  This helps show intent of the function, and helps the compiler keep track of when variables are getting modified.

I hope this helps with jq's maintainability.  Thanks for a great tool.

(Aside: I was inspired to look into the internals of jq because I just wrote an article where I suggested people use it: https://blog.newrelic.com/engineering/7-things-never-code/)

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -232,7 +232,7 @@ static jv f_endswith(jq_state *jq, jv a, jv b) {
   const char *bstr = jv_string_value(b);
   size_t alen = jv_string_length_bytes(jv_copy(a));
   size_t blen = jv_string_length_bytes(jv_copy(b));
-  jv ret;;
+  jv ret;
 
   if (alen < blen ||
      memcmp(astr + (alen - blen), bstr, blen) != 0)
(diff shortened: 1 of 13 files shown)


nicowilliams:
Thank you for this contribution!  We have a couple of big PRs ahead of this one, so as not to disrupt their integration I think we'll hold back before we merge this one, but we'll merge it.

petdance:
Thanks.  I'm glad you can use it.

The other thing I've been playing around with on a branch is GCC function attributes, specifically `__attribute__((const))__` and `__attribute__((pure))__` to mark const and pure functions, to help the compiler optimize.

I see you have some other attributes sprinkled about, like

    __attribute__((warn_unused_result)) void* jv_mem_realloc(void*, size_t);

Is the `pure` and `const` something you'd find helpful, too?

nicowilliams:
I guess I'd have to see a proposed list of functions to mark `const` or `pure`, and whether it helps.

Most `jv` and `jq` functions are neither pure nor const in actuality, since the `jv` API, though copy-on-write, does modify reference counts, which means most `jv` functions have side-effects, therefore they can be neither pure nor const.  Notionally, however, much of the `jv` API *is* const or pure.

PR #1883, opened 2019-04-04
jq_util_input_init: Zero memory using calloc

orzen:
Calloc will zero the allocated memory which makes one memset and a
number of explicit zero assignments redundant.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -208,20 +208,11 @@ jq_util_input_state *jq_util_input_init(jq_util_msg_cb err_cb, void *err_cb_data
     err_cb = fprinter;
     err_cb_data = stderr;
   }
-  jq_util_input_state *new_state = jv_mem_alloc(sizeof(*new_state));
-  memset(new_state, 0, sizeof(*new_state));
+  jq_util_input_state *new_state = jv_mem_calloc(1, sizeof(*new_state));
   new_state->err_cb = err_cb;
   new_state->err_cb_data = err_cb_data;
-  new_state->parser = NULL;
-  new_state->current_input = NULL;
-  new_state->files = NULL;
-  new_state->nfiles = 0;
-  new_state->curr_file = 0;
   new_state->slurped = jv_invalid();
-  new_state->buf[0] = 0;
-  new_state->buf_valid_len = 0;
   new_state->current_filename = jv_invalid();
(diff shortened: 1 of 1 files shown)


nicowilliams:
Ah, so, the C standard doesn't require that zero-valued pointers actually all-zero-bits in memory.  There are no platforms I care about where zero-valued pointers aren't zero-bits in memory, and also we're not necessarily disciplined about not assuming that zero-valued pointers are zero in memory, but in recent years I've taken to actually not making that assumption when I write code.  It's something I got from @vdukhovni (who co-maintains Postfix, and Viktor and Wietse apparently do have to care about platforms where zer-valued pointers aren't zero in memory).

So, even when using `calloc()` I think I still want to explicitly assign `0` to pointer fields of structs.

orzen:
Fair enough. Do you have any examples of such platforms so I know what to take into account in future commits, since I only develop for Linux?

vdukhovni:
No recent platforms, but see <http://c-faq.com/null/machexamp.html>.

nicowilliams:
Thanks @vdukhovni.  It's good to know all of those are basically dead.  I suspect there will not be a new such platform -- it would be a huge mess, but since I've picked up this habit, I'll stick with it, just not with any serious discipline.

nicowilliams:
@orzen if you change your patch to leave the pointer field zero'ing, I'll take it.

vdukhovni:
OpenSSL does not zero pointers after calloc(), but there is a build-time assertion that checks that the platform we're building for has zero-valued NULL pointers. This was the team's compromise on pointer representation portability. Oddball platforms are not supported, but there's a compile-time check. The test probably does cover potential differences between (void *), (char *) and function pointers... :-(

nicowilliams:
@vdukhovni ah, I like that compromise.  I take it you don't?

vdukhovni:
The compromise is ok. In Postfix CPU is not the bottleneck, so we don't bother with micro-optimizations. We need fast I/O, not peak CPU. In OpenSSL there is arguably some benefit from avoiding all the explicit member assignments, though a sufficiently good compiler should be able to optimize them out.

nicowilliams:
OK, thanks.  I've merged this.  I'll think about adding a configure check for zero-valued pointer behavior.

PR #1900, opened 2019-05-12
Optimise size of docker file

jamesgawn:
A minor tweak to the docker file to maintain the same image for building the jq binary, while changing the execution environment to scratch. This will reduce the download size of the resultant docker image from 74 MB to 2 MB.


jsok:
This is a very useful change.
The auto-builds on DockerHub appear to not be working anymore either, can they be re-enabled to have up to date image created?

PR #1907, opened 2019-05-21
Unify/simplify the MultiByteToWideChar() code and add wrappers for open()/stat()

bit2shift:
This is the patch I mentioned on https://github.com/stedolan/jq/issues/1066#issuecomment-494579743.
Patches 2 subtle buffer overruns in:
```c
// cchWideChar is the size, in characters, of the buffer indicated by lpWideCharStr.
// It was erroneously being passed twice that origina value, the corresponding number of bytes.
MultiByteToWideChar(CP_UTF8, 0, fname, -1, wfname, sz);
                                                   ^^
MultiByteToWideChar(CP_UTF8, 0, mode, -1, wmode, sz);
                                                 ^^
```
by way of unifying duplicate code into a separate function: `utf8_to_utf16()`.
Also added a corresponding wrapper for `open() -> _wopen()`.

~Pending: dealing with the unsolved problem with modules located in Unicode paths.~
Edit: dealt with.

### DIFF
--- a/src/linker.c
+++ b/src/linker.c
@@ -35,7 +35,9 @@ static int path_is_relative(jv p) {
   const char *s = jv_string_value(p);
 
 #ifdef WIN32
-  int res = PathIsRelativeA(s);
+  wchar_t *ws = utf8_to_utf16(s);
+  int res = PathIsRelativeW(ws);
+  jv_mem_free(ws);
 #else
   int res = *s != '/';
 #endif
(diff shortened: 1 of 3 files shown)


bit2shift:
Noteworthy: I switched from `alloca()` to `jv_mem_calloc()` since UNC paths have no length restriction.
Besides, `alloca()`'d memory in `utf8_to_utf16()` wouldn't be accessible to parent functions.

Issue #1624, opened 2018-03-10
substring of index/rindex doesn't work for utf8 inputs


ilovenwd:
```
echo  '正xyz' | jq -Rsr '.[:rindex("x")]'
正xy
while it should output:
正
```
It seems substring thinks length(正xy)==3 
but rindex thinks length(正xy)==5
this issue make jsonp parse example in FAQ fails for utf8 inputs

```
Q: How can I convert JSON-P (JSONP) to JSON using jq?
A: Assuming that the padding takes the form of a function call:
$ jq -s -R  '.[1+index("("): rindex(")")] | fromjson'
```

pkoppstein:
You're right and I've updated the FAQ so that it uses `match`.  In the example you give, we would have:

```
echo  '正xyz' | jq1.5 -Rsr '.[: (match("x").offset)]'
正
```

ilovenwd:
match works, but it's really confusing that index and slice use different string model (bytes vs strings)
it's very different from common languages like c,python, etc.
I suggest add another pair of functions like indexu/rindexu to behave exactly as substring slice does.

itchyny:
I believe this behavior is worth changing its default. How much is byte index important in jq? Defining index of type string by `explode | .[$x|explode]` will work with utf8 strings. If someone needs conversion between string and bytes, how about adding byte-version of explode and implode?

itchyny:
Same issue: #1430.

pkoppstein:
@itchyny - Changing the semantics of `index` would only be possible in a "Major Release" of jq, and might never happen.

Rather than tilting at that particular windmill, I would suggest adding a new C-coded built-in function with the desired semantics, not least because the existing implementation of `index` is ill-suited for finding the first index of anything.  

Although there is something to be said for a function with a narrow domain (e.g. `codepointOf` for JSON strings), it would be more in keeping with jq's existing builtins to be polymorphic, which would suggest a name such as `indexOf`, though a more distinctive name would no doubt be preferable.

itchyny:
Okay, thanks for detail explanation.

Issue #1430, opened 2017-06-19
index() function returns wrong offset for non-ascii chars


atschabu:
I'm trying to strip away some text from part of a text. Trying to use something like `sub("!.*"; "")` doesn't work, as it is giving me a Segmentation fault when text is too long. So I tried to go this route:

`$ jq '.msg | .[0:index("!")]'`

which works fine with input like:
`{"msg": "hello world!"}
`
but fails when text contains wide characters:
`{"msg": "здравствуй мир!"}
`
```
$ echo '{"msg": "здравствуй мир!"}' | jq '.msg | index("!")'
27
```

```
$ echo '{"msg": "hello world!"}' | jq '.msg | index("!")'
11
```
```
$ jq --version
jq-1.5
$ uname -a
Darwin atschabu-C02SF0UTG8WM 15.6.0 Darwin Kernel Version 15.6.0: Tue Apr 11 16:00:51 PDT 2017; root:xnu-3248.60.11.5.3~1/RELEASE_X86_64 x86_64
```

pkoppstein:
There is some documentation about this on the "Pitfalls" page (https://github.com/stedolan/jq/wiki/How-to:-Avoid-Pitfalls)

In brief, you can use `match/1`:

    echo '{"msg": "здравствуй мир!"}' | jq '.msg | match("!").offset'
    14

This works in jq 1.5 and later.

atschabu:
My bad. I haven't even realized there is a wiki. I took all the information from the manual, which didn't mention anything about index being byte wise. I'll give match a go.

I still haven't figured out when exactly the Segmentation fault is happening, as I couldn't find the input yet which is producing it. But I went by the assumption it is related to issue [922](https://github.com/stedolan/jq/issues/922) until I can proof the opposite.

I guess we can close this one, and I'll open a new ticket, in case my segmentation fault issue is not related to 922.

nicowilliams:
No, this is a bug.  We should fix it.

PR #1916, opened 2019-06-04
improve index, rindex and indices against string to count the index by utf8 characters

itchyny:
This is a fix for  #1430 and #1624 to resolve confusing byte index.


nicowilliams:
@itchyny why did you close this?

itchyny:
Because I think this will not be merged in the near future due to this comment https://github.com/stedolan/jq/issues/1624#issuecomment-498491794 (I still think it's worth changing the behavior though).

nicowilliams:
Ah, yes, of course.  Thanks.

emanuele6:
This is a fix for #1430.

Do we want to merge this, or close that issue?

nicowilliams:
> This is a fix for #1430.
> 
> Do we want to merge this, or close that issue?

Great question.  String indexing is fraught; maybe we shouldn't have it at all!  But if we'll have it we should definitely have:

 - index by nth glyph
 - index by nth codepoint

and maybe:

 - index by nth byte (because it's all UTF-8, and indexing by byte in Unicode-aware apps may make sense)

That we have indexing by nth byte is essentially a design bug born of not having enough Unicode functionality.

emanuele6:
@nicowilliams 

I think we should change the behaviour of `index` to make `.[index("a"):]` work as expected, since string slice indexing is by nth unicode codepoint, otherwise `index`/`rindex`/`indices` are pretty much useless for strings. That is also how `index` functions work in `gojq`/`fq`.

PR #1929, opened 2019-06-17
Fix @sh escaping to properly handle apostropher (single-quote).

cfraizer:
For a jq string with a value `foo " bar 'baz'`, that is:
- the word `foo`
- `<space>`
- `&quot;`
- `<space>`
- the word `bar`
- `<space>`
- the word `baz` enclosed within single quotes (`&apos;`)

The current `@sh` escape spits out `'foo " bar '\''baz'\'''` the part after the last space is:
- `&apos;<backslash>&apos;&apos;`
- the word `baz`
- `&apos;<backslash>&apos;&apos;&apos;`

Bourne Shell and Bash do not allow escaping of single quotes inside single quotes.
Instead, you must end the enclosing string and then add the single quote as its own string (automatically appended by juxtaposition).

The correct output should be: `'foo " bar '"'"'baz'"'"''`, that is:
- the word `foo`
- `<space>`
- `&quot;`
- `<space>`
- the word `bar`
- `<space>`
- `&quot;&apos;&quot;`
- the word `baz`
- `&quot;&apos;&quot;`
- [optional, but hurts nothing], `&apos;&apos;` (an empty string appended to the end). Just makes implementation easier.


cfraizer:
I'm a n00b. It took me a minute to understand how the test worked.

cfraizer:
I don't know how to fix the travis-ci failure noted above. I fixed the test and now `make check` passes when I run it.

nicowilliams:
@cfraizer If you look in the failed builds' logs you'll see that `mantest` failed, and if you look at the `test-suite.log` file you'll see something like this:

```
Testing '@sh "echo \(.)"' at line number 630
*** Expected "echo 'O'\\''Hara'\\''s Ale'", but got "echo 'O'\"'\"'Hara'\"'\"'s Ale'" for test at line number 632: @sh "echo \(.)"
```

which relates to this in `docs/content/manual/manual.yml`:

```
          - program: '@sh "echo \(.)"'
            input: "\"O'Hara's Ale\""
            output: ["\"echo 'O'\\\\''Hara'\\\\''s Ale'\""]
```

cfraizer:
The problem I had was not related to jq but to I was using to deal with the output.

The escaping jq does is fine. Instead of outputting `&quot;&apos;&quot;`, it outputs `<backslash>&apos;`, which is fine. The latter is perfectly acceptable when not enclosed in a pair of `&apos;`.

This change can be rejected. My apologies for the distraction.

cfraizer:
I wonder if it's okay for me to just close this myself?

nicowilliams:
Yes, you can close it yourself :)  And now you know how to make contributions to jq, which is great!  Looking forward to any future contributions!  Thanks!

PR #1944, opened 2019-07-09
clarify filter usage

tessus:
The current paragraph is not complete, since a digit is not a special character.

Changing it to:

    If the key contains special characters or starts with a digit,
    you need to surround it with double quotes like this:
    `."foo$"`, or else `.["foo$"]`.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -317,8 +317,9 @@ sections:
           is, keys that are all made of alphanumeric characters and
           underscore, and which do not start with a digit.
 
-          If the key contains special characters, you need to surround
-          it with double quotes like this: `."foo$"`, or else `.["foo$"]`.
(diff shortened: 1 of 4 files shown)


wtlangford:
This looks great, thanks! Should we also demonstrate a version like `."2foo"` for extra clarity?

tessus:
IMO there's no need, and I wouldn't even know where to add it an how.

It would make it more complex and confusing. Since you've already used both forms, you'd have to use both for the 2nd example as well:

e.g.:

```
`."foo$"` or `.["foo$"]`, and `."2oo"` or `.["2foo"]`
```

For me this makes it less readable, but maybe that's just me.

tessus:
@wtlangford is antthing else blocking this from being merged?

wtlangford:
Nope, I was just busy. Thanks!

tessus:
Thanks.

PR #1948, opened 2019-07-21
docs: fix seriously dangerous download instructions for Arch Linux

eli-schwartz:
The current instructions tell users to perform two actions:
- update the package database
- install the jq package

The only thing users need to or should be doing is actually installing the jq package -- regardless of which version is being installed. Guidelines on how to perform system updates are massively out of scope.

In the case of partially performing a system update as a prerequisite for installing jq, the official guidance from Arch Linux is not to do this: partial updates are not supported, we refuse to support them, and anyone who does try to perform them anyway is assumed to know so much about their system that they clearly do not ever need help from anyone else (which is a good thing since they won't get it). The result is a frankensteined system that can only ever be supported by the person who frankensteined it to begin with. The only reason the package manager even allows it to occur in the first place is because other distributions using pacman might have different LTS policies, and because it would prevent expert users from being in control of their system, as per the
traditional Unix philosophy:

"Unix was not designed to stop you from doing stupid things, because that would also stop you from doing clever things."

Consequences of performing partial updates without understanding the ramifications in extensive detail can include breaking the partially updated application (jq), breaking any application that shares a mutual dependency with the partially updated application (which jq is *lucky* to only depend on the ever-backwards-compatible glibc), or breaking the entire operating system by leaving armed traps behind for the next time a `pacman -S new-package` is executed and thereby breaks *its* cascading
dependencies.

See: https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported

### DIFF
--- a/docs/content/download/default.yml
+++ b/docs/content/download/default.yml
@@ -29,7 +29,7 @@ body:
 
        * jq 1.5 is in the official
          [Arch](https://www.archlinux.org/packages/?sort=&q=jq&maintainer=&flagged=)
-         repository.  Install using `sudo pacman -Sy jq`.
+         repository.  Install using `sudo pacman -S jq`.
 
        * jq 1.6 binaries for
          [64-bit](https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64)


eli-schwartz:
I will never understand why this memetic hazard of `pacman -Sy` keeps appearing everywhere, but somehow it's incredibly common for users to submit PRs adding installation instructions for some software on Arch Linux, that contain this issue.

Sorry you got caught up in it. ;)

wtlangford:
I'm not familiar with pacman/arch, but my reading of the documentation implies that the `-y` flag simply refreshes the package database (similar to what yum/dnf/apt do regularly).
Why is that harmful on arch, but not on other platforms?

I do see in their documentation that they recommend against things like `pacman -Sy jq`, so I'm happy to merge this PR, but I'd like to understand why this is so different on arch.

eli-schwartz:
Because other Linux distributions have a policy to support versioned dependencies -- if you `apt update` to refresh the repositories, then `apt install foo`, program "foo" will depend on the sonames it requires in order to run. You cannot install a new version of "foo" which is incompatible with the other programs on your system, because of package versioning conflicts, and in fact there will be multiple versions of foo (libfoo4, libfoo5, libfoo6) that can coexist in order to ensure that program bar (which links to an older libfoo release) can still be installed and run while using newer versions of libfoo for other programs.

This is, mostly, the difference between a stable release where there is a strong guarantee of backwards-compatibility ("we guarantee you can continue to use the old version that you are accustomed to") vs. rolling release distros ("Arch Linux assumes you want the latest version of everything, so we will rebuild every program as needed against new libraries -- new features are assumed to be good, but if you really don't like the changes, try discussing the feature with upstream").

tl;dr
- On stable releases, system stability means packages depend on the pinned versions they need.
- On Arch Linux, system stability means that if you ensure all packages are updated together, we guarantee it works, otherwise you are assumed to be Mr Expert Debugger (and can fix your own problems).
- People who go outside of the sandbox provided by the distro, whether that is via Arch Linux's `pacman -Sy` or by using `dpkg -i` (which does not install dependencies) with a carefully curated list of packages+versions as *.deb files, should either be doing so intentionally as a very low-level debugging tool, or be strongly warned to not do so at all.

wtlangford:
Thanks, that all makes a bit more sense to me now! (I'd missed the part about arch being rolling release)

I'll merge this, then, and make a point to update the site docs soon.

PR #1960, opened 2019-08-14
Define range/2 by jq-coded function

itchyny:
Implementation of builtin functions in bytecode occasionally  results into some complex bugs. As for range/2, there was #886, which was fixed by a47b32999b8a6a754b91909548700efdad3c2dd9. But I think the implementation is complicated enough and it's worth make it simple by jq-coded function. The RANGE is unnecessary and reducing the instructions simplifies the interpreter loop.


pkoppstein:
Unfortunately, defining `range/2 `in this way is significantly slower than having it as a builtin.

The slow-down is especially relevant because `range` is often nested.

Also, `range/2` gives a useful error message if the args are not as they should be.  With the proposed def,  `range(0;null)` yields empty, and `range(0;"")` produces an infinite stream.  Of course it would be easy to address this issue, but the slow-down, it seems to me, renders it moot.

itchyny:
Yeah, it is actually slower, but I'm not sure this is significantly slower. This is fast enough to iterate to 10000000 which may be too large for normal use case isn't it?
```sh
 % time jq -n 'range(10000000)' > /dev/null
jq -n 'range(10000000)' > /dev/null  7.26s user 0.03s system 98% cpu 7.426 total
 % time ./jq  -n 'range(10000000)' > /dev/null
./jq -n 'range(10000000)' > /dev/null  11.68s user 0.03s system 99% cpu 11.816 total
```
Also I'm not sure how it is important to keep the behavior of `range(0; "")`, which is inconsistent with `range(0; ""; 1)`.
Anyway, the comment `Note that we can now define `range` as a jq-coded function` should be replaced with the reason why range/2 is implemented as bytecoded function.

PR #1977, opened 2019-09-12
Override field color with JQ_COLORS

rbolsius:
This is a small enhancement to make the field color configurable via the existing JQ_COLORS environment variable by allowing the field color to be specified as the eighth color.  If JQ_COLORS contains only seven colors, then the field color will continue to default to bold blue (1;34).

Below is an example JQ_COLORS string that makes the field names bold yellow:

export JQ_COLORS='1;30:1;32:1;32:1;32:1;31:0;37:0;37:1;33'

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -3284,9 +3284,10 @@ sections:
         - color for strings
         - color for arrays
         - color for objects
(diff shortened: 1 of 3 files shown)


itchyny:
Implemented by #2703.

PR #1993, opened 2019-10-22
Better dumping

### LINKED ISSUE #1788: jq-1.6 - Segmentation fault when combining options -a and -r
**Describe the bug**
On jq 1.6, using the `-a` (ascii-output) and `-r` (raw) options conflict. 
This behaviour doesn't happen on jq 1.5

**To Reproduce**

**1.5**

```
$ jq --version
jq-1.5

$ echo '{"key1":"value1"}' | jq -r -a .key1
value1
```

**1.6**

```
$ jq --version
jq-1.6

$ echo '{"key1":"value1"}' | jq -r -a .key1
"value1"
```

Notice how output still contain quotes despite the `-r` flag.

Removing the `-a` flag produces the correct behaviour

```
$ echo '{"key1":"value1"}' | jq -r .key1
value1
```

This problem is reproducible with both the `jq-linux64` and `jq-osx-amd64` builds.

The linux build even shows a `Segmentation fault` error:

```
$ echo '{"key1":"value1"}' | ./jq.1-6 -r -a .key1
"value1"
Segmentation fault

$ echo '{"key1":"value1"}' | ./jq.1-6 -r .key1
value1
```

bb010g:
A fair amount of refactoring went into this PR to make each step clear and prevent issues like those in #1789, where escaping wasn't correctly handled. I'll summarize the non-bugfix commits, and the commit messages for the fixes follow. `jvp_dump_string` changed from consuming a boolean-use `int ascii_only` to `enum jv_print_flags` `int flags`. This makes `jvp_dump_string` more capable of responding to input formats, reducing pressure on consumers to get it right. `jvp_dump_char` is factored out from `jvp_dump_string`, making future increases in complexity to both clearer. `JV_PRINT_RAW` and `JV_PRINT_RAWCOLOR` flags are introduced to `enum jv_print_flags`, consolidating prints in `main.c`'s `process` to `jv_dump(result, dumpopts)`. The `JV_PRINT_RAW` flag is not passed down to recursive `jv_dump*` calls for arrays or objects. `JV_PRINT_RAWCOLOR` is unset by default and has no API currently, but is there if colored, raw output is desired in the future. `JV_PRINT_RAWCOLOR` avoids problems with colored output by noticing when a type would be raw printed and disabling `JV_PRINT_COLOR` before that's a problem.

### Fix `stderr/0` builtin to match documentation

> It's supposed to print raw, compact, and with no decoration or trailing newline. Previously, it would print JSON formatted, compact, with no trailing newline and no easy way to obtain one, and ignoring the `--ascii-output` / `-a` flag.

`f_stderr`, instead of calling `jv_dumpf(jv_copy(input), stderr, 0)`, now has a `stderr_cb` setup similar to `f_debug` that it uses identically. The only functional difference is `main.c`'s `stderr_cb`:

```c
static void stderr_cb(void *data, jv input) {
  int dumpopts = *(int *)data;
  jv_dumpf(input, stderr, (dumpopts & ~JV_PRINT_PRETTY) | JV_PRINT_RAW);
  // debug_cb prints a newline here.
}
```

### Bring `--raw-output` and `--ascii-output` to agreement

> Escapes are still printed whenever characters outside the ASCII plane are encountered. To avoid ambiguity, backslash is the only ASCII character escaped (as `\\`).
>
> Fixes #1788, properly this time. Closes #1789.

`jvp_dump_char` gains a boolean-use `int raw` parameter that is used to always print ASCII characters (save backslash) and turn on escapes for everything else. `jvp_dump_string` learns to handle raw strings in the non-`(raw && !ascii_only)` path. Due to the earlier changes, every other use of raw strings is immediately corrected.

---
This change is [Reviewable](https://reviewable.io/reviews/stedolan/jq/1993)

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1189,7 +1189,11 @@ static jv f_debug(jq_state *jq, jv input) {
 }
 
 static jv f_stderr(jq_state *jq, jv input) {
-  jv_dumpf(jv_copy(input), stderr, 0);
+  jq_msg_cb cb;
+  void *data;
+  jq_get_stderr_cb(jq, &cb, &data);
+  if (cb != NULL)
+    cb(data, jv_copy(input));
   return input;
 }
 
--- a/src/execute.c
+++ b/src/execute.c
@@ -49,6 +49,8 @@ struct jq_state {
   void *input_cb_data;
   jq_msg_cb debug_cb;
   void *debug_cb_data;
+  jq_msg_cb stderr_cb;
+  void *stderr_cb_data;
 };
 
 struct closure {
@@ -1237,6 +1239,16 @@ void jq_set_debug_cb(jq_state *jq, jq_msg_cb cb, void *data) {
 void jq_get_debug_cb(jq_state *jq, jq_msg_cb *cb, void **data) {
   *cb = jq->debug_cb;
   *data = jq->debug_cb_data;
+}
+
+void jq_set_stderr_cb(jq_state *jq, jq_msg_cb cb, void *data) {
+  jq->stderr_cb = cb;
+  jq->stderr_cb_data = data;
+}
+
+void jq_get_stderr_cb(jq_state *jq, jq_msg_cb *cb, void **data) {
+  *cb = jq->stderr_cb;
+  *data = jq->stderr_cb_data;
 }
 
 void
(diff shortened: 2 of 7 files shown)


bb010g:
Rebased and force pushed to run Travis after the fixes.

bb010g:
Is there anything I can do to help this get reviewed & in-tree?

nicowilliams:
I'll take a look as soon as I can, which might not be till next week :(

bb010g:
bump again

bb010g:
Rebump.

bb010g:
Rebased with manual review over new commits (24564b2d345398a818eb74a194faf3733655d73a got close to this patch's lines), and still passes tests.

vergenzt:
This PR addresses #2063

PR #1995, opened 2019-10-22
Rework pipenv requirement to be more relaxed

wtlangford:
`brew install jq --HEAD` doesn't currently work because of our pipenv requirement.
This reworks the way we handle the pipenv requirement to make it less required for running all the tests- now you only _need_ it if you're making changes to the manpage yml.

We keep a prebuilt copy of the mantests in place for this purpose- this should be kept up to date as changes are made to the man tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -130,27 +130,46 @@ endif
 
 ### Tests (make check)
 
-if ENABLE_DOCS
 TESTS = tests/optionaltest tests/mantest tests/jqtest tests/onigtest tests/shtest tests/utf8test tests/base64test
-else
-TESTS = tests/optionaltest tests/jqtest tests/onigtest tests/shtest tests/utf8test tests/base64test
-endif
 TESTS_ENVIRONMENT = NO_VALGRIND=$(NO_VALGRIND)
 
+# This is a magic make variable that causes it to treat tests/man.test as a
+# DATA-type dependency for the check target. As a result, it will attempt to
+# run any defined targets for tests/man.test as a dependency for check. This
+# allows us to ensure that the tests are up-to-date if the manual has been updated
+check_DATA = tests/man.test
+
+### Building the man tests
+
+# We use the examples in the manual as additional tests, to ensure they always work.
+# As a result, we need to rebuild the tests if the manual has been updated.
+# Making changes to the manpage without having the python deps means your
+# tests won't run. If you aren't making changes to the examples, you probably
+# don't care. But if you are, then you need to run the tests anyway.
+tests/man.test: $(srcdir)/docs/content/manual/manual.yml
+if ENABLE_DOCS
+	$(AM_V_GEN) ( cd ${abs_srcdir}/docs; $(PIPENV) run python build_mantests.py ) > $@
+else
+	@echo Changes to the manual.yml require docs to be enabled to run the tests
+	@false
+endif
 
 ### Building the manpage
 
+# We build the docs from the manpage yml. If no changes have been made to the
+# manpage, then we'll end up using the cached version. Otherwise, we need to
+# rebuild it.
 man_MANS = jq.1
+jq.1.prebuilt: $(srcdir)/docs/content/manual/manual.yml
 if ENABLE_DOCS
-jq.1: $(srcdir)/docs/content/manual/manual.yml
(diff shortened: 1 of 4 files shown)


wtlangford:
@pkoppstein: moving our conversation from https://github.com/stedolan/jq/issues/2011 over to here to avoid noise in the other user's email.

Would you mind sharing the set of commands you ran to clone and attempt to build this? I haven't had any issues anywhere I've tried, so I'm suspecting there's something one of us is missing about the process.

pkoppstein:
@wtlangford - As I mentioned, I used SourceTree to clone.  A copy of the .git/config file is appended.  Then I checked out dd87c3b9d7acf4efaced0a8eccb534df82c6cf27

The actual commands:
    cd jq; make clean; autoreconf -if ; ./configure ; make
```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
	ignorecase = true
	precomposeunicode = true
[remote "origin"]
	url = https://github.com/wtlangford/jq.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
	remote = origin
	merge = refs/heads/master
[branch "fix-brew-head-build"]
	remote = origin
	merge = refs/heads/fix-brew-head-build
[submodule "modules/oniguruma"]
	active = true
	url = https://github.com/kkos/oniguruma.git
```

wtlangford:
Hm, okay, so there are a couple of things going on:
1) I'm running `pipenv check` to detect the presence of your python env, which will try to set up your env for you. That's not what I thought it did. There's another flag I can use, which does what I meant. (Interestingly, blowing away your jq folder doesn't necessarily remove the env pipenv created for you- which was why I never noticed the issue)
2) I set up the make rules for generating the manpage incorrectly

I'll push a fix in a bit

wtlangford:
@pkoppstein: That commit should fix the issues. Let me know how it goes for you. To make sure you have nothing from pipenv, run this: `cd docs; pipenv --rm`. (Or just make sure pipenv isn't in your path, I guess)

pkoppstein:
@wtlangford - I'm not quite sure what counts as a success here.   Using the latest commit and after running `pipenv --rm` as suggested, jq builds successfully:
    ./jq --version
    jq-1.3-997-g9e8a9dd-dirty

But that was also true before.  Also as before `make` ends with ominous error messages, this time:
```
Changes to the manual.yml require docs to be enabled to generate an updated manpage
make[2]: *** [jq.1.prebuilt] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

Perhaps that's what you intend, but I think that most users (including myself) would prefer to see more soothing messages at the end of the process if it has completed "successfully".

Are there instructions somewhere for updating the manpage?  I used pipenv to create a "virtual environment (`✔ Successfully created virtual environment! `) but it does not seem to make a difference (i.e., the same error messages as shown above are generated).

Thanks!

wtlangford:
That's an error. It failed to rebuild the manpage, because it thinks it needs to, but didn't detect a python environment at configure.

I'm surprised, because it shouldn't have tried to rebuild the manpage, so long as the manual.yml was older than the prebuilt version.

I get the expected behavior on my machine (where it doesn't try to rebuild the manpage), but I'll do some more testing tomorrow to try and figure it out.

Makefile.am
```diff
@@ -130,27 +130,36 @@ endif
...
-endif
 TESTS_ENVIRONMENT = NO_VALGRIND=$(NO_VALGRIND)
 
+# We build the docs from the manpage yml. If no changes have been made to the
+# manpage, then we'll end up using the cached version. Otherwise, we need to
+# rebuild it- making changes to the manpage without having the docs means your
+# tests won't run. If you aren't making changes to the examples, you probably
+# don't care. But if you are, then you need to run the tests anyway.
+check_DATA = tests/man.test
+tests/man.test: $(srcdir)/docs/content/manual/manual.yml
```
nicowilliams: Do we need a dependency on this target?

wtlangford: Yep, this forces `make check` to update `tests/man.test` if the `manual.yml` has been updated. It's intended to prevent changes to the manual not getting reflected in the prebuilt `man.test`.

wtlangford:
I'm going to merge this, because I think it's an improvement to the existing experience. We'll address any other issues as they come up.

PR #2000, opened 2019-10-26
Fix assertion when using multiple data imports

wtlangford:
```jq
import "a" as $a;
import "b" as $b;
def f: $a,$b;
f
```

This fails an assertion: `Assertion failed: (var < fr->bc->nlocals), function frame_local_var, file src/execute.c, line 104.`

The issue _appears_ to be that we don't properly increment the number of local vars defined in the frame where we `STORE_GLOBAL`, which means we can easily have more vars stored than we expect in our assertions. The code for `STOREV` _does_ increment this, so I duplicated that logic into `STORE_GLOBAL`. I can't think of any ways that this might cause issues. @nicowilliams can you?

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -1347,6 +1347,7 @@ static int compile(struct bytecode* bc, block b, struct locfile* lf, jv args, jv
       code[pos++] = nesting_level(bc, curr->bound_by);
       uint16_t var = (uint16_t)curr->bound_by->imm.intval;
       code[pos++] = var;
+      if (var > maxvar) maxvar = var;
     } else if (op->flags & OP_HAS_CONSTANT) {
       code[pos++] = jv_array_length(jv_copy(constant_pool));
       constant_pool = jv_array_append(constant_pool, jv_copy(curr->imm.constant));


irgeek:
I just ran into the same issue. It'd be really awesome if this were merged so I didn't have to resort to the hack of concatenating all my data files so I can import them all at once and access them as array members. Alternatively, if importing JSON data is _supposed_ to be restricted to one file, that should be documented.

lorenzleutgeb:
I also ran into this just now, consider the following snippet:

```jq
import "./a" as $a;
import "./b" as $b;

def a: $a;
def b: $b;
```

When completed with either 

```jq
{a: a}
```
or
```jq
{b: b}
```

it will evaluate. However, when completed with

```jq
{a: a} * {b: b}
```

it fails with

```
jq: src/execute.c:104: frame_local_var: Assertion `var < fr->bc->nlocals' failed.
```

That's quite confusing, and I spent quite a lot of time figuring out what's going on. Please fix this!

itchyny:
LGTM, I will merge this patch with the following test, thank you.
```sh
jq -n -c -L tests/modules 'import "data" as $a; import "data" as $b; def f: {$a, $b}; f'
```

PR #2004, opened 2019-11-01
Rewrite Dockerfile to use Alpine Linux, and produce a minimal image

thaJeztah:
This pull request contains two changes (described below). With this change in place, it should be possible to enable auto-builds on Docker Hub, as it looks like the current image has been [last updated four years ago](https://hub.docker.com/r/stedolan/jq), is likely outdated, and probably contains vulnerabilities;

Screenshot 2019-11-01 at 11 12 26

### Change Dockerfile to be a multistage build

The existing Dockerfile produced a broken binary (static
compile caused the binary to segfault).

This patch changes the Dockerfile to

- Build a non-static binary
- Use a multi-stage build: separate stages are used for build and
  deploy; the build stage contains the tools needed for building
  the binary, after which the build artefacts are copied to the
  final stage. which produces the actual image.
  Using this approach, the build-tools can be kept in the Docker
  build-cache, which improves performance of repeated builds and
  does not require cleanup steps in the Dockerfile itself, which
  can simplify the steps.
- The official Python image is used for the build stage, so that
  there is no need to install python manually. The version of python
  is set to the same version as the version used in CI, but can be
  overridden using at build-time using a build-arg:
    ```console
    docker build --build-arg PYTHON_VERSION=x ..
    ```
- Both build- and deploy stages use the debian "stretch" variants,
  which matches what's used in the current version of the Dockerfile,
  but could be updated to the current stable ("buster") variant if
  there is a need.
- The final stage uses the "slim" variant to reduce the size. For
  comparison, below are sizes of the final image when using the
  regular, or "slim" version of the image:

    ```console
    REPOSITORY  TAG                 IMAGE ID            CREATED             SIZE
    jq          stretch-slim        4b7f380970f0        About a minute ago  70.3MB
    jq          stretch             c5fa8b766cd4        7 minutes ago       119MB
   ```

A `.dockerignore` file is also added to prevent "busting" the Docker
build-cache if not needed, and steps in the Dockerfile are optimized
for BuildKit, but remain compatible with the legacy builder. To build,
and verify the image:

```console
DOCKER_BUILDKIT=1 docker build -t jq .
```

And, test the image, use it (for example) to pretty-print its own
"docker image inspect" output:

```console
docker image inspect jq | docker run -i --rm jq .
[
  {
    "Id": "sha256:4b7f380970f0d06d4ca9137fbbb3fa3f6554da996147d1d956cda6f44be25650",
    "RepoTags": [
      "jq:latest",
      "jq:stretch-slim"
    ],
```

### Dockerfile: use Alpine image to build static binary

This patch switches the Dockerfile to use an Alpine Linux
base image to build the static binary. Given that the binary
is now fully static, a minimal image can be produced that
only contains the essential artefacts.

The static version of the Docker image does not contain the man-
pages, so the existing (debian based) Dockerfile is kept, but
renamed to Dockerfile.debian, and can be built by specifying
the alternative Dockerfile using the `-f` flag:

```console
DOCKER_BUILDKIT=1 docker build -t jq:stretch -f Dockerfile.debian .
```

With this patch applied, the image size is reduced drastically:

```console
DOCKER_BUILDKIT=1 docker build -t jq .

docker image ls

REPOSITORY  TAG                 IMAGE ID            CREATED             SIZE
jq          latest              7da2ea1c016d        5 seconds ago       800kB
jq          stretch-slim        4b7f380970f0        About a minute ago  70.3MB
jq          stretch             c5fa8b766cd4        7 minutes ago       119MB
```

The image can be tested, for example, by inspecting itself and pretty-
printing the json:

```console
docker image inspect jq | docker run -i --rm jq .
[
  {
    "Id": "sha256:7da2ea1c016d97e7b52b09d5a323a1e7c0e4dbdbc77ce2715aedd3188721e4da",
    "RepoTags": [
      "jq:latest",
      "jq:static"
    ],
...
```

Or a simple test;

```console
echo '{"a":1, "b":2}' | docker run -i --rm jq '.a+.b'
```


thaJeztah:
ping @stedolan @wtlangford  ptal

Nicceboy:
Any updates about this? The image size change alone would make this really useful.

kolyshkin:
@wtlangford @nicowilliams PTAL

Dockerfile
```diff
@@ -1,54 +1,54 @@
...
+
+COPY .gitmodules .
+COPY .git        ./.git/
+RUN git submodule init
+RUN git submodule update
+RUN sed -i.bak '/^AM_INIT_AUTOMAKE(\[-Wno-portability 1\.14\])$/s/14/11/' modules/oniguruma/configure.ac
+
+COPY . .
+RUN autoreconf -if
+RUN ./configure --disable-valgrind --with-oniguruma=builtin YACC=/usr/bin/bison --prefix=/build/
```
kolyshkin: It's better to do `./configure --prefix=/usr/local` ...

kolyshkin: ...which is the default if we're dealing with classic autoconf, so just drop the `--prefix`.

The reason why `--prefix=/build/` could be bad is this path _could_ end up in a compiled binary (e.g. it will look for config files under `/build/etc/` and so on) which is definitely not what we want here.

Dockerfile
```diff
@@ -1,54 +1,54 @@
...
+RUN make BISON_PKGDATADIR=/usr/bin/bison src/parser.c || make src/parser.c
+RUN make -j8
+RUN make check -j8
+RUN make install
```
kolyshkin: ... and `make install DESTDIR=/build`

This way everything will be installed under `/build`, i.e. `/build/usr/local/bin/jq` etc.

To summarize, assuming we're dealing with classing autoconf/automake here, instead of

`./configure --prefix /temp/install/path && make && make install`

one should use

`./configure && make && make DESTDIR=/temp/install/path install`.

Dockerfile
```diff
@@ -1,54 +1,35 @@
-FROM debian:9
+FROM alpine:3.10 AS build
```
tao12345666333: we can update to use alpine:3.11 :cat:

tao12345666333 (approved):

Dockerfile
thaJeztah: I updated with your recommendation (but realised that these changes were replaced by the second commit, which switches to alpine; still, didn't hurt to have it in there I guess)

Dockerfile
thaJeztah: Updated to use alpine 3.12 👍

thaJeztah:
@stedolan @kolyshkin @tao12345666333 forgot about this one; updated with your recommendations; PTAL

tao12345666333 (approved):
Great！

PR #2009, opened 2019-11-13
Change Homebrew version in docs to 1.6

voke:
Stable version is now 1.6 for [jq formula](https://formulae.brew.sh/formula/jq)


wtlangford:
Thanks!

PR #2010, opened 2019-11-14
Fix to incorrect --stream paths

cefn:
Thought something was wrong with the example provided against the --stream option. Proved that the index of the path to "b" was incorrect by running the example on the terminal.  Instead of `[[1,0],"b"]` it should read `[[2,0],"b"]` . 

Additionally, this isn't the whole stream. Given it's a depth-first traversal, after visiting descendants, it will visit the paths of the ancestors too. However I guess these have been omitted deliberately for brevity? 

If you prefer the full sequence I'll force push a fix to this PR branch which contains the ancestor paths traversed too.  

## Example in terminal

```
echo '[[],"a",["b"]]' | jq --stream --compact-output
[[0],[]]
[[1],"a"]
[[2,0],"b"]
[[2,0]]
[[2]]
```

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -113,10 +113,11 @@ sections:
 
       * `--stream`:
 
-        Parse the input in streaming fashion, outputing arrays of path
-        and leaf values (scalars and empty arrays or empty objects).
-        For example, `"a"` becomes `[[],"a"]`, and `[[],"a",["b"]]`
-        becomes `[[0],[]]`, `[[1],"a"]`, and `[[1,0],"b"]`.
+        Parse the input in depth-first, streaming fashion, outputting 
+        arrays of path and leaf values (scalars and empty arrays or 
+        empty objects). For example, `"a"` becomes `[[],"a"]`, and 
+        `[[],"a",["b"]]` becomes `[[0],[]]`, `[[1],"a"]`, and 
+        `[[2,0],"b"]`.
 
         This is useful for processing very large inputs.  Use this in
         conjunction with filtering and the `reduce` and `foreach` syntax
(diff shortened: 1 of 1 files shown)


cefn:
I noticed also https://github.com/stedolan/jq/issues/1279 which reports the issue and https://github.com/stedolan/jq/pull/1946 which is an alternate fix.

docs/content/manual/manual.yml
```diff
@@ -113,10 +113,11 @@ sections:
...
 
-        Parse the input in streaming fashion, outputing arrays of path
-        and leaf values (scalars and empty arrays or empty objects).
-        For example, `"a"` becomes `[[],"a"]`, and `[[],"a",["b"]]`
-        becomes `[[0],[]]`, `[[1],"a"]`, and `[[1,0],"b"]`.
+        Parse the input in depth-first, streaming fashion, outputting 
+        arrays of path and leaf values (scalars and empty arrays or 
+        empty objects). For example, `"a"` becomes `[[],"a"]`, and 
+        `[[],"a",["b"]]` becomes `[[0],[]]`, `[[1],"a"]`, and 
+        `[[2,0],"b"]`.
```
Alanscut: `[[],"a",["b"]]` becomes `[[0],[]]`, `[[1],"a"]`, `[[2,0],"b"]`,
       `[[2,0]]` and `[[2]]`.

Alanscut (changes requested):

docs/content/manual/manual.yml
cefn: Agreed and committed to branch.

itchyny:
The original path error was fixed by #2387 (0171a9eb6029fa326b1439b76ddef2f18f0ab2b7). I think this section explains this option in briefly, and there is a full explanation in the Streaming section. I think this is okay, so closing the PR. Thank you.

PR #2024, opened 2019-12-09
Differentiate WIN32 / Cygwin in configure script

rhlee:
Because shlwapi.h is present on both systems, as Cygwin system is detected as a WIN32. This patch adds the detection of cygwin.h .


nicowilliams:
Thanks!

PR #2028, opened 2019-12-13
Update main.c

toralf-hlag:
spelling in help text

greetings from a user (BTW the AWS mentions a https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-4074 in the used Linux Alpine Docker image we do use)


pkoppstein:
This "correction" is incorrect in that the occurrence of "that that" that it changes is correct.

nicowilliams:
The usage is correct!

> The simplest filter is ., which copies jq's input to its output unmodified (except for formatting, but note that IEEE754 is used for number representation internally, with all that that implies).

"all that that implies" is the correct formulation in English of what this is trying to say: that IEEE754 implies a bunch of issues.

PR #2032, opened 2019-12-17
correct the default JQ_COLORS to use 37, not undefined color 39

DavidBiesack:


nicowilliams:
Huh, you're quite right that `39` is not a valid color for vt100, but that is what's used in the source, so correcting the manual is not enough!

```
$ git grep 0\;39
docs/content/manual/manual.yml:      `"JQ_COLORS=1;30:0;39:0;39:0;39:0;32:1;39:1;39"`.
docs/content/manual/v1.6/manual.yml:      `"JQ_COLORS=1;30:0;39:0;39:0;39:0;32:1;39:1;39"`.
jq.1.prebuilt:The default color scheme is the same as setting \fB"JQ_COLORS=1;30:0;39:0;39:0;39:0;32:1;39:1;39"\fR\.
src/jv_print.c:  {COL("1;30"),    COL("0;39"),      COL("0;39"),     COL("0;39"),
```

nicowilliams:
Thanks!

saeedazimi-5717:
Very good

PR #2041, opened 2020-01-07
Add information on how to install on OS X with MacPorts

bn3t:
Maybe you would want to name OS X, MacOS now as it is the official name.

Small caveat, the change is simple enough and I didn't run the build to try it locally. I hope it's ok.


nicowilliams:
Thanks!

PR #2049, opened 2020-01-21
Updating Homebrew jq version on website

fishcharlie:
According to [Homebrew](https://formulae.brew.sh/formula/jq#default) it will now install version 1.6 of `jq`. This PR updates the website to reflect this change to make it clear to users that installing from [Homebrew](https://formulae.brew.sh/formula/jq#default) will get you the latest version of `jq`.

### DIFF
--- a/download/index.html
+++ b/download/index.html
@@ -109,7 +109,7 @@
 
 <ul>
 <li>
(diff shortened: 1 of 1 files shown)


nicowilliams:
Thanks!

Issue #2033, opened 2019-12-19
pad function


elarrarte-d:
Hi!

I wrote myself a pad(n) function to jq.
Maybe there is a chance to make it an official jq function?

def pad(n): if n == 0 then (.) else ["0", .] | join("") | pad(n - 1) end;

Congratulations for this great tool!
Thanks in advance!

davidfetter:
Maybe there could be a left pad (lpad) as above and a corresponding right pad (rpad) function.

elarrarte-d:
Any updates about this?
Regards!

davidfetter:
How about something along this line? It's not recursive, but it's likely pretty fast.
```jq
def lpad(string;len;fill):
  if len == 0 then string else (fill * len)[0:len] + string end;
def rpad(string;len;fill):                                                                                                                                                                                                                                             
  if len == 0 then string else string + (fill * len)[0:len] end;
```

itchyny:
If you want to pad string to make the result have the expected length, `def lpad(string; len; fill): fill * (len - (string|length)) + string;`  (or `def lpad(len; fill): fill * (len - length) + .;`) would be correct.

pkoppstein:
@DavudFetter: those defs are not idiomatic.  As in itchyny's second example, it would be more idiomatic for the string to be presented as input to the filter. It would also be in keeping with jq to use `tostring` on the input.  So, allowing for the possibility that (fill|length) > 1, one might consider:

```
def lpad($len; $fill): ($len - length) as $l | tostring | ($fill * $l)[:$l] + .;
```

itchyny:
By the way what's the difference of `($str * $l)[:$l]` and `($str * $l)`?

itchyny:
Ahhhh, sorry, ignore my comment! It is different when `($str | length) != 1`.

davidfetter:
Adjusted per comments, and thanks for looking at this!

itchyny:
By the way it doesn't work for multi-width characters, if you really want to pad spaces to strings to make them fit to a fixed width in terminals (wrap text by a frame, for example). String length and displayed width are different things.

elarrarte-d:
OK! I 've seen a lot of options here jeje. Are we going to have a pad() function or similar in the next release of jq?

Thanks!

elarrarte-d:
Add some tests about performance:
Testing 1000 times [def pad(n): if n==0 then (.) else "0" + (.) | pad(n - 1) end]
[...]

PR #2060, opened 2020-02-14
Added lpad and rpad

davidfetter:
#2033

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -203,6 +203,11 @@ def ascii_downcase:
 def ascii_upcase:
   explode | map( if 97 <= . and . <= 122 then . - 32  else . end) | implode;
 
+def lpad(len;fill):
+  tostring | ($len - length) as $l | ($fill * $l)[:$l] + .;
+def rpad(len;fill):
+  tostring | ($len - length) as $l | . + ($fill * $l)[:$l];
+
 # Streaming utilities
 def truncate_stream(stream):
   . as $n | null | stream | . as $input | if (.[0]|length) > $n then setpath([0];$input[0][$n:]) else empty end;


MultisampledNight:
heyo @itchyny, could we get this looked at/share what would need to change? this is very useful and rather small in impact

src/builtin.jq
```diff
@@ -203,6 +203,11 @@ def ascii_downcase:
 def ascii_upcase:
   explode | map( if 97 <= . and . <= 122 then . - 32  else . end) | implode;
 
+def lpad(len;fill):
+  tostring | ($len - length) as $l | ($fill * $l)[:$l] + .;
+def rpad(len;fill):
+  tostring | ($len - length) as $l | . + ($fill * $l)[:$l];
```
MultisampledNight, line 209: To comply with the formatting in the rest of the file as well as (new?) syntax:

```suggestion
def lpad($len; $fill):
  tostring | ($len - length) as $l | ($fill * $l)[:$l] + .;
def rpad($len; $fill):
  tostring | ($len - length) as $l | . + ($fill * $l)[:$l];
```

itchyny:
We need clear explanation on the use case and agreement on how it handles multi-width characters.

MultisampledNight:
One use case I had right now was handling timecodes in `HH:MM:SS.FFF` format:

- For parsing, the fractional format may be e.g. just `.2` (denoting 2 tenths of a second = 200 milliseconds) or also `.222` (denoting 222 milliseconds). I `rpad` the fractional part with zeroes, then `tonumber`: https://codeberg.org/MultisampledNight/core/src/commit/5cf1ccb293a986f89e688ee7d46e6d62b868e895/path/script/calc-runtime-to-ms#L12
    - This would be possible to express somewhat similarly with `tonumber * pow(10; 3 - length)`, albeit then e.g. `1000` (which in my impl should roll over to 1s, hence cap out at ms) would end up in `100`.
    - So the true (semantic) equivalent would be: `tonumber * pow(10; [3 - length, 0] | max)`.
 - For formatting, I `lpad` each field with zeroes so even 0 milliseconds end up in the neat pattern of `00:00:00.000`: https://codeberg.org/MultisampledNight/core/src/commit/5cf1ccb293a986f89e688ee7d46e6d62b868e895/path/script/calc-ms-to-runtime#L16-L23
 - Due to the semantics of the rollover, the script for normalizing any given runtime can be [literally just a direct pipe](https://codeberg.org/MultisampledNight/core/src/commit/5cf1ccb293a986f89e688ee7d46e6d62b868e895/path/script/calc-runtime-rebalance#L5) with little effort (whereas the mathematical solution would be more wordy)

As for the multi-width characters, I'd expect behavior consistent with the rest of jq. Which isn't grapheme-aware whatsoever, only codepoint-aware, so evaluating the same length as `length` does ([which is just codepoint count](https://jqlang.org/manual/#length)).

PR #2067, opened 2020-02-24
Fedora link returns 404

calpeconsulting:
Updated with correct path

### DIFF
--- a/docs/content/download/default.yml
+++ b/docs/content/download/default.yml
@@ -24,7 +24,7 @@ body:
          `sudo apt-get install jq`.
 
        * jq 1.5 is in the official
-         [Fedora](http://pkgs.fedoraproject.org/cgit/jq.git/) repository.
+         [Fedora](https://src.fedoraproject.org/rpms/jq) repository.
          Install using `sudo dnf install jq`.
 
        * jq 1.4 is in the official [openSUSE](https://software.opensuse.org/package/jq)


docs/content/download/default.yml
```diff
@@ -24,7 +24,7 @@ body:
          `sudo apt-get install jq`.
 
        * jq 1.5 is in the official
```
calpeconsulting: It's actually at 1.6.4 now in Fedora 32

wtlangford: We haven't released any revisions to 1.6- that `-4` is referring to the version of the RPM itself. That said, this _should_ say 1.6 instead of 1.5. Mind updating it?

ravster: @wtlangford , since this seems to be a dead PR.  I've confirmed here https://src.fedoraproject.org/rpms/jq/blob/master/f/sources that Fedora is using jq v1.6

```suggestion
       * jq 1.6 is in the official
```
Can you accept the GH suggestion and we get this closed, or would you prefer I make a new PR?

calpeconsulting: > We haven't released any revisions to 1.6- that `-4` is referring to the version of the RPM itself. That said, this _should_ say 1.6 instead of 1.5. Mind updating it?

Apologies, I thought I had already done the update and you had to accept the PR?

ravster: > Apologies, I thought I had already done the update and you had to accept the PR?

@calpeconsulting The change in the PR still refers to jq 1.5.  It needs to be jq 1.6.

itchyny (approved):
LGTM, thank you. This PR aims to fix the dead link, and versions can be fixed separately (we may drop mentioning versions each package managers).

PR #2075, opened 2020-03-03
Makefile: prepend srcdir to jq.1.prebuilt to fix out of source builds

1480c1:
Seems https://github.com/stedolan/jq/commit/50a7022ea68fb37faefdcd7a35661df72fbfd655 reverted a change and broke jq out of source builds
https://github.com/m-ab-s/media-autobuild_suite/issues/1567


wtlangford:
Thanks!

PR #2079, opened 2020-03-12
Correct object construction with variables example

kennethallen:
Trunk `jq` gives the behavior described in the manual text, but the example is wrong, using the value of `$foo` as the key instead of the name of the variable.


itchyny:
This documentation error was corrected by merging #2212 (b7511b9b1f98161326a4159ebec443f05d7ec3ea), so closing the PR. Thanks anyway.

PR #2081, opened 2020-03-13
use homebrew addons

stephengroat:


stephengroat:
> Change the files

could you elaborate at all?

itchyny:
We have switched to GitHub Actions, so closing this PR, thanks.

PR #2093, opened 2020-04-02
Fix the default search paths and documentation

itchyny:
~I fixed the documentation for default search paths.~
Ref: https://github.com/jqlang/jq/blob/f88c4e5888d6d125695444d044df4bb55ad75888/src/main.c#L591-L596

EDIT: Based on the discussion, we decided to fix the implementation to comply the documentation.

Also clarify that `$ORIGIN` is replaced by the directory where the executable is located. This is the intention of `JQ_ORIGIN` in the source code.
Ref: https://github.com/jqlang/jq/blob/f88c4e5888d6d125695444d044df4bb55ad75888/src/main.c#L604

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -592,7 +592,7 @@ int main(int argc, char* argv[]) {
     // Default search path list
     lib_search_paths = JV_ARRAY(jv_string("~/.jq"),
                                 jv_string("$ORIGIN/../lib/jq"),
-                                jv_string("$ORIGIN/lib"));
+                                jv_string("$ORIGIN/../lib"));
   }
   jq_set_attr(jq, jv_string("JQ_LIBRARY_PATH"), lib_search_paths);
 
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -592,7 +592,7 @@ int main(int argc, char* argv[]) {
     // Default search path list
     lib_search_paths = JV_ARRAY(jv_string("~/.jq"),
                                 jv_string("$ORIGIN/../lib/jq"),
-                                jv_string("$ORIGIN/lib"));
+                                jv_string("$ORIGIN/../lib"));
```
wader, line 595: So i understand: let's say the jq binary is at /usr/bin/jq then `$ORIGIN` will be replace with `/usr/bin` so search path will be `/usr/bin/../lib` abs path `/usr/lib`?

wader:
Should the old documentation be updated with the buggy beaviour?

wader (approved):

pkoppstein:
@wader asked:
 > Should the old documentation be updated with the buggy beaviour?

It often happens that the documentation is interpreted as a specification, so if documentation is ever updated to describe behavior which is later viewed as buggy, it should clearly say so.

itchyny:
Thanks, I'll include this.

PR #2094, opened 2020-04-03
manual: mention 0-based indexes in array slicing

siers:


itchyny (changes requested):
I think it looks obvious as mentioned in the previous section, but I agree the proposed change makes it more clear. I suggest two things; zero-based not zero based, and needs applied to the development version manual as well (otherwise the manual of the next release will not contain the sentence).

PR #2097, opened 2020-04-10
Add minimal shell.nix to support nix-shell

vrinek:
## Rationale

`nix-shell` is essentially a fast way to get from _"I just cloned this repo"_ to _"I just built the binary"_ without having to deal with tracking down build dependencies. It has helped me in the past (eg when building https://github.com/qmk/qmk_firmware) and I hope this will help others.

The one prerequisite to get this to work is to install https://nixos.org/nix/.

## Tested on

- macOS Mojave 10.14.6
- [Pop!_OS](https://system76.com/pop) (pretty much Ubuntu) 19.10

## Tested with

```sh
nix-shell --pure # this brings up a shell with only the build dependencies available

git submodule update --init
autoreconf -fi
./configure --with-oniguruma=builtin # and with --disable-maintainer-mode
make -j8 # and with LDFLAGS=-all-static
make check
```

## Todo

- [x] Add `shell.nix`
- [ ] Wait for a green light from one of the maintainers to proceed with the following
- [ ] Mention on README
- [ ] Add to CI

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -3,6 +3,6 @@ with (import <nixpkgs> {});
 mkShell {
   name = "jq";
   buildInputs = [
-    git cacert cmake automake autoconf libtool
+    git cacert cmake automake autoconf libtool flex bison
   ];
 }
--- a/shell.nix
+++ b/shell.nix
@@ -3,6 +3,6 @@ with (import <nixpkgs> {});
 stdenv.mkDerivation {
   name = "jq";
   buildInputs = [
-    git cacert cmake automake autoconf oniguruma libtool
+    git cacert cmake automake autoconf libtool
   ];
 }
--- a/shell.nix
+++ b/shell.nix
@@ -1,7 +1,6 @@
 with (import <nixpkgs> {});
 
 mkShell {
-  name = "jq";
   buildInputs = [
     git cacert cmake automake autoconf libtool flex bison
   ];
--- a/shell.nix
+++ b/shell.nix
@@ -1,6 +1,6 @@
 with (import <nixpkgs> {});
 
-stdenv.mkDerivation {
+mkShell {
   name = "jq";
   buildInputs = [
     git cacert cmake automake autoconf libtool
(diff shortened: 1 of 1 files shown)


nprindle:
Why not use `pkgs.mkShell` instead of `pkgs.stdenv.mkDerivation`? `mkShell` is meant for `nix-shell`, and will stop you from trying to incorrectly instantiate the derivation.

Also, why not include bison/flex in the inputs?

vrinek:
@nprindle because I'm new to Nix 😅 

I'll try what you suggested and see how it goes. Thanks for the tips!

vrinek:
```diff
--- a/shell.nix
+++ b/shell.nix
@@ -1,8 +1,8 @@
 with (import <nixpkgs> {});
 
-stdenv.mkDerivation {
+mkShell {
   name = "jq";
   buildInputs = [
-    git cacert cmake automake autoconf libtool
+    git cacert cmake automake autoconf libtool flex bison
   ];
 }
```

The above change seems to work just fine. Thanks @nprindle for the suggestions.

lheckemann:
This may be superfluous, since nixpkgs has an expression for building jq — you can enter an environment appropriate for building jq with `nix-shell '<nixpkgs>' -A jq`

vrinek:
@lheckemann I gave what you suggested a try and it does not work for me. For example, it does not install `git` or the necessary `cacert` to be able to pull submodules from github.

Have you gotten it to work?

lheckemann:
Ah, I guess it's only suitable for building for a release tarball then. This `shell.nix` could still be based on nixpkgs's jq expression, but given that the jq expression itself isn't sufficient without modifications I guess this approach is just as good.

vrinek:
I'll see if I can somehow reuse the existing expression in `shell.nix`.

vrinek:
It's been a while. I'm closing this until I (or someone else) feel the need to pick up work on it again.

PR #2098, opened 2020-04-10
Adds "toboolean" builtin function for converting strings to booleans

brahmlower:
This PR adds a builtin `tobool` function which converts string values to booleans, similar to `tonumber`.

This was motivated by the need to create a boolean value based on a value provided through an environment variable. All environment variables are strings, so I needed some mechanism to convert from a string to a boolean. A contrived example of what I wanted (and what this PR implements) is:

```shell
$ TEST=true jq --null-input '{"mytest": env.TEST | tobool}'
{
  "mytest": true
}
```

This boolean conversion is already achievable using a function defined in the filter:

```shell
$ TEST=true jq --null-input 'def tobool: {"true":true, "false":false}[ . | tostring ]; {"mytest": env.TEST | tobool}'
```

The inline function definition is rather unwieldily for my use case, making the availability of the builtin function more appealing.

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -1476,6 +1476,18 @@ sections:
             input: '[1, "1"]'
             output: ['1', '1']
 
+      - title: "`tobool`"
+        body: |
+
+          The `tobool` function parses its input as a boolean. It
+          will convert correctly-formatted strings to their boolean
+          equivalent, leave booleans alone, and give an error on all other input.
+
+        examples:
+          - program: '.[] | tobool'
+            input: '["true", "false", true, false]'
+            output: ['true', 'false', 'true', 'false']
+
       - title: "`tostring`"
         body: |
 
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -1476,15 +1476,15 @@ sections:
             input: '[1, "1"]'
             output: ['1', '1']
 
-      - title: "`tobool`"
+      - title: "`toboolean`"
         body: |
 
-          The `tobool` function parses its input as a boolean. It
+          The `toboolean` function parses its input as a boolean. It
           will convert correctly-formatted strings to their boolean
           equivalent, leave booleans alone, and give an error on all other input.
 
         examples:
-          - program: '.[] | tobool'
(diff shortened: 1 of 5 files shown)


itchyny:
I like this proposal, so I rebased your PR onto the master branch and added a performance improvement commit. We have `tonumber` and why not `tobool`.

pkoppstein:
@itchyny asked:

> We have tonumber and why not tobool.

The question should be: We have tonumber and why not toboolean?

That is, jq has tonumber and tostring, not tonum and tostr.

itchyny:
Makes sense. I renamed to `toboolean`.

pkoppstein:
@brahmlower @itchyny - I think it would be worthwhile discussing whether toboolean should also map TRUE and FALSE to JSON boolean values.  That would honor the generosity that `with_entries` shows by allowing both 'value' and 'VALUE';  indeed it allows any typographical-case variant of 'value'.  

It certainly wouldn't be bizarre.  E.g. consider DuckDB:
```
D select 'TRUE'::boolean;
┌─────────────────────────┐
│ CAST('TRUE' AS BOOLEAN) │
│         boolean         │
├─────────────────────────┤
│ true                    │
└─────────────────────────┘
D select 'FALSE'::boolean;
┌──────────────────────────┐
│ CAST('FALSE' AS BOOLEAN) │
│         boolean          │
├──────────────────────────┤
│ false                    │
└──────────────────────────┘
```

Of course I recognize that there is potentially a principle at stake here: should toTYPE filters ever generalize `fromjson`?

itchyny:
I tend to follow the existing principles or consistency, so I'm in favor of accepting only JSON format, I would suggest using `ascii_downcase` if asked for supporting `TRUE` or `True`.

brahmlower:
Hey @itchyny and @pkoppstein, glad to see yall are reviving this PR! 😁 Looks like you two are taking point on getting this aligned with the project's current goals and standards, but if you need anything from me just let me know 👍

> I think it would be worthwhile discussing whether toboolean should also map TRUE and FALSE to JSON boolean values.

I really can't give any useful input here. I've completely forgotten the context for opening this PR in the first place 🤷

itchyny (approved):
I think this is good.

Issue #2115, opened 2020-05-06
Heap use-after-free in string multiplication


whentze:
## Describe the bug

When multiplying a string that is already the result of a string multiplication,
jq will access memory in `jvp_utf8_next` that was previously freed in `jv_string_append_buf`.

## To Reproduce

```
echo '"abc"' | jq '63666*.*3'
```

This invocation results in a segmentation fault on my computer.

## Expected behavior

jq should repeat "abc" many times.

### OS
- debian unstable on Linux 5.6.7 (x86_64)

### jq versions tested

- 64-bit binary version 1.6 from [github release](https://github.com/stedolan/jq/releases/tag/jq-1.6)
- debian package version 1.6-1
- build from source using git master (5b9e63e4af339bc5867603f14441b4b4cbb9e175)

## Additional context

This behavior was detected using techniques developed by the SYMBIOSYS research project at COMSYS, RWTH Aachen University. This research is supported by the European Research Council (ERC) under the EU's Horizon 2020 Research and Innovation Programme grant agreement n. 647295 (SYMBIOSYS).

itchyny:
This is already fixed in the master.

PR #2118, opened 2020-05-08
Fix multiple string multiplication

### LINKED ISSUE #2117: String multiplication produces unexpected result
**Describe the bug**
Multiple string multiplication produces unexpected result on mac osx

**To Reproduce**
`echo 0 | jq '"x" * 2 * 36'`
output: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\u0004\u0000"

**Expected behavior**
output: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"

**Environment:**
 - OS and Version: macos 10.15.4
 - jq version: 1.6

**Additional context**
Tried on centos, no such problem

itchyny:
This pull request fixes #2117. And maybe #2115 as well (I can't confirm yet).

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -322,14 +322,15 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
     }
     int n;
     size_t alen = jv_string_length_bytes(jv_copy(str));
-    jv res = str;
+    jv res = jv_copy(str);
 
     for (n = jv_number_value(num) - 1; n > 0; n--)
       res = jv_string_append_buf(res, jv_string_value(str), alen);
 
+    jv_free(str);
     jv_free(num);
     if (n < 0) {
-      jv_free(str);
+      jv_free(res);
       return jv_null();
     }
     return res;
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -320,19 +320,17 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
       str = b;
       num = a;
     }
-    int n;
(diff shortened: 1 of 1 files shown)


whentze:
I just tried the reproducing example from #2115 against this branch and it doesn't crash (and prints the correct result). Sanitizers and valgrind are looking clean too.

wtlangford:
Good find! I'll suggest a slightly more efficient solution:
```c
jv res = jv_null();
int n = jv_number_value(num);
if (n > 0) {
  size_t alen = jv_string_length_bytes(jv_copy(str));
  res = jv_string_empty(alen * n);

  for (; n > 0; n--) {
    res = jv_string_append_buf(res, jv_string_value(str), alen);
  }
}
jv_free(str);
jv_free(num);
return res;
```

itchyny:
@wtlangford I update the patch, thanks. But considering the performance here seriously, we should reduce the number of `jvp_utf8_is_valid` checks.

wtlangford:
> @wtlangford I update the patch, thanks. But considering the performance here seriously, we should reduce the number of `jvp_utf8_is_valid` checks.

Agreed, we should- but for now, I think this is good. Thanks for tracking this down!

itchyny:
I noticed that `0.5 * "abc"` is changed by this pull request. Should we keep the original behavior?

wtlangford:
Yes, we definitely should

PR #2137, opened 2020-06-06
allow MS Windows \r\n style line endings and it's (broken) "UTF-8 BOM"

smason:
have been bitten by this issue a few times, as apparently have a few others, see:

  https://stackoverflow.com/a/61116743/1358308

not sure if the parser/lexer are the right place to fix this, but it's an easy change and "works for me"!

if this is the right approach happy to add some tests as well, but the SO answer has a couple of examples of things that it's trying to fix (and seems to for me)

haven't included recompiled `.c` and `.h` files mostly because I've got a much newer version of Bison and it's quite a big change


itchyny:
Resolved by #3274. BOM will not be supported. Don't use it.

PR #2142, opened 2020-06-12
Fix string multiplication with a value between 0.0 and 1.0

itchyny:
This commit fixes a regression of 6306ac89667c (#2118). ~`jq -n '0.5 * "abc"'` should print `"abc"` not `null`.~ Based on the discussion, we change the behavior of `string * number` to emit `""` when `0 ≦ number < 1`.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -360,7 +360,8 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
       num = a;
     }
     jv res = jv_null();
-    int n = jv_number_value(num);
+    double d = jv_number_value(num);
+    int n = 0.0 < d && d < 1.0 ? 1 : d;
     if (n > 0) {
       size_t alen = jv_string_length_bytes(jv_copy(str));
       res = jv_string_empty(alen * n);
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -359,10 +359,12 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
       str = b;
       num = a;
     }
-    jv res = jv_null();
+    jv res;
     double d = jv_number_value(num);
-    int n = 0.0 < d && d < 1.0 ? 1 : d;
-    if (n > 0) {
+    if (d < 0 || isnan(d)) {
+      res = jv_null();
+    } else {
+      int n = d;
       size_t alen = jv_string_length_bytes(jv_copy(str));
       res = jv_string_empty(alen * n);
       for (; n > 0; n--) {
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1286,7 +1286,11 @@ indices(", ")
 
 [.[] * "abc"]
 [-1.0, -0.5, 0.0, 0.5, 1.0, 1.5, 3.7, 10.0]
-[null,null,null,"abc","abc","abc","abcabcabc","abcabcabcabcabcabcabcabcabcabc"]
+[null,null,"","","abc","abc","abcabcabc","abcabcabcabcabcabcabcabcabcabc"]
+
+[. * (nan,-nan)]
+"abc"
+[null,null]
 
 [.[] / ","]
 ["a, bc, def, ghij, jklmn, a,b, c,d, e,f", "a,b,c,d, e,f,g,h"]
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1284,6 +1284,10 @@ indices(", ")
 ["a", "ab", "abc"]
 ["aaa", "ababab", "abcabcabc"]
 
+[.[] * "abc"]
+[-1.0, -0.5, 0.0, 0.5, 1.0, 1.5, 3.7, 10.0]
+[null,null,null,"abc","abc","abc","abcabcabc","abcabcabcabcabcabcabcabcabcabc"]
+
 [.[] / ","]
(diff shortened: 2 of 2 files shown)


itchyny:
This is makes it behave as same as 1.6, but I personally prefer `string * number` emits `""` when `0 ≦ number < 1`.

pkoppstein:
@itchyny - There are certainly good arguments for evaluating String * 0 as "", so much so that I would be happy to open a “bug report” to that effect if you would be willing to make the change. As things are, I often define a helper function to work around the problem.  Of course, it’s easy enough to use // but it’s so easy to forget that jq somehow got this one wrong….

nicowilliams:
I'd like to remove string multiplication, and any arithmetic operations on strings other than addition.  Too much polymorphism in a dynamically typed language is a mistake ***@***.*** convinced me of this years ago).

pkoppstein:
@nicowilliams wrote:
> I'd like to remove string multiplication

Well, that particular horse is out of the barn, grazing in some
distant field in another country with which we no longer have
an extradition treaty.

Anyway, STRING * NUMBER is well-understood and quite widely adopted.(*)

The thing that is really questionable is `NUMBER | length`,
which evidently comes from mathematical notation (`|...|`) but still,
even mathematicians don't pronounce that as "length".

Since that horse too is out of the barn, I would propose that (at least pending jq 2.0):

a) `NUMBER | length` be officially deprecated.
b) `def abs:` be added

q.v. #2672 

---

(*) Courtesy of ChatGPT:

Python:
"ab" * 2

Ruby:
"ab" * 2

JavaScript:
"ab".repeat(2);

Perl:
my $result = "ab" x 2;

PHP:
$result = str_repeat("ab", 2);

Elixir:
result = String.duplicate("ab", 2)

nicowilliams:
> Well, that particular horse is out of the barn, grazing in some
> distant field in another country with which we no longer have
> an extradition treaty.

This is one reason I'd like to have builtin modules, and then we could have `jq::v1_6`, `jq::v1_7`, etc., and we could change builtins' behaviors in backwards incompatible ways.

src/builtin.c
```diff
@@ -359,7 +359,8 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
       num = a;
     }
     jv res = jv_null();
-    int n = jv_number_value(num);
+    double d = jv_number_value(num);
+    int n = 0.0 < d && d < 1.0 ? 1 : d;
```
nicowilliams: Questions while we're here:

 - if `("string" * 0) == null`, why should `("string" * 0.1) == "string"`?
 - shouldn't `("string") * 0 == ""`?  `null` seems very wrong
 - currently multiplying by a negative number, `nan`, `-nan`, `infinite`, or `-infinite` yields `null` -- why not an error, or `empty`?  (well, `empty` would be harder to arrange here, so let's not go there)

pkoppstein:
> Questions while we're here:
@nicowilliams wrote:
> shouldn't ("string") * 0 == ""? null seems very wrong

Yes!!! I think we're all (?) agreed about that. Let's do it!

> currently multiplying by a negative number, nan, -nan, infinite, or -infinite yields null -- why not an error, or empty? (well, empty would be harder to arrange here, so let's not go there)

Based on the results (i.e. jq), I think @stedolan sought to avoid raising errors, preferably by giving some meaning to otherwise meaningless expressions (this would help explain the liberal doses of polymorphism), or preferring `null` to `empty` (illustrated by `index/1`).  

In the present case, it probably doesn't make much practical difference, except for backwards compatibility.  Ergo, I'd stick with `null` for now.

But let's not make the perfect be the enemy of the good.

nicowilliams:
> Based on the results (i.e. jq), I think @stedolan sought to avoid raising errors, preferably by giving some meaning to otherwise meaningless expressions (this would help explain the liberal doses of polymorphism), or preferring null to empty (illustrated by index/1).

Well, that was before we had a way to catch errors, but sure.

itchyny:
Thanks, I fixed to emit an  empty string when repeating a string by 0 (and less than 1). The logic here can be optimized by copying the written string to grow twice and twice, but could lead to another bug so let's do later.

nicowilliams:
Thanks!

pkoppstein:
@nicowilliams wrote:

> then we could have jq::v1_6, jq::v1_7, etc., and we could change builtins' behaviors in backwards incompatible ways.

One feature (or per @itchyny, either a misfeature or a non-feature) of jq is that jq allows `::` prefixes even in the absence of modules.  E.g.:
```
def v1_6::abs: fabs;
def v1_7::abs: if . < 0 then - . else . end;
```
This seems to be somewhat relevant here, as I'll illustrate in the case of `walk`, which has a (feeble) implementation in jq 1.6, and for which the proposed improvement for jq 1.7 (see (#2611) entails a backwards-compatibility issue. 

Let's suppose that we want to make it easy for someone using jq 1.7 to invoke (or have their existing code use) the 1.6 version.  Then we could retain the 1.6 version by adding the `jq1_6::` prefix, and point out that users can:

1. use `jq1_6::walk`; and/or

2 (a). add `def walk: jq1_6::walk(f);` early enough to any code that assumes the old implementation; or
2 (b). add the two lines:
```
def walk: jq1_6::walk(f);
def jq1_7::walk(f): walk(f);
```

This last allows users to have their cake (existing code does not need to be modified) and eat it (use the new walk at the same time, too).

One important question, though, remains: Is this ability to use :: prefixes in this way a bug, a feature, or something else?

PR #2159, opened 2020-07-11
Fix build on aarch64 Darwin

saagarjha:
Apple's platforms want _REENTRANT defined before they declare the *_r functions in math.h. On x86_64 this merely caused a warning, but on aarch64 -Wimplicit-function-declaration has been bumped up to an error.


wtlangford:
This should be fixed on master now

PR #2176, opened 2020-08-21
inline example fix

mjcherbert:
if-then-else code example now works as described in the rest of the paragraph


wtlangford:
This was fixed in a different PR #2271 that got merged.

PR #2183, opened 2020-09-20
Build out of tree

bb010g:
Tests weren't happy with a setup like `mkdir -p build && cd build && ../configure && make`. `make check` now works on out-of-tree-builds.

While I was making tests work, one of them failed for me due to Valgrind output, so that's fixed (I think with virtually no runtime costs; the biggest cost would be `strncpy()` instead of `strcpy()`, with lengths we already were computing).

While I was making that test work, I noticed a function pointer cast error, so that's also received what should be a zero-cost patch (`always_inline`d function that exists solely to cast a parameter type and tail call the actual function).

Another test failed due to my normal environment exporting `JQ_COLORS`, so that's now unset before all tests that try to use it.

---
This change is [Reviewable](https://reviewable.io/reviews/stedolan/jq/2183)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,10 @@ m4/ltversion.m4
 m4/lt~obsolete.m4
 tests/*.trs
 
+# Clang junk
+.clangd
+compile_commands.json
+
 cscope.in.out
 cscope.out
 cscope.po.out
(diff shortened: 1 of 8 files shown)


bb010g:
(AppVeyor is experiencing pacman `mingw32.db` issues, unrelated to this patchset.)

itchyny:
The main issue of this PR `make check` in sub-directory works now (maybe fixed in #2700), so closing this PR. This PR contains various fixes to be separated into multiple PRs.

PR #2191, opened 2020-10-07
fix: missing function lgamma_r on macos

churib:
See bug https://github.com/stedolan/jq/issues/2190

Functions lgamma and lgammaf are not defined in math.h without
_REENTRANT symbol on macOS.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -191,7 +191,7 @@ static jv f_modf(jq_state *jq, jv input) {
   return jv_array_append(ret, jv_number(i));
 }
 #endif
-#ifdef HAVE_LGAMMA_R
+#if defined(HAVE_LGAMMA_R) && !(defined(__APPLE__) && defined(__GNUC__) && !defined(_REENTRANT))
 static jv f_lgamma_r(jq_state *jq, jv input) {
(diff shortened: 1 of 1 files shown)


raimue:
But if these functions are only declared when you set `_REENTRANT`, why did `AC_CHECK_MATH_FUNC(lgamma_r)` define `HAVE_LGAMMA_R` during configure? There seems to be some inconsistency with the check here?

danchr:
> But if these functions are only declared when you set `_REENTRANT`, why did `AC_CHECK_MATH_FUNC(lgamma_r)` define `HAVE_LGAMMA_R` during configure? There seems to be some inconsistency with the check here?

The root cause of this is in `config/m4/check-math-func.m4` which defines  `AC_CHECK_MATH_FUNC`; it only invokes `AC_CHECK_LIB()`. I posted #2196 to fix that.

wtlangford:
I ended up addressing this as part of some other autoconf cleanup in e660003abf9bdb9f9e6959d5ebe0a536862960e7 and 77417c1335a12c4ceef469caf38c0cbfb6315b45. Thanks, though!

PR #2196, opened 2020-10-21
Fix Autoconf checks on Darwin

danchr:
The Autoconf checks merely verify that the symbol is available and linkable, but not that it's actually defined and usable. This breaks compilation on recent versions of macOS, where `lgamma_r()` is unavailable unless `_REENTRANT` is defined. Since this is rather difficult to detect, I strengthened the check so that it simply omits lgamma_r on macOS.

While at it, I added support for `exp10()` on macOS, which is available, but named `__exp10` due to its nonstandard nature.


wtlangford:
I ended up resolving this as part of a larger set of autoconf improvements in 77417c1335a12c4ceef469caf38c0cbfb6315b45 and e660003abf9bdb9f9e6959d5ebe0a536862960e7, but thanks for the contribution!

dave-r12:
Hey, I also ran into this when attempting to build from the 1.6 release. Would you have a recommendation for how to resolve it for that? Also worth noting -- when I build it with `homebrew` I don't run into this issue. I'm inexperienced building things from source, so am a bit confused how `homebrew` doesn't seem to run into this for the 1.6 release. Thanks.

Issue #1912, opened 2019-05-28
Wrong timezone names in strftime/strflocaltime for DST


trantor:
**Describe the bug**
jq  does not print the timezone correctly expanding the %Z in the time format for strftime/strflocaltime when it comes to DST.
**To Reproduce**
Using the example from the wiki...
```
$ TZ=Europe/Rome ./jq-linux64 -cn 'now|strflocaltime("%Y-%m-%dT%H:%M:%S %Z")'
"2019-05-28T17:10:54 CET"

$ TZ=Europe/Rome date +"%Y-%m-%dT%H:%M:%S %Z"
2019-05-28T17:10:54 CEST
```
`date` reports, correctly, `CEST` as the timezone taking DST into account. `jq` does not.

**Expected behavior**
The output of `jq` to specify the correct timezone following DST

**Environment (please complete the following information):**
 - Linux Ubuntu 16.04
 - jq version 1.6

nicowilliams:
Huh, I think there's no real difference between `strftime` and `strflocaltime`...

Also, `strftime(3)` in glibc should be doing this for us, but evidently it's not, and the `date(1)` command is not calling `strftime(3)` either!

This looks like a bug in glibc, honestly, though we might be able to work around it.

To get at the DST TZ name we need to use `localtime_r()` and expect a `tm_zone` name from the broken-down time struct (`struct tm`), then use that in the formatting of time.  So to work around this we'll need an autoconf test for `tm_zone`, and when it's present... use either a private `strftime()` or a wrapper that replaces `%Z` with the `tm_zone`.  Ugh.

krzyk:
This issue exists also in Debian testing in 2021.

DehanLUO:
Same error on MacOS Monterey version 12.6.
This looks like a bug in [function static·jv·tm2jv(struct·tm·*tm)](https://github.com/stedolan/jq/blob/cff5336ec71b6fee396a95bb0e4bea365e0cd1e8/src/builtin.c#L1234) and [function static·int·jv2tm(jv·a,·struct·tm·*tm)](https://github.com/stedolan/jq/blob/cff5336ec71b6fee396a95bb0e4bea365e0cd1e8/src/builtin.c#L1452). The daylight saving time flag (tm_isdst), which is greater than zero if DST is in effect, is lost. 
[...]

PR #2202, opened 2020-10-29
Fix strflocaltime on daylight saving time values

wjlroe:
This PR changes the way time values get serialised and deserialised to the internal format when passing them between stages. Previously the internal value didn't retain the dst value so time values would lose that field and when printed using `strflocaltime`, the fact the value was in DST was lost. This seems to me to be the only way to have `strflocaltime` printing the correct time zone information on values during a daylight saving time.

The following behaviour demonstrates how the daylight savings value gets
lost when a time value gets turned into an internal array of numbers and
then back into a (struct tm):

        $ TZ=Europe/London ./jq -r -n '1504803635 | strflocaltime("%Y-%m-%d %H:%M:%S (%Z)")'
        2017-09-07 18:00:35 (GMT)

After this change however, this works as expected:

        $ TZ=Europe/London ./jq -r -n '1504803635 | strflocaltime("%Y-%m-%d %H:%M:%S (%Z)")'
        2017-09-07 18:00:35 (BST)

(the only difference is in the `%Z` value - "BST" instead of the incorrect "GMT").


pkoppstein:
@wtlangford - I cannot say whether this PR fixes all TZ-related bugs in jq, but I hope that you will keep in mind that the serious TZ-related bugs in jq are "silent" and as such really deserve to be prioritized, even over a release of jq 1.7.  

If you're too busy to attend to these "silent" TZ-related bugs yourself, perhaps we could recruit some volunteers to review this particular PR, but if we do so, would you be able at least to try to "pull" the PR if the reviewers attest to its being an improvement?  Thanks.

itchyny:
Issue: #1912.

DehanLUO:
Branch [439cf72](https://github.com/stedolan/jq/commit/439cf72afd6a758e3f1366fc26d5d6224213662d) works for me, thx.

trantor:
@nicowilliams was this fixed by the recent locale-related patches?

nicowilliams:
> @nicowilliams was this fixed by the recent locale-related patches?

No, sadly.

chrismo:
> > @nicowilliams was this fixed by the recent locale-related patches?
> 
> No, sadly.

I installed 1.7 yesterday and it seemed to have fixed one of my scripts that uses it? /shrug

itchyny:
The DST issue has been resolved by #3203, so closing the PR.
```sh
 $ TZ=Europe/London ./jq -r -n '1504803635 | strflocaltime("%Y-%m-%d %H:%M:%S (%Z)")'                                              
2017-09-07 18:00:35 (BST)
```

PR #2212, opened 2020-11-15
Manual error -- key should be "foo" without spaces

### LINKED ISSUE #2177: A small mistake in the development manual: bad example for object construction with variables
The development version of the manual says this under the "Object Construction: `{}`" header:

>Variable references as keys use the value of the variable as the key. Without a value then the variable's name becomes the key and its value becomes the value,
>`"f o o" as $foo | "b a r" as $bar | {$foo, $bar:$foo}`
> produces
> `{"f o o":"f o o","b a r":"f o o"}`

The result should be `{"foo": "f o o", "b a r": "f o o"}` instead, because the first key is given by the name of `$foo`, which is `"foo"`.

By the way, the expression does not compile at all with v1.6 — the `$bar` needs to be parenthesized. I'm not sure if it's OK, but maybe it is, because this feature is mentioned only in the development manual.

mattiast:
Resolves  #2177.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -616,7 +616,7 @@ sections:
 
           produces
 
-              {"f o o":"f o o","b a r":"f o o"}
+              {"foo":"f o o","b a r":"f o o"}
 
         examples:
           - program: '{user, title: .titles[]}'


itchyny (approved):
Thank you!

PR #2216, opened 2020-11-21
Add an example usage for the inputs builtin

hexagonrecursion:
It took me too long to figure out a simple trick: you can use `., inputs` to get a stream of all the inputs including the first one. I want to add this to the manual to make this more discoverable.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -2899,7 +2899,16 @@ sections:
       - title: "`inputs`"
         body: |
 
-          Outputs all remaining inputs, one by one.
+          Outputs all remaining inputs, one by one. For example:
+          
+          ```sh
+          $ echo 1 2 3 | jq '[., inputs]'
+          [
+            1,
+            2,
+            3
+          ]
+          ```
 
           This is primarily useful for reductions over a program's
           inputs.
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -140,11 +140,11 @@ sections:
         Execute the filter exactly once using `null` as the input
         instead of automatically reading every input and passing it
         through the filter. The inputs are still available to be
-        read manually using the `input` and `inputs` builtins.
+        read programmatically using the `input` and `inputs` builtins.
         
-        This is useful when using jq as a simple calculator, to
-        construct JSON data from scratch or to do recuctions over
-        the inputs (see `inputs`).
+        This option is useful, for example, when using jq as a simple
+        calculator, to construct JSON data from scratch or
+        to do reductions over the inputs (see `inputs`).
 
       * `--compact-output` / `-c`:
 
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -137,9 +137,14 @@ sections:
 
       * `--null-input`/`-n`:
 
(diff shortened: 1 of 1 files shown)


wtlangford:
This is a great addition- But I think I'd recommend a slightly different example, and possibly show off the use of the `-n` flag to pass `null` as the first input and leave the input stream untouched (this is the way I typically use `inputs`).

```sh
$ echo 1 2 3 | jq -n 'reduce inputs as $i (0; . + i)'
6
```

This shows off the purpose a bit better- it lets you deal with a sequence of JSON values that you don't want to slurp up all at once into a single array. Imagine you were processing structured log output and you had a file with a large number of (potentially also quite large) log entries you wanted to reduce over- you'd want to avoid slurping all of that up into an array for memory reasons.

Regardless, this is definitely a gap in our documentation- If you want to update your PR to use an example like the one I provided, I'd be happy to merge it!

hexagonrecursion:
> show off the use of the `-n` flag to pass `null` as the first input and leave the input stream untouched

You can do that? I did not know. The documentation for the -n option needs an improvement too.

> Imagine you were processing structured log output and you had a file with a large number of (potentially also quite large) log entries you wanted to reduce over- you'd want to avoid slurping all of that up into an array for memory reasons.

Something like that. Before I figured out the `., inputs` trick to grab all the inputs I used
```
jq 'select just the info I care about' | jq --slurp 'process it some more e.g. de-duplicate'
```
when I finally figured out the above trick I switched to
```
jq '[., inputs | select just the info I care about] | process it some more e.g. de-duplicate'
```

hexagonrecursion:
Thank you @pkoppstein I'm sorry this PR fell through the cracks and I completely forgot about it. I have incorporated your feedback.

hexagonrecursion:
CI failed for strange reasons. Close and reopen to try again.

hexagonrecursion:
I am 99.8% certain this is an issue on your end (or upstream)
```
error: mingw32: key "5F944B027F7FE2091985AA2EFA11531AA0AA7F57" is unknown
:: Import PGP key 4096R/5F944B027F7FE2091985AA2EFA11531AA0AA7F57, "Christoph Reiter (MSYS2 development key) <reiter.christoph@gmail.com>", created: 2020-06-22? [Y/n] error: mingw32: signature from "Christoph Reiter (MSYS2 development key) <reiter.christoph@gmail.com>" is unknown trust
error: failed to update mingw32 (invalid or corrupted database (PGP signature))
```
```
Error: flex: no bottle available!
```
```
Error: bison: no bottle available!
```
```
The command "brew install flex bison" failed and exited with 1 during .
```

hexagonrecursion:
@stedolan @wtlangford I have received no feedback on my last patch which I have submitted almost a month ago.

Any of the following would be preferable to complete silence:
* I an busy right now. I will be able to review your patch by `<some time in the reasonably near future>`
* Sorry, but we are not interested in fixing this
* I like the idea, but your current implementation needs more work. `<please specify what needs improvement>`

hexagonrecursion:
Close and reopen to restart CI

hexagonrecursion:
ping?

itchyny (changes requested):
This looks like a nice addition. However, in jq manual you need to use four space indentation instead of a code block for the example.

PR #2227, opened 2020-12-16
Fix error in --stream example

ajasmin:
The manual contains an error in regard to `--stream`
> `[[],"a",["b"]]` becomes `[[0],[]]`, `[[1],"a"]`, and `[[1,0],"b"]`.

This PR fix the last value to be `[[2,0],"b"]` (wrong index)

This is in fact followed by `[[2,0]]` and `[[2]]`. We should also mention this and explain why that is.


ajasmin:
already fixed by https://github.com/stedolan/jq/pull/2010

PR #2229, opened 2020-12-17
Fix extended regex pattern example

ajasmin:
`test( "a\sb", "x" )`  as it appears in the manual was doubly erroneous 
- `\s` in a string literal should be escaped as `\\s`
- The argument delimiter should be `;` instead of `,`


wtlangford:
Good catch, thanks!

PR #2240, opened 2021-01-03
Added repeat/2 and repeated/2

2x-1:
Instead of having to use `limit/2`, repeat can now trunctuate the output to a specific number. Examples:

```
repeat([1,2,3];3) ->
[
  1,
  2,
  3
]
[
  1,
  2,
  3
]
[
  1,
  2,
  3
]
```

`repeated/2` joins the output of `repeat/2` with `|add`, just like in JavaScript.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -195,6 +195,8 @@ def repeat(exp):
      def _repeat:
          exp, _repeat;
      _repeat;
+def repeat(exp,reps):
+    limit(reps ; repeat(exp))
 def inputs: try repeat(input) catch if .=="break" then empty else error end;
 # like ruby's downcase - only characters A to Z are affected
 def ascii_downcase:
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -198,7 +198,7 @@ def repeat(exp):
 def repeat(exp ; reps):
     limit(reps ; repeat(exp))
 def repeated(exp ; reps):
-    repeat(exp ; reps) | add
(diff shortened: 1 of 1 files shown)


pkoppstein:
Not sure why all the automatic checks failed, but you might want to wait for feedback from the jq maintainers before spending more time on this proposal:

1.  Using `limit` and `repeat` together for this purpose is significantly slower than the "jq-esque" alternative.  Details are below.
2. There is currently a performance issue with adding new built-in functions, and for this reason the maintainers have recently been very reluctant to do so, especially if the new defs do not add new functionality.

In my experience, the designers and maintainers are in any case generally opposed to adding new built-ins that duplicate functionality that is readily achieved using pipes.  For example, a proposal to add 

    def not(value): value | not; 

for the sake of those used to thinking of `not` in a "conventional" (prefix) way, was rejected.

## Performance Comparison
```
#!/bin/bash
N=1000
ITERATIONS=100000

function r1 {
    jq -n --argjson iterations $ITERATIONS --argjson n $N '[range($n)] as $r | range($iterations) | $r | empty'
}
echo r1
time r1

function r4 {
    jq -n --argjson iterations $ITERATIONS --argjson n $N '[range(0;$n)] as $r | limit($iterations;repeat($r)) | empty'
}
echo r4
time r4
```

Results (seconds):
```
                  u+s 
ITERATIONS   10000  100000
r1           0.051  0.078
r4           0.059  0.156
```

PR #2241, opened 2021-01-03
Added base/1 and unbase/1

ghost:
Base conversion functions get re-implemented quite frequently, thus they deserve their own built-ins.

`base/1` takes an operand `x` and converts the input to base `x`. Example usage:
```
14 | base(2)  -> [1, 1, 1, 0]
```

`unbase/1` is the opposite of `base/1`: it takes an operand `x` and converts the input from base `x`. Example usage:
```
[1, 1, 1, 0] | unbase(2) -> 14
```

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -14,8 +14,8 @@ def _assign(paths; $value): reduce path(paths) as $p (.; setpath($p; $value));
 def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
 def map_values(f): .[] |= f;
 
-def base($a): [while(. > 0; (. / $a) | floor) % $a] | reverse;
-def unbase($a): reduce .[] as $i (0; . * $a + $i);
+def tobase($a): [while(. > 0; (. / $a) | floor) % $a];
+def frombase($a): reduce (.[] | reverse) as $i (0; . * $a + $i);
 
 # recurse
 def recurse(f): def r: ., (f | r); r;
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -14,8 +14,8 @@ def _assign(paths; $value): reduce path(paths) as $p (.; setpath($p; $value));
 def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
 def map_values(f): .[] |= f;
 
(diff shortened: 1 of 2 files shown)


src/builtin.jq
```diff
@@ -14,6 +14,9 @@ def _assign(paths; $value): reduce path(paths) as $p (.; setpath($p; $value));
 def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
 def map_values(f): .[] |= f;
 
+def base(a): [while(. > 0; (. / a) | floor) | . % a] | reverse;
```
itchyny, line 17: Use `$a` instead of `a` to use the argument as a value. Also, `f | . % a == f % a`.
```bash
❯ jq -nc 'def base(a): [while(. > 0; (. / a) | floor) % a] | reverse; 14 | base(2,3,4)'
[1,1,3,1,1,1,2,0,1,1,2,1,1,3,3,2,1,1,0,1,1,1,2,1,1,1,2,1,1,0,1,2,1,1,1,1,1,1,0,0,1,1,0,1,1,1,1,0]

❯ jq -nc 'def base($a): [while(. > 0; (. / $a) | floor) % $a] | reverse; 14 | base(2,3,4)'
[1,1,1,0]
[1,1,2]
[3,2]
```

2x-1: Fixed.

odnoletkov:
In my experience it is more convenient to work with this representation of numbers – array of base digits – when order of digits is reversed. So the lowest order digit is the first and highest is the last in the array:

    14 | base(2)  -> [0, 1, 1, 1]

That way when implementing operations with multiple numbers same order digits match naturally, for example:

    def bitwiseand($a; $b): [$a, $b] | transpose | map(if all(. > 0) then 1 else 0 end);

---

For the names what do you think about `tobase/1` and `frombase/1`? This follows convention set by other APIs in the standard library.

---

Lastly, we need to document these in the manual.

itchyny:
> Lastly, we need to document these in the manual.

And tests, too.

2x-1:
@odnoletkov Sorry for the inactivity. These are all good suggestions, I'll be trying to fullfill them.

2x-1:
@itchyny Not sure where the tests are though - I'm relatively new to jq's source code.

tst2005:
I think there is a bug in frombase.

```diff
-def frombase($a): reduce (.[] | reverse) as $i (0; . * $a + $i);
+def frombase($a): reduce (reverse | .[]) as $i (0; . * $a + $i);
```

PR #2255, opened 2021-01-26
Added a first fuzzer for integration with OSS-Fuzz.

DavidKorczynski:
Hi Maintainers,

I would like to set up continuous fuzzing of jq by way of OSS-Fuzz. Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

In this PR https://github.com/google/oss-fuzz/pull/4980 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate jq. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

### DIFF
--- a/tests/jq_fuzz_parse.c
+++ b/tests/jq_fuzz_parse.c
@@ -1,0 +1,21 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "jv.h"
+
+int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
+  // Creat null-terminated string
+  char *null_terminated = (char *)malloc(size + 1);
+  memcpy(null_terminated, (char *)data, size);
+  null_terminated[size] = '\0';
+
+  // Fuzzer entrypoint
+  jv res = jv_parse(null_terminated);
+  jv_free(res);
+
+  // Free the null-terminated string
+  free(null_terminated);
+
+  return 0;
+}


nicowilliams:
~How does this work?  Why wouldn't we just use AFL?~  EDIT: Because Google will run the fuzzer for us.

nicowilliams:
Thanks!  I'll rebase this and play with it.

DavidKorczynski:
> ~How does this work? Why wouldn't we just use AFL?~ EDIT: Because Google will run the fuzzer for us.

Yeah, Google will run the fuzzers by way of OSS-Fuzz and this includes both AFL++, libFuzzer, Honggfuzz and Centipede. This will be done continuously.

nicowilliams:
> Yeah, Google will run the fuzzers by way of OSS-Fuzz and this includes both AFL++, libFuzzer, Honggfuzz and Centipede. This will be done continuously.

Yeah, I'm very keen on adding support for all the fuzzing libraries that only need us to provide the same function -- it will be very easy to support all of them.

DavidKorczynski:
> > Yeah, Google will run the fuzzers by way of OSS-Fuzz and this includes both AFL++, libFuzzer, Honggfuzz and Centipede. This will be done continuously.
> 
> Yeah, I'm very keen on adding support for all the fuzzing libraries that only need us to provide the same function -- it will be very easy to support all of them.

Great! If you'd like I can fix this PR up again (am not sure what has changed since it was made, on the JQ side) and ensure things are working on the OSS-Fuzz side? Once that is done we just merge in the pull requests and you will shortly after have code coverage reports available as well as emails with any potential bugs found.

nicowilliams:
> Great! If you'd like I can fix this PR up again (am not sure what has changed since it was made, on the JQ side) and ensure things are working on the OSS-Fuzz side? Once that is done we just merge in the pull requests and you will shortly after have code coverage reports available as well as emails with any potential bugs found.

That would be fantastic!  Thanks for your help!

DavidKorczynski:
@nicowilliams this is ready now.

The corresponding PR is https://github.com/google/oss-fuzz/pull/4980

In order to complete the OSS-Fuzz integration:
1) merge this PR
2) I will adjust https://github.com/google/oss-fuzz/pull/4980 to use fuzzers from here upstream, rather than the one provided from that PR
3) The PR on OSS-Fuzz will be merged
4) OSS-Fuzz will start running the fuzzer using the engines libFuzzer, AFL++, Honggfuzz, Centipede

Could you share an email address that I can attach to the OSS-Fuzz integration? This is used to e.g. receive bug reports. It shuold be affiliated with a Google account due to the authenticate measures used. This email will receive bug reports and also be able to log on to oss-fuzz.com where there will be detailed bug reports, stack traces, reproducer inputs etc.

nicowilliams:
~No changes needed to `Makefile.am`?~  EDIT: No, the `OSS-Fuzz` PR builds this.

tests/jq_fuzz_parse.c
```diff
@@ -0,0 +1,21 @@
...
+#include "jv.h"
+
+int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
+  // Creat null-terminated string
+  char *null_terminated = (char *)malloc(size + 1);
+  memcpy(null_terminated, (char *)data, size);
+  null_terminated[size] = '\0';
+
+  // Fuzzer entrypoint
+  jv res = jv_parse(null_terminated);
```
nicowilliams, line 14: BTW, we do have a `jv_parse_sized()` so you don't have to add a `NUL` terminator.

nicowilliams:
I've sent you an email, so now you'll have my email.

nicowilliams:
I sent three gmail-related emails: for me, for @emanuele6, and for @pkoppstein.  Other maintainers/owners will get forwarded copies from us.

Issue #2190, opened 2020-10-07
build on macOS


churib:
Hi,

the ruby gem 'ruby-jq' uses jq as external lib to provide jq syntax in ruby (which is quite cool).

Unfortunately jq does not build on newer macOSes when the normal `autreconf && configure.sh && make`-process with gcc is used like on all other platforms and which is also used by ruby-jq:
```
src/builtin.c:200:31: error: implicit declaration of function 'lgamma_r' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  jv ret = JV_ARRAY(jv_number(lgamma_r(jv_number_value(input), &sign)));
```
The gamma functions are not thread-safe and are not defined without the `_REENTRANT`-makro since [version 315 of math.h](https://opensource.apple.com/source/Libm/Libm-315/Source/Intel/math.h.auto.html) (CTRL-F gamma) on Apple's plattforms.

bobnautic:
Same Problem here

wtlangford:
I have a patch in progress that will fix this in the next release (I see ruby-jq uses the 1.6 release tarball, though). You'll want to reach out to the developer of ruby-jq and have them allow you to pass through some cflags. The flag in particular you'll want to add is `-pthread`.

sudeep-ib:
Same issue here on macos bigsur

churib:
@wtlangford can you foresee when the next release including the patch will come?

itchyny:
I think #2196 is acceptable solution for this issue.

wtlangford:
The current master now has fixes for these particular build issues, and hopefully I'll be able to release a new version soon.

PR #2265, opened 2021-02-06
MacOSX fix for lgamma_r function.

jacksonps4:
The lgamma_r needs to be requested explicitly on OS X.

From the man page:
> In order to use the lgamma_r() function, define the macro _REENTRANT before including
<math.h>
This fixes the build on OS X, which is particularly useful for those wanting to build
a version to run natively on Apple Silicon.


itchyny:
Please refer to the discussion of #2190, #2191, and #2196.

jacksonps4:
Ah - thanks.

PR #2292, opened 2021-03-28
Disable some tests when building without regex support

KAction:


itchyny:
I understand the tests fail without regex support, but I'm against skipping the manual test entirely. This is written in README.md as `note that jq's tests require regexp support to pass`. Closing as unacceptable patch, but if you have any other solution, please comment below.

KAction:
Just to be sure I understand correctly, with my distribution maintainer hat on -- if I am building  `jq` version without regex support, I am not supposed to run tests, correct?

itchyny:
Depends on what you want to achieve. I found Homebrew recipe checks a minimal test case that the executable can be invoked ([ref](https://github.com/Homebrew/homebrew-core/blob/1d647932d0c46f64891d6494a6191791d8e51326/Formula/jq.rb#L57)). This allows end users to install the command even when some bugs or flakiness were introduced to bundled test cases. But if you still want to run the entire tests, we need a better solution than skipping man.test entirely. Maybe splitting the man.test into two, or improving `--run-tests` to skip regex tests if not available.

KAction:
Then I am lost. This pull request is supposed doing exactly that -- improving tests so regex tests are skipped when regex is not built.

itchyny:
I created a PR #2644 to the solution `improving --run-tests to skip regex tests if not available`.

PR #2293, opened 2021-03-30
with_entries and from_entries improvement to support array

tst2005:


gcr:
could you elaborate more on the use case for this? as written, this appears to allow one to omit elements of an array by setting `.key` to null, but one can already do that with `select`. Another question is that I'd expect to be able to change the key to reorder the elements of the array, but I don't think this patch supports that.

One thing i like to do is to use `with_entries` to quickly convert from an array with a key field to an object, like this: `with_entries(.key=.value.key)` takes `[{"key": "abc", ...}, {"key": "def", ...}]` to `{"abc": {"key": "abc", ...}, "def": {"key": "def", ...}}` One way you could support this use case is by dispatching on the type of the `.key` element after the transformation and not before. Then, object results could be `|add`ed together and arrays could be recollected into their proper order, all without any compatibility breakage

gcr:
(i'm not any official maintainer of the `jq` team, just adding my 2¢ as an interested user)

tst2005:
Hello @gcr

Did you read https://github.com/stedolan/jq/issues/2294 ? there more information about the cases I tried to support.
I also just made an comment to explain my new approach.

Do not hesitate to create an issue to discuss about your idea/needs, I will close this obsolete PR.

PR #2311, opened 2021-05-10
Replace bareword error

jsoref:


jsoref:
@wtlangford is there something I need to do to move this along?

PR #2320, opened 2021-05-26
Move the IRC channel to Libera.Chat

ncfavier:
As you might be aware, a lot of open source communities have begun migrating away from freenode to [Libera.Chat](https://libera.chat/) (same staff, different administrative foundation) after recent events. In particular, the most active helpers in `#jq` (`nf` (me), `geirha` (@geirha) and `earnestly` (@Earnestly)) have moved out.

This PR updates the documentation to point at the new network.

P.S. @wtlangford, @nicowilliams, ping me on Libera if you'd like founder access to `#jq`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ about: Create a report to help us improve
 <!--
 READ THIS FIRST!
 
-If you have a usage question, please ask us on either Stack Overflow (https://stackoverflow.com/questions/tagged/jq) or in the #jq channel (http://irc.lc/freenode/%23jq/) on Freenode (https://webchat.freenode.net/).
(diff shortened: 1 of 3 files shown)


dngray:
I'd like to see this move to libera. I recently tried using Freenode and it's basically unusable now since the new administration took over.

I couldn't login because my IP was apparently restricted, couldn't login via Tor even when connecting with SASL and creating an account via their SSO domain. I couldn't use the web client either as apparently my IP of my VPN and ISP were both on dronebl. On top of that email cloaking services seemed to be blocked as well.

In the end I just ended up giving up. Please move to a sane IRC network.

emanuele6:
The channel is already on libera.chat.

nicowilliams:
Thanks! And thanks @owenthereal for curating a list of PRs for me to merge.

PR #2324, opened 2021-06-09
docs: Document repeat(exp)

wader:


docs/content/manual/manual.yml
```diff
@@ -1664,6 +1664,22 @@ sections:
...
+          The `repeat(exp)` function allows you to repeatedly
+          apply expression `exp` to `.` until an error is raised.
+
+          Note that `repeat(exp)` is internally defined as a
+          recursive jq function.  Recursive calls within `repeat` will
+          not consume additional memory if `exp` produces at most one
+          output for each input.  See advanced topics below.
+
+        examples:
+          - program: '[repeat(.*2, error)?]'
```
wader, line 1679: Suggestion of better example is welcome

nicowilliams:
Thanks!

pkoppstein:
> Suggestion of better example is welcome

Here's an example that's at least slightly interesting as it will stop (without producing an error) on non-JSON input:

    jq -n 'repeat(input)? | length'

Another reasonable though not very inspired example would be: `[limit(100; repeat(0)]`

PR #2350, opened 2021-09-18
Document the potential security issues when using `-r`/`-0`.

pabs3:
Also document how `-r` does separators.

Reported-by: @pcworld
Reported-in: https://github.com/stedolan/jq/issues/1271#issuecomment-920835007

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -196,6 +196,11 @@ sections:
         making jq filters talk to non-JSON-based systems. The output
         values will be separated by a newline (`\n` aka LF) character.
 
+        Please note a **potential security issue** when using this option.
+        Please note that if the selected data of the input JSON contains
+        newline characters then processing of jq output will incorrectly
+        split a single item containing an newline character into two items.
+
       * `--join-output` / `-j`:
 
         Like `-r` but jq won't print a newline after each output.
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -193,7 +193,8 @@ sections:
         With this option, if the filter's result is a string then it
         will be written directly to standard output rather than being
         formatted as a JSON string with quotes. This can be useful for
-        making jq filters talk to non-JSON-based systems.
+        making jq filters talk to non-JSON-based systems. The output
+        values will be separated by a newline (`\n` aka LF) character.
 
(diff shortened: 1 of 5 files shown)


pabs3:
Hmm, the AppVeyor failure seems unrelated to the changes in this PR.

docs/content/manual/manual.yml
```diff
@@ -193,7 +193,13 @@ sections:
...
         will be written directly to standard output rather than being
         formatted as a JSON string with quotes. This can be useful for
-        making jq filters talk to non-JSON-based systems.
+        making jq filters talk to non-JSON-based systems. The output
+        values will be separated by a newline (`\n` aka LF) character.
+
+        Please note a **potential security issue** when using this option.
+        Please note that if the selected data of the input JSON contains
+        newline characters then processing of jq output will incorrectly
+        split a single item containing an newline character into two items.
```
vdukhovni, line 202: Frankly, I'm not convinced this is a *security issue*, for example, when CSV output is generated with `@csv`, multi-line output works correctly, because CSV handles multi-line strings correctly.  Similar observations can be made about @html, etc.

Raw strings only have as much structure as is ensured in their construction, and there's nothing `jq`-specific here.  If one wants structured output that won't confuse some downstream parser, one has ensure the right encoding is used, which may or may be JSON.

Thus, if one wants to use `jq` as an `awk` or `sed` replacement working with line-oriented data, one can do that:
```
$ printf -- 'Do it with %s!\n' "sed" |
      jq -Rr 'sub("\\bsed\\b"; "jq"; "g")'
Do it with jq

$ cat /etc/passwd |  # Averaging uids from /etc/passwd is of course nonsense!
     jq -Rrn 'reduce (inputs | [limit(3;splits(":"))] | .[2] | tonumber) as $n ([0, 0]; [.[0]+$n, .[1] + 1])
              | "The average uid is: \(.[0]*1000/.[1] | rint | ./1000)"'
The average uid is: 1715.038
```

So what we have here is that when the output is not JSON it needs to be appropriately encoded for the syntax expected by the consumer. I don't think this really deserves any sort of security warning.

Even with JSON output, if the elements encoded as JSON are sloppily constructed, they may already mediate some sort of "injection" attack.

pabs3:
It is clear that this particular PR isn't acceptable, so closing.

PR #2352, opened 2021-09-29
Cppcheck fix: Missing "va_end" call. 

llansky3:
This was found by running the cppcheck static analysis where it shows as error.

There should be "va_end" call for every "va_start" according to  http://en.cppreference.com/w/c/variadic/va_end.


owenthereal:
LGTM

nicowilliams:
Thanks! And thanks @owenthereal for curating a list of PRs for me to merge.

PR #2357, opened 2021-10-12
Validate module metadata to be an object

itchyny:
Module metadata should be an object so we should validate module directive just as import directive.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -514,6 +514,7 @@ jv block_list_funcs(block body, int omit_underscores) {
 }
 
 block gen_module(block metadata) {
+  assert(block_is_const(metadata) && block_const_kind(metadata) == JV_KIND_OBJECT);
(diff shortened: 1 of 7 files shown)


owenthereal:
LGTM

nicowilliams:
Ah, can you rebase and re-gen the bison outputs?

itchyny:
Do I need to use older version of Bison?

nicowilliams:
> Do I need to use older version of Bison?

No.  Any version will do provided it works.

itchyny:
Thank you. I rebased and confirmed `make check` passes on Ubuntu environment.

itchyny:
Sorry, I rebased again to generate by the latest Bison (3.8.2).

wader (approved):

itchyny:
Thank you.

PR #2360, opened 2021-10-21
No quadratic modify

nicowilliams:
@vdukhovni please review.

@pkoppstein I bet you'll enjoy this.

PLEASE DO NOT REMIND ME WE NEED A RELEASE.  I have very little time for this.

What I need to make a release: jq to pass tests on Windows.


src/parser.c
```diff
@@ -1,9 +1,8 @@
-/* A Bison parser, made by GNU Bison 3.3.2.  */
+/* A Bison parser, made by GNU Bison 3.0.4.  */
```
vdukhovni, line 1: Did you mean to check in the generated code?  Do some builds start with parser.c, rather than parser.y?

vdukhovni (changes requested):
Looks good with nits.

Not sure whether you meant to check in new parser.c/parser.h, but given that they're under revision control, perhaps so...

src/builtin.jq
```diff
@@ -11,7 +11,13 @@ def min_by(f): _min_by_impl(map([f]));
...
 def del(f): delpaths([path(f)]);
 def _assign(paths; $value): reduce path(paths) as $p (.; setpath($p; $value));
-def _modify(paths; update): reduce path(paths) as $p (.; label $out | (setpath($p; getpath($p) | update) | ., break $out), delpaths([$p]));
+def _modify(paths; update):
+    reduce path(paths) as $p (.;
+        getpath($p) as $v
+      | setpath($p; null)
+      | [($$$$v) | update] as $v2
+      | if ($v2|length)==0 then delpaths([$p]) else setpath($p; ($$$$v2)[0]) end
+    );
```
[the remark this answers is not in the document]
nicowilliams: So, whether this particular commit is needed is unclear.  It does mean not holding on to a reference to the `.` of the `update`, which is good if you have lots of large values to pass in to `update`, but it seems unlikely, but then, minimizing reference taking is a good thing.  I'll test a version of this w/o the additional allocation (which probably slows things down a fair bit).

src/execute.c
```diff
@@ -562,6 +562,7 @@ jv jq_next(jq_state *jq) {
       }
       jv_free(stack_popn(jq));
       stack_push(jq, *var);
+      var = frame_local_var(jq, v, level);
```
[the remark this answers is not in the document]
nicowilliams: Yes, I'll write a comment for that.

src/parser.c
nicowilliams: Yes, I did.  We commit Bison outputs for users who build from source on systems that lack a recent Bison.  I should not change the Bison version though...

vdukhovni (approved):
Hooray.  Looks great. Thanks!

src/parser.y
```diff
@@ -749,6 +749,25 @@ FORMAT {
...
+ * This is used in `_modify`, in src/builtin.jq, to avoid holding on to a
+ * reference to `.`.
+ *
+ * We could just have the compiler emit bytecode for `_modify` so it can use
+ * LOADVN w/o needing jq syntax for LOADVN.
+ *
+ * This syntax, `$$$$varname`, violates referential transparency: it has
+ * side-effects that are surprising.
+ *
+ * DO NOT USE!!  I will break your jq code if you do use this outside
```
nicowilliams, line 764: Hmmm, we should confirm that this does not introduce a security vulnerability when running other people's jq code.

nicowilliams: Ah, this does not introduce a security vulnerability, duh.  I hadn't paused to swap this all in when I wrote the above.

nicowilliams: Now I'm wondering just how dangerous this is at all.  It doesn't seem that crazy to allow out-of-tree jq code to use this...

vdukhovni: Can it result in dangling references if retained in an unexpected/unsafe way?

nicowilliams: No, what `$$$$var` does is a) produce `$a`'s value, b) cause future references to `$s` (or `$$$$a`) to produce `null` instead of whatever the original value was.  This feels very non-functional / impure, but it doesn't break the VM, leak values, nor cause use-after-free errors.

nicowilliams: > This feels very non-functional / impure

This is essentially an assignment operator for `$binding`s, which are supposed to be constant.  It's not a generic assignment because all you get to assign is `null` (which is an immediate value in C, so there's no heap object involved).

nicowilliams: And `LOADVN` long pre-existed, and it does the right thing regarding reference counts.

vdukhovni: You're right, nothing bad happens:
```
$ ./jq -Mcn '42 as $answer | $answer as $foo | [$foo] | . + [$foo] | . + [$answer]'
[42,42,42]
$ ./jq -Mcn '42 as $answer | $$$$answer as $foo | [$foo] | . + [$foo] | . + [$answer]'
[42,42,null]
```

nicowilliams: And with `valgrind`:

```
valgrind ./jq -Mcn '"42" as $answer | $$$$answer as $foo | [$foo] | . + [$foo] | . + [$answer]'
==3408529== Memcheck, a memory error detector
==3408529== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3408529== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==3408529== Command: ./jq -Mcn "42"\ as\ $answer\ |\ $$$$answer\ as\ $foo\ |\ [$foo]\ |\ .\ +\ [$foo]\ |\ .\ +\ [$answer]
==3408529==
["42","42",null]
==3408529==
==3408529== HEAP SUMMARY:
==3408529==     in use at exit: 0 bytes in 0 blocks
==3408529==   total heap usage: 7,263 allocs, 7,263 frees, 1,005,890 bytes allocated
==3408529==
==3408529== All heap blocks were freed -- no leaks are possible
==3408529==
==3408529== For lists of detected and suppressed errors, rerun with: -s
==3408529== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

Issue #2374, opened 2021-11-23
Reduce over raw files - EOF doesn't terminate `inputs`


eddyashton:
This looks like a bug, but perhaps I'm just missing something obvious. I'm trying to combine multiple raw (non-JSON) files and embed them in a JSON object. These files are ultimately found by a glob, so I don't think I can use `--rawfile` without extensive bash machinery on top; instead I need to reduce them with `-R` to read each line, and recombine them based on `input_filename`.

```
$ jq -Rn 'reduce inputs as $line ({}; .[input_filename] += [$line])' foo.txt bar.txt baz.txt
```

This almost works, but the last line of each file is inserted into the wrong list:
```
{
  "foo.txt": [
    "First line of foo.",
    "I'm a text file with multiple lines!"
  ],
  "bar.txt": [
    "Last line of foo.First line of bar.",
    "I'm also a text file!"
  ],
  "baz.txt": [
    "Last line of bar.First line of baz.",
    "I'm a third text file.",
    "Last line of baz."
  ]
}
```

It took me a while to work out, but this is because there's no bare newline at the end of my files. If I add that to the test files, it works correctly, but I can't guarantee that the real files will be terminated by an empty line. Surely `inputs` should 'split' at EOF, as well as each newline? It clearly does for the final file, since we get a final entry for the last line there, so why does it combine the entries from earlier files, across EOF marks?

Just to confirm, this also happens with direct invocations of `input`:
```
$ jq -Rn '[input] + [input] + [input] + [input] + [input] + [input] + [input]' foo.txt bar.txt baz.txt
[
  "Last line of baz.",
  "I'm a third text file.",
  "Last line of bar.First line of baz.",
  "I'm also a text file!",
  "Last line of foo.First line of bar.",
  "I'm a text file with multiple lines!",
  "First line of foo."
]
```

eddyashton:
This is the behaviour on both 1.5 and 1.6, afaict.

eddyashton:
Another data point - this weirdness on the last line of a file also affects `input_line_number`, which is off-by-one on the final line of the final file:

```
$ jq -Rn 'reduce inputs as $line ({}; .[input_filename] += [input_line_number, $line])' foo.txt bar.txt baz.txt
{
  "foo.txt": [
    1,
    "First line of foo.",
    2,
    "I'm a text file with multiple lines!"
  ],
  "bar.txt": [
    1,
    "Last line of foo.First line of bar.",
    2,
    "I'm also a text file!"
[...]

PR #2375, opened 2021-11-23
Fix for #2374

eddyashton:
See #2374 for explanation of the issue.

I've poked around the code and found this possible fix. It's not very neat (checking `state->current_input` and modifying `state->current_line` outside of `jq_util_input_read_more`), but it works and doesn't break any other tests. Please feel free to propose a neater fix if you can see one.

I've also added a regression test. This uses a bunch of functions, so perhaps shouldn't be in `shtest`? But I don't see `input_line_number` tested anywhere else.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -411,6 +411,11 @@ jv jq_util_input_next_input(jq_util_input_state *state) {
       is_last = jq_util_input_read_more(state);
       if (state->buf_valid_len == 0)
         continue;
+      if (state->current_input && feof(state->current_input))
+      {
+        state->current_line++;
+        is_last = 1;
+      }
       if (jv_is_valid(state->slurped)) {
         // Slurped raw input
         state->slurped = jv_string_concat(state->slurped, jv_string_sized(state->buf, state->buf_valid_len));
--- a/tests/shtest
+++ b/tests/shtest
@@ -150,6 +150,12 @@ cmp $d/out $d/expected
 printf "[1,2][3,4]\n" | $JQ -cs add > $d/out 2>&1
 cmp $d/out $d/expected
 
+# Regression test for #2374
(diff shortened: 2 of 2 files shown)


xguo-prestolabs (approved):

owenthereal:
LGTM

nicowilliams:
This needs to be a command-line option.  As @pkoppstein explained in #2374, this would be an incompatible change.

nicowilliams:
You can probably use `input_filename` in a reduction (`reduce` or `foreach`) over `inputs` to detect when jq has moved from reading from one file to the next.

eddyashton:
As @pkoppstein explained in #2374, this change breaks backwards compatibility. Closing for now, will revisit as an additional commandline if I get time.

emanuele6:
I don't think it matters that fixing this bug "breaks backward compatibility".
The current behaviour is not good, and lets `input_filename/0` and `input_line_number/0` return values that are wrong.

I think this fix should be merged.

The fix looks good, but there a merge conflict in shtest.

emanuele6:
Actually, this patch does not fix the issue fully correctly.

If you use `--slurp --raw-input`, it will stop slurping after the end of the first non-empty file that does not have a final newline, and start reading the rest of the inputs line by line (though still not merging content of different files into one string):
```bash
$ printf foo\\n > foo; printf bar > bar; printf baz1\\nbaz2\\n > baz; printf xyz > xyz; printf hello > hello
$ jq -sR . foo bar baz xyz hello
"foo\nbar"
"baz1"
"baz2"
"xyz"
"hello"
```

emanuele6:
And this problem also occurs for files in non-raw mode:

```bash
$ jq . <(printf 4) <(echo 5)
45
```

nicowilliams:
Let's fix this in a separate PR.

Issue #2318, opened 2021-05-25
travis-ci.org is shutting down


martinvonwittich:
I clicked on the [![Build Status](https://travis-ci.org/stedolan/jq.svg?branch=master)](https://travis-ci.org/stedolan/jq) button in the readme. The site says:

>  Please be aware travis-ci.org will be shutting down by end of May 2021. Please consider migrating to travis-ci.com.

More details can be found here: https://blog.travis-ci.com/2021-05-07-orgshutdown

itchyny:
We have switched to GitHub Actions in #2596. But still, icon and CI configuration remain in the repository.

PR #2376, opened 2021-11-24
Fix AppVeyor

eddyashton:
Noticed on my previous PR that the CI's red, so thought I'd try to fix it. The AppVeyor setup is trying to pull a version of bison that's no longer available on the msys2 repo (http://repo.msys2.org/msys/x86_64/ no longer contains `bison-3.5.4-1`).

Note also that Travis is no longer running Linux builds because of #2318.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,10 +17,10 @@ install:
   - C:\msys64\usr\bin\bash -lc "pacman-key --verify msys2-keyring-r21.b39fb11-1-any.pkg.tar.xz.sig"
   - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm msys2-keyring-r21.b39fb11-1-any.pkg.tar.xz"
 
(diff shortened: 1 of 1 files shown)


eddyashton:
Abandoning this for now as I don't have time to keep hunting for potential fixes. Findings if anyone else looks at this:

- http://repo.msys2.org no longer hosts `bison-3.5.4`, so this dependency must be upgraded or found elsewhere
- All versions of bison are now packaged as `.zst`, which the installed version of msys2 can't read
- Trying to update msys2, or install a `.zst`  reader, or even `pacman -Su` is tricky, because all of the recent packages are `.zst`s!
- Updating the AppVeyor image to `Visual Studio 2019` gets past this (can't tell what the default/original value was?), but then hits some pip version because the docs' `Pipfile.lock` demands versions that it can't fulfil

PR #2387, opened 2021-12-22
fixes incorrect example in man page

gwkunze:
Might also want to add the "closing" paths at the end (`[[2,0]]` and `[[2]]`)


owenthereal:
LGTM

PR #2391, opened 2021-12-28
Add fixes for manual

bewuethr:
This adds a few minor fixes for the manual:

- Use inline code markup instead single or double quotes where appropriate in the modules, update-assignment, and plain assignments sections
- Add examples from plain assignment text as actual examples, removing one of the examples entirely from the text itself as it was adding no additional information beyond what can be gleaned from the example
- Use indented code blocks instead of normal text for code examples in text


itchyny (approved):
LGTM, thank you!

Issue #2394, opened 2022-01-06
`inputs` skips the first file specified


stanio:
I have a number (few hundred) of `.json` files which content is array (possibly empty) of objects, and I'm using the following command to merge them:

    jq '[inputs | .[]]' *.json > merged.json

I've discovered the content of one of the files is missing from the result.

**To Reproduce**

Have three JSON files in a directory:

    foo.json
    bar.json
    baz.json

Issue a command:

    jq '[inputs | input_filename]' foo.json bar.json baz.json

The result I'm getting:

    [
      "bar.json",
      "baz.json"
    ]

**Expected behavior**

A result of:

    [
      "foo.json",
      "bar.json",
      "baz.json"
    ]

**Environment (please complete the following information):**

 - Windows 10 64-bit (21H1 build 19043.1415)
 - jq version 1.6

**Additional context**

I'm running in a Git-bash console:

```
$ jq '[inputs | input_filename]' foo.json
[]
```

```
$ jq '[inputs | input_filename]' foo.json foo.json bar.json baz.json
[
  "foo.json",
  "bar.json",
  "baz.json"
]
```

```
$ jq '[inputs | input_filename]' bar.json foo.json bar.json baz.json
[
  "foo.json",
  "bar.json",
  "baz.json"
]
```

`input` also appears to act weird:

```
$ jq '[input | input_filename]' foo.json
jq: error (at <unknown>): break
```

```
$ jq '[input | input_filename]' foo.json bar.json
[
  "bar.json"
]
```

```
$ jq '[input | input_filename]' foo.json bar.json baz.json
[
  "bar.json"
]
jq: error (at <unknown>): break
```

```
$ jq '[input | input_filename]' bar.json foo.json bar.json baz.json
[
  "foo.json"
]
[
  "baz.json"
]
```

(I'm not really sure what's the expected outcome with the last one.)

---

https://stedolan.github.io/jq/manual/v1.6/#IO

wader:
Hey, try to add `-n`, otherwise jq will implicitly read one input.
```sh
$ jq ".,inputs" <(echo 1) <(echo 2)
1
2
$ jq -n "inputs" <(echo 1) <(echo 2)
1
2
```
Might feel be a bit unintuitive but i think it makes sense.

stanio:
Thank you very much, @wader!  Yeah, it's pretty unintuitive to me – not a jq-guru, and it gets even more confusing after reading the doc:

> - `--null-input`/`-n`:
>
[...]

PR #2395, opened 2022-01-07
Mention -n in IO-section and for input/inputs

wader:
Related to #2394

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -2924,7 +2924,9 @@ sections:
       are provided for this, `input` and `inputs`, that read from the
       same sources (e.g., `stdin`, files named on the command-line) as
       jq itself.  These two builtins, and jq's own reading actions, can
-      be interleaved with each other.
+      be interleaved with each other.  They are commonly used in combination
+      with the null input option `-n` to prevent one input from being read
+      implicitly.
 
       Two builtins provide minimal output capabilities, `debug`, and
       `stderr`.  (Recall that a jq program's output values are always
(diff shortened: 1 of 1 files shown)


docs/content/manual/manual.yml
```diff
@@ -2944,13 +2946,15 @@ sections:
...
+          Commonly used in combination with the `-n` option.
+
       - title: "`inputs`"
         body: |
 
           Outputs all remaining inputs, one by one.
 
           This is primarily useful for reductions over a program's
-          inputs.
+          inputs.  Commonly used in combination with the `-n` option.
```
iainelder, line 2957: I think this should be worded more strongly. If you omit `-n`, you lose the first item of the stream. That almost definitely produces incorrect results.

Perhaps you could quote @pkoppstein 's excellent [A stream oriented introduction to jq](https://github.com/pkoppstein/jq/wiki/A-Stream-oriented-Introduction-to-jq)?

> Note that when using inputs in this way, it will generally be necessary to invoke jq with the -n command-line option, otherwise the first entity will be lost.

iainelder:
@wader, did you get any more feedback on this from the maintainers?

The CI is failing for reasons unrelated to your change.


https://ci.appveyor.com/project/stedolan/jq/builds/42116887

```text
C:\msys64\usr\bin\bash -lc "pacman-key --verify bison-3.5.4-1-x86_64.pkg.tar.xz.sig"
==> Checking bison-3.5.4-1-x86_64.pkg.tar.xz.sig...
==> ERROR: The signature identified by bison-3.5.4-1-x86_64.pkg.tar.xz.sig could not be verified.
Command exited with code 1
cat config.log
```

docs/content/manual/manual.yml
wader: Good idea, reworded a bit

wader:
@iainelder Nope haven't heard anything

iainelder:
It seems we're not alone with the CI failure. It was reported at least 16 months ago in #2250. I don't see any response from the maintainers there either.

PR #2398, opened 2022-01-17
Clarify use of --argjson option

ldnicolasmay:


itchyny:
I think the original document is much clear in that the variable is available with the name given to this option.

PR #2400, opened 2022-01-20
Fix iterration problem for non decimal string

thalman:
When the string transformation to number failed, all following
transformation failed too.

This happened because the status in decNumberFromString function is
updated just in case of error. See the line

     if (status!=0) decStatus(dn, status, set);

Reusing the DEC_CONTEXT that failed before results into error even if the string is valid number.

I realize that this is not good solution. This PR is meant more to start the discussion on how to fix the issue correctly. I'm not familiar with the jq code.

With this patch all tests works. So maybe just making decStatus() available and calling it here would be good enough. Maybe setting the status to 0 at the beginning of decNumberFromString()?

What do you think?


thalman:
Issue https://github.com/stedolan/jq/issues/2406

thalman:
I force pushed small change. Using

    decContextClearStatus(ctx, DEC_Conversion_syntax);

is probably better than

     ctx->status = 0;

thalman:
Fixes issue #2400 too.

nicowilliams:
Thank you!!!

PR #2432, opened 2022-05-25
[docs] Use yaml.safe_load() in build_*.py scripts

nmulcahey:


itchyny (approved):
Although the original error has been resolved by specifying the loader argument at https://github.com/jqlang/jq/commit/89caf46a5fb0a97274f034affbe5ec05297107b1#diff-9bd5cad8c70843c69fa119d40bbf39d8f6fdbb09818752a44b5eb8683e4c8eaaR190, I prefer using `yaml.safe_load` which looks cleaner. Thank you.

PR #2468, opened 2022-08-26
Use autotools standard ${docdir}

h3xx:
Specified with `--docdir=<path>`, the `$(docdir)` variable is already available, and has the same default as what it was being set to.

By specifying it where it was, it was overriding what the `--docdir` option specified.

With the change, the value from the configure script's `--docdir` option will be respected instead.


itchyny (approved):
LGTM, thank you.

PR #2469, opened 2022-08-26
Fix CI build instructions

h3xx:


h3xx:
Closing; I gave up on this.

Not even really sure what sort of insanity running `pacman` from Arch Linux in a Windows environment really entails, but luckily it's relegated to a temporary CI environment.

I would look into duplicating some other projects' VSCode build process, such as https://github.com/CelestiaProject/Celestia, or https://github.com/Aleph-One-Marathon/alephone to name a couple.

PR #2473, opened 2022-08-30
Clarify in the manual how map_values() behaves differently from map().

### LINKED ISSUE #2466: Docs: unclear explanation of `map_values()`
I love jq. Thank you.

I read the section of the Manual on `map()` and `map_values()`, and although I eventually believe I understand it, I found it hard to grasp the meaning of the text and only understood once I saw the example. I would like to propose a change to improve the text, but I don't want to do that until I better understand what you folks had had in mind when you wrote the text.

If you'd rather I merely propose a change in a PR, please comment saying so and I'll do it.

This is the text that confused me:

> Similarly, map_values(x) will run that filter for each element, but it will return an object when an object is passed.

I _think_ this means the following. Please tell me if I have it right or wrong.

- if `x` is not an object, then `map_values()` behaves exactly like `map()`.
- if `x` is an object, then `map_values()` transforms only the values exactly like `map()` would, but then returns an object with the same keys and the new values.

Of course, those explanations are also terrible, but it's the best I can draft at this moment while my brain wants to get back to my current task. :)

Questions:

- How well do I understand was `map_values()` does?
- Is this what you meant to say by "...but it will return an object when an object is passed"? (I think so, but I'd like to clarify.)
- Instead of answering this second question, would you prefer that I just submit a PR after I've written that text in a much nicer way? ;)

jbrains:
Fixes https://github.com/stedolan/jq/issues/2466

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -892,8 +892,12 @@ sections:
           element of the input array, and return the outputs in a new
           array. `map(.+1)` will increment each element of an array of numbers.
 
-          Similarly, `map_values(x)` will run that filter for each element,
-          but it will return an object when an object is passed.
+          To transform an object using filter `x`, use `map_values(x)`.
+          When an object is passed to `map_values(x)`,
+          it will run that filter for each value in the object,
+          leaving the keys unchanged.
+          When any other type of value is passed to it, `map_values(x)`
+          behaves the same as `map(x)`.
 
           `map(x)` is equivalent to `[.[] | x]`. In fact, this is how
           it's defined. Similarly, `map_values(x)` is defined as `.[] |= x`.


wader:
LGTM 👍

itchyny:
>           When any other type of value is passed to it, `map_values(x)`
>           behaves the same as `map(x)`.

Not always. At least I can give a counterexample as follows.

```bash
 $ jq -n -c '[1,2,3] | map_values(.,.+10)'
[1,2,3]
 $ jq -n -c '[1,2,3] | map(.,.+10)' 
[1,11,2,12,3,13]
```

pkoppstein:
Closing as this issue was specifically dealt with subsequently in 8345a8a62b7e3ea37974ee14bf368073e6faca7a

@jbrains - If the current version of manual.yml is still not to your liking, please feel free to submit a new PR. Thanks.

PR #1527, opened 2017-11-03
Update the man page to note that the sort functions are stable

hawicz:
See also Issue #343

PR #2474, opened 2022-08-30
Document how sort_by() can sort by multiple keys, including an example.

### LINKED ISSUE #2467: Docs: `sort_by()` doesn't mention sorting by multiple keys of an object
I wanted to sort an array of objects by multiple keys (to break ties in the first key). The Manual doesn't describe this at all, but a Stack Overflow article showed that it works with `sort_by(key1, key2, ...)`.

I propose to add this to the Manual.

Based on a comment from https://github.com/stedolan/jq/issues/2466 I'm just going to submit a PR and wait patiently until someone merges it. No rush, since anyone who needs to know this will almost certainly eventually find the information.

jbrains:
Fixes https://github.com/stedolan/jq/issues/2467

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -1368,6 +1368,11 @@ sections:
           `sort_by(foo)` compares two elements by comparing the result of
           `foo` on each element.
 
+          Also note that `sort_by(.k1, .k2, .k3, ...)` behaves as you would
+          expect: it sorts first by comparing elements by key `k1`, then
+          comparing elements by key `k2` to "break ties" left by `k1`,
+          then comparing elements by key `k3`, and so on.
+
         examples:
           - program: 'sort'
             input: '[8,3,null,6]'
@@ -1375,6 +1380,9 @@ sections:
           - program: 'sort_by(.foo)'
             input: '[{"foo":4, "bar":10}, {"foo":3, "bar":100}, {"foo":2, "bar":1}]'
             output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":100}, {"foo":4, "bar":10}]']
+          - program: 'sort_by(.foo, .bar)'
+            input: '[{"foo":4, "bar":10}, {"foo":3, "bar":100}, {"foo":2, "bar":1}, {"foo":3, "bar":200}]'
+            output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":100}, {"foo":3, "bar":200}, {"foo":4, "bar":10}]']
 
       - title: "`group_by(path_expression)`"
         body: |


wader:
LGTM 👍

docs/content/manual/manual.yml
```diff
@@ -1368,13 +1368,21 @@ sections:
...
         examples:
           - program: 'sort'
             input: '[8,3,null,6]'
             output: ['[null,3,6,8]']
           - program: 'sort_by(.foo)'
             input: '[{"foo":4, "bar":10}, {"foo":3, "bar":100}, {"foo":2, "bar":1}]'
             output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":100}, {"foo":4, "bar":10}]']
+          - program: 'sort_by(.foo, .bar)'
+            input: '[{"foo":4, "bar":10}, {"foo":3, "bar":100}, {"foo":2, "bar":1}, {"foo":3, "bar":200}]'
+            output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":100}, {"foo":3, "bar":200}, {"foo":4, "bar":10}]']
```
itchyny, line 1385: The input value can be improved by flipping the foo:3 elements to make it clear the program sorts by bar (program `sort_by(.foo)` yields the same output currently).

php-coder: BTW, it would also be great to have another complicated example where the keys have transformations. For example:
```console
$ echo '[{"foo":"20", "bar":10}, {"foo":"3", "bar":100}, {"foo":"2", "bar":1}, {"foo":"3", "bar":200}]' | jq '. | sort_by([.foo | tonumber], .bar)'
[
  {
    "foo": "2",
    "bar": 1
  },
  {
    "foo": "3",
    "bar": 100
  },
  {
    "foo": "3",
    "bar": 200
  },
  {
    "foo": "20",
    "bar": 10
  }
]
```

itchyny: @php-coder I think such a complicated example with filter combination can be listed in the [Cookbook](https://github.com/jqlang/jq/wiki/Cookbook), and actually there is an [example using sort_by with tonumber](https://github.com/jqlang/jq/wiki/Cookbook#sort-by-numeric-values-extracted-from-text). You can freely edit the wiki page if this example is not enough. BTW I think there should be an example for version sorting. Thanks.

pkoppstein:
@jbrains - Since #1527 has been closed, could you please add a sentence to the effect that the sort functions are stable.  E.g. 
```
As of jq 1.6, the sorting functions are stable.
This means that one way to sort by multiple
criteria is to create a suitable pipeline.
```

Thanks.

PR #2479, opened 2022-09-03
feat: `dig(<string>)` extracts from a nested index

yertto:
`dig("foo")` behaves almost the same as `.["foo"]` but it
can dig down through a sequence of keys obtained from
spliting `<string>` on `"."`.

ie. so `dig("foo.bar")` is equivalent to `.["foo"]["bar"]`.
(although not the same as `.["foo.bar"]`)

(Nb. named after ruby's [dig](https://apidock.com/ruby/Hash/dig))

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -42,6 +42,7 @@ def to_entries: [keys_unsorted[] as $k | {key: $k, value: .[$k]}];
 def from_entries: map({(.key // .Key // .name // .Name): (if has("value") then .value else .Value end)}) | add | .//={};
 def with_entries(f): to_entries | map(f) | from_entries;
 def reverse: [.[length - 1 - range(0;length)]];
+def dig($path): reduce ($path | split("."))[] as $key (.; .[$key]);
 def indices($i): if type == "array" and ($i|type) == "array" then .[$i]
   elif type == "array" then .[[$i]]
   elif type == "string" and ($i|type) == "string" then _strindices($i)
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -393,6 +393,28 @@ sections:
           `.["foo"]` (.foo above is a shorthand version of this, but
           only for identifier-like strings).
 
+      - title: "Dig for Generic Object Index: `dig(<string>)`"
+
+        body: |
+
+          `dig("foo")` behaves almost the same as `.["foo"]` but it
+          can dig down through a sequence of keys obtained from
+          spliting `<string>` on `"."`.
+
+          ie. so `dig("foo.bar")` is equivalent to .["foo"]["bar"]`.
+          (Which is not the same as `.["foo.bar"]`)
+
+        examples:
+          - program: 'dig("a.b")'
+            input: '{"a": {"b": 42}}'
+            output: ['42']
+          - program: 'dig("a")'
+            input: '{"a": {"b": 42}}'
+            output: ['{"b": 42}']
+          - program: 'dig("a.b")'
+            input: '{"a.b": 42}'
+            output: ['null']
+
       - title: "Array Index: `.[2]`"
         body: |
 


wader:
Curious about the use case for this, when the path expression you want to index comes from some user input etc? to support something like this i think it also would have to support indexing, key quoting and escaping i guess.

For [fq](https://github.com/wader/fq) i have `expr_to_path` and `path_to_expr` that can be used to build things like this. So in fq `dig` could be defined as `def dig($expr): getpath($expr | expr_to_path)`. Would be great if jq had similar functions to work with path expressions.

src/builtin.jq
```diff
@@ -42,6 +42,7 @@ def to_entries: [keys_unsorted[] as $k | {key: $k, value: .[$k]}];
 def from_entries: map({(.key // .Key // .name // .Name): (if has("value") then .value else .Value end)}) | add | .//={};
 def with_entries(f): to_entries | map(f) | from_entries;
 def reverse: [.[length - 1 - range(0;length)]];
+def dig($path): reduce ($path | split("."))[] as $key (.; .[$key]);
```
emanuele6, line 45: I would avoid using `split/1` since I consider it basically deprecated (all the other `split` and `splits` functions take a regex, but `split/1` doesn't and it is kept like that for backward compatibility). I would especially avoid it with `"."` since in regexps, `.` matches any characters and if `split/1` aligns with the other `split*` functions in the future, this will not work anymore.

So, using string division, it would be:

```suggestion
def dig($path): reduce ($path / ".")[] as $key (.; .[$key]);
```

But also notice that this is a really silly way of writing:
```jq
def dig($path): getpath($path / ".");
```
So maybe use that instead. To be honest I don't think this function makes much sense as a builtin.

pkoppstein:
@ emanuele6 wrote:

> To be honest I don't think this function makes much sense as a builtin.

Indeed.  Another consideration is that functions defined in builtins.jq have an overhead.  

It would help the jq maintainers focus on important enhancements if this PR were retracted.

yertto:
Thanks everyone!

 > Curious about the use case for this, when the path expression you want to index comes from some user input etc?
  
@wader - yes - exactly - the path expression is coming from user input.


> ```getpath($path / ".")```

@emanuele6 - perfect - thanks for that! 🙏 


> It would help the jq maintainers focus on important enhancements if this PR were retracted.

@pkoppstein - Yer very happy to close this.

Looks like i was trying to solve my problem with a shovel, when a perfectly good hammer & nail existed.

For anyone else who ends up going down this wrong path,
I'll just make a couple of notes from my learnings that I didn't manage to pick up from the [manual](https://stedolan.github.io/jq/manual/#Builtinoperatorsandfunctions).

1. Strings can be split using the `division` (`/`) operator:
    `echo '"a/bb/ccc"' | jq -c '. / "/"'`
    ```
    ["a","bb","ccc"]
     ```

    `echo '"a.bb.ccc"' | jq -c '. / "."'`
    ```
    ["a","bb","ccc"]
    ```

2. this *single* array can then be passed into `getpath(PATHS)` 
    `echo '{"a":{"bb":{"ccc": 42, "ddd": 99}}}' | jq --arg path 'a.bb.ccc' 'getpath($path / ".")'`
    ```
    42
    ```

wader:
👍  as the path is just an array with values you can also do something like this:
```sh
$ echo '{"a":[{"bb":{"ccc": 42, "ddd": 99}}]}' | jq --arg path '["a",0,"bb","ccc"]' 'getpath($path | fromjson)'
42
```
or even
```sh
$ echo '{"a":[{"bb":{"ccc": 42, "ddd": 99}}]}' | jq --argjson path '["a",0,"bb","ccc"]' 'getpath($path)'
42
```

yertto:
>```
> $ echo '{"a":[{"bb":{"ccc": 42}}]}' | jq --arg     path '["a",0,"bb","ccc"]' 'getpath($path | fromjson)'
> 42
> $ echo '{"a":[{"bb":{"ccc": 42}}]}' | jq --argjson path '["a",0,"bb","ccc"]' 'getpath($path           )'
> 42
> ```

@wader - ahh that's neat too!

Although my preferred option would've been your first suggestion of `expr_to_path`,
but *also* with a way of passing expressions in as arguments.

So, following on from your second suggestion I was thinking this is a closer fit to the existing naming scheme:
```
$ echo '{"a":[{"bb":{"ccc": 42}}]}' | jq --arg     path '.a[0].bb.ccc' '($path | fromexpr) * 10'
420
$ echo '{"a":[{"bb":{"ccc": 42}}]}' | jq --argexpr path '.a[0].bb.ccc' '$path * 10'
420
```

Then with a `--argexpr` could you also do things like?
```
$ echo '[2, 7, 1, 8, 2]' | jq --argexpr user_input_expr 'last' '(. | $user_input_expr) * 10'
20
$ echo '[2, 7, 1, 8, 2]' | jq --argexpr user_input_expr 'max'  '(. | $user_input_expr) * 10'
80
```

---
Oh and @wader I had a look at your [fq presentation](https://www.youtube.com/watch?v=GJOq_b0eb-s) - wow!
What an amazing little world you've built there, with your _"binary indenter"_

(Although, I think its a shame so many people - me included ... for *many* years - don't even realize `jq` is a first-class programming language and not just a few extra bits & bobs tacked on to command line arguments. I had a colleague at work - who's a fan & has spent many hours using it - only just discover that they could define their own functions.)

BTW, there was a very quick transition between the `fq` script and the `json` it produced at [34:55](https://www.youtube.com/watch?v=GJOq_b0eb-s&t=2083s).
I reckon it'd be great for folks like me coming to this with no idea of what it does, if we could seee the `fq` filter script and the `json` output side-by-side to get a clearer understanding of how the script produces its `json` output.

Ahh and I reckon the slides you had on `jq` were great - not too much or too little - just really easy to understand examples.
And I think [that](https://www.youtube.com/watch?v=GJOq_b0eb-s&t=940s) was the best example I've ever seen for `foreach`.
(ie. much clearer than the one currently in the manual - which totally lost me.)

It seems like `jq` is like playing with lego, and `fq` provides a similar experience, but takes it to the sub-atomic level! :atom: 

And then, that extensively featured repl, and how skilled you are at using it, was just amazing to watch. ✨ 

Not sure if I'll ever be spending much time down at that sub-atomic level to make any (direct) use of it myself, although I'm sure there's a lot that flows back from it (like your [jq-lsp](https://github.com/wader/jq-lsp)), but mostly its just a delight to know that it exists.

Inspirational work!

wader:
> Although my preferred option would've been your first suggestion of `expr_to_path`, but _also_ with a way of passing expressions in as arguments.
> 
> So, following on from your second suggestion I was thinking this is a closer fit to the existing naming scheme:
> 
> ```
> $ echo '{"a":[{"bb":{"ccc": 42}}]}' |
>   jq --arg     path '.a[0].bb.ccc' '($path | fromexpr) * 10'
> 420
> 
> $ echo '{"a":[{"bb":{"ccc": 42}}]}' |
>   jq --argexpr path '.a[0].bb.ccc' '$path * 10'
> 420
> ```
> 
> Then with a `--argexpr` could you also do things like?
> 
> ```
> $ echo '[2, 7, 1, 8, 2]' |
>   jq --argexpr user_input_expr 'last' '(. | $user_input_expr) * 10'
> 20
> 
> $ echo '[2, 7, 1, 8, 2]' |
>   jq --argexpr user_input_expr 'max'  '(. | $user_input_expr) * 10'
> 80
> ```

Yes some kind of functions for working with path index expressions would be nice but not sure if it should be part of standard jq or not.

> Oh and @wader I had a look at your [fq presentation](https://www.youtube.com/watch?v=GJOq_b0eb-s) - wow! What an amazing little world you've built there, with your _"binary indenter"_
> 
> (Although, I think its a shame so many people - me included ... for _many_ years - don't even realize `jq` is a first-class programming language and not just a few extra bits & bobs tacked on to command line arguments. I had a colleague at work - who's a fan & has spent many hours using it - only just discover that they could define their own functions.)

Thanks, same for me, have used jq a lot for simple things and then one day i read the man page and was amazed and got a bit obsessed.

> BTW, there was a very quick transition between the `fq` script and the `json` it produced at [34:55](https://www.youtube.com/watch?v=GJOq_b0eb-s&t=2083s). I reckon it'd be great for folks like me coming to this with no idea of what it does, if we could seee the `fq` filter script and the `json` output side-by-side to get a clearer understanding of how the script produces its `json` output.

That is good feedback, maybe it would be better to demonstrate using a simpler script also?

> Ahh and I reckon the slides you had on `jq` were great - not too much or too little - just really easy to understand examples. And I think [that](https://www.youtube.com/watch?v=GJOq_b0eb-s&t=940s) was the best example I've ever seen for `foreach`. (ie. much clearer than the one currently in the manual - which totally lost me.)

Actually have been thinking about writing two jq tutorials, one something like "practical jq" with common usage examples, and one "jq the hard way" describing some internals how some expressions get evaluated by "stedolan"-jq or gojq.

> It seems like `jq` is like playing with lego, and `t` provides a similar experience, but takes it to the sub-atomic level! :atom:

Yes one thing that really makes jq nice is how "composable" it is, both the pipe syntax and other things, very ergonomic to work with. When i started working on fq and played around with some other languages i noticed that most of them were not that composable and also more verbose when recusing/transforming/filtering etc.

> And then, that extensively featured repl, and how skilled you are at using it, was just amazing to watch.. ✨

😊  i spend quite a lot of time in the REPL digging into broken media files and for other jq needs so really want to make it comfortable. Really happy about how the sub REPL:s with "slurp" support turned out.

> Not sure if I'll ever be spending much time down at that sub-atomic level to make any (direct) use of it myself, although I'm sure there's a lot that flows back from it (like your [jq-lsp](https://github.com/wader/jq-lsp)), but mostly its just a delight to know that it exists.

If you  take a look at the internals of fq you will noticed that large part of it is written jq so i felt it is probably a good investment to have at least basic syntax and completion support in some IDE.

> Inspirational work!

Thanks again and lots of credit should be passed on to @itchyny for gojq which made fq possible.

If you want to talk more or have questions about fq, feel fee to email me or open an issue in the fq repo. Maybe good to not spam people here :)

PR #2480, opened 2022-09-06
Use right headers and flags in configure

kraj:


itchyny:
Since #2519 is merged, closing this PR.

PR #2482, opened 2022-09-09
Fix double-installed doc

h3xx:
Remove redundant README

The README symlink was causing the file (a regular file) to be installed twice in <DOCDIR>, once as `README` and once as `README.md` with the exact same contents.


itchyny (commented):
I think we don't need to list `COPYING` and `AUTHORS` here, because `automake` automatically includes the files if found. Also, we will not need to list `README.md` here because it includes automatically (when `README` does not exist), but this is a new feature introduced in automake 1.16.4 in Jul 2021, so let's keep this (I believe we don't even need this line in the future). Correct me if my understanding is wrong.

itchyny (approved):
Considering the adoption of .md detection in automake, no one would against the distribution with only README.md. So I'll include this, thanks.

PR #2487, opened 2022-09-14
Add !x as syntactic sugar for x | not

### LINKED ISSUE #2486: Feature request: boolean not operator
Sometimes, I just want to select elements in an array based on their boolean attributes, and it gets tiring to write `== false` all the time.

It would be nice if I could just do something like:
```
.[] | select(.attr1 and !.attr2)
```
instead of:
```
.[] | select(.attr1 and .attr2 == false)
```

I understand I could do something like:
```
.[] | select(.attr2 | not)
```
but that seems unnecessarily wordy and doesn't look great when incorporated as part of complex boolean expressions:
```
.attr1 and ((.attr2 and (.attr3 | not)) | not)
```
This is difficult to understand rather than:
```
.attr1 and !(.attr2 and !.attr3)
```
I know I can use De Morgan's law to reduce parts of the expression, but I would like it if the tool could lower the cognitive burden.

Is there a reason why the boolean not operator does not exist in the language? Or am I perhaps missing it?

alkim0:
Fix #2486

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -250,6 +250,17 @@ static jv f_negate(jq_state *jq, jv input) {
   return ret;
 }
 
+static jv f_not(jq_state *jq, jv input) {
+  jv ret;
+  if (jv_get_kind(input) == JV_KIND_NULL || jv_get_kind(input) == JV_KIND_FALSE) {
+    ret = jv_true();
+  } else {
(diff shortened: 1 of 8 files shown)


itchyny:
Based on the comments in #2486, we're unlikely to accept this syntax.

PR #2492, opened 2022-10-16
Show missing backslash in manual

zstadler:
The 1.6 manual about [`sub(regex;tostring)`](https://stedolan.github.io/jq/manual/#sub(regex;tostring)sub(regex;string;flags)) currently says:

> a reference to a captured variable named "x" would take the form: "(.x)".

Using `"(.x)"` produces this string literally rather than the captured variable. Instead, the manual should say:

> a reference to a captured variable named "x" would take the form: `"\(.x)"`.

This PR may also need to be back-ported to earlier versions of the jq manual as well as ported to the development version of the jq manual.

### DIFF
--- a/docs/content/manual/v1.6/manual.yml
+++ b/docs/content/manual/v1.6/manual.yml
@@ -2361,7 +2361,7 @@ sections:
           be a jq string, and may contain references to named captures. The
           named captures are, in effect, presented as a JSON object (as
           constructed by `capture`) to `tostring`, so a reference to a captured
-          variable named "x" would take the form: "\(.x)".
+          variable named "x" would take the form: `"\(.x)"`.
 
         example:
           - program: 'sub("^[^a-z]*(?<x>[a-z]*).*")'


zstadler:
As far as I can tell, the [CI failure](https://ci.appveyor.com/project/stedolan/jq/builds/45082987) has nothing to do with the contents of this PR

```
C:\msys64\usr\bin\bash -lc "pacman-key --verify msys2-keyring-r21.b39fb11-1-any.pkg.tar.xz.sig"
==> Checking msys2-keyring-r21.b39fb11-1-any.pkg.tar.xz.sig...
==> ERROR: The signature identified by msys2-keyring-r21.b39fb11-1-any.pkg.tar.xz.sig could not be verified.
Command exited with code 1
cat config.log
Command exited with code 1
```

zstadler:
@stedolan, @nicowilliams 

Please look at this documentation fix PR.

The failure in the CI check is unrelated to the change in the `README.md` file,

wader:
Fix LGTM but have no access to merge

Issue #2305, opened 2021-04-26
About the jq's release process (Was: Is jq is still alive/maintained ?)


tst2005:
Hello,

`jq` is a already powerfull enough for most of use.
it does not need new release too often but ...
There are lot of PR for bugfixes and interesting improvements.

is @stedolan still maintaining jq ?
if it is a "dead" project, Am I alone to think about a fork ?

Regards,

carlsmedstad:
Also wondering this. It would be a shame if it stopped being developed and maintained. Maintainers, is there anyway to help you out?

wtlangford:
Hey, so I'm still around- the main issue is that I've been super busy with work and dealing with life this last year that it's been especially hard to find the time/energy to devote to jq.

We'd definitely appreciate help- 
I've got a couple small changes floating around that I need to finish up which should improve some build behaviors (I think I can make the CI builds more reliable- they keep breaking. Should also be able to fix some things around the math functions that got brought up "recently"). I'll also try and dig up that list of things that needed finishing in order to cut the next release and see if I can set up some issues for it and a milestone so that it's clear what needs doing. (Depending on what's left on that, I might just say we should cut a release and schedule myself some time to do it).

Beyond that, I need to go review all the open PRs- I know I've seen some good stuff come through my email- and also try to clean up the issue tracker some.

tst2005:
Thanks for your answers.

@carlsmedstad : IMHO, I don't think it is a shame if @stedolan can not or don't want maintain jq anymore. I prefer saying a big thank for the work already done.

Today
I see the [stedolan's jq repository](https://github.com/stedolan/jq) has 19.3k stars, 1.1k fork, 121 PR, 574 issues.
121/1100 --> ~10% of forks are PR
574/1100 --> ~50% of forks equal to an issue (bug/ask/feature request/...)

There different solutions to get a projet maintained, :
- the original author is still here, he can do stuff to transfert the role to another guy/community.
- the original author is not here. The repository is forked to a new "official/maintained" one, it is difficult to make people create issue/PR on the new repo because all current search result still anwer the original repository.

For now I don't read people answering "I forked and I can maintain jq",
@wtlangford you are the closest one but you already told us you are busy.

I'm in the same case. I want a maintained jq, I can contribute but I don't have enough time to be the only one to maintain jq.

When nobody has the time to maintain the projet, we can try to make a github "organisation" like "jq-community" or "jq-maintainers". It imply to work together (it is not always easy).

[...]

Issue #2284, opened 2021-03-11
Please document all command-line options in --help


rrthomas:
**Describe the bug**

--help does not document all command-line options. This breaks bash-completion, as well as confusing the user.

See https://github.com/scop/bash-completion/pull/507#pullrequestreview-609207835

PR #2504, opened 2022-11-22
doc: Added information on --indent to --help

EvanCarroll:
Simple doc patch when I noticed functionality I already need was there just undocumented.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -81,6 +81,8 @@ static void usage(int code, int keep_it_short) {
       "  -M               monochrome (don't colorize JSON);\n"
       "  -S               sort keys of objects on output;\n"
       "  --tab            use tabs for indentation;\n"
+      "  --indent n       use $n spaces for indentation.\n"
+      "                   between -1 (default of 4 spaces) and 7;\n"
       "  --arg a v        set variable $a to value <v>;\n"
       "  --argjson a v    set variable $a to JSON value <v>;\n"
       "  --slurpfile a f  set variable $a to an array of JSON texts read from <f>;\n"


src/main.c
```diff
@@ -81,6 +81,8 @@ static void usage(int code, int keep_it_short) {
       "  -M               monochrome (don't colorize JSON);\n"
       "  -S               sort keys of objects on output;\n"
       "  --tab            use tabs for indentation;\n"
+      "  --indent         use indent amount of spaces for indentation.\n"
```
wader, line 84: Maybe should be `--indent n` to be clearer?

EvanCarroll: @wader fixed

wader: 👍 hopefully will get merged eventually, see #2305 for details

pkoppstein:
Since --indent requires an argument, an argument should be mentioned in usage(). Following the manual, `--indent n` would be appropriate.

The larger point here is that documenting all the options in usage() in this manner would increase the size of the executable quite a lot, and goes against the idea of keeping the executable small by providing details online.

I would propose that usage() simply provide a synopsis of *all* the options (in the usual man page style), leaving the descriptive details of the options to the online documentation.

EvanCarroll:
@pkoppstein the patch was already amended 12 hours before your comment to include `n`. And you're talking about a couple of bytes of text. Hardly something that will "increase the size of the executable quite a lot."

pkoppstein:
@EvanCarroll - You apparently overlooked the phrase “ … documenting all the options in usage() …”.

(The screen I was looking at only showed the line without the “n”.)

wader:
I would not worry about binary size either, possibly that the text is getting long, but maybe then we should cut down on some other usage text instead? maybe the stuff about IEEE754 numbers etc

itchyny (commented):
I'm not worried about the executable size due to this help, we should list all the option #2284 (we should rather consider replace the builtins by C-coded internals). The only concern about this PR is about the newline in the description; can Bash's `_parse_help` parse this form?

itchyny:
Looks OK.
```sh
 $ _parse_help jq -h | xargs
-c -n -e -s -r -R -C -M -S --tab --arg --argjson --slurpfile --rawfile --args --jsonargs --
 $ _parse_help ./jq -h | xargs
-c -n -e -s -r -R -C -M -S --tab --indent --arg --argjson --slurpfile --rawfile --args --jsonargs --
```

itchyny (approved):

src/main.c
```diff
@@ -81,6 +81,8 @@ static void usage(int code, int keep_it_short) {
       "  -M               monochrome (don't colorize JSON);\n"
       "  -S               sort keys of objects on output;\n"
       "  --tab            use tabs for indentation;\n"
+      "  --indent n       use $n spaces for indentation.\n"
+      "                   between -1 (default of 4 spaces) and 7;\n"
```
itchyny, line 85: Oops, `-1` is not `4 spaces`, but the tab character.

itchyny (changes requested):

PR #2505, opened 2022-11-26
Build and test using GitHub actions workflow

dave-r12:
Also update to oniguruma 6.7.1 to fix the build on MacOS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,100 @@
+name: build
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  ubuntu:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: sudo apt-get install -y gcc libtool make autoconf automake flex bison
+      - name: CI-Build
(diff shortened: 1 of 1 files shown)


wader:
Hey, about CI i know @nicowilliams has been looking into adding at https://github.com/jqlang/jq

Oniguruma update looks resonable to me. Would it make sense to update to latest? 6.9.8?

The pow/log2 issue for cygwin i don't know much about. Could it be some float as string representation issue? how does it handle  `jq -n 2251799813685248` and `jq -n '2251799813685248|log2'`, maybe can give some clue.

mrienstra:
Re: log2: You've probably gotten this far already, but here are a few links just in case it's helpful. Looks like others have encountered this issue before:
1. https://www.gnu.org/software/gnulib/manual/html_node/log2.html
2. https://cygwin.cygwin.narkive.com/NGBO7wVB/log2-function-from-c-standard-library-is-inaccurate
3. https://hg.octave.org/octave/log?rev=log2 (<-- not specific to this issue, but you might find some ideas on avoiding it here)

itchyny:
We have switched to GitHub Actions in  #2596, so closing this PR. Thank you.

Issue #1271, opened 2016-11-06
Feature request: NUL-delimited output


charles-dyfis-net:
Right now, the standard-practice way to read an array from `jq` into a shell-script is to use raw output and parse on newlines.

However, JSON strings can contain literal newlines; this makes such parsing error-prone.

NUL-delimited output, allowing `IFS= read -r -d '' string` to read exactly one C string unambiguously, would resolve this.

eric-brechemier:
@charles-dyfis-net is it not simpler in this case to keep newline escaping, instead of using raw output? This allows to keep a single item per line, which is easier to loop over in a shell script:

### input.json

```
[
  "LF\nLF",
  "TAB\tTAB",
  "FF\fFF"
]
```

### Filter

```
.[]
```

### Command Line

```sh
$ jq '.[]' input.json
```

### Output

```
"LF\nLF"
"TAB\tTAB"
"FF\fFF"
```

Otherwise, you can actually add a character of your choice at the end of each line, directly from your `jq` filter:

### Filter + NUL

```
.[]
| ( . + "\u0000")
```

### Command Line + NUL

```
$ jq '.[] | ( . + "\u0000")' input.json
```

### Output + NUL

```
"LF\nLF\u0000"
"TAB\tTAB\u0000"
"FF\fFF\u0000"
```

### Command Line + NUL as Raw (View as Hex)

```
$ jq -r '.[] | ( . + "\u0000")' input.json | xxd
```

### Output + NUL as Raw (Viewed as Hex)

```
0000000: 4c46 0a4c 4600 0a54 4142 0954 4142 000a  LF.LF..TAB.TAB..
0000010: 4646 0c46 4600 0a                        FF.FF..
```

charles-dyfis-net:
Thank you -- I actually have a few StackOverflow answers I'm going to want to amend in light of the patterns suggested in this ticket.

That said, this still would be a desirable feature to have.

[...]

PR #2535, opened 2023-02-03
Remove deprecated `--nul-output` documentation

peter-dolkens:
`--nul-output` / `-0` was removed RE: #1271 and #2350

This updates the documentation so one doesn't go crazy trying to figure out why it's not working 😂

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -199,11 +199,6 @@ sections:
 
         Like `-r` but jq won't print a newline after each output.
 
-      * `--nul-output` / `-0`:
-
-        Like `-r` but jq will print NUL instead of newline after each output.
-        This can be useful when the values being output can contain newlines.
-
       * `-f filename` / `--from-file filename`:
 
         Read filter from the file rather than from a command line, like


wader:
Hi, what version is this? with jq master i see this.
```shell
$ jq --version
jq-1.6-159-gcff5336

$ jq -n --nul-output '1,2,3' | hexdump -C
00000000  31 00 32 00 33 00                                 |1.2.3.|
00000006

$ jq -n --null-output '1,2,3' | hexdump -C
jq: Unknown option --null-output
Use jq --help for help with command-line options,
or see the jq manpage, or online docs  at https://stedolan.github.io/jq

# -0 treated as number
$ jq -n -0 '1,2,3' | hexdump -C
00000000  2d 30 0a                                          |-0.|
00000003
```

Confusd, so #1271 was not reverted? and -0 was not added?

peter-dolkens:
Sorry if I was mistaken. I'm running on M2 Macbook Air

```
➜  ~ jq --version
jq-1.6
➜  ~ jq --help
jq - commandline JSON processor [version 1.6]

Usage:  jq [options] <jq filter> [file...]
        jq [options] --args <jq filter> [strings...]
        jq [options] --jsonargs <jq filter> [JSON_TEXTS...]

jq is a tool for processing JSON inputs, applying the given filter to
its JSON text inputs and producing the filter's results as JSON on
standard output.

The simplest filter is ., which copies jq's input to its output
unmodified (except for formatting, but note that IEEE754 is used
for number representation internally, with all that that implies).

For more advanced filters see the jq(1) manpage ("man jq")
and/or https://stedolan.github.io/jq

Example:

        $ echo '{"foo": 0}' | jq .
        {
                "foo": 0
        }

Some of the options include:
  -c               compact instead of pretty-printed output;
  -n               use `null` as the single input value;
  -e               set the exit status code based on the output;
  -s               read (slurp) all inputs into an array; apply filter to it;
  -r               output raw strings, not JSON texts;
  -R               read raw strings, not JSON texts;
  -C               colorize JSON;
  -M               monochrome (don't colorize JSON);
  -S               sort keys of objects on output;
  --tab            use tabs for indentation;
  --arg a v        set variable $a to value <v>;
  --argjson a v    set variable $a to JSON value <v>;
  --slurpfile a f  set variable $a to an array of JSON texts read from <f>;
  --rawfile a f    set variable $a to a string consisting of the contents of <f>;
  --args           remaining arguments are string arguments, not files;
  --jsonargs       remaining arguments are JSON arguments, not files;
  --               terminates argument processing;

Named arguments are also available as $ARGS.named[], while
positional arguments are available as $ARGS.positional[].

See the manpage for more options.
➜  ~
```

I assumed from the issues linked above that this functionality had been reverted, as it is in documentation, but not in the latest official release as far as I could tell.

pkoppstein:
Unfortunately, the versioning information on the online documentation is not always exactly correct.

wader:
@peter-dolkens No worries, i'm confused also what the state of master is. Looking at the code in master https://github.com/stedolan/jq/blob/master/src/main.c#L402 seems like `-0` and `--nul-output` is there... but `-0` is broken for some reason, i get a feeling it ends up here https://github.com/stedolan/jq/blob/master/src/main.c#L325 as `isoptish` check if the option `isalpha`. But not sure it's a good idea to allow `-0` as it's a bit ambigous.

peter-dolkens:
For what it's worth - whilst I understand the reasoning behind the discussions I linked - I'd still support `-0` and similar options.

The challenge I was trying to solve was I had a collection of "json-like" blobs serialized in a larger json object.

I was trying to output them raw using `-r` which worked fine, but then I wanted to run them through `hjson` or similar to sanitize them. Unfortunately I was struggling to find a separator that I could use to pass to `xargs` to achieve this as `-0` wasn't working, and the blobs contained new lines themselves.

I personally don't think the security argument holds much weight, as ultimately it's not the parsing or separator that's causing the security vulnerability, but rather their use on unsanitized/unverified inputs. Javascript has `eval`, Dotnet has `Assembly.Load`, Typescript has `require` - these aren't security vulnerabilities - it's the *use* of these features that may constitute a vulnerability.

itchyny:
Ref: #2235.

itchyny:
Now that #2235 is merged to the master branch, I think the master version fully support what is written in the development version of documentation. Note (again) that this option is written in the development version document not 1.6, and will be shipped in the next version.

PR #2538, opened 2023-02-08
fix: tarball fails distcheck: add files missing from manifest

chickenandpork:
The distributed tarball needs `autoreconf` to run before it can be built; otherwise the docs say that you don't need autotools.  It's helpful to include enough to avoid autotools because the latest autotools gives a bunch of warnings.  Including, say, `configure` and `Makefile.in` allows someone porting the code to eliminate/replace these files without much effort, but enables the person who isn't replacing them to continue with minimal fuss as well.

Additionally, the lack of a `darwin/amd64` or `osx-arm-jq` is causing some users to try to overlay the missing tool so we should provide an update that includes that arch.


wader:
Hi, not a auto* expert at all, i see this in the 1.6 source tar:
```
$ curl -sL https://github.com/stedolan/jq/releases/download/jq-1.6/jq-1.6.tar.gz | tar t | grep -E '(configure|Makefile)'
jq-1.6/configure
jq-1.6/configure.ac
jq-1.6/Makefile.am
jq-1.6/Makefile.in
jq-1.6/modules/oniguruma/configure.ac
jq-1.6/modules/oniguruma/Makefile.am
jq-1.6/modules/oniguruma/test/Makefile.am
jq-1.6/modules/oniguruma/src/Makefile.am
jq-1.6/modules/oniguruma/src/Makefile.windows
jq-1.6/modules/oniguruma/sample/Makefile.am
```
The github release page has a "Source code (tar.gz)" link that i think is an automatic tar.gz of the repo, maybe you downloaded that one? or is this an issue just in master?

But the change in the PR seems to another problem?

(Im not a jq maintainer but the project is a bit stalled atm, see #2305, so trying to help out anyways)

itchyny (approved):
Thank you. Without this fix, `make rpm` just fails.

nicowilliams:
Thanks!

Issue #2545, opened 2023-03-01
Segfault in multi-threaded environment in jq 1.6


thalman:
**Describe the bug**
When jq library is used in multiple threads it fails with coredump

**To Reproduce**
following code reproduces the issue. It is based on what mod_auth_openidc does

    #include <stdlib.h>
    #include <jq.h>
    #include <stdio.h>
    #include <string.h>
    #include <unistd.h>
    #include <pthread.h>
    
    int my_jq_parse(jq_state *jq, struct jv_parser *parser)
    {
        int rv = 0;
        jv value;
    
        value = jv_parser_next(parser);
        while (jv_is_valid(value)) {
            jq_start(jq, value, 0);
            jv result;
    
            while (jv_is_valid(result = jq_next(jq))) {
                jv dumped = jv_dump_string(result, 0);
                const char *str = jv_string_value(dumped);
                printf("dumped: %s\n", str);
            }
    
            jv_free(result);
            value = jv_parser_next(parser);
        }
    
        if (jv_invalid_has_msg(jv_copy(value))) {
            jv msg = jv_invalid_get_msg(value);
            printf("invalid: %s\n", jv_string_value(msg));
            jv_free(msg);
            rv = 1;
        } else {
            jv_free(value);
        }
    
        return rv;
    }
    
    void *run(void *not_used) {
        int rv;
        jq_state *jq;
        const char *prg = ".data";
    
        jq = jq_init();
        printf("jq_init jq: %p prg: %s\n", jq, prg);
        if (jq_compile(jq, prg) == 0) {
            jq_teardown(&jq);
            return NULL;
        }
        printf("compiled\n");
    
        struct jv_parser *parser = jv_parser_new(0);
        const char *buf = "{ \"data\": 1 }";
        jv_parser_set_buf(parser, buf, strlen(buf), 0);
        rv = my_jq_parse(jq, parser);
        if (rv != 0) {
            printf("my_jq_parse failed!\n");
        }
    
        jv_parser_free(parser);
        jq_teardown(&jq);
        return NULL;
    }
    
    #define THREADS 2
    /* calling run() twice works fine */
    /* calling them in threads causes core-dump */
    int main (int argc, char *argv[])
    {
        pthread_t threads[THREADS];
        int createerror;
        int a;
    
         memset(&threads, 0, sizeof(threads));
        for (a = 0; a < THREADS; ++a) {
            // sleep(1); // < if you want to run threads sequentionally 
[...]

Issue #2483, opened 2022-09-09
RFC: Shared-nothing multithreaded usage of libjq.so


edwardb96:
**Describe the bug**
Shared-nothing multi-threading crashes/segfaults due to a bug similar / to that described by #2056, the bug is only present when using a library compiled to use the pthreads implementation of thread-local storage (`pthread_setspecific`, `pthread_getspecific` and `pthread_create_key`) `USE_TLS` must be false. This isn't the default for the stated environment below when compiling manually but is in the distro packages.

**To Reproduce**
Compile and run the program from this gist https://gist.github.com/edwardb96/47aa1c0b0416846f6292a3818058624d which uses
*completely separate jq_state and jv instances* on separate threads.

I am not sure to what extent this library usage is supported so I'm not sure if I'm justified in referring to it as a bug. If multi-threaded usage is ill-advised/unsupported even in this specific 'shared-nothing' kind of scenario perhaps its worth making this more explicit in the documentation.

**Expected behavior**
The program should run to completion, exiting normally with code 0 and outputting the following to stdout
```
Running multi-threading tests
Passed multi-threading test
```

**Environment (please complete the following information):**
 - OS and Version: Fedora Linux 36, using standard jq and jq-devel packages.
 - jq version 1.6

**Additional context**
The documentation / wiki seem to warn off multi-threaded usage due to the non-atomic reference counters
in `jv` values. Despite this I get the impression from some of the [pull request comments](https://github.com/stedolan/jq/pull/109#issue-13742876) that this kind of multi-threading is quietly 'supported' - is this a correct assesment?

For my use-case this kind of 'shared-nothing' multithreading seems to work fine when `USE_TLS` (or my fix) is enabled but I haven't yet performed extensive testing.

Anyway, my analysis of the bug is as follows:
1) The first thread to get there calls `jq_compile_args` which calls `jv_nomem_handler` with data=`NULL`, handler=`NULL` both of these pointers originate from the jq_state object obtained by `jq_init` which sets them to `NULL` by default.
2) `jv_nomem_handler` calls `tsd_init` via `pthread_once` which sets up the thread specific key, allocates a `nomem_handler` object and stores a pointer to it in the thread specific storage. This thread happily continues it's business...
[...]

PR #2546, opened 2023-03-01
Fix segmentation fault when using jq in threads

thalman:
In previous code nomem_handler in jv_alloc.c is called only once and therefore the structure is not initialized for second and following threads.

This leads to segmentation fault in multi-threading environment.

This patch moves initialization of nomem_handler out of the pthread_once call.

This patch fixes the  issue #2545

### DIFF
--- a/src/jq_test.c
+++ b/src/jq_test.c
@@ -2,12 +2,17 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#ifdef HAVE_PTHREAD
+#include <pthread.h>
+#endif
 #include "jv.h"
 #include "jq.h"
 
 static void jv_test();
 static void run_jq_tests(jv, int, FILE *, int, int);
-
+#ifdef HAVE_PTHREAD
+static void run_jq_pthread_tests();
+#endif
 
 int jq_testsuite(jv libdirs, int verbose, int argc, char* argv[]) {
   FILE *testdata = stdin;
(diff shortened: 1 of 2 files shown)


edwardb96:
I think this is related to #2483 and #2484 - but given it been 6 months I'm not sure I'd hold out much hope.

thalman:
@edwardb96 I see, you are right. This is the same issue. I do not mind if my patch is not selected, but what prevents merging one of those?

I'm about to include one of them to Fedora and RHEL anyway.

Tom

wader:
Hi, thanks both for the PRs, see #2305 for current state of the project

nicowilliams:
Let's make sure we review this and related issues/PRs before we release, no?  I'll see if I can find time this week.

wader:
Maybe this one should be added to https://github.com/jqlang/jq/milestones/1.7%20release? it's the list i've picked things from.

@itchyny wrote about a release plan here https://github.com/jqlang/jq/issues/2623#issuecomment-1620860311

itchyny:
@nicowilliams Thank you. Very appreciated, and I added this to the milestone.

thalman:
Thank you, guys for taking care of this issue. Let me know if you need anything from my side.

wader:
@thalman Hi, can you rebase on master? the CI should work better now

thalman:
> @thalman Hi, can you rebase on master? the CI should work better now

@wader, rebased. Thanks for the CI work!

wader:
Do you know if this is a regression? i haven't linked with libjq myself but looking at the jq code it seems to have quite a lot of code to be thread safe, but was is never really tested or it broke at some point?

Thanks for adding a test btw!

thalman:
The bug is not present in 1.5. It was introduced later on in commit 3f86e97f709d222fc79f4e7b096af40782ea8775 

I would not call that regression. It brings a new functionality that allows different `no_mem_handler` for different threads, it just did not work properly and it was not tested in more threads.

nicowilliams:
Thanks!!

PR #2547, opened 2023-03-02
add unicode `upcase`, `downcase`

### LINKED ISSUE #492: Add toUpper and toLower methods
Looking through the system, support for a toUpper and toLower might make some folks lives easier. In my case I have some keys that are upper case, some that are mixed case and others that are lower case and I'd like to use a single match on them by being able to use the same case.

liquidaty:
See https://github.com/stedolan/jq/issues/492


pkoppstein:
@liquidaty - Thanks for keeping this issue alive.

Given that jq already has ascii_upcase and ascii_downcase, wouldn't it be preferable to use the names `upcase` and `downcase` for the new filters?

Also, perhaps you could trigger a restart of the automated "checks"?

liquidaty:
thank you @pkoppstein -- absolutely, will do. Your suggested names sound fine to me.

I realized after submitting the last PR state that the autoconf did not properly work with regards to [libutf8proc](https://github.com/JuliaStrings/utf8proc). I've already done the work (outside of the PR) to handle that so that will be easy to update. 

The approach it will take is the same as for libutf8proc as it does for onigurama, which is to include it as a submodule in `modules/`, with the same options and choices and defaults. Feel free to let me know if you / other maintainers prefer a different approach.

pkoppstein:
> Feel free to let me know if you / other maintainers prefer a different approach.

What would the main “cons” be in including libutf8proc unconditionally?  Do we have any numbers at this point?

liquidaty:
> What would the main “cons” be in including libutf8proc unconditionally? Do we have any numbers at this point?

No cons in my view, unless someone had some preference to use their own version of libutf8proc or anyone wants the option to have a (very slightly) lighter-weight `jq` that did not support these functions. No numbers I'm aware of

At the moment the updated PR I'm prepping already makes it optional, defaulting to yes, and is basically ready to go (I just want to add a simple test before re-submitting), so easy to change it to unconditional

pkoppstein:
Assuming we’re ok with the license, I’d suggest NOT making its inclusion conditional, partly because the functionality that will be enabled is so central to Unicode and thus JSON, and because it’s already such a pain having even one external library optional (e.g. w.r.t. man.test).

wader:
I agree with @pkoppstein to make it non-optional as it feels fundamental. Only cons i can come up with is if it makes libjq more complicated to link? should we namespace the symbols somehow?

Nice addition!

liquidaty:
OK should be all set now, utf8proc is included unconditionally (hence no need to modify autoconf), filter names changed to `upcase` and `downcase`, couple tests have been added, license notice added, all check pass. LMK if any further changes wanted in order to merge

nicowilliams:
I've not reviewed this, but yeah, we need this!

(We probably also need normalization and form-insensitive string comparisons, but that's for another issue/PR and another time.)

liquidaty:
Actually, please do not merge this just yet. It occurs to me that the current way of unconditionally including libutf8 is to simply include the utf8proc.c into the source code. However, this may cause issues if jq is used as a library together with other code that links the utf8proc library. I will update this PR to unconditionally include libutf8 but still providing an option for the user to specify their own

nicowilliams:
> Actually, please do not merge this just yet. It occurs to me that the current way of unconditionally including libutf8 is to simply include the utf8proc.c into the source code. However, this may cause issues if jq is used as a library together with other code that links the utf8proc library. I will update this PR to unconditionally include libutf8 but still providing an option for the user to specify their own

_Can of worms opening alert_ :)

There was an issue asking for us to make sure all non-static symbols are named `jq_...` or `_jq_...` or `jv_...`.  The idea is that when _static linking_ one has no control over what non-static symbols are "exported" by libraries.  I was not, and still am not sympathetic because there is no reason that static linking should be stuck with 1980 semantics while dynamic linking has much better semantics.  But the reality is that static linking still exists, it is still stuck with 1980 semantics, and we probably still want to support it, so we probably need to do some symbol renaming.

So I'm inclined to agree that any external library sources that we fork right into `jq`'s should have symbols renamed.

Whereas if we use some external library and import it as a submodule like we did with Oniguruma, then users who want static linking would probably not use the submodules when building `jq`, and we can leave any symbol conflict issues in those submodules to the user.

So I'm further inclined to believe that if we can treat libutf8 like Oniguruma, that's probably better, and we might be able to then avoid opening this can of worms.  Though we should rename our non-static symbols that don't start with `jq_`, `_jq_`, or `jv_`.

pkoppstein:
> asking for us to make sure all non-static symbols are named jq_... or _jq_... or jv_.... 

Couldn't we just add `utf8proc_` to the list of "reserved prefixes"?

See https://juliastrings.github.io/utf8proc/doc/utf8proc_8h.html

Having recently tried and failed to make a "-static" version of jq, my sense is that the benefits of including the library unconditionally outweigh the drawbacks.  If it's good enough for Julia, it's good enough for jq.

nicowilliams:
> > asking for us to make sure all non-static symbols are named jq_... or _jq_... or jv_....
> 
> Couldn't we just add `utf8proc_` to the list of "reserved prefixes"?

Anyone building a busybox-like thing will probably want jq's internal libutf8 to not conflict with other utilities'.  Is libutf8 something we can link with and -if need be- include as a submodule, just like Oniguruma?

> See https://juliastrings.github.io/utf8proc/doc/utf8proc_8h.html
> 
> Having recently tried and failed to make a "-static" version of jq, my sense is that the benefits of including the library unconditionally outweigh the drawbacks. If it's good enough for Julia, it's good enough for jq.

True for you, and true for me.  I'm not very sympathetic to users who want to link statically, but I am willing to do a bit for them.  We could have a header that can be included to rename all the relevant symbols, say.

I really wish that the various binutils-like tools out there could learn ELF semantics for static linking...  (Specifically:

 - `.a` archive files should have a well-known `.o` whose contents is the set of relevant link-editor options used to produce the archive, such as all the `-L...`, `-l...`, `-rpath ...` options, including `global`/`local` symbol demotions from mapfiles
 - when doing a final link-edit of an executable, `ld` should
    - not require the full, flattened dependency tree to be provided on the command-line, instead
    - `ld` should learn the non-flattened dependency tree from inspecting the direct dependencies' dependencies and so on
    - `ld` should resolve symbol conflicts like dynamic linking does (pick the symbol from direct dependencies)
    - "map files"

Then we'd not have to consider these silly symbol naming issues.)

nicowilliams:
Anyways, for now I'll take the inclusion of this utf8 library and forget about symbol renaming.

nicowilliams:
Looking at the header, it looks like we can also get normalization and other Unicode functionality with this!

nicowilliams:
I've not reviewed the code in `src/builtin.c` yet, but assuming we get a review of that (possibly mine) I'm willing to just take this.  Is there any reason this is a Draft PR and not just a PR?

liquidaty:
I've updated the PR with a proposed way to thread the needle as follows (apologies for the weird path of merging master into the branch, that was a workaround for an originally ill-targeted commit in the first attempt to update the PR):
* the JQ executable is unconditionally compiled with a statically linked `libutf8proc.a` library (irrespective of ENABLE_ALL_STATIC)
* moved libutf8proc code to modules and added all of that repo into this repo as normal files
* User can specify `--with-utf8proc=prefix` configure option, else the default prefix will be searched for utf8proc.h and used if that is found, else built-in will be used
* the JQ library does _not_ internally bundle the utf8proc library. However, `Libs.private` is added to `libjq.pc`, and contains the path to the libutf8proc.a that was used

This way:
* the jq executable has no additional shared lib dependencies
* No symbol renaming is needed (libutf8proc currently does not provide a way to automatically do this, so this avoids having to apply some additional symbol renaming processing)
* the jq library can be linked without any issues via standard pkg-config using `--libs --static`

Some downsides:
* If the installation proceeds with the built-in utf8proc, then, at least as currently implemented, the location of utf8proc.a will be {srcdir}/modules/utf8proc, rather than a "normal" location such as /usr/local. This issue can be mitigated by providing notice/warning when configure is run (e.g. "Using built-in utf8proc. If you will be using jq as a library, you may want to consider installing utf8proc and using that location, which you can do by running `make -C modules/utf8proc install`") (we could even include a configure option `--with-utf8proc=install` to auto-install if we wanted)
* libjq linking is now a bit more complicated, requiring a libutf8proc.a link, than if we had bundled libutfproc with renamed symbols

As an aside, I believe this will address any issues building a `-static` version, at least insofar as they are related to libutf8proc, but please send details if not.

nicowilliams:
libjq.so needs to be a thing though (I use it at $WORK!).

nicowilliams:
I'd be OK with the UTF-8 builtins being added by the jq executable and not being in libjq, but I'd rather they be in libjq.

liquidaty:
> I'd be OK with the UTF-8 builtins being added by the jq executable and not being in libjq, but I'd rather they be in libjq.

OK, maybe we can have our cake and eat it. Updated configure.ac and Makefile.am to support `--with-utf8=bundled` which will simply include the bundled modules/utf8proc/utf8proc.c into jqlib_src, thereby adding it directly into `jq` as well as `libjq` binaries. Did not add any tests specific to this option but looks like it's working as expected by doing a `grep utf8proc_version .libs/*.*`

pkoppstein:
@liquidaty - I'm not sure what exactly is holding up merging this PR, but I noticed: "This branch has conflicts that must be resolved".

liquidaty:
OK, appears to have been the result of some other merge that took place after the last time this PR was modified, resulting in git not knowing how to merge some additions to Makefile.am. I've gone ahead and try to resolve, by keeping the changes from both; apparently the CI / dist check is failing because, from what I gather in the logs, it is trying to run `make distdir` in all SUBDIRS and no such make target exists for utf8proc. I'm not familiar with this kind of issue but will take a look at it in a bit and if it isn't too hairy, should have it resolved soon.

liquidaty:
@pkoppstein think this bad boy is good to go now

nicowilliams:
I'll try to review this sometime in the next few days, but not tomorrow, and I owe other people reviews too, and I've to create a new release key.  But I think this would be really good to have in 1.7.

pkoppstein:
@liquidaty - I couldn't find any entries for the new filters in manual.yml.  

Also, would it help the review process if the commits were "squashed"?

liquidaty:
@pkoppstein yes, apologies for not squashing, for this last flurry of changes, in some cases (like the manual.yml commit) I should have been more thorough in running the local tests first but in other cases (e.g. the prior failures from `make distcheck`), I have not figured out how to replicate the same entire set of CI checks in my dev environment (x86 macos) as are performed on github (have also tried with docker linux container but still get different distcheck results at some point). 

If there is a way to run the github ci checks in some sort of staging environment, such that intermediate failures  (and their related commits) do not need to pollute the final PR record, then I'd be happy to do that.

Alternatively, if you'd like me to clean it up and squash with e.g. `merge` and/or `rebase`, just lmk.

pkoppstein:
@liquidaty - Hopefully someone more knowledgeable than I am will chime in to answer your questions.  In this case, I think "squashing" would be helpful; if you think there's too much noise in the PR history, you could close the PR, and start afresh, couldn't you?

itchyny:
I would argue that `unicode_` prefix would help users understand the difference against `ascii_*`.

There are use cases that we really want to avoid unicode case conversions. Non-prefixed filters will bring us some review comments such as _Since upcase converts cases of unicode characters, in this case we should use ascii_upcase_.

Makefile.am
```diff
@@ -14,7 +14,7 @@ LIBJQ_SRC = src/builtin.c src/bytecode.c src/compile.c src/execute.c    \
         src/jv_dtoa.c src/jv_file.c src/jv_parse.c src/jv_print.c       \
         src/jv_unicode.c src/linker.c src/locfile.c src/util.c          \
         src/decNumber/decContext.c src/decNumber/decNumber.c            \
-        src/jv_dtoa_tsd.c                                               \
+        src/jv_dtoa_tsd.c 			                        \
```
itchyny, line 17: [nitpick] Unexpected tab spaces.

Makefile.am
```diff
@@ -175,13 +175,37 @@ endif
...
+AM_CPPFLAGS += -I$(srcdir)/modules/utf8proc
+
+$(builddir)/libutf8proc.a:
+	$(MAKE) $(AM_MAKEFLAGS) -C $(srcdir)/modules/utf8proc builddir=$(shell cd $(builddir) && pwd) libutf8proc.a
+else
+jq_LDADD += ${utf8proc_LIBS}
+endif
+endif
+
 CLEANFILES += jq.1
```
itchyny, line 206: [nitpick] Put this line above the added lines.

pkoppstein:
@itchyny wrote:
> I would argue that unicode_ prefix would help users understand the difference against ascii_*.

As the person who (implicitly) suggested not using a prefix, let me explain why.

The primary reason is that the `ascii_` prefix was used to emphasize that the functionality was limited to ASCII case conversion.  The new functions have no such restrictions, and indeed (so far as I'm aware), really no restrictions within the relevant realm (i.e. of JSON strings).  The absence of a prefix nicely conveys that idea as well.  

The secondary reason, which is really much less important and which you may wish to ignore, is that when the `ascii_*` case conversion functions were introduced, it was because we were not yet ready to support unrestricted case conversion functions, so we wanted to reserved the `upcase`/`downcase` names for the future, making it possible for the `ascii`-prefixed functions to be deprecated, leaving just the `upcase`/`downcase` functions.

Another consideration is that the specific prefix `unicode_` makes the whole thing rather long and tedious.

Makefile.am
```diff
@@ -175,13 +175,37 @@ endif
...
+if BUILD_UTF8PROC
+BUILT_SOURCES += $(builddir)/libutf8proc.a
+CLEANFILES += $(builddir)/libutf8proc.a
+jq_LDADD += $(builddir)/libutf8proc.a
+
+AM_CFLAGS += -I${srcdir}/modules/utf8proc
+AM_CPPFLAGS += -I$(srcdir)/modules/utf8proc
+
+$(builddir)/libutf8proc.a:
+	$(MAKE) $(AM_MAKEFLAGS) -C $(srcdir)/modules/utf8proc builddir=$(shell cd $(builddir) && pwd) libutf8proc.a
```
itchyny, line 200: Using non-POSIX feature may cause portability problems. Hope FreeBSD users could test this.
```
Makefile.am:196: warning: shell cd $(builddir: non-POSIX variable name
Makefile.am:196: (probably a GNU make extension)
```

itchyny:
I would like to hear Turkish people to their opinion about `upcase | downcase` isn't an identity operation for their lower case alphabets.
```sh
 $ ./jq -n '"cgiosu", "çğıöşü" | (upcase | downcase) == .'
true
false
```

pkoppstein:
@itchyny wrote:
> upcase | downcase isn't an identity operation for their alphabet [Turkish]

Consider also the case of `mu` and `micron`:

```
./jq -ncM '"µ" | ( . == (upcase | downcase))'
false
```

Should the fact that `upcase` is not invertible be mentioned in the manual?  (@liquidaty)

---- 

```
./jq -ncM '[181,956] | implode'
"µμ"
```

nicowilliams:
> I would like to hear Turkish people to their opinion about `upcase | downcase` isn't an identity operation for their lower case alphabets.
> 
> ```shell
>  $ ./jq -n '"cgiosu", "çğıöşü" | (upcase | downcase) == .'
> true
> false
> ```

This is a well-known problem.  The solution to this is to have the `tolower`/`toupper` functions be locale-aware: either use the current locale, or have an explicit locale input.  Thus `i` should upcase to `İ` and `İ` should downcase to `i` in a Turkish locale, but otherwise `i` should upcase to `I`, and `İ` should downcase to `i`.  (Or something like this -- I don't remember the exact capitalization rules here.

There is no other solution.

This also applies to sorting -- collations need to be implicit from the locale or explicitly requested.

The same sort of consideration does not apply to normalization, fortunately.

In the interest of being able to make progress, I would advise that we take on Unicode functionality piecemeal, designing functions that don't paint us into corners.  In general we can add more arguments to specify locales explicitly, so I don't think we're painting ourselves into corners.

So I'd go ahead with this PR, though with a note in the docs and release notes about this particular issue and any others that we can identify, and that we open issues for these so that we can track them in the future.

Besides, there's a lot more to do than to add Unicode tolower/toupper functions:

 - normalization
 - form-insensitive string comparison
 - form-insensitive string hashing (for object keys)
 - iteration by code unit (we don't have this), codepoint (we have this with `explode` and string slices), character (we don't have this), and glyph (we don't have this either)
 - table of glyph width and other attributes (e.g., color, for emoticons)
 - etc.

emanuele6:
I think you meant that `(upcase | downcase)` is not the same as just `downcase`, `(upcase | downcase)` is not an identity even when only using ascii character (e.g. `"aBa"`). ^^

Anyway interesting, I knew the turkish alphabet had a dotted `i` (ascii U+0069) and an undotted `ı` (U+0131), but I didn't know that for the turkish locale, they use a special `İ` (U+0130) character for the uppercase version of dotted i, and the regular ascii `I` (U+0049) for the uppercase version of non-ascii undotted `ı`. Fun!

itchyny:
> ./jq -n '"cgiosu", "çğıöşü" | (upcase | downcase) == .'

That wouldn't be an issue that this conversion isn't identity for some cases. But the point is In Turkish the upper case of `i` is `İ` and the lower case of `I` is `ı`. I'm reading the Unicode Specification Section 1.18 Case Mapping to learn how this topic is difficult.

pkoppstein:
> Fun!

Yes, but I don't think any of this bears on jq's `upcase`/`downcase` naming or functionality, because as e.g. RFC 4627 says, in JSON:

> A string is a sequence of zero or more Unicode characters

No doubt there should, one day, be LOCALE-aware case-conversion functions, but I think it would be preferable if they were accordingly named to indicate as much, e.g. perhaps with the "locale_" prefix.  And/or we could have upcase/1 and downcase/1 to avoid dependence on environment variables.

itchyny:
Sorry for raising the discussion, I just wanted to express my concern about the conversion ambiguity. As a user with language with no alphabet cases, the case conversion topic is very unclear to us, and hopefully the manual is clear to refer to well-defined document about the conversion. In general, we tend to avoid such features to avoid unexpected conversions, so that's why I think it should be clear that they are named with `unicode_` prefix.

pkoppstein:
From Unicode Specification Section 1.18:

> To make case mapping language sensitive, the Unicode Standard specificially [sic] allows implementations to tailor the mappings for each language, but does not provide the necessary data. 

So perhaps (one day) we could have `downcase({"locale": "tr"})` # per ISO 639-1
etc

pkoppstein:
@itchyny wrote:
> hopefully the manual is clear to refer to well-defined document about the conversion.

Yes, hopefully @liquidaty will update the manual, hopefully with pointers to the specific standards document(s) and/or links to the relevant utf8proc pages. 

I believe two relevant citations are:

(1) The file: https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt

(2) The explanations of the relevant fields in https://www.unicode.org/L2/L1999/UnicodeData.html

For the uppercase mapping, the latter explains:

> 12 | Uppercase mapping | informative | Upper case equivalent mapping. If a character is part of an alphabet with case distinctions, and has an upper case equivalent, then the upper case equivalent is in this field.

nicowilliams:
> > Fun!
> 
> Yes, but I don't think any of this bears on jq's `upcase`/`downcase` naming or functionality, because as e.g. RFC 4627 says, in JSON:

You must mean RFC 8259 ;)

> > A string is a sequence of zero or more Unicode characters

But it says nothing about functions in jq or any programming language.  Saying "Unicode" is enough to imply the full panoply of Unicode semantics for anyone who dares to implement them :)

> No doubt there should, one day, be LOCALE-aware case-conversion functions, but I think it would be preferable if they were accordingly named to indicate as much, e.g. perhaps with the "locale_" prefix. And/or we could have upcase/1 and downcase/1 to avoid dependence on environment variables.

We don't need to do that.  We can add arguments and be done, just as we have `match/1` and also `match/2` and not `match/1` and `mode_match/2`.

nicowilliams:
> Sorry for raising the discussion, [...]

Please don't be.  It was actually a very good point you raised.  EDIT: And, really, I should have noticed it myself earlier, as I have been keenly aware of the need for locale-aware case conversion functionality.

> [...] I just wanted to express my concern about the conversion ambiguity. As a user with language with no alphabet cases, the case conversion topic is very unclear to us, and hopefully the manual is clear to refer to well-defined document about the conversion. In general, we tend to avoid such features to avoid unexpected conversions, so that's why I think it should be clear that they are named with `unicode_` prefix.

I don't think we need a `unicode_` prefix unless implementation details force us to have a build configuration option to exclude this functionality (but we don't have that problem).

jq deals in JSON.  JSON deals in Unicode.  If we want to have tolower/toupper functions only for ASCII, then those should get a prefix `ascii_`.

If we're going to have more Unicode functionality than 1.6 had, it should feel natural.  That's my opinion anyways.

nicowilliams:
> Fun!

That's the best way to think of it.  Many people look at the complexity of Unicode and curse the Unicode Consortium, but the reality is that most of the complexity in Unicode arises naturally from the complexity of the scripts we use for human languages.  Even w/o pre-compositions Unicode would still need normalization.  Even with the best attempts to unify common glyphs in scripts, Unicode would still have confusables (just as plain old US-ASCII does).  These difficulties are natural, and if we see them as challenges, then they can be just that: fun.  Though I'm not proposing that jq contributors and maintainers actually write low-level Unicode functionality!  Just that we build jq interfaces that make sense for dealing with Unicode.

liquidaty:
@nicowilliams @itchyny @pkoppstein I have no strong opinion on any of these matters so please go ahead and modify however you see fit, and/or lmk if you prefer that I do so

pkoppstein:
@liquidaty wrote:
> lmk

I'm not familiar with the utf8proc documentation, so I'm not sure what the best way to update manual.yml so that it's easy to find the relevant descriptions, specifications, and/or caveats.  In my opinion, it would be appropriate to mention that we're using the one-to-one maps defined by Unicode, and that round-tripping is not guaranteed by these maps.

src/builtin.c
```diff
@@ -523,6 +523,89 @@ static jv f_utf8bytelength(jq_state *jq, jv input) {
...
+  if(jv_get_kind(input) != JV_KIND_STRING) {
+    return input;
+  } else {
+    const char *jstr = jv_string_value(input);
+    size_t len = jv_string_length_bytes(jv_copy(input));
+    if(len) {
+      int malformed = 0;
+      unsigned char *changed = changeCase((const unsigned char *)jstr, &len, &malformed);
+      if(malformed || !len)
+        free(changed);
```
leonid-s-usov, line 585: shouldn't we report an error here? or a warning?

liquidaty: I don't think so. A better and more flexible alternative would be to add two built-ins along the lines of `utf8_valid` (return true iff input is valid utf8) and `utf8_replace_malformed` (replace any malformed utf8 in some specified manner). In any case it seems to me this issue is more related to https://github.com/jqlang/jq/issues/2660 than to these particular `upcase` and `downcase` functions.

liquidaty: That said I don't really feel strongly and defer to whatever you maintainers decide

Makefile.am
```diff
@@ -175,13 +175,37 @@ endif
 jq.1: jq.1.prebuilt
 	$(AM_V_GEN) cp $(srcdir)/jq.1.prebuilt $@
 
+SUBDIRS =
+
+AM_CFLAGS += ${utf8proc_CFLAGS}
+
+if BUNDLE_UTF8PROC
+LIBJQ_SRC += modules/utf8proc/utf8proc.c
```
itchyny, line 187: `modules/utf8proc/**` should be marked as vendored in .gitattributes (ref: #2705)

docs/content/manual/manual.yml
```diff
@@ -1726,6 +1726,17 @@ sections:
...
+      - title: "`downcase`, `upcase`"
+        body: |
+
+          Emit a copy of the input string with its characters (unicode) converted to the
+          specified case.
+
+        example:
+          - program: 'upcase'
+            input: '"useful for é"'
+            output: '"USEFUL FOR É"'
```
itchyny: `output` of manual example needs to be an array (make sure to `make tests/man.test`)

liquidaty: I was wondering why they did not show up in jq.1. These examples were just copied from the existing examples for `ascii_downcase` and `ascii_upcase`, so I suppose this comment would apply to those as well?

itchyny: Yes, those examples should be fixed as well.

emanuele6: Those `ascii_upcase`/`ascii_downcase` examples as well as a bunch of other examples are not even generated because tehy are using `example:` instead of `examples:`.

You should put examples in the `examples:`, not `example:`

liquidaty: is there any definition of the schema that manual.yml should conform to? If so (or if not, and it's easy to create), I will run a validator on it to check for any other issues (and the validation steps would be available for future reuse as well...)

src/builtin.c
pkoppstein: @leonid-s-usov, @itchyny, @nicowilliams, @emanuele6 - I'm inclined to think that it's too much to expect @liquidaty to resolve the larger issue (#2660) here, and would therefore propose that the particular issue regarding `malformed` not stand in the way of this PR being accepted, though I think it would be appropriate to add documentation to the manual regarding the resolution of the anomaly, even if it's only to say the behavior is "undefined".

While on the subject of handling anomalies - is there a precedent that should be followed regarding "warning" messages on stderr?

nicowilliams: #2660 is out of scope here.  Importing utf8proc will help resolve #2660, for sure, but #2660 has a lot to think about.

docs/content/manual/manual.yml
itchyny: I added a schema for manual files in #2745, and validate both on building the manual locally and on GitHub Actions.

pkoppstein:
@liquidaty - I hope you haven't run out of steam!   Is there some way I can help push this over the finishing line?  It would be a great shame if all your work didn't make it into jq 1.7.

liquidaty:
@pkoppstein have been getting pulled away but will take a look this week, hopefully on the earlier side (I'm also entirely fine with maintainers making any of the contemplated changes, in case anyone prefers that which sometimes is less work than chasing the change to begin with...)

pkoppstein:
@liquidaty - I noticed that you recently merged `master` into the PR (https://github.com/jqlang/jq/pull/2547/commits/88a102f5f88d1547a51e0ceb092dfdbdad66334e).  My understanding is that the jq administrators have a very strong preference for rebasing.  That's just a "heads-up" as I would not hazard to say what your best option at this point would be.

liquidaty:
My apologies, 
I am new to multi-external-party merging with GitHub and I thought that was the way we were "supposed" to resolve the merge conflicts. I definitely tried to explore whether the GitHub UI provided alternative ways, and afaict GitHub only seemed to offer that 1 and only 1 way for me to resolve the conflict (i.e. click "Resolve", make changes, click "Save")

I assume from your note however that there are other ways that I'm not aware of. I will have to ask for assistance next time the "Resolve conflict" button pops up for me then, as I don't know how else to resolve. If there are further steps I should take at this time to get it to the way you need/want it to be, please feel free to lmk.

nicowilliams:
@liquidaty just do this in the future `git pull --rebase` or `git fetch origin && git rebase origin/master`.

pkoppstein:
@nicowilliams - What would you suggest @liquidaty do to undo the damage?   Start a new branch?

nicowilliams:
> @nicowilliams - What would you suggest @liquidaty do to undo the damage? Start a new branch?

No need to start a new branch.  Just `git pull --rebase`, or `git fetch origin; git rebase origin/master`.  I also wrote a lengthy comment in a reply to you (I think on https://github.com/jqlang/jq/pull/2731) about this same sort of issue on another PR, and linked to [a gist I wrote about git](https://gist.github.com/nicowilliams/a6e5c9131767364ce2f4b3996549748d).

liquidaty:
OK I can do that. I also though it preferable to consolidate all the changes for this PR into the following 3 chunks. that sound ok?

I used the following commands to rebase:
```
git fetch origin pull/2547/head:pr2547
git checkout pr2547
git rebase -i origin/master
# do the rebase-- see below for end result

#### if all is ok, the next cmds will be:
  git remote add liquidaty https://github.com/liquidaty/jq
  git remote set-url --push liquidaty git@github.com:liquidaty/jq
  git push -f liquidaty HEAD:2547-resolve-jq.test
####
```

and the rebase consolidated the commit history into:

```
commit 235cd5e0c69e3a9b0e1d53235752df9dc968d469 (HEAD -> pr2547)
Author: liquidaty <info@liquidaty.com>
Date:   Sat Jul 15 10:37:10 2023 -0700

    add upcase/downcase to manual.yml

    passing tests

    update manual.yml to correct `example` vs `examples`

    updated jq.1.prebuilt from prior commit

    updated tests/man.test

    restore modules/oniguruma, from prior inadvertent modification and commit

commit c6830a7036d072351a87c4cd914d8a3bf19626e2
Author: liquidaty <info@liquidaty.com>
Date:   Sun Jul 9 07:27:35 2023 +0200

    add --with-utf8proc=bundled option

commit 290ef0d20c7b3d1435efa2b6469d2c4ddcee594b
Author: liquidaty <info@liquidaty.com>
Date:   Thu Jun 9 14:15:29 2022 -0700

    add utf8tolower, utf8toupper

    change new filter names to `upcase` and `downcase`
    add one simple test to jq.test for each
    include utf8proc files directly in repo as an unconditional part of the build
    update COPYING to include utf8proc license

    statically link utf8proc
    allow custom utf8proc location; search default prefix if not found, else use builtin
    omit utf8proc from libjq
    changed original modules/utf8proc/Makefile to omit checks that require ruby to be installed

```
(I'm still figuring out how to clean up the multi-line comments...)

nicowilliams:
Tomorrow I'll try to rebase and clean up the branch's history and force-push to the PR, and I'll include a log of what I did.

src/builtin.c
leonid-s-usov: ack

pkoppstein:
@nicowilliams - I'm looking at the revised documentation for `encoding`, which says simply:

    Outputs the encoding set with `encodeas`.

This is a bit too terse.  First, it makes no mention that the input is expected to be a string; more important, it suggests that e.g. `"abc" | encoding` would be an error or empty.  

So perhaps the text should read s.t. like:

     Assuming the input is a string, it outputs its encoding as set with `encodeas` or else  `"UTF-8"`.

ryenus:
Possible to include this in the upcoming 1.7 release? 
Hopefully not too late, as it's been a few years since 1.6 :-)

liquidaty:
@itchyny @nicowilliams @pkoppstein et al: great job re release 1.7.

Coming back to this issue: I had (apparently mistakenly) thought it would be in 1.7 given the prior discussion and the state of the pull which afaik is totally ready to go. Could we put this issue on a definitive roadmap to be merged?

pkoppstein:
@liquidaty - Rebasing might be necessary.

@itchyny - Once this PR has been prepared for merging, could you please help ensure it’s merged before it becomes outdated again?  Many of us have been looking forward to the inclusion of this functionality in a master version of jq, and now that 1.7 has been released, this would be an excellent time, not least because it would allow for any imperfections to be resolved calmly. Thank you.

bitsondatadev:
ufff, this one is soooo close. 

Great work @liquidaty! Hopefully a maintainer can get some time to bash out the last few changes needed.

liquidaty:
@bitsondatadev it was fully ready to go well before 1.7 and at this point I'm resigned to just using my own modified version of jq and not waiting for it to ever be officially merged. I'd be happy to get it in release-ready shape again if the maintainers are willing to provide some indication that the effort won't just be ignored, but otherwise I'm probably just never going to use anything past my own modified version 1.6

wader:
Hello i did a rebase, squash, cleanup and resolved a bunch of review comment and pushed a variant of this here https://github.com/wader/jq/commits/2547-resolve-jq.test-cleanup/ should i push it to the PR branch?

liquidaty:
FWIW I don't believe it's my decision to make but if I'm wrong about that then yes please

wader:
> FWIW I don't believe it's my decision to make but if I'm wrong about that then yes please

👍 could you take a quick peek before i push to see if things looks sane? i'm mostly concern about configure.ac and Makefile.am which has quite a lot of messy conflicts.

liquidaty:
sure will do. imv the configure.ac changes generally should be fine if it still passes all tests; a separate related question would be, do we want to add more tests to verify the different options that configure.ac now offers related to this feature? I generally prefer a "don't write it if you don't use / test it" principle, but in this case I would override that and vote "no" for various reasons I won't expand on unless folks disagree...

wader:
TODO:
- [ ] Update to latest utf8proc https://github.com/JuliaStrings/utf8proc and maybe make sure no changes are needed to build to make it easier to update in the future. I think current vendored does some changes to use ${builddir}?
- [ ] Different link modes: builtin or external utf8proc, bundled or dynamic for libjq? more?
- [ ] Make sure `make dist` works. Wonder if this should be some kind of CI test?
- [ ] More tests?

git-developer:
Is this PR alive?

liquidaty:
@wader apologies for the belated response. looks good to me. thank you!

wader:
@liquidaty no worries!

@git-developer I think the PR it stalled/blocked on the TODOs i mentioned above. Someone has to do some work/decide about those points. I'm leaning towards somehow embedding the utf8proc source for simplicity, less cmake, `make dist` and submodules mess. But in that case we should probably somehow prefix all utf8proc symbols so that someone using libjq and their own utf8proc code would not get a symbol conflict.

liquidaty:
@itchyny @pkoppstein @nicowilliams @wader so now this isn't going to make it into 1.8 either? The full feature code has been available for over 2 years now complete with tests. The actual changes are minimal. Many users have asked for this. As other core maintainers have pointed out, the utf8proc lib brings other benefits for other purposes as well. In light of that, I hope you will reconsider, but I suppose I can live with my custom 1.6 forever if not.

PR #2563, opened 2023-03-30
Enable non-single-line mode

weeble:
Regular expressions are always evaluated in single line mode. This is because Oniguruma's PERL_NG syntax defaults to single line mode and must instead be forced _out_ of it by passing ONIG_OPTION_NEGATE_SINGLELINE. By changing this we now default to non-single line mode, and can opt in to single line mode as desired.

Ref: https://github.com/stedolan/jq/issues/2562

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -887,6 +887,9 @@ static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
 
   jv_free(modifiers);
 
+  if (!(options & ONIG_OPTION_SINGLELINE)) {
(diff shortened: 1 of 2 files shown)


wader:
Hi, thanks for look into this. jq's maintenance is a bit in limbo at the moment, see #2305

jsoref:
This seems like a fairly sizable behavior change. (The referenced issue is more or less resolved, this repository is now managed by the jslang team.) As yet, I don't think jslang has started talking about compatibility promises. I'd personally expect things not to change incompatibly within jq 1.x, but I haven't checked carefully to ensure that it has never broken a behavior.

PR #2564, opened 2023-03-30
Attempt to fix gsub with zero length matches

weeble:
Problem: If a regex given to gsub can match a zero length string, it will never terminate and eventually crash. This is because after matching once it keeps checking for the same match at the same location and finding it.

Solution: We can't skip ahead because there might be a non-zero length match at the same location. Instead, keep track of whether the last match was zero length, and when it was, ignore any zero length match at the same offset.

Remaining problem: This fails one of the new tests I added. For some reason, match doesn't seem to correctly return (some) zero length matches at the end of the input string. I'm going to report that as a separate bug, I can't see why it behaves like that.

Ref: https://github.com/stedolan/jq/issues/2148

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -132,20 +132,30 @@ def sub($re; s; flags):
   def subg: [explode[] | select(. != 103)] | implode;
   # "fla" should be flags with all occurrences of g removed; gs should be non-nil if flags has a g
   def sub1(fla; gs):
-    def mysub:
+    def find_first_new_match($lastzero):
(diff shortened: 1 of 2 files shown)


wader:
Hi, thanks for look into this. jq's maintenance is a bit in limbo at the moment, see #2305

weeble:
Thanks for the heads-up!

itchyny:
The untermination issue of gsub with empty matches has been fixed by #2641.

PR #2571, opened 2023-04-15
docs: parenthesize key expression to compile

maxwell-k:
Before this change the example does not compile:

```
jq: error: syntax error, unexpected ':', expecting '}' (Unix shell quoting issues?) at <top-level>, line 1:
"f o o" as $foo | "b a r" as $bar | {$foo, $bar:$foo}                                               
jq: error: May need parentheses around object key expression at <top-level>, line 1:
"f o o" as $foo | "b a r" as $bar | {$foo, $bar:$foo}                                     
jq: 2 compile errors
exit status 3
```

After this change the output is produced as expected.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -599,9 +599,8 @@ sections:
           If the keys are "identifier-like", then the quotes can be left
           off, as in `{a:42, b:17}`.  Variable references as key
           expressions use the value of the variable as the key.  Key
-          expressions other than constant literals, identifiers, or
-          variable references, need to be parenthesized, e.g.,
-          `{("a"+"b"):59}`.
+          expressions other than constant literals or identifiers need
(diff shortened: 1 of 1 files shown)


emanuele6:
That doesn't work only because you are using `jq 1.6` (the latest official release).

This works without `()` in the version on the git repo:
```bash
$ ./jq -n '"f o o" as $foo | "b a r" as $bar | {$foo, $bar:$foo}'
{
  "foo": "f o o",
  "b a r": "f o o"
}
```

So I don't think it makes much sense to update the documentation on the git repository for this.

Also note that this example is not present on the documentation for jq 1.6, only on git-master's documentation.

maxwell-k:
Thanks for pointing out my mistake @emanuele6, I appreciate your help.

You're right I was using the latest release and reading the manual for the development version.

PR #2582, opened 2023-05-08
docs: Update Flex URL

t-dub:


jsoref (approved):
I'm not an org member, but as of today, https://web.archive.org/web/20230523211131/https://flex.sourceforge.net/ says:

> The project repository is now hosted at github: [https://github.com/westes/flex](https://web.archive.org/web/20230523211131/https://github.com/westes/flex).

owenthereal:
Thanks for the contribution @t-dub and thanks for reviewing it @jsoref!

pkoppstein:
@owenthereal, @itchyny - it’s only a minor issue, but this change, even though well-intentioned, is actually incorrect. The original construction depends on a subtlety of English that might easily seem incorrect to non-native English speakers. The larger point here is that the jq documentation is generally written in a terse style with a view to correctness, so triple-checking a proposed change is often advisable.

itchyny:
@pkoppstein You're talking about #2588, not this PR?

pkoppstein:
@itchyny - yes,#2588. Not sure how my comment ended up in the wrong place. Tx.

PR #2588, opened 2023-05-22
Typo in basic filters manual

bew:


jsoref (approved):
(I'm not a member of the org)

I agree with this change, although technically it isn't a typo.

owenthereal:
Thanks for the contribution @bew and thanks for reviewing it @jsoref!

PR #2603, opened 2023-06-05
Add a GitHub Actions workflow to build and update website

itchyny:
I created a GitHub Actions workflow to automate website updates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,9 @@ m4/ltversion.m4
 m4/lt~obsolete.m4
 tests/*.trs
 
+# Docs output
+docs/output
+
 cscope.in.out
 cscope.out
 cscope.po.out
--- a/scripts/update-website
+++ b/scripts/update-website
@@ -1,16 +1,17 @@
-#!/bin/sh
+#!/bin/bash
 
 # This script builds the website from the docs directory of
 # the current branch and copies it over to the gh-pages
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -47,6 +47,9 @@ m4/ltversion.m4
 m4/lt~obsolete.m4
 tests/*.trs
 
+# Docs output
+docs/output
+
```
itchyny, line 52: We need to move this here to avoid adding docs/output in gh-pages branch.

scripts/update-website
```diff
@@ -20,7 +22,7 @@ cp -r docs/output/* "$tmpdir"
 cp .gitignore "$tmpdir"
 
 # copy to gh-pages
-git checkout gh-pages
+git switch gh-pages
```
jsoref: I'd argue against this. As of `Git 2.40.1`, git-switch is experimental and thus subject to change when `git` changes. Since you're using `ubuntu-latest`, you don't have any guarantees about the version of `git` and thus the behavior of this command may change randomly at some unexpected time.

jsoref (changes requested):

scripts/update-website
itchyny: Reverted, thanks.

wader (approved):

PR #2604, opened 2023-06-05
Improve docs of optional indexing syntax to make it clear (ref #2588)

itchyny:
This reverts commit 908f5d54a7897bc31c85ddcb9737463c1df30faf.

Reason: https://github.com/jqlang/jq/pull/2582#issuecomment-1576085857.

### DIFF
--- a/docs/content/manual/v1.6/manual.yml
+++ b/docs/content/manual/v1.6/manual.yml
@@ -332,7 +332,7 @@ sections:
       - title: "Optional Object Identifier-Index: `.foo?`"
         body: |
 
-          Just like `.foo`, but does not output an error even when `.`
+          Just like `.foo`, but does not output even an error when `.`
           is not an array or an object.
 
         examples:
--- a/docs/content/manual/v1.6/manual.yml
+++ b/docs/content/manual/v1.6/manual.yml
@@ -332,8 +332,8 @@ sections:
       - title: "Optional Object Identifier-Index: `.foo?`"
         body: |
 
-          Just like `.foo`, but does not output even an error when `.`
-          is not an array or an object.
+          Just like `.foo`, but does not output an error when `.` is not an
+          object.
 
         examples:
           - program: '.foo?'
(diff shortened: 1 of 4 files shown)


bew:
> The original construction depends on a subtlety of English that might easily seem incorrect to non-native English speakers.

As a non-native English I definitely don't understand what you're talking about.

But I'd argue that using subtle English phrasing might not be a good idea for a documentation that is to be read by many people, probably mostly non-native.

docs/content/manual/v1.6/manual.yml
```diff
@@ -332,7 +332,7 @@ sections:
       - title: "Optional Object Identifier-Index: `.foo?`"
         body: |
 
-          Just like `.foo`, but does not output an error even when `.`
+          Just like `.foo`, but does not output even an error when `.`
```
jsoref: I'm a native speaker and this construction is lousy. Just dropping the word would be clearer

```suggestion
          Just like `.foo`, but does not output an error when `.`
```

itchyny:
Waiting for comments from @pkoppstein, but I want to revert the changes anyway. Documentation fixes like this should be applied to all versions.

jsoref:
> But I'd argue that using subtle English phrasing might not be a good idea for a documentation that is to be read by many people, probably mostly non-native.

I'd argue straight out that it's a horrible idea for exactly this reason. Documentation should be clear, concise, and unambiguous. Using `even` here fits none of those criteria.

owenthereal:
👍 to dropping the`even`. This revert wouldn't make the doc more straightforward.

docs/content/manual/v1.6/manual.yml
itchyny: Done 87c657ddc5a1ffaae866493b638f6ca7cc3b1c37

itchyny:
Noticed the original sentence ```when `.` is not an array or an object.``` is a mistake, so I also fixed it. `.foo` output an error if `.` is an array, and `.foo?` suppresses the error. That's Optional **Object** Identifier-Index.

PR #2608, opened 2023-06-11
Fix version scripts for some shallow clone situations (fix #1549)

itchyny:
I noticed that in some shallow clone situations `autoreconf -fi` fails on detecting the version. In some situations tags are reachable but `git describe` fails. Further reference: https://github.com/jqlang/jq/pull/1549#issuecomment-1585987160. Resolves #1549.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 m4_define([jq_version],
-          m4_esyscmd_s([(git rev-parse --verify -q jq-1.0 > /dev/null &&
-                        (git describe --tags --dirty --match 'jq-*'|sed 's/^jq-//')) ||
+          m4_esyscmd_s([git describe --tags --match 'jq-*' >/dev/null 2>&1 &&
+                        git describe --tags --match 'jq-*' --dirty | sed 's/^jq-//' ||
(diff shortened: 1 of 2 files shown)


wader (approved):
Based on your comment in #1549 it looks good it think. Wonder if it would be possible to use `scripts/version` for `jq_version`?  maybe for the future if so

btw i noticed this while poking around how the verison string is generated https://github.com/jqlang/jq/blob/master/docs/site.yml#L7 the `-dirty` should probably not be there?

itchyny:
The version variable in `site.yml`  is not used anymore so we can clean up including the default template and `Makefile.am`. The commit https://github.com/jqlang/jq/commit/81f328f1aa153462d67bc79e97e35c98960cf61b#diff-45b410e29e049ce4fc88cfc7a4e4c518107fe97593820edcf316591536f202c7L54 removed the `$JQ_VERSION` placeholder from the manual source.

itchyny:
Is there any objection to use `scripts/version` in `configure.ac`?

wader:
I think it makes sense if they use the same version format already?

wader (approved):

itchyny:
Thank you.

PR #2610, opened 2023-06-13
scanbuild: use ubuntu-latest

jsoref:
As is, this workflow fails to run because ubuntu-18.04 is no longer available


itchyny (approved):
Thank you!

PR #2614, opened 2023-06-14
manual.yml - Perl NG

pkoppstein:
Explain which "flavor" is used. Also drop reference to Ruby as Ruby has moved on.  Also explain that certain special characters must  be escaped in REGEX.


itchyny (approved):
LGTM!

pkoppstein:
@itchyny - Tx.  What’s the next step?

itchyny:
@pkoppstein You can merge by yourself.

pkoppstein:
@itchyny - GitHub said I did not have permission to merge, without giving any further details. The permission system seems to be insanely complex, at least in its full glory.

Issue #2354, opened 2021-10-03
gsub lookahead cannot allocate memory


stoat1:
**Describe the bug**
When using regex lookahead `jq` freezes, eats up all system memory, prints an error message and eventually aborts

**To Reproduce**

    $ jq -n '"qux" | gsub("(?<=u)"; "u")'
    "quux"

    $ free -g
                  total        used        free      shared  buff/cache   available
    Mem:             15           5           6           0           3           8
    Swap:             1           0           1

    $ jq -n '"qux" | gsub("(?=u)"; "u")'
    error: cannot allocate memory
    [1]    123456 abort (core dumped)  jq -n '"qux" | gsub("(?=u)";  #"u")'

**Expected behavior**
`jq` outputs the result and exit

**Environment**
 - Ubuntu 20.04.2 LTS
 - jq-1.6

**Additional context**
I'm aware that this may be an issue in Oniguruma but I wasn't able to reproduce it due to my lack of experience with C

pkoppstein:
I was not able to reproduce the problem on a Mac OS, though it's interesting that gojq doesn't like the RE and prints it out in a curious way:
```
for jq in jq1.5 jq1.6 jqMaster gojqMaster ; do
     $jq --version
    echo gsub
    $jq -n '"qux" | gsub("(?<=u)"; "u")'
    echo sub
    $jq -n '"qux" | sub("(?<=u)"; "u")'
    echo
done
```
Output:
```
jq-1.5
gsub
"quux"
sub
"quux"

jq-1.6
gsub
"quux"
sub
"quux"

jq-1.6-129-g80052e5-dirty
gsub
"quux"
sub
"quux"

gojq 0.12.5 (rev: 4ada50f/go1.17)
gsub
gojq: invalid regular expression "(?P<=u)": error parsing regexp: invalid named capture: `(?P<=u)`
sub
gojq: invalid regular expression "(?P<=u)": error parsing regexp: invalid named capture: `(?P<=u)`

```

stoat1:
@pkoppstein thank you for your findings.

I'd like to add that the same regex works as expected in tools using another regex flavor, e.g.

    $ awk 'gsub("^", "")' <<<a
    a

    jshell> "a".replaceAll("^", "")
    $1 ==> "a"

    > console.log('a'.replaceAll(/^/g, ''))
    a

    $ python -c '
    import re
    print(re.sub("^", "", "a"))
    '
    a

Issue #1425, opened 2017-06-06
gsub  breaks \b for checking word boundaries


thedward:
When I run the following program with `jq -n`:

```
"hello world"|"\\b(?<i>[[:lower:]])" as $rx|gsub($rx;.i|ascii_upcase)
```

Expected Output:

> `"Hello World"`

Actual Output:

> `"HELLO WORLD"`

I've got a workaround for my particular use case:

```
"hello world"|"\\b(?<fst>[[:lower:]])(?<rst>[[:lower:]]*)\\b" as $rx|gsub($rx;"\(.fst|ascii_upcase)\(.rst)")
```

pkoppstein:
@thedward - As you've discovered, jq's interpretation of the "g" in gsub is currently different from the one you have in mind.  Here's one way to achieve the effect you want:
 
    def capitalize:
      def c: sub("\\b(?<i>[[:lower:]])";.i|ascii_upcase);
      c as $c | if . == $c then $c else ($c|c) end;

Example:

    "hello world" | capitalize

thedward:
It's also currently different than documented. :-) 

I'm trying to write a correct implementation — if I succeed, I'll make a pull request.

itchyny:
I'm uncertain but this looks the same instance of `gsub` bugs, like #1206, #2148.

pkoppstein:
@itchyny wrote:
> I'm uncertain but this looks the same instance of gsub bugs, like https://github.com/jqlang/jq/issues/1206, https://github.com/jqlang/jq/issues/2148.

With respect to the example given at the top of this thread, please note that gsub is working as intended, that is, in accordance with the idea that `gsub(re; _)` should repeatedly apply the transformation specified by `sub(re;_)` to non-overlapping substrings of the original input from left to right, until quiescence.  That's not to say that this "specification" is desirable, but it is defensible.

thedward:
@pkoppstein where is this behavior specified?

pkoppstein:
@thedward - I wrote `this "specification"` to make it clear that I was referring to the previous sentence.  
If you think that the existing documentation of gsub precludes the interpretation as iterative application, please just say so.  

In any case, my hope is that whatever issues there may regarding the "iterative application" interpretation of gsub should not get in the way of fixing the infinite looping behavior described in #1206 and #2148.

thedward:
@pkoppstein, I was just hoping there was a source other than the manual to shed more light on this.

[...]

Issue #1206, opened 2016-08-11
gsub with empty regex doesn't terminate


4z3:
Calling `gsub` with an empty regex will run forever. Example:

``` console
$ jq -n '"" | gsub(""; "")'
```

Not sure if this is a bug or expected behavior, but terminating with an error might be more appropriate :)

pkoppstein:
The bug was fixed by ad8d1a8 around Dec 9, 2015; see #1033 

That is, it was fixed AFTER the release of jq 1.5

4z3:
I can reproduce this with current master (0b8218515eabf1a967eba0dbcc7a0e5ae031a509).

4z3:
I can reproduce this bug with current master (0b8218515eabf1a967eba0dbcc7a0e5ae031a509).

agguser:
Seems that empty matches also cause this bug.
```
$ jq -n -r '"abc\nabc" | gsub("^"; "+")'   # not terminate!

$ jq -n -r '"abc\nabc" | sub("^"; "+")'   # work if not 'g'
+abc
abc

$ jq -n -r '"abc\nabc" | gsub("^."; "+")'   # weird!
+++
abc
```

itchyny:
Related issue: #2148

PR #2621, opened 2023-06-19
fix some gsub bugs (incremental changes only); clarify gsub documentation

pkoppstein:
1) Let sub/2 stand
2) Incremental enhancements of sub/3 to resolve some important gsub bugs including #2148 #2354 #1206
3) Augment tests in onig.test
4) Tweak documentation of gsub

Note that these updates do not address the issue raised (for example) in #1425


pkoppstein:
better fix is in the works

PR #2607, opened 2023-06-09
create security policy

eslerm:
Security Policy based on [OpenSSF's example](https://github.com/ossf/oss-vulnerability-guide/blob/main/templates/security_policies/email_intake.md).

This would help security researchers who want to privately report vulnerabilities to jq like https://github.com/jqlang/jq/issues/2496

Please take out whatever does not apply to your project and replace $EMAIL. The last two sentences are fairly standard (and encouraged), but *could* be removed.

PR #2622, opened 2023-06-19
Create SECURITY.md

nicowilliams:
Supersede #2607 with a security policy that refers to GitHub's private reporting mechanism.


nicowilliams:
Although this hasn't pushed yet, you can start using the GitHub private security vulnerability reporting mechanism immediately, as it has been enabled.

PR #2625, opened 2023-06-20
Document `length` for numbers and bools

alexjurkiewicz:
This was added in ae7a042876130c471dce28a4396abc215192eaa9 but not explicitly documented.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -814,16 +814,20 @@ sections:
             codepoints it contains (which will be the same as its
             JSON-encoded length in bytes if it's pure ASCII).
 
+          - The length of a **number** is its absolute value.
+
           - The length of an **array** is the number of elements.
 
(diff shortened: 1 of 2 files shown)


wader (approved):

wader:
As we're nearly listing all types, should we also add that length of boolean is an error?

wader (approved):
Maybe update PR title?

itchyny (changes requested):
This needs to applied to the development version manual too. Also, the period is missing (sorry for nitpick).

wader (approved):
@itchyny do we care to update < 1.6 verisons too?

itchyny (approved):
LGTM, thank you.

@wader
Document improvements are good enough if been applied to the latest released & development versions. But talking of typo fixes, they keep sending patches with spell checkers even if it is in older version manuals, so I'd like to fix them all.

PR #2627, opened 2023-06-23
Clarify variable binding a little

chx:


emanuele6:
This is a bit misleanding, it sounds like `.foo as $bar` is an expression that assigns the result of `.foo` to `$bar`, and is otherwise equivalent to `.`, but really you cannot use it in that way.
For example, `jq '.foo as $bar'` or `jq '(.foo as $bar) | .baz'` are not valid, and nor is `jq '.foo as $bar, .baz as $boo | .xyz'`.

`EXPR as $VAR` is not an expression by itself, so it does not result in a value, it is a special form that has to be followed by `|` and an expression: `EXPR as $VAR | EXPR`.

chx:
I am open to reword this but it was really not clear to me this doesn't change the input and that's what I would like to clarify here. The closest concept I could map to was a variable assignment.

Perhaps "This works because the variable binding is not an expression but a special form which does not change the input and must be followed by | EXPR"?

itchyny:
This is already mentioned isn't it; `The expression exp as $x | ... means: for each value of expression exp, run the rest of the pipeline with the entire original input, and with $x set to that value.`

chx:
I completely have missed that. May I then recommend moving that paragraph to the very beginning of `as` ?

itchyny:
>  But I'd remove the foreach part.

Why? Considering some cases like `range(10) as $x | [$x]`, this looks like a loop.

chx:
Because that leads back to the original confusion? I mean, in say PHP `foreach ([1, 2, 3] as $a) { foo($a)` will call `foo()` with 1 first then 2 then 3. Which is exactly what is _not_ happening here, if I gather correctly.

chx:
Because that leads back to the original confusion? I mean, in say PHP foreach ([1, 2, 3] as $a) { foo($a) will assign 1 first to $a then 2 then 3. Which is exactly what is _not_ happening here, if I gather correctly.

itchyny:
For me iteration is happening here; `jq -nc 'range(3) as $x | [$x]'` and `python -c 'for i in range(3): print([i])'` yields the same result.

chx:
But that's because you started with `range(3)`. 

This whole thing started out from https://stackoverflow.com/a/76535899/308851

chx:
OK I put back that `foreach` bit,

docs/content/manual/v1.6/manual.yml
```diff
@@ -2421,6 +2421,11 @@ sections:
       - title: "Variable / Symbolic Binding Operator: `... as $identifier | ...`"
         body: |
 
+          The expression `exp as $x | ...` means: for each value of expression
```
nicowilliams, line 2424: `s/expression/form/`

nicowilliams: The rest of this text looks fine, though perhaps a lot of the text following this new text needs to be altered or deleted.

itchyny:
The section looks very redundant and moving a sentence to the beginning does not clarify what was confusing so closing.

PR #2630, opened 2023-06-26
Ignore error of outdated manual test when docs disabled (fix #2310)

### LINKED ISSUE #2310: make check always fails
Whether ./configure is invoked with --disable-docs or --enable-docs, make && make check always emits:

"Changes to the manual.yml require docs to be enabled to generate an updated manpage"
"As a result, your manpage is out of date."

..and make check returns 2

itchyny:
This PR fixes  #2310, it was caused by `@false`, also tweaks the error messages.


wader (approved):

itchyny:
Thank you.

PR #2634, opened 2023-06-28
Fix bison warnings on incompatibilities with POSIX Yacc

itchyny:
This PR fixes bison warnings on incompatibilities with POSIX Yacc (e.x. `warning: POSIX Yacc does not support string literals`), also fixes deprecation warning (e.x. `warning: deprecated directive: ‘ %error-verbose’ , use ‘ %define parse.error verbose’`). See https://www.gnu.org/software/bison/manual/html_node/Diagnostics.html for reference, `--warnings=all` is still important.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,10 +44,10 @@ BUILT_SOURCES = src/builtin.inc src/version.h
 	$(AM_V_LEX) echo "NOT building lexer.c!"
 endif
 
-# Tell YACC (bison) autoconf macros that you want a header file created.
-# If the --warnings=all fails, you probably have an old version of bison
-# OSX ships an old bison, so update with homebrew or macports
-AM_YFLAGS = --warnings=all -d
+# Tell YACC (Bison) autoconf macros that you want a header file created.
+# If the --warnings=all fails, you probably have an old version of Bison
+# macOS ships an old Bison, so update with Homebrew or MacPorts.
+AM_YFLAGS = --warnings=all -Wno-yacc -d
 
 ### libjq
 
--- a/src/parser.y
+++ b/src/parser.y
@@ -26,7 +26,7 @@ struct lexer_param;
 }
 
 %locations
-%error-verbose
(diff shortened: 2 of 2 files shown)


Makefile.am
```diff
@@ -44,10 +44,10 @@ BUILT_SOURCES = src/builtin.inc src/version.h
...
 endif
 
-# Tell YACC (bison) autoconf macros that you want a header file created.
-# If the --warnings=all fails, you probably have an old version of bison
-# OSX ships an old bison, so update with homebrew or macports
-AM_YFLAGS = --warnings=all -d
+# Tell YACC (Bison) autoconf macros that you want a header file created.
+# If the --warnings=all fails, you probably have an old version of Bison
+# macOS ships an old Bison, so update with Homebrew or MacPorts.
+AM_YFLAGS = --warnings=all -Wno-yacc -d
```
wader, line 50: Wondering is we should at some point add some kind of dev docs as i always forget how set this, ex things like `YACC="$(brew --prefix)/opt/bison/bin/bison -y" ./configure ...` for brew/macOS

wader (approved):

Makefile.am
itchyny: I have `export PATH="$HOMEBREW_PREFIX/opt/bison/bin:$PATH"` in `.zprofile`. Maybe [wiki](https://github.com/jqlang/jq/wiki/Installation#from-github) can be improved (I don't like the symlink command).

wader: Ah yeap wiki could be good, want some place to dump it so it's easy to find and again

itchyny: I edited this page [like this](https://github.com/jqlang/jq/wiki/Installation/_compare/6ed0b47f03c857742f119c6caa959be08c5fd9fa...3e99f901ba8b8098f9da8e85636526c7b578c9bc), but you can edit further if you have better suggestion.

wader: 👍

PR #2637, opened 2023-06-28
Remove deleted README file from jq.spec

itchyny:
The rerundant README was removed in #2482, which is good, but building rpm package currently fails due to this change. So I removed the entry from `jq.spec`.


wader (approved):

itchyny:
Thank you.

PR #1791, opened 2019-01-18
Set field color using JQ_COLORS

haguenau:
This should address #1739.

As far as I can tell, this works and passes the test suite (which I updated to test this specific addition). There is a chance that the code changes are not perfect; for example I'm not sure how much sense it makes to have introduced an enum member `JV_KIND_FIELD` for this (although it keeps the code mostly regular).

PR #2638, opened 2023-06-28
Make field colors configurable

ericpruitt:
Allow object key colors to be configured with the "JQ_COLORS" environment variable. This is a updated version of a [GitHub pull request created by David
Haguenauer](https://github.com/stedolan/jq/pull/1791). Since the prebuilt manual already needed to be updated, this change also fixes in omission in <https://github.com/stedolan/jq/commit/cf61515> in which "39" in "JQ_COLORS" should have been replaced with "37".

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -13,7 +13,8 @@ typedef enum {
   JV_KIND_NUMBER,
   JV_KIND_STRING,
   JV_KIND_ARRAY,
-  JV_KIND_OBJECT
+  JV_KIND_OBJECT,
+  JV_KIND_FIELD
 } jv_kind;
 
 struct jv_refcnt;
--- a/tests/shtest
+++ b/tests/shtest
@@ -293,20 +293,44 @@ if [ "$($VALGRIND $Q $JQ -n '{"a":"xyz"} | halt_error(1)' 2>&1)" != '{"a":"xyz"}
 fi
 
 # Check $JQ_COLORS
+## Default colors, null input
 $JQ -Ccn . > $d/color
 printf '\033[1;30mnull\033[0m\n' > $d/expect
 cmp $d/color $d/expect
+
+## Set non-default color, null input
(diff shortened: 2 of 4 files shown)


ericpruitt:
I also wrote a script called jqcolors that's analagous to dircolors that I could include in a separate PR if you'd accept it: <https://github.com/ericpruitt/emus/blob/master/scripts/jqcolors>.

Example file:
```
~$ cat .jqcolors
STRING  0           # Default
ARRAY   0           # Default
OBJECT  0           # Default
FIELD   1;37        # Bold White

# XTerm colors are used for these tokens to make the text bright -- having the
# red, green and/or blue channels maxed out -- without being bold.
NULL    0;38;5;226  # Yellow
FALSE   0;38;5;196  # Red
TRUE    0;38;5;46   # Green
NUMBER  0;38;5;201  # Magenta
```

tests/shtest
```diff
@@ -293,20 +293,44 @@ if [ "$($VALGRIND $Q $JQ -n '{"a":"xyz"} | halt_error(1)' 2>&1)" != '{"a":"xyz"}
 fi
 
 # Check $JQ_COLORS
+## Default colors, null input
 $JQ -Ccn . > $d/color
```
itchyny, line 297: Please prepend `JQ_COLORS= ` this line and L307. `make check` fails when exporting the environment variable.

src/jv.h
```diff
@@ -13,7 +13,8 @@ typedef enum {
   JV_KIND_NUMBER,
   JV_KIND_STRING,
   JV_KIND_ARRAY,
-  JV_KIND_OBJECT
+  JV_KIND_OBJECT,
+  JV_KIND_FIELD
```
itchyny, line 17: Please see https://github.com/jqlang/jq/pull/1791#issuecomment-1613218512.

itchyny (changes requested):

PR #2639, opened 2023-06-28
Move an example to the manual and regenerate man.test

itchyny:
The PR #2400 added a test to `man.test`, but this file is generated from the manual. There're two options; move this test to `jq.test` or to the manual. I think this is worth adding to the `?` operator section, so added there and regenerated the test file.


wader:
Makes sene to me. The other changes to man.test are because we haven't regenerated it a while?

wader (approved):

itchyny:
Yeah, we could generate it and make CI fail if there is a diff.

PR #2644, opened 2023-07-01
Skip regex tests when built without regexp support

itchyny:
This PR catches the exception of regex library absence, and gracefully skip the test when built without oniguruma. This resolves #2292, but is better running all the manual tests except for regex ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,7 @@ homepage and on the github release page, https://github.com/jqlang/jq/releases
 If you're building directly from the latest git, you'll need flex,
 bison (3.0 or newer), libtool, make, automake, and autoconf installed.
 To get regexp support you'll also need to install Oniguruma or clone it as a
-git submodule as per the instructions below.
-(note that jq's tests require regexp support to pass).  To build, run:
+git submodule as per the instructions below. To build, run:
 
     git submodule update --init # if building from git to get oniguruma
     autoreconf -fi              # if building from git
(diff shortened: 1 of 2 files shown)


wader:
Could we use the fact that (nearly?) all regexp related test are in onig.test? if so maybe skip could be done in the build system instead of in the run tests code?

The macOS build failure looks like something temporary:
```
error: Could not read ffac597fdb6ce8236b5e2da77ab6fbae3d9d9cb7
fatal: revision walk setup failed
error: https://github.com/Homebrew/homebrew-cask did not send all necessary objects

Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask failed!
```

itchyny:
But we need this code to make man.test pass without regexp support after all.

wader:
> But we need this code to make man.test pass without regexp support after all.

Aha i see 👍 

Maybe if we end up with more conditional tests in the future the .test format could be extended with some kind of feature gating?

wader (approved):

itchyny:
Hmm, let me reconsider merging and explore about improving the .test format. How about adding `%%REGEX` sign to indicate the test case requires the regex support, including the `regex: true` flag in the manual source file, and skip them unless `HAVE_LIBONIG`?

wader:
Maybe use the config vars and have `%%HAVE_LIBOING` or just `%%LIBOING` would make it similar to what we do in c code. But  i like `%%REGEX` also, more feature name than implementation detail... but i guess jq is not changing regex implementation anytime soon :)

PR #2648, opened 2023-07-03
Create dist and upload on release also disable docs on build

### LINKED ISSUE #1872: Instructions to build tarball unclear
It was difficult to build the 1.6 tarball from source with builtin-onigurama, the expected ./configure && make fails.

I had to do this extra step:

```
cd modules/onigurama
autoreconf -i
./configure
```

then jq-1.6 built correctly.

I had to guess the command `autoreconf -i`, so this info could be added to the README.

### LINKED ISSUE #2372: version 1.6 tar compile fails on builtin oniguruma
i had to do the following to make it work

```
tar -xvf jq-1.6.tar.gz
cd jq-1.6
cd modules/oniguruma
autoreconf -fi
cd ../..
./configure --with-oniguruma=builtin
```

### LINKED ISSUE #2448: tar ball jq-jq-1.6.tar.gz miss file "configure"
according the readme.md for complie from tar ball, it seem that lack of file "configure".
and in old version, "configure" file exist, like in jq-1.5.tar.gz

itchyny:
- I added new `dist` job to create the dist in CI, and upload the artifact on release.
- I disabled building docs because Python dependency is not correctly configured, and we also have another workflow to build the website.
- I also dropped `-f` flag from `autoreconf` because it overwrites the submodule files resulting into git diffs and -dirty suffix in dist file.
- Maybe we could build jq.1.prebuild somewhere but tackle later.

I downloaded the artifacts [here](https://github.com/jqlang/jq/actions/runs/5439907844?pr=2648) and confirmed that both .tar.gz and .zip can be built with `./configure && make`. Closes #1872, closes  #2372, also closes #2448.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,12 +39,7 @@ jobs:
       - name: Install packages
         run: |
           sudo apt-get update -qq
-          sudo apt-get install -y \
-            automake \
-            autoconf \
-            bison \
-            flex \
-            gdb
+          sudo apt-get install -y automake autoconf
       - name: Build
         run: |
           autoreconf -i
@@ -54,8 +49,7 @@ jobs:
             --disable-maintainer-mode \
             --disable-valgrind \
             --with-oniguruma=builtin \
-            ${{ matrix.configure_flag }} \
-            YACC="$(which bison) -y"
+            ${{ matrix.configure_flag }}
           make
           strip jq
       - name: Test
@@ -97,12 +91,7 @@ jobs:
         run: |
           # brew update sometimes fails with "Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask failed!"
           brew update || brew update-reset
-          brew install \
-            autoconf \
-            automake \
-            libtool \
-            flex \
-            bison
+          brew install autoconf automake libtool
       - name: Build
         run: |
           autoreconf -i
@@ -113,8 +102,7 @@ jobs:
             --disable-valgrind \
             --with-oniguruma=builtin \
             --enable-static \
-            --enable-all-static \
-            YACC="$(brew --prefix)/opt/bison/bin/bison -y"
+            --enable-all-static
           make
           strip jq
       - name: Test
@@ -162,8 +150,6 @@ jobs:
             autoconf
             automake
             libtool
-            bison
-            flex
       - name: Build
         shell: msys2 {0}
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -44,13 +44,13 @@ jobs:
             autoconf \
             bison \
             flex \
-            gdb \
-            python3
+            gdb
```
wader: Do we use gdb for something?

.github/workflows/ci.yml
```diff
@@ -199,17 +201,49 @@ jobs:
...
+      - name: Create dist
+        run: |
+          autoreconf -i
+          ./configure --disable-dependency-tracking \
+            --disable-silent-rules \
+            --disable-docs \
+            --disable-maintainer-mode \
+            --disable-valgrind \
+            --with-oniguruma=builtin
+          git diff --exit-code
```
wader: Wonder if something similar is good to do for the builds as a sanity check?

.github/workflows/ci.yml
```diff
@@ -44,13 +44,13 @@ jobs:
             autoconf \
             bison \
             flex \
```
wader: Unrelated to this: I wondered before why bison and flex is installed, should not be needed? guess it was inherited from the old CI config somehow?

wader (approved):

.github/workflows/ci.yml
itchyny: Considering we pass `YACC` explicitly, originally there was an intention to generate the code, but somehow maintainer-mode is disabled thus they are unused. But I think it's better re-generating the parser/lexer codes.

itchyny: Hmm, but I think it's better to use the source code included in repo for release artifacts.

wader: Intension was to always generate and not have flex/yacc output in the repo at all? but maybe it makes sense to run flex/yacc as a check/lint for CI?
For release i agree we should not run flex/yacc

.github/workflows/ci.yml
```diff
@@ -147,6 +138,10 @@ jobs:
       CC: ${{ matrix.compiler }}
       SUFFIX: windows-${{ matrix.image }}-${{ matrix.compiler }}
     steps:
+      - name: Prepare git
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
```
wader: Need in addition to .gitattributes or it's not used for some reason?

wader:
Wonder what is going on with "CI / release (pull_request) Skipped" is because it uses `if`?

wader (approved):

wader (approved):

itchyny:
Thank you!

PR #2649, opened 2023-07-03
Improve Dockerfile: update base image and optimize image size

### LINKED ISSUE #845: Dockerfile cleanup
Following up on #842:
- split into layers on natural boundaries (get dependencies, get code, build code and manual) to reduce rebuilding time
- factor long steps into shell scripts, maybe even reusable in .travis.yml
- enable valgrind if possible

itchyny:
This PR improves Dockerfile to resolve endless reports about the Docker image in the next release. I update the base image to Debian 12, and use multi-stage build to optimize the image size. We stop re-building the parser/lexer codes since that's what we do on building the release executables. We no longer re-build docs and manpage in this image, so drop dependency on Python. Releasing multi-arch image to GHCR will be in the next PR.
This PR closes #845, closes #1631, closes #1900, closes #2004, and closes #2617.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,6 @@ WORKDIR /app
 COPY . /app
 RUN autoreconf -i \
  && ./configure \
-      --disable-dependency-tracking \
-      --disable-silent-rules \
       --disable-docs \
       --disable-maintainer-mode \
       --disable-valgrind \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,54 +1,38 @@
-FROM debian:9
+FROM debian:12-slim AS builder
 
 ENV DEBIAN_FRONTEND=noninteractive \
     DEBCONF_NONINTERACTIVE_SEEN=true \
     LC_ALL=C.UTF-8 \
     LANG=C.UTF-8
 
+RUN apt-get update \
+ && apt-get install -y \
+      build-essential \
+      autoconf \
+      libtool \
+      git \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
 COPY . /app
+RUN autoreconf -i \
+ && ./configure \
+      --disable-dependency-tracking \
+      --disable-silent-rules \
+      --disable-docs \
+      --disable-maintainer-mode \
+      --disable-valgrind \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,54 +1,38 @@
...
+ && rm -rf /var/lib/apt/lists/*
 
-# get dependencies, build, and remove anything we don't need for running jq.
-# valgrind seems to have trouble with pthreads TLS so it's off.
+WORKDIR /app
+COPY . /app
+RUN autoreconf -i \
+ && ./configure \
+      --disable-dependency-tracking \
+      --disable-silent-rules \
```
wader: Got a curious about these options, seems `--disable-silent-rules` is the default? if so maybe can skip specify it? `--disable-dependency-tracking` seems to speed up one-time builds by not generate make dependency files, i suspect not much build speed up for jq?

Dockerfile
```diff
@@ -1,54 +1,38 @@
...
-        git \
-        flex \
-        python3 \
-        python3-pip && \
-    apt-get autoremove -y && \
-    rm -rf /app/modules/oniguruma/* && \
-    rm -rf /app/modules/oniguruma/.git && \
-    rm -rf /app/modules/oniguruma/.gitignore && \
-    rm -rf /var/lib/apt/lists/* /var/lib/gems
+FROM scratch
```
wader, line 32: In dockerfiles like this i usually have a `RUN ["/path/to/binary", "--version"]` just as a sanity test that the binary can run correct in a scratch container, have saved me a couple of times

Dockerfile
```diff
@@ -1,54 +1,38 @@
...
 
-ENTRYPOINT ["/usr/local/bin/jq"]
-CMD []
+COPY --from=builder /app/AUTHORS /app/COPYING /usr/local/bin/jq /
+ENTRYPOINT ["/jq"]
+CMD ["--help"]
```
wader: Maybe skip or leave empty to make it behave as normal cli?

Dockerfile
itchyny: I agree. I'll drop both flags.

wader (approved):

itchyny:
Thank you!

PR #2651, opened 2023-07-04
Fix release workflow tag trigger pattern

itchyny:
We use `jq-` prefix for tags and `scripts/version` relies on this, so keep using the prefix.


wader (approved):

itchyny:
Thank you!

PR #2655, opened 2023-07-05
revamp walk/1

### LINKED ISSUE #2584: Deleting objects using walk has side effects
**Describe the bug**
Deleting objects using walk seems to cause side effects. I'm not sure of the details, so I'll just have to provide an example.

**To Reproduce**
`walk(select(IN({}, []) | not))`

input:
```
{
  "a": 1,
  "b": []
}
```

**Expected behavior**
```
{
  "a": 1
}
```

**Actual behavior**
```
null
```

 - OS and Version: MacOS 11.7.4
 - jq version 1.6

This seems to be reproducible in every environment I've tried, though.

**Additional context**
In addition, if the input is the semantically equivalent:

```
{
  "b": [],
  "a": 1
}
```

then the output is as-expected.

### LINKED ISSUE #2611: ER: simpler, faster `walk/1`
A well-known efficiency enhancement for `walk/1` is to use an inner function along the lines of:
```
def walk(f):
  def w: ...;
  w;
```

In addition to this efficiency enhancement, I would like to propose a very small change
that both greatly simplifies the definition and resolves some other issues, notably a discrepancy with the semantics of 
assignments of the form `.b = empty`.

For example, the expression:
```
[ { a: {b: 2, c: 3} } ] | map(.b = empty)
```
yields []

so one would expect (and perhaps require) that:
```
walk(if type == "object" then .b = empty]
```
should also yield []. Using the current def of `walk/1`, however, it yields [null] instead.

Here then is the proposed new version of `walk/1`, which is 
faster, simpler and more conformant as described above:
```
def walk(f):
  def w:
    if type == "object"
    then map_values(w)
    elif type == "array" then map(w)
    else .
    end
    | f;
  w;
```
Please also note that the above proposed definition is sufficient to resolve the issue raised in #2584.

pkoppstein:
Resolves #2584; also resolves #2611


itchyny:
The output of `{x:0} | walk(.,1)` changes by this PR. Is it intended?

pkoppstein:
It is not the primary intention, which was to make the definition in terms of `map` and `map_values` as simple as possible while ensuring correct behavior when the arg was "simple" (no top-level commas). The current behavior was unintended.

Perhaps the manual should mention this example?

itchyny:
I just wanted to confirm that you're aware of the behavior change. Even if not mentioned in the manual, we need reasonable explanations to any behavior changes. Typical reason is that the original behavior was unintentional bug, but making the implementation simpler is also a good reason. This time, I think changing this behavior would be Okay. By the way, [in gojq I use `last`](https://github.com/itchyny/gojq/blob/v0.12.13/builtin.jq#L50-L57) to make it work like jq but I will drop it...

nicowilliams:
Hmm, I think we want the original behavior, yeah.

nicowilliams:
Or maybe I should read @2584 before commenting...

pkoppstein:
@nicowilliams - Perhaps it would help to review the situation w.r.t. `walk/1`.

1) The old def of `walk(f)` was written without regard to the
   possibility that f could be a sequence of expressions.  I have not
   double-checked but hopefully the official documentation
   never came remotely close to specifying what the semantics should be
   in such cases.

2)  In the case in question, I think it's generally agreed that one would want

   `{x:0} | walk(.,1)` to be equivalent to `{x:0} | walk(.), walk(1)`

    In this and other cases, jaq and the new walk are in agreement,
    and as you have no doubt noticed,  @itchyny indicated that he had to go out of his way
    to align gojq's `walk` behavior with jq's.

3) The new def elegantly resolves several important documented issues; this consideration,
   in conjunction with the fact that the new def of walk/1 is as close to a "declarative" specification of
   the desired functionality as one could achieve or want in jq, seem to me very compelling.

pkoppstein:
@itchyny wrote:
> By the way, [in gojq I use last](https://github.com/itchyny/gojq/blob/v0.12.13/builtin.jq#L50-L57) to make it work like jq but I will drop it...

So what should I do to get the ball rolling again?  For jq 1.7, we should at the bare minimum address the inefficiency of the 1.6 version.

I still think the version in this PR (with perhaps some elaboration in manual.yml) is the way to go, but please let me know if I should add a call to `last()`. Thanks.

src/builtin.jq
```diff
@@ -249,17 +249,18 @@ def bsearch($target):
       else .[2]
       end
   end;
-
+#
```
itchyny: I prefer this line to be an empty line rather than an empty comment.

itchyny:
Please rebase this PR, thanks.

src/builtin.jq
pkoppstein: OK, but someone should probably run: `s/^# *$//`

pkoppstein:
@itchyny wrote:
> Please rebase this PR, thanks.

I did, but somehow a "merge" was added as well.  Feel free to "squash" as desired.

itchyny:
Could you rebase the commit onto the latest remote master branch? This PR includes the commits ahead and I can't review the diff now.

pkoppstein:
Given the difficulty merging this PR, I'm closing it in favor of https://github.com/jqlang/jq/pull/2795

nicowilliams:
@pkoppstein if you would only just use your terminal application and the git command you wouldn't have such difficulties and you wouldn't cause so many duplicate PRs.

pkoppstein:
@nicowilliams wrote:
> use your terminal application and the git command 

In brief, I did, but itchyny said he could not use the result of a simple "Rebase" and "Push".

In long - well, let me know if and when you would like further details.

Issue #2578, opened 2023-05-03
No (documented) shorthand way to copy nested objects in Object Construction


garethhumphriesgkc:
**Describe the bug**
We can easily copy fields from the root of the input object into our output by using `{ field }` syntax, but there is no way to copy nested fields with the same shorthand.  I can't say `{ fields.field }`, I must use the cumbersome `{ fields: { field: .fields.field } }`.

This is especially tedious when your desired field is many layers deep and you want to keep it in context.

While I appreciate there can be questions around what the expected output is when navigating nested objects, I think in this case the same short hand logic makes unambiguous sense - it is still "create a field called _this_, with the contents of _this_", it's just that now _this_ has multiple layers to it.

**To Reproduce**
```
$ echo '{"a":{"b":1}}' | jq '{ a }'
{
  "a": {
    "b": 1
  }
}
$ echo '{"a":{"b":1}}' | jq '{ a.b }'
jq: error: syntax error, unexpected FIELD, expecting '}' (Unix shell quoting issues?) at <top-level>, line 1:
{a.b}  
jq: 1 compile error
$
```

**Expected behavior**
Output should include the value of the selected field
```
$ echo '{"a":{"b":1,"junk":0},"useless":0}' | jq '{ a.b }'
{
  "a": {
    "b": 1
  }
}
$
```

**Environment (please complete the following information):**
jq-1.6

**Additional context**

wader:
Hi, yeap something like that would be nice. jq development and maintenance is current a bit on hold, see s#2305 

The closest i think of with current version of jq is a `pick` function described here https://github.com/stedolan/jq/issues/2062#issuecomment-1165579448

pkoppstein:
Here's a version of `pick` that works with both jq and gojq, and allows multiple dot-paths to be specified, but does NOT check whether the incoming entity has the specified paths:
```
# stream should be a stream of dot-paths
def pick(stream):
  . as $in
  | reduce path(stream) as $a (null;
      setpath($a; $in|getpath($a)) );
```
Example:
```
{a: {b: 1, c: 2, d: {e: 2}}} | pick(.a.b, .a.d.e)
```
yields
```
{"a":{"b":1,"d":{"e":2}}}
```

itchyny:
`pick/1` is implemented by #2656 and merged so closing.

PR #2656, opened 2023-07-05
builtin: add pick(stream)

pkoppstein:
pick(stream) works on both arrays and objects

Restrictions on the dot-paths are specified in manual.yml

See #2578


wader:
I wonder if it would make sense to extend the object construction syntax to allow things like `{a, b.c, x}` to mean `{a: .a, b: {c: .b.c}, x: .x}`? but would that also work for arrays somehow? would `{a[1].b}` mean? `{a: [.a[1].b] }` or `{a: [null.a[1].b]}`? and what would `{a: [.a[1,2].b] }` mean? should be in sync with destructing syntax 🤔 

Sorry for rambling, maybe i should create a proposal issue to collect ideas if it's something we want to consider? ... also i have no idea about implementation difficulty

pkoppstein:
@wader - the linked issue already refers to such a proposal, which of course involves an extension to the language, whereas this PR only proposes a builtin, which I think is simple and useful enough to be included regardless of any such extension.

wader:
@pkoppstein Aha, i see. A nice thing with `pick/1` is that things like this works `pick(.. | select(...))`

src/builtin.jq
```diff
@@ -260,6 +260,12 @@ def walk(f):
   else f
   end;
 
+# stream should be a stream of dot-paths
+def pick(stream):
```
nicowilliams: Rather than call the thunk `stream` I'd call it `paths` or `pathexp`.  After all, it's not used as a stream but rather passed to `path/1` which then generates a stream of paths.

docs/content/manual/manual.yml
```diff
@@ -925,6 +925,24 @@ sections:
             input: '{"a": 1, "b": 2, "c": 3}'
             output: ['{"a": 2, "b": 3, "c": 4}']
 
+      - title: "`pick(stream)`"
```
nicowilliams: I think we want `s/stream/pathexp/` here and below, since the closure passed here is passed to `path/1`, so it's really a path expression.

nicowilliams:
Apart from proposing `s/stream/pathexp/` for the name of the closure passed to `pick/1`, I like this a lot!

pkoppstein:
@nicowilliams wrote:
> s/stream/pathexp/ 

How about s/stream/pathexps/ ?  I wanted to emphasize the "reduction" aspect of the def, i.e. that the argument is NOT regular.

nicowilliams:
> How about s/stream/pathexps/ ? I wanted to emphasize the "reduction" aspect of the def, i.e. that the argument is NOT regular.

Works for me!

docs/content/manual/manual.yml
nicowilliams: Don't forget to update the docs to say `pathexps`.

pkoppstein:
@nicowilliams - Done. Thanks for reminder.

nicowilliams:
Thanks!

PR #2661, opened 2023-07-06
Fix dumping large floating numbers (fix #2367)

### LINKED ISSUE #2367: Inconsistent dump results of jq
Hi there,

I found that for large floating numbers, when compiling with **gcc** and **clang**, `jq` would emit different outputs.

For example, for `echo '{"a": 1E9999999999}' | jq .`, 
**gcc** compiled `jq` would emit: 
```
{
  "a": ((double)1.79769313486231570814527423731704357e+308L)
}
```
while **clang** compiled `jq` would emit:
```
{
  "a": 1.7976931348623157e+308
}
```

I think the cause is probably different interpretation of `DBL_MAX` / `DBL_MIN` in gcc and clang. But to ensure the correctness, robustness, and consistency of `jq`, it's important to keep programs' behaviour consistent across main-stream compilers.

itchyny:
Due to the differences of `DBL_MAX` and `DBL_MIN` definition between compilers, large floating number literals in input JSON are dumped incorrectly as reported by #2367. This PR fixes #2367.

### DIFF
--- a/src/jv.c
+++ b/src/jv.c
@@ -216,16 +216,11 @@ enum {
 #define JVP_FLAGS_NUMBER_NATIVE_STR   JVP_MAKE_FLAGS(JV_KIND_NUMBER, JVP_MAKE_PFLAGS(JVP_NUMBER_NATIVE, 1))
 #define JVP_FLAGS_NUMBER_LITERAL      JVP_MAKE_FLAGS(JV_KIND_NUMBER, JVP_MAKE_PFLAGS(JVP_NUMBER_DECIMAL, 1))
 
-#define STR(x) #x
-#define XSTR(x) STR(x)
-#define DBL_MAX_STR XSTR(DBL_MAX)
-#define DBL_MIN_STR "-" XSTR(DBL_MAX)
-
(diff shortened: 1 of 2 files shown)


src/jv.c
```diff
@@ -658,12 +653,12 @@ static const char* jvp_literal_number_literal(jv n) {
...
-      return DBL_MIN_STR;
-    } else {
-      return DBL_MAX_STR;
-    }
+    // We cannot preserve the literal data of numbers outside the limited
+    // range of exponent. Since `decNumberToString` returns "Infinity"
+    // (or "-Infinity"), and to reduce stack allocations as possible, we
+    // normalize infinities in the callers instead of printing the maximum
+    // (or minimum) double here.
+    return NULL;
```
wader, line 661: By returning `NULL` here will we end up here https://github.com/jqlang/jq/blob/master/src/jv_print.c#L243-L252 and inf gets clamped to DBL_MAX?

wader (approved):

src/jv.c
itchyny: Exactly.

PR #2666, opened 2023-07-07
Remove deprecated filters: leaf_paths, recurse_down

### LINKED ISSUE #1163: `paths(scalars)` bug?

itchyny:
This PR removes two builtin filters marked as **deprecated** and will be removed in the next major release. Both of `leaf_paths` and `recurse_down` filters were deprecated by #426 (dde43f796ef6aca9942e4c5c69478fc5f64367ed) in 2014, and it's already 9 years passed since then. I think we have waited too long and can remove them in the coming next release. Closes #1163 (which is not actually a bug but mentions about the quite confusing definition of `leaf_paths` so let me close).


nicowilliams:
Thanks!

uncle-fed:
@itchyny : could you kindly elaborate why "#1163 is not a bug"  and whether the PR here changes the behaviour of paths(scalars) reported in #1163. Thanks a lot!

nicowilliams:
> @itchyny : could you kindly elaborate why "#1163 is not a bug" and whether the PR here changes the behaviour of paths(scalars) reported in #1163. Thanks a lot!

@uncle-fed the issue is that `scalars` isn't a boolean predicate, so `false|scalars` emits `false`, but `paths/1` wants a boolean predicate, so `paths(false)` will not output anything.  What you want is `paths(scalars|true)`.

uncle-fed:
@nicowilliams : thanks a lot for the clear explanation, finally :-)

My confusion (and you can see that there are quite a few others) comes from encountering the `paths(scalars)` behaviour, then scratching your head and wondering whether it is a bug (perhaps because the current docs are not quite clear in that regard) coming to GitHub, finding this exact issue and then seeing how the issue is not addressed on the bug tracker for years, assume that 1) it _is_ a bug, and 2) nobody cares.  :-)

There are now "bug workarounds" on #1163 such as `paths([scalars] != [])` , which of course are more complex than the correct usage suggested now, i.e., `paths(scalars|true)`

And what's worse, there are other bug reports open (also beyond the jq main repository but in other projects that rely on jq) which simply treat #1163 as a bug that needs fixing on the jq side.

I think it would be of a great help if both, the #1163 and the docs were updated with this vital information.

Thank you, as always, for the fantastic tool.

uncle-fed:
OK, I see there was now further development of the #1163. Thanks a lot for at least setting the record straight in the original report, finally! ;-)

PR #2670, opened 2023-07-08
Drop comments in builtin filters on build

itchyny:
I think we can drop comments of `builtin.inc` to minimize the file for runtime performance improvement. I used `s/^ *#.*//` instead of `/^ *#/d` to keep the line numbers. What do you think?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -114,7 +114,7 @@ src/main.c: src/version.h
 
 src/builtin.inc: $(srcdir)/src/builtin.jq
 	mkdir -p src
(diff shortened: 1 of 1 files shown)


nicowilliams:
It makes a big difference?  A more involved thing we could do would be to pre-compile all the jq-coded builtins into `block` form -- that would probably bloat jq's size but speed up compilation.

itchyny:
Yeah, maybe not much difference so closing. If we could pre-parse the builtins on build, that would be great. Actually that's what gojq does.

nicowilliams:
I envision making a bootstrap jq that can emit a C representation of the builtins' blocks, then build the final jq using that, and at run-time use those blocks instead of having to compile the builtins.

nicowilliams:
As it happens I've a branch that dumps blocks in JSON form.  I could change it to emit C instead.

PR #2671, opened 2023-07-08
Fix has(nan) on arrays to output false

itchyny:
I fixed `has/1` to make `[0] | has(nan)` emit `false` not `true`.

### DIFF
--- a/src/jv_aux.c
+++ b/src/jv_aux.c
@@ -214,10 +214,14 @@ jv jv_has(jv t, jv k) {
     jv_free(elem);
   } else if (jv_get_kind(t) == JV_KIND_ARRAY &&
              jv_get_kind(k) == JV_KIND_NUMBER) {
-    jv elem = jv_array_get(t, (int)jv_number_value(k));
-    ret = jv_bool(jv_is_valid(elem));
+    if (jvp_number_is_nan(k)) {
+      ret = jv_false();
+    } else {
+      jv elem = jv_array_get(t, (int)jv_number_value(k));
+      ret = jv_bool(jv_is_valid(elem));
+      jv_free(elem);
+    }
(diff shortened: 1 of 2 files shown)


emanuele6:
`[0,1,2] | has(nan)` already returns `false` for me.

```
$ jq --version    # jq 1.6
jq-1.6
$ jq -n '[0,1,2] | has(nan, (0 / 0), (100 | asin))'
false
false
false
$ ./jq --version  # jq master
jq-1.6-269-g14e5e63-dirty
$ ./jq -n '[0,1,2] | has(nan, (0 / 0), (100 | asin))'
false
false
false

emanuele6:
Still, the result of casting nan from a floating point type to integer type is unspecified in C, so it makes sense to check `!is_nan` before doing that.

emanuele6 (approved):

itchyny:
On my macOS (M1 Mac), your command emits three true. Likely caused by the difference of signed bit of NaN between architectures (ref: https://twitter.com/robknight_/status/1610638557118349317)

nicowilliams:
> Still, the result of casting nan from a floating point type to integer type is unspecified in C, so it makes sense to check `!is_nan` before doing that.

Hmm, that explicit cast was not even necessary, and C does convert from `double` to integral types -- these implicit conversions are well-defined, though I don't have the spec in front of me and I'm not sure what an implicit conversion of `NaN` to `int` does.

So, yeah, let's merge this.

nicowilliams:
> On my macOS (M1 Mac), your command emits three true. Likely caused by the difference of signed bit of NaN between architectures (ref: https://twitter.com/robknight_/status/1610638557118349317)

TIL!

nicowilliams:
Thanks!

PR #2673, opened 2023-07-09
Fix the latest manual on string multiplication by zero

itchyny:
Based on the discussion in #2142, we changed the behavior of `"x" * 0`. So I fixed the manual.


nicowilliams:
Thanks!

PR #2677, opened 2023-07-09
Fix empty regular expression matches (fix #2565)

### LINKED ISSUE #2565: Zero length regular expression match misbehaviour
## Description
Some regular expressions behave weirdly when matches are zero length - they miss matches or match too often.

### Case 1
Input:
```
jq --null-input --compact-output '"xx"|match("";"g")'
```
Output:
```
{"offset":0,"length":0,"string":"","captures":[]}
{"offset":1,"length":0,"string":"","captures":[]}
```
Expected:
```
{"offset":0,"length":0,"string":"","captures":[]}
{"offset":1,"length":0,"string":"","captures":[]}
{"offset":2,"length":0,"string":"","captures":[]}
```
There should be a match at the end of the string too.

### Case 2
Input:
```
$ jq --null-input --compact-output '"xx"|match("$";"g")'
```
Output:
```
{"offset":2,"length":0,"string":"","captures":[]}
{"offset":2,"length":0,"string":"","captures":[]}
```
Expected:
```
{"offset":2,"length":0,"string":"","captures":[]}
```
The same match shouldn't show up twice. (Extra matches seem to be returned in number equal to the length of the string, for some reason. But they're all the same match at the same offset.)

## Environment
 - OS and Version: Ubuntu
 - jq version: master branch, commit cff5336ec71b6fee396a95bb0e4bea365e0cd1e8

## Additional context
I stumbled on this trying to fix https://github.com/stedolan/jq/issues/2148 and found that my [attempted solution](https://github.com/stedolan/jq/pull/2564) didn't behave like I expected it to. I tried to isolate the problem and ended up with these.

itchyny:
As reported by #2565, `"ab" | match(""; "g")` should yield 3 matches at offset 0, 1, and 2. Fixing this behavior also fixes `"a" | gsub(""; "a")` to emit `"aaa"` not `"aa"`. Fixes #2565.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1002,7 +1002,7 @@ static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
             jv_string((char*)ebuf)));
       break;
     }
-  } while (global && start != end);
+  } while (global && start <= end);
   onig_region_free(region,1);
   region = NULL;
   if (region)
--- a/tests/onig.test
+++ b/tests/onig.test
@@ -1,11 +1,15 @@
 # match builtin
 [match("( )*"; "g")]
 "abc"
-[{"offset":0, "length":0, "string":"", "captures":[]},{"offset":1, "length":0, "string":"", "captures":[]},{"offset":2, "length":0, "string":"", "captures":[]}]
+[{"offset":0, "length":0, "string":"", "captures":[]}, {"offset":1, "length":0, "string":"", "captures":[]}, {"offset":2, "length":0, "string":"", "captures":[]}, {"offset":3, "length":0, "string":"", "captures":[]}]
 
 [match("( )*"; "gn")]
 "abc"
 []
+
+[match(""; "g")]
+"ab"
+[{"offset":0,"length":0,"string":"","captures":[]},{"offset":1,"length":0,"string":"","captures":[]},{"offset":2,"length":0,"string":"","captures":[]}]
 
 [match("a"; "gi")]
 "āáàä"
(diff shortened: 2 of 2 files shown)


pkoppstein:
@itchyny - Congratulations!  

You could remove the disclaimer at onig.test:91
("The following is a regression test ...")

nicowilliams:
Thanks!

emanuele6:
> Fixing this behavior also fixes "a" | gsub(""; "a") to emit "aaa" not "aa".

I think that is a bug, it should emit `"aa"`, no?

itchyny:
Why? Empty regex matches every character boundaries. Easier example is `"xyz" | gsub(""; "a")` should be `axayaza`

emanuele6:
@itchyny 
Oh, you are right, I got confused.

pkoppstein:
> I think that is a bug, it should emit "aa", no?

No.  @itchyny and (not surprisingly :-) gojq are correct.

In any case, gsub relies (and ought to rely) on `match(_;"g")`.  In this case, the call to `match` produces two results, ergo two insertions.

emanuele6:
I was used to the sed behaviour where `sed 's/1*/2/g' <<< 11'` outputs `2`, but it looks like in javascript and other languages, if the pattern is able to match the empty string, it matches the empty string at the end of the input even if the previous match goes up to the last character of the input: `node -p '"1111".replaceAll(/1*/g, "2")'` outputs `22`

emanuele6:
`jq`'s `gsub` used to behave like sed's `s///g`, but now it behaves more like javascript's `replaceAll` after this patch:
```console
$ jq -n '"1111" | gsub("1*";"2")'
2
$ ./jq -n '"1111" | gsub("1*";"2")'
22
```

Fine, I guess

PR #2678, opened 2023-07-09
Fix manual on paths/1 to use boolean filter for its argument

### LINKED ISSUE #2288: Error selecting values using paths(node_filter)
**Describe the bug**
Was trying to delete all null objects/values from my JSON, but using the `paths(node_filter)` function to generate their paths was unsuccessful.

**To Reproduce**
Using an example from the Docs, and filter as used on jqplay (using // for comment just over here)

    Input : [[],{},1,"foo",null,true,false]
    paths(nulls) // Empty o/p -
    paths(nulls | not) // o/p - [4]
    paths(values) // o/p - [0, 1, 2, 3, 5]
    paths(values | not) // o/p - [6]
    paths(booleans) // o/p - [5]
    paths(booleans | not) // o/p - [6]
    paths(. == null) // o/p - [4] | filter is similar to nulls but this works unlike nulls

The same issue seems to extend to all other `node_filter` it seems, since only true values are produced, for `scalar`, and false or null values, for `scalar | not`.

**Expected behavior**
For any given pre-defined `node_filter` the output should consist of the values satisfying them.

    Input : [[],{},1,"foo",null,true,false]
    paths(nulls) // Expected o/p - [4]
    paths(nulls | not) // o/p - [0, 1, 2, 3, 5, 6]
    paths(values) // o/p - [0, 1, 2, 3, 5, 6]
    paths(values | not) // o/p - [4]
    paths(booleans) // o/p - [5, 6]
    paths(booleans | not) // o/p - [0, 1, 2, 3, 4]
    paths(. == null) // o/p - [4] | filter is now similar to nulls

 - OS and Version:
   - Windows 10 (WSL2 Ubuntu 20.04)
   - Linux
     - Ubuntu 20.04
     - Mint 19.3
 - jq version - 1.6

itchyny:
The manual of `paths/1` explains its usage the argument filter yields a boolean. Nevertheless the following sentence uses non-boolean filter `numbers` and also the definition of `leaf_paths` filter was wrong (#1163).  The filter `leaf_paths` was deprecated 9 years ago and was removed in #2666. There is still remaining confusing manual so I fixed it in this PR. Along with #2666, this PR resolves #2288.


wader (approved):

docs/content/manual/manual.yml
```diff
@@ -1148,15 +1148,15 @@ sections:
           (except it does not output the empty list, representing .
           itself).
 
-          `paths(f)` outputs the paths to any values for which `f` is true.
-          That is, `paths(numbers)` outputs the paths to all numeric
+          `paths(f)` outputs the paths to any values for which `f` is `true`.
+          That is, `paths(type == "number")` outputs the paths to all numeric
```
wader, line 1152: This got me thinking that i several times wanted to have `is_number` etc functions, for fq i have these https://github.com/wader/fq/blob/d15a41f9430f869dc6d11ac360f56c29e3a1b45e/pkg/interp/internal.jq#L147C1-L153 because i kept needing it

nicowilliams: `def paths(node_filter): . as $dot|paths|select(. as $p|$dot|getpath($p)|node_filter);`

`f` doesn't have to output `true` -- it just has to not be `empty` or `null`.

nicowilliams: I mean, maybe that's a bug...

But separately from that, jq considers any non-null, non-empty [EDIT: and non-`false`] value to be "trueish".

emanuele6: and non-false. :^)

emanuele6: Also empty is not really a falsish value, it really stops execution for that expression. e.g. `if empty then "yes" else "no" end` outputs nothing, not `"no"`. Anyway, this is all intended in my opinion. `false` and `null` are the only false values, and `empty` works as it should, and when you use it inside `select()` it is effectively the same as false.

wader: Do want to start using wording like "... for which `f` is considered true" or "... for which `f` is trueish" and have some separate section about jq and truthiness? "truthiness" i currently mentioned in the docs for if-then-else

itchyny: As a user manual, it is enough to mention that the filter works well with boolean filter arguments. No need to mention all its behaviors.

emanuele6 (approved):

docs/content/manual/manual.yml
nicowilliams: Yes, speaking of "empty values" is not exactly accurate.  It's a bit of a colloquialism in jq speak.  In the context of `select/1` though, `empty` really is falseish :)

nicowilliams: @wader I'm not sure that it's a good idea to document all the places where trueish values work, yeah.

nicowilliams:
Thanks!

docs/content/manual/manual.yml
wader: @nicowilliams  true :)

nicowilliams: ish

PR #2680, opened 2023-07-09
clarify explanation of map_values, including description of new behavior

pkoppstein:
#2466

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -911,25 +911,31 @@ sections:
           to the values of `.[]`.
           
           In the absence of errors, `map(f)` always outputs an array
-          whereas `map_values(f)` outputs an array or an object
-          depending on the type of the input.
+          whereas `map_values(f)` outputs an array if given an array,
+          or an object if given an object.
 
-          One important difference is that whereas `map(f)` simply
-          forms an array from all the values of `($x|f)` for each value,
-          $x, of the input, `map_values(f)` only uses `first($x|f)`.
+          When the input to `map_values(f)` is an object, the output
+          object has the same keys as the input object except for
+          those keys whose values when piped to `f` produce no values
+          at all.
+
+          The key difference between `map(f)` and `map_values(f)` is
+          that the former simply forms an array from all the values of
+          `($x|f)` for each value, $x, in the input array or object,
+          but `map_values(f)` only uses `first($x|f)`.
 
           Specifically, for object inputs, `map_value(f)` constructs
-          the output object by examining in turn the values of
+          the output object by examining in turn the value of
           `first(.[$k]|f)` for each key, $k, of the input.  If this
-          expression produces no values (i.e., is equivalent to
-          `empty`), the corresponding key will be dropped; otherwise,
-          the output object will have this value at the key, $k.
+          expression produces no values, then the corresponding key
+          will be dropped; otherwise, the output object will have that
+          value at the key, $k.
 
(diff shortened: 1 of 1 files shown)


docs/content/manual/manual.yml
```diff
@@ -925,6 +929,15 @@ sections:
             input: '{"a": 1, "b": 2, "c": 3}'
             output: ['{"a": 2, "b": 3, "c": 4}']
 
+          - program: 'map(., .)'
+            input: '[1,2]'
+            output: ['[1,1,2,2]']
+
+          - program: 'map_values(. // empty)'
+            input: '{"a": null, "b": true, "c": false}'
+            output: ['{"b":true}']
```
wader, line 967: Too much to have object with `map` and array with `map_values` examples?

wader (approved):

docs/content/manual/manual.yml
```diff
@@ -908,14 +908,18 @@ sections:
...
+          array. For example, `map(.+1)` will increment each element
+          of an array of numbers, and `map(.,.)` will duplicate each
+          element of the input array.
 
-          `map(f)` is equivalent to `[.[] | f]`. In fact, this is how
-          it's defined. Similarly, `map_values(f)` is defined as `.[] |= f`.
+          `map(f)` is equivalent to `[.[] | f]`. In fact, this is how it's defined.
 
+          `map_values(f)` is similarly defined as `.[] |= f`.  This
+          essentially means that each value, $v, in the input array or
```
nicowilliams: `s/, [$]v,//`

docs/content/manual/manual.yml
```diff
@@ -908,14 +908,18 @@ sections:
...
+          object is replaced by `first($v|f)`, it being understood
+          that if the input is an object, then keys for which this
```
nicowilliams: Array values for which `f` returns `empty` also get deleted.

I liked the original text better.  Staring at the definitions of `map/1` and `map_values/1` the difference between them isn't quite obvious, but the original text here makes it obvious: `map/1` always returns an array, while `map_values/1` returns an array if its input was an array, or an object if its input was an object.

pkoppstein: Tx. Hopefully the subsequent commit addresses all these concerns.  It is more systematic about the various possibilities.

docs/content/manual/manual.yml
```diff
@@ -906,15 +906,42 @@ sections:
...
-          For any filter `f`, `map(f)` will run that filter for each
-          element of the input array, and return the outputs in a new
-          array. `map(.+1)` will increment each element of an array of numbers.
+          For any filter `f`, `map(f)` and `map_values(f)` apply `f`
+          to each of the values in the input array or object, that is,
+          to the values of `.[]`.
+          
+          In the absence of errors, `map(f)` always outputs an array
+          whereas `map_values(f)` outputs an array or an object
+          depending on the type of the input.
```
nicowilliams: `In the absence of errors, ` is not needed.

Can you add something about the objects' keys being preserved?  Something like:

```
          `map(f)` always outputs an array whereas `map_values(f)` outputs
          an array if given an array, or an object if given an object.  In the
          case of `map_values(f)` applied to an object, the object output
          has all the same keys as the input object minus any for whose
          values `f` produced no new values.
```

something like that.

pkoppstein: Done. Tx.

nicowilliams:
Thanks!

PR #2684, opened 2023-07-09
Rename --nul-output to --raw-output0, abort on string containing NUL

### LINKED ISSUE #2683: Consider removing `-0`/`--nul-output` before releasing 1.7
See #1271.  @vdukhovni [believes](https://github.com/jqlang/jq/issues/1271#issuecomment-1627791117) we should remove it, and explains [why](https://github.com/jqlang/jq/issues/1271#issuecomment-945051603) in the comments.

itchyny:
The option naming --nul-output was confusing, especially when we have a
similar option for input stream in the future (--nul-input vs --null-input).
Based on the observation of other command line tools, we rename the option
to --raw-output0. We also drop the short option -0 to avoid confusion on
introducing the NUL-delimited input option.

Unlike the other command line tools outputting file names with NUL delimiter,
jq deals with JSON, and its strings may contain NUL character. To protect
users from the risk of injection attacks, we abort the program and print an
error message before outputting strings including NUL character. Closes #2683.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -165,12 +165,20 @@ cmp $d/out $d/expected
 printf "[1,2][3,4]\n" | $JQ -cs add > $d/out 2>&1
 cmp $d/out $d/expected
 
-# Regression test for -0 / --nul-output
+# Regression test for --raw-output0
 printf "a\0b\0" > $d/expected
-printf '["a", "b"]' | $JQ -0 .[] > $d/out 2>&1
+printf '["a", "b"]' | $VALGRIND $Q $JQ --raw-output0 .[] > $d/out
 cmp $d/out $d/expected
-printf '["a", "b"]' | $JQ --nul-output .[] > $d/out 2>&1
-cmp $d/out $d/expected
+printf "a\0" > $d/expected
+if printf '["a", "c\\u0000d", "b"]' | $VALGRIND $Q $JQ --raw-output0 .[] > $d/out; then
+  echo "Should exit error on string containing NUL with --raw-output0" 1>&2
+  exit 1
+elif [ $? -ne 5 ]; then
+  echo "Invalid error code" 1>&2
+  exit 1
+else
+  cmp $d/out $d/expected
+fi
 
 ## Test streaming parser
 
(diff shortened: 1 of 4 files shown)


nicowilliams:
We should first decide whether to keep `-0`.

nicowilliams:
Can you remove `-0` while you're at it?

itchyny:
There were some proposed options but I'm ok with this plan. What do you think?

tests/shtest
```diff
@@ -165,13 +165,21 @@ cmp $d/out $d/expected
 printf "[1,2][3,4]\n" | $JQ -cs add > $d/out 2>&1
 cmp $d/out $d/expected
 
-# Regression test for -0 / --nul-output
+# Regression test for --raw-output0
 printf "a\0b\0" > $d/expected
-printf '["a", "b"]' | $JQ -0 .[] > $d/out 2>&1
```
wader, line 170: Do we want to keep a `-0` test but now that it's not an option?

wader:
`NUL` meaning zero byte comes from ASCII? the man page for find calls it "ASCII NUL character (character code 0)" when describing `-print0`

itchyny:
Ah, I remember the concern is that it's not abort, but just skip. It exits with non-zero only when the last output contains NUL.

tests/shtest
nicowilliams: I'll leave that up to @itchyny.  I don't think there's a need for a test `-0` is now not an option.

nicowilliams:
LGTM!  Merge at your convenience.  I think this gets us very close to release ready.  I need to change the release key's name.  And I need to figure quite how to do the signing in the actions, unless you already know.

tests/shtest
itchyny: I don't think it's important to test that.

itchyny:
> Ah, I remember the concern is that it's not abort, but just skip. It exits with non-zero only when the last output contains NUL.

Sorry for pushing after reviewing, but I reconsidered that this should be an error. So I `break` on such case, and assert that it fails with non-zero status (actually 5), no further output is printed. The patch is now somehow smaller and readable IMO.

nicowilliams:
Thanks!

PR #2686, opened 2023-07-10
Build Docker image using pre-built executables

itchyny:
As suggested by @userdocs in #2665, I fixed the Docker build job to use the pre-built executables. Also I cleaned up the workflow syntax. The entire workflow was tested [here](https://github.com/itchyny/jq/actions/runs/5502969516) and published [here](https://github.com/itchyny/jq/pkgs/container/jq). The docker job runs within 30 seconds (compared to 40 minutes with entire emulation in #2652). BTW, I noticed `ld: warning: option -s is obsolete and being ignored` in macOS build steps. Do we explicitly use the strip or just drop the option? Related PRs: #2665, #2652.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
             target: arm64-apple-darwin
     runs-on: macos-13
     env:
-      LDFLAGS: -s
+      LDFLAGS: -dead_strip
       SUFFIX: macos-${{ matrix.arch }}
     steps:
       - name: Clone repository
@@ -162,6 +162,7 @@ jobs:
             --enable-static \
             --enable-all-static
           make -j"$(nproc)"
+          strip ./jq
           file ./jq
           cp ./jq jq-${{ env.SUFFIX }}
       - name: Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,60 +13,58 @@ jobs:
       fail-fast: false
       matrix:
         arch:
-          [
-            amd64,
-            arm64,
-            armel,
-            armhf,
-            i386,
-            mips,
-            mips64,
-            mips64el,
-            mips64r6,
-            mips64r6el,
-            mipsel,
-            mipsr6,
-            mipsr6el,
-            powerpc,
-            ppc64el,
-            s390x,
-            riscv64,
-          ]
+          - amd64
+          - arm64
+          - armel
+          - armhf
+          - i386
+          - mips
+          - mips64
+          - mips64el
+          - mips64r6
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -13,60 +13,58 @@ jobs:
...
           - arch: ppc64el
-            CC: "powerpc64le-linux-gnu"
-          - arch: s390x
-            CC: "s390x-linux-gnu"
+            CC: powerpc64le-linux-gnu
           - arch: riscv64
-            CC: "riscv64-linux-gnu"
+            CC: riscv64-linux-gnu
+          - arch: s390x
+            CC: s390x-linux-gnu
```
itchyny, line 67: Sorts alphabetically.

.github/workflows/ci.yml
```diff
@@ -13,60 +13,58 @@ jobs:
...
+          - mips64el
+          - mips64r6
+          - mips64r6el
+          - mipsel
+          - mipsr6
+          - mipsr6el
+          - powerpc
+          - ppc64el
+          - riscv64
+          - s390x
```
itchyny, line 32: Preference of YAML array.

.github/workflows/ci.yml
```diff
@@ -97,7 +95,7 @@ jobs:
             --with-oniguruma=builtin \
             --enable-static \
             --enable-all-static
-          make -j$(nproc)
+          make -j"$(nproc)"
```
itchyny, line 98: For actionlint (shellcheck).

.github/workflows/ci.yml
```diff
@@ -115,10 +113,10 @@ jobs:
           path: |
             test-suite.log
             tests/*.log
-      - name: Upload Artifacts
+      - name: Upload artifacts
         uses: actions/upload-artifact@v3
         with:
-          name: jq-${{ env.SUFFIX }}
+          name: jq-linux
```
itchyny, line 119: No need to create artifacts separately.

owenthereal (approved):
:shipit:

owenthereal:
> Do we explicitly use the strip or just drop the option?

As mentioned [here](https://github.com/jqlang/jq/pull/2665#issuecomment-1627629038), `LDFLAGS=-s` already strip symbols which has the same result as an explicit `strip`. You could compare the binary size.

owenthereal:
> BTW, I noticed ld: warning: option -s is obsolete and being ignored in macOS build steps. Do we explicitly use the strip or just drop the option?

Hmm...I think we may need an explicit `strip` for MacOS if that's the case.

itchyny:
Thank you!

PR #2689, opened 2023-07-10
Allow using  { $__loc__ }  as shorthand for  { __loc__: $__loc__ }

emanuele6:
n.b.: `{ $__loc__: "foo" }`  is still not allowed.

`$__loc__` cannot be overwritten, and it is always an object.
Using `{ $x: "foo" }` where `$x` is not a string is an error, so there is not much of a point in supporting that.


nicowilliams:
Thanks!

nicowilliams:
image

image

Funny.  `u20AC` is the euro sign, `€`.  What's "hidden" about `€`?  Maybe if the font one is using doesn't have that glyph?

Issue #2692, opened 2023-07-10
Website build deprecation warnings


nicowilliams:
```
: ; (source venv/bin/activate && (cd ../docs && pipenv install && pipenv run python build_website.py))
Installing dependencies from Pipfile.lock (62a1ce)…
Ignoring importlib-metadata: markers 'python_version < "3.8"' don't match your environment
Ignoring typing-extensions: markers 'python_version < "3.8"' don't match your environment
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 8/8 — 00:00:09
/home/nico/ws/jq/docs/build_website.py:29: DeprecationWarning: 'contextfunction' is renamed to 'pass_context', the old name will be removed in Jinja 3.1.
  env.globals['unique_id'] = contextfunction(
/home/nico/ws/jq/docs/build_website.py:26: DeprecationWarning: 'jinja2.Markup' is deprecated and will be removed in Jinja 3.1. Import 'markupsafe.Markup' instead.
  env.filters['markdownify'] = lambda input: Markup(markdown(input))
/home/nico/ws/jq/docs/build_website.py:27: DeprecationWarning: 'jinja2.Markup' is deprecated and will be removed in Jinja 3.1. Import 'markupsafe.Markup' instead.
  env.filters['no_paragraph'] = lambda input: Markup(re.sub(r"</?p>", '', input))
```

itchyny:
#2628 will resolve the issue.

PR #2693, opened 2023-07-10
Update docs/Pipfile.lock

emanuele6:
This fixed my `pipenv install` issue with python 3.11. I simply ran `pipenv lock` to bump the dependencies.


emanuele6:
After bumping the dependencies, the deprecation warnings of #2692, become `ImporError`s :(

```
$ (source venv/bin/activate && (cd ../docs && pipenv install && pipenv run python build_website.py))
Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project. You can set PIPENV_IGNORE_VIRTUALENVS=1 to force pipenv to ignore that environment and create its own instead. You can set PIPENV_VERBOSITY=-1 to suppress this warning.
Installing dependencies from Pipfile.lock (62a1ce)...
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project. You can set PIPENV_IGNORE_VIRTUALENVS=1 to force pipenv to ignore that environment and create its own instead. You can set PIPENV_VERBOSITY=-1 to suppress this warning.
Traceback (most recent call last):
  File "/home/emanuele6/.source_code/jq/docs/build_website.py", line 4, in <module>
    from jinja2 import Environment, FileSystemLoader, Markup, select_autoescape, contextfunction
ImportError: cannot import name 'Markup' from 'jinja2' (/home/emanuele6/.source_code/jq/build/venv/lib/python3.11/site-packages/jinja2/__init__.py)
```

nicowilliams:
Thanks!

nicowilliams:
> After bumping the dependencies, the deprecation warnings of #2692, become `ImporError`s :(

Oh no.  Maybe I shouldn't have merged this :/

nicowilliams:
Well, if we don't fix these in time for 1.7 we can revert this locally to make the website.

nicowilliams:
Hmm, I've reverted this as it did break the GHA workflow we have for website updates.

PR #2695, opened 2023-07-10
Quiet some warnings

nicowilliams:


src/jv_dtoa.c
```diff
@@ -1723,6 +1724,7 @@ gethex(struct dtoa_context* C,  CONST char **sp, U *rvp, int rounding, int sign)
 		  case '-':
 			esign = 1;
 			/* no break */
+			JQ_FALLTHROUGH;
```
itchyny, line 1727: We no longer need `no break` comment?

nicowilliams: Well, this isn't our code, and the `no break` comments never had any effect as far as compilers and linters go, and I think leaving them there as a bread-crumb on what to look for in history might be useful.  Though, yeah, they serve no purpose now _and_ you can still find them in the history, so I could drop those.  Your call.

itchyny: Hmm, maybe we shouldn't edit jv_dtoa.c for any purpose.

nicowilliams: That is definitely an option, though it's been edited some -- first of all by @stedolan.

src/jv.h
```diff
@@ -5,6 +5,20 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#if defined(DISPATCH_FALLTHROUGH)
+# define JQ_FALLTHROUGH DISPATCH_FALLTHROUGH
+#else
+# if defined(__GNUC__)
+#  if __GNUC__ >= 7
```
itchyny: Can we just check `__has_attribute(__fallthrough__)`?

src/jv.h
```diff
@@ -5,6 +5,20 @@
...
+#   define JQ_FALLTHROUGH __attribute__((fallthrough))
+#  else
+#   define JQ_FALLTHROUGH do {} while (0) /* fallthrough */
+#  endif
+# else
+#  define JQ_FALLTHROUGH do {} while (0) /* fallthrough */
```
itchyny: I found `fallthrough` macro in Linux code and the code using it looks nice, but should we use JQ_ prefix to avoid conflict?

src/jv_dtoa.c
itchyny: I couldn't find the original base code of dtoa.c when included into the repository, but I think the original diff was only `jvp_` prefix and `MALLOC`/`FREE` definitions. The latest version is available at https://www.netlib.org/fp/ so I downloaded dtoa.c and g_fmt.c to compare against jv_dtoa.c, and there seems to be some bug fixes. I don't confirm jq will be affected by the fixes, but hopefully we could follow the upstream. Since it's just compiler warnings, it's better to avoid edits, report to the original author (maybe along with 6151f204776db83047bfad1b76d5db0617a535f3), and then include the updates.

emanuele6:
This seems ok, but this PR doesn't silence all warnings:
```
src/util.c:3: warning: "_GNU_SOURCE" redefined                                       
    3 | #define _GNU_SOURCE               
      |                                                                              
<command-line>: note: this is the location of the previous definition
```

Since we are using `AC_USE_SYSTEM_EXTENSIONS` in `configure.ac`, on GNU/Linux, `-D_GNU_SOURCE=1` will be passed to the compiler.

I think either wrapping that `#define` in `#ifndef _GNU_SOURCE` or just removing it should fix that.

src/jv.h
nicowilliams: @itchyny well, we need one that is portable to not-Linux, including BSDs, OS X, and Windows, as well as different compilers.  I did struggle with: what header file to put this in, and what to call this macro.

It's very annoying that there isn't a universally available way to do this...

src/jv.h
nicowilliams: Sure.  I'll update this PR tomorrow.

src/jv_dtoa.c
nicowilliams: I remember finding the original and seeing that @stedolan made some modifications to the default representation of numbers.  But it's been years since I looked.

nicowilliams: https://www.google.com/search?q=dtoa.c+david+m+gay

nicowilliams:
> This seems ok, but this PR doesn't silence all warnings:

I have updated the caption!  :)

src/jv.h
emanuele6: @nicowilliams ping :)

nicowilliams:
```C
#if defined(DISPATCH_FALLTHROUGH)
# define JQ_FALLTHROUGH DISPATCH_FALLTHROUGH
#else
# if defined __has_attribute
#  if __has_attribute(__fallthrough__)
#   define JQ_FALLTHROUGH __attribute__((fallthrough))
#  endif
# else
#  define JQ_FALLTHROUGH do {} while (0) /* fallthrough */
# endif
#endif
```
look ok?

src/jv.h
```diff
@@ -5,6 +5,18 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#if defined(DISPATCH_FALLTHROUGH)
+# define JQ_FALLTHROUGH DISPATCH_FALLTHROUGH
```
itchyny: Came from [libdispatch](https://github.com/apple/swift-corelibs-libdispatch/blob/9566a131de09a472790f7b69ae661326cdedf306/dispatch/base.h#L144-L154)? Are you sure?

itchyny:
I thought `__has_attribute(__fallthrough__)` is the correct way of checking the availability of attributes, but after some observations in [code search](https://github.com/search?q=__has_attribute%28fallthrough%29&type=code), I noticed some complicated situations of clang. [Some](https://github.com/ludocode/ghost/blob/f582c521c8fba10a0beca83e505622efc1785207/include/ghost/language/ghost_fallthrough.h#L92-L99) [implementations](https://github.com/mcr/tcpdump/blob/6f6a66d390622f2c175835096f6154acfbea87b8/netdissect-stdinc.h#L418) [suggests](https://gitlab.isc.org/isc-projects/bind9/-/issues/3371) me that on Clang < 10, `__has_attribute(fallthrough)` produces 1 even if it does not support C-style (but only supports C++-style) `fallthrough` attribute. Clang 10 is released on Mar 24, 2020, so we may need to support older versions. Since `__has_attribute` check is not reliable, I end up concluded that checking the version is the stable way.

```c
#if (defined(__GNUC__) && __GNUC__ >= 7 && !defined(__clang__) || \
    defined(__clang__) && __clang_major__ >= 10)
# define JQ_FALLTHROUGH __attribute__((fallthrough))
#else
# define JQ_FALLTHROUGH
#endif
```

What do you think?

src/jv.h
nicowilliams: I copied this from [Heimdal](https://github.com/heimdal/heimdal) :)  The link is helpful.  I'll update this.

nicowilliams:
> [...] I noticed some complicated situations of clang.

Bummer.  ~Your link shows use of `JE_COMPILABLE` to deal with this -- that strikes me as the best option, and I'll integrate that now.~  I'll integrate a version of the code you proposed.

BTW, thank you for that research!  I should have done that myself.

src/jv_dtoa.c
```diff
@@ -2367,10 +2372,12 @@ jvp_strtod
 		case '-':
 			sign = 1;
 			/* no break */
+			JQ_FALLTHROUGH;
 		case '+':
 			if (*++s)
 				goto break2;
 			/* no break */
```
emanuele6, line 2379: redundant comment

src/jv_dtoa.c
```diff
@@ -3699,6 +3707,7 @@ jvp_dtoa
 		case 2:
 			leftright = 0;
 			/* no break */
```
emanuele6, line 3709: redundant comment

src/jv_dtoa.c
```diff
@@ -3707,6 +3716,7 @@ jvp_dtoa
 		case 3:
 			leftright = 0;
 			/* no break */
```
emanuele6, line 3718: redundant comment

src/jv_dtoa.c
```diff
@@ -2367,10 +2372,12 @@ jvp_strtod
 		case '-':
 			sign = 1;
 			/* no break */
```
emanuele6, line 2374: redundant comment

src/jv_dtoa.c
```diff
@@ -1723,6 +1724,7 @@ gethex(struct dtoa_context* C,  CONST char **sp, U *rvp, int rounding, int sign)
 		  case '-':
 			esign = 1;
 			/* no break */
```
emanuele6, line 1726: redundant comment

nicowilliams:
@emanuele6 I wanted to leave the original `no break` commentary in `src/jv_dtoa.c`.

emanuele6:
Ok, that's fine then

emanuele6:
Thanks!

PR #2696, opened 2023-07-10
Do not remove manpage prebuild file on build failure

itchyny:
I'm annoyed by the target removal here, since the target file is managed by git. Just like `tests/man.test`, I don't think we should remove this file on build failure.


emanuele6 (approved):

nicowilliams:
Thanks!

PR #2710, opened 2023-07-13
debug/1

### LINKED ISSUE #2111: Feature request: Add 1-arity built-in function debug(f) for more control over debug output

### LINKED ISSUE #2709: ER: debug/1
There have been several proposals for adding debug/1 for convenience.

In particular, @nicowilliams mentioned:

    def debug($msg): ("\($msg): \(.)|debug|empty), .;

This has good memory-management characteristics, but includes "\(.)" unconditionally, which reduces flexibility and can lead to very long `debug` messages that can potentially cause problems.  Thus I would like to propose a similar but simpler definition:

    def debug(msg): (msg | debug | empty), .;

This would allow tailored messages such as: `debug("entering foo(\(arg1); \(arg2))")`

while still allowing . to be prepended with a string, at a cost of just a few extra keystrokes:

    debug("hello: \(.)")

Or one could go with:

    debug("hello", .)

---
Footnote:

https://github.com/jqlang/jq/pull/2112 proposes:

    def debug(f): . as $val | f | debug | $val;

but this def effectively presupposes that f is a singleton, for if
one were to use it and wrote `debug("msg1"; "msg2")`, an undesirable fork would result.

The two defs above avoid this issue, and indeed make it easy to split up a debug message over more than one line, while keeping them together.

pkoppstein:
This def ensures the output of debug(m1,m2) is kept together.

Resolves #2709, resolves #2111, closes #2112

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -3075,13 +3075,32 @@ sections:
           to invoke jq with the -n command-line option, otherwise
           the first entity will be lost.
 
-      - title: "`debug`"
+      - title: "`debug`, `debug(msgs)`"
         body: |
 
-          Causes a debug message based on the input value to be
-          produced.  The jq executable wraps the input value with
-          `["DEBUG:", <input-value>]` and prints that and a newline on
-          stderr, compactly.  This may change in the future.
+          These two filters are like `.` but have as a side-effect the
+          production of one or more messages on stderr.
+
+          The message produced by the `debug` filter has the form
+
+              ["DEBUG:",<input-value>]
+
+          where `<input-value>` is a compact rendition of the input
+          value.  This format may change in the future.
+
+          The `debug(msgs)` filter is defined as `(msgs | debug | empty), .`
+          thus allowing great flexibility in the content of the message,
+          while also allowing multi-line debugging statements to be created.
+
+          For example, the expression:
+
+              1 as $x | 2 | debug("Entering function foo with $x == \($x)", .) | (.+1)
+
+          would produce the value 3 but with the following two lines
+          being written to stderr:
+
+              ["DEBUG:","Entering function foo with $x == 1"]
+              ["DEBUG:",2]
 
       - title: "`stderr`"
         body: |
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -266,6 +266,9 @@ def pick(pathexps):
   | reduce path(pathexps) as $a (null;
       setpath($a; $in|getpath($a)) );
 
+# ensure the output of debug(m1,m2) is kept together:
+def debug(msgs): (msgs | debug | empty), .;
+
 # SQL-ish operators here:
 def INDEX(stream; idx_expr):
   reduce stream as $row ({}; .[$row|idx_expr|tostring] = $row);
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -3353,8 +3353,41 @@ Outputs all remaining inputs, one by one\.
 .P
 This is primarily useful for reductions over a program\'s inputs\. Note that when using \fBinputs\fR it is generally necessary to invoke jq with the \-n command\-line option, otherwise the first entity will be lost\.
 .
-.SS "debug"
-Causes a debug message based on the input value to be produced\. The jq executable wraps the input value with \fB["DEBUG:", <input\-value>]\fR and prints that and a newline on stderr, compactly\. This may change in the future\.
+.SS "debug, debug(msgs)"
+These two filters are like \fB\.\fR but have as a side\-effect the production of one or more messages on stderr\.
+.
+.P
+The message produced by the \fBdebug\fR filter has the form \fB["DEBUG:", <input\-value>]\fR where \fIinput\-value\fR is a compact rendition of the input value\. This format may change in the future\.
+.
+.P
+The \fBdebug(msgs)\fR filter is defined as \fB(msgs | debug | empty), \.\fR thus allowing great flexibility in the content of the message, while also allowing multi\-line debugging statements to be created\.
+.
+.P
+For example, the expression:
+.
+.IP "" 4
+.
+.nf
+
+`1 as $x | 2 | debug("Entering function foo with $x == \e($x)", \.) | (\.+1)`
+.
+.fi
+.
+.IP "" 0
+.
+.P
+would produce the value 3 but with the following two lines being written to stderr:
+.
+.IP "" 4
+.
+.nf
+
+["DEBUG:","Entering function foo with $x == 1"]
+["DEBUG:",2]
+.
+.fi
+.
+.IP "" 0
 .
 .SS "stderr"
 Prints its input in raw and compact mode to stderr with no additional decoration, not even a newline\.
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -3357,7 +3357,20 @@ This is primarily useful for reductions over a program\'s inputs\. Note that whe
 These two filters are like \fB\.\fR but have as a side\-effect the production of one or more messages on stderr\.
 .
 .P
-The message produced by the \fBdebug\fR filter has the form \fB["DEBUG:", <input\-value>]\fR where \fIinput\-value\fR is a compact rendition of the input value\. This format may change in the future\.
+The message produced by the \fBdebug\fR filter has the form
+.
+.IP "" 4
+.
+.nf
+
+["DEBUG:",<input\-value>]
+.
+.fi
+.
+.IP "" 0
+.
+.P
+where \fB<input\-value>\fR is a compact rendition of the input value\. This format may change in the future\.
 .
 .P
 The \fBdebug(msgs)\fR filter is defined as \fB(msgs | debug | empty), \.\fR thus allowing great flexibility in the content of the message, while also allowing multi\-line debugging statements to be created\.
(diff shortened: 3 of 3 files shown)


docs/content/manual/manual.yml
```diff
@@ -3075,13 +3075,29 @@ sections:
...
-          `["DEBUG:", <input-value>]` and prints that and a newline on
-          stderr, compactly.  This may change in the future.
+          These two filters are like `.` but have as a side-effect the
+          production of one or more messages on stderr.
+          
+          The message produced by the `debug` filter has the form
+          `["DEBUG:", <input-value>]` where <input-value> is a compact rendition of 
+          the input value.  This format may change in the future.
+
+          The `debug(msgd)` filter is defined as `(msgs | debug | empty), .`
```
emanuele6: msgd => msgs

src/builtin.jq
```diff
@@ -265,7 +265,8 @@ def pick(pathexps):
   . as $in
   | reduce path(pathexps) as $a (null;
       setpath($a; $in|getpath($a)) );
-
+# ensure the output of debug(m1,m2) is kept together:
+def debug(msgs): (msgs | debug | empty), .;
 # SQL-ish operators here:
```
emanuele6: add blank lines to separate it from `pick/1` and SQL-like functions?

emanuele6:
You also need to regenerate `jq.1.prebuilt` after having modified `manual.yml`.
`workflows/manpage.yml` was incorrectly configured to only run on push, and not on pull requests so the CI for this PR is green even though it is red on your fork https://github.com/pkoppstein/jq/actions/runs/5538611539/jobs/10108730569

55249e2 should fix that.

docs/content/manual/manual.yml
```diff
@@ -3075,13 +3075,29 @@ sections:
...
 
-          Causes a debug message based on the input value to be
-          produced.  The jq executable wraps the input value with
-          `["DEBUG:", <input-value>]` and prints that and a newline on
-          stderr, compactly.  This may change in the future.
+          These two filters are like `.` but have as a side-effect the
+          production of one or more messages on stderr.
+          
+          The message produced by the `debug` filter has the form
+          `["DEBUG:", <input-value>]` where <input-value> is a compact rendition of
```
emanuele6: There is an extra space after `of` at the end of this line, it will result in a double space in the manual when the site/manpage are generated, so it should be removed

docs/content/manual/manual.yml
```diff
@@ -3075,13 +3075,29 @@ sections:
...
+      - title: "`debug`", "`debug(msgs)`"
         body: |
 
-          Causes a debug message based on the input value to be
-          produced.  The jq executable wraps the input value with
-          `["DEBUG:", <input-value>]` and prints that and a newline on
-          stderr, compactly.  This may change in the future.
+          These two filters are like `.` but have as a side-effect the
+          production of one or more messages on stderr.
+
```
emanuele6: This line has some unnecessary spaces

docs/content/manual/manual.yml
```diff
@@ -3075,13 +3075,29 @@ sections:
           to invoke jq with the -n command-line option, otherwise
           the first entity will be lost.
 
-      - title: "`debug`"
+      - title: "`debug`", "`debug(msgs)`"
```
emanuele6: Actually, the CI (on your fork) is red because this is invalid YAML syntax, so the man page cannot even be generated; not only because jq.1.prebuilt is not up to date ;) 

https://github.com/pkoppstein/jq/actions/runs/5538611539/jobs/10108730569

It should be: <code>"&#96;debug&#96;, &#96;debug(msgs)&#96;"</code>

itchyny:
Some tests could be added to shtest?

docs/content/manual/manual.yml
pkoppstein: Thanks for the hint about the extra space.  As for jq.1.prebuilt - I've tried many ways and many times to get the python scripts to work, but with no success.  E.g. just now, after successfully installing lxml:

$ pip3 list | grep lxml
lxml              4.9.2

$ pip list | grep lxml
lxml              4.9.2

$ pipenv run ./build_manpage.py
ModuleNotFoundError: No module named 'lxml'

$ pipenv run python3 build_manpage.py
ModuleNotFoundError: No module named 'lxml'

#### Similar problems with man.test, by the way:
$ pipenv run ./build_mantests.py
ModuleNotFoundError: No module named 'yaml'

It had been my impression that updating these downstream files could be postponed e.g. until the site was regenerated.  But again, please feel free.

pkoppstein:
> Some tests could be added to shtest?

There aren't any for `debug` so that's one reason I didn't reach out for help.  I don't know how to test shtest - it uses VALGRIND.  Please feel free to add.

docs/content/manual/manual.yml
emanuele6: @pkoppstein 
You don't need to have `yaml`, `lxml`, etc installed globally on your system, it only matters they are installed in the `pipenv`.

I think you just need to run `(cd ./docs && pipenv install)` then run `./configure --enable-docs #something else if you want`, and finally you should be able to just run `make jq.1.prebuilt` and `make tests/man.test`.

EDIT: I tested on a fresh clone of jqlang/jq and those steps worked for me

pkoppstein: Thanks, that worked. `man.test` did not change.  It's missing some tests though.

Any hints on how to avoid the Oniguruma submodule problem when attempting to "squash" commits?   
(GitHub Desktop says: "Error: Unable to squash when changes are present on your branch. The following files would be overwritten: modules/onigurma ")

emanuele6: @pkoppstein 
I think you just need to run `git submodule update`, you have probably accidentally manually pulled newer commits in `modules/oniguruma`.

pkoppstein: @emanuele6 - Thanks, that worked. I hope that the squashed commit won't cause anyone any grief.

Btw, I am quite sure that I didn't do anything (deliberately or otherwise) to the modules/oniguruma repository apart from whatever was done when I first cloned jqlang/jq.  (I did notice that "modules/oniguruma" was marked as having changed when I created the clone.)

emanuele6: Ah, maybe then you have not run `git submodule update` after pulling 7c1efd6 (2 weeks ago) from master?
If you don't run `git pull` with `--recurse-submodule`, and you have not set the `submodule.recurse` gitconfig setting to `true`, you need to do that manually after you pull a commit that updates the submodule I think.

nicowilliams:
I'm ok with the continued lack of tests for `debug` -- partly it's because its output form shouldn't really be considered stable, and partly because it is meant to have no impact on the jq program itself and what it does.  But if you'd like to write a test go ahead, otherwise we should merge this.

src/builtin.jq
emanuele6: @pkoppstein can you fix this please?

docs/content/manual/manual.yml
emanuele6: @pkoppstein I just to pull 092fef7 (that upgrades `modules/oniguruma`), with just `git pull` without `submodule.recurse` set to true, and indeed I needed to run `git submodule update` manually.

src/builtin.jq
pkoppstein: Done.

docs/content/manual/manual.yml
```diff
@@ -3075,13 +3075,29 @@ sections:
...
+          `["DEBUG:", <input-value>]` where <input-value> is a compact rendition of
+          the input value.  This format may change in the future.
+
+          The `debug(msgs)` filter is defined as `(msgs | debug | empty), .`
+          thus allowing great flexibility in the content of the message,
+          while also allowing multi-line debugging statements to be created.
+
+          For example, the expression:
+
+              `1 as $x | 2 | debug("Entering function foo with $x == \($x)", .) | (.+1)`
```
emanuele6: I think this should not be wrapped in <code>&#96;</code>. Other than that LGTM

emanuele6: Yes, otherwise it is rendered like this on the man page.

file

And with literal <code>&#96;</code> on the website

emanuele6 (approved):

PR #2712, opened 2023-07-13
Parse  nan  in JSON as NaN instead of triggering a parse error

### LINKED ISSUE #2021: Inconsistent behavior with NaN
**Describe the bug**
Jq does not accept `nan`, but accepts `-nan`.

**To Reproduce**
```
echo nan | jq .
parse error: Invalid literal at line 2, column 0

echo -nan | jq .
null

echo NaN | jq .
null

echo -NaN | jq .
null
```

**Expected behavior**
Not sure if jq should accept `nan`, or reject `-nan`. I would vote for rejecting `-nan`, as some popular applications like MongoDB do.

 - OS: Kubuntu 19.04
 - jq version: jq-1.5-1-a5b5cbe

emanuele6:
Fixes #2021

### DIFF
--- a/src/jv_aux.c
+++ b/src/jv_aux.c
@@ -215,6 +215,7 @@ jv jv_has(jv t, jv k) {
   } else if (jv_get_kind(t) == JV_KIND_ARRAY &&
              jv_get_kind(k) == JV_KIND_NUMBER) {
     if (jvp_number_is_nan(k)) {
+      jv_free(t);
       ret = jv_false();
     } else {
       jv elem = jv_array_get(t, (int)jv_number_value(k));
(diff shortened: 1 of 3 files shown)


emanuele6:
I wrote a test that uses `fromjson/1`, and it failed because it triggers a memory leak, I am investigating

emanuele6:
Hmm, actually this was unrelated to my patch and the use of `fromjson` in the test; b5c4c3d67decec22d34f494a200af59bbcadcc80 introduced a memory leak.

I am not sure why the CI didn't catch it. Does the CI run tests without `valgrind`?

emanuele6:
Yes, CI configures with `--disable-valgrind`; it should probably not be disabled, no?

itchyny (approved):
LGTM

nicowilliams:
> Yes, CI configures with `--disable-valgrind`; it should probably not be disabled, no?

I would make one workflow not do that.  Not all of them, because there's no need to waste that much build capacity.

emanuele6:
@nicowilliams Makes sense; also only the `amd64` CI runs tests anyway.

emanuele6:
Hmm, afaict `--disable-valgrind` only matters if you run `make check`. So the fix is to remove `--disable-valgrind` for all workflows anyway. There is no need to conditionally disable it with something like `${{ matrix.arch != 'amd64' && "--disable-valgrind" }} \`.

PR #2714, opened 2023-07-13
On linux amd64, run tests with valgrind

emanuele6:
This flag only matters if you run  make check  and we only run tests (with make check) on linux amd64, macos amd64, and windows amd64.

We want to run tests using valgrind on those target to avoid introducing memory leaks, so let's not use `--disable-valgrind`.

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -1,0 +1,42 @@
+name: valgrind
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  linux:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Install packages
+        run: |
+          sudo apt-get update
+          sudo apt-get upgrade
+          sudo apt-get install -y automake autoconf libtool valgrind
+      - name: Build
+        run: |
+          autoreconf -i
+          ./configure \
+            --disable-maintainer-mode \
+            --disable-docs \
+            --with-oniguruma=builtin
+          make -j"$(nproc)"
+          file ./jq
+      - name: Test
+        run: |
+          make check
+          git diff --exit-code
+      - name: Upload Test Logs
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-logs-valgrind-linux
+          retention-days: 7
+          path: |
+            test-suite.log
+            tests/*.log


emanuele6:
> This flag only matters if you run make check and we only run tests (with make check) on linux amd64, macos amd64, and windows amd64.

On all the other systems we just build `jq` without running tests. Wouldn't it be better to run tests on all workflows, using valgrind only on amd64 systems?

Only building `jq` with `gcc` on different systems without running it, doesn't seem that useful.

itchyny:
Running tests of different arch needs emulation with lots of duration so skipping is okay. You will need to install valgrind though (`checking for valgrind... no`).

.github/workflows/ci.yml
```diff
@@ -147,6 +147,10 @@ jobs:
           # brew update sometimes fails with "Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask failed!"
           brew update || brew update-reset
           brew install autoconf automake libtool
+      - name: Install valgrind
+        if: ${{ matrix.arch == 'amd64' }}
+        run: |
+          brew install valgrind
```
nicowilliams: valgrind works on OS X?

emanuele6: No, I deleted the wrong hunk!
But initially I thought `valgrind` worked also on macOS, apparently not. :/

emanuele6:
Hmm, CIs also configure with `--enable-static --enable-all-static`, and that makes valgrind report tonnes of "Conditional jump or move depends on uninitialised value(s)" errors

emanuele6:
It found a memory leak in parser.y =)

.github/workflows/ci.yml
pkoppstein: @emanuele6 - On Mac, `brew install valgrind` claims it only works on Linux, but there is a valgrind variant that works on Mac OS:
```
brew tap LouisBrunner/valgrind; brew install --HEAD LouisBrunner/valgrind/valgrind
```
When I ran test/shtest, it reveals the following:
```
+ valgrind --error-exitcode=1 --leak-check=full --suppressions=/Users/user/github/jqlang/tests/onig.supp --suppressions=/Users/user/github/jqlang/tests/local.supp -q /Users/user/github/jqlang/tests/../jq -Rse '. == "a\u0000b\nc\u0000d\ne"' /var/folders/15/wd4cntyn53x87v4x86wxysyc0000gn/T/jqXXXXXX.koOzeC6M/input
--57220-- run: /usr/bin/dsymutil "/Users/user/github/jqlang/tests/../jq"
==57220== Conditional jump or move depends on uninitialised value(s)
==57220==    at 0x7FF80BB3EA3F: ???
==57220==    by 0x7FF80BA1A8ED: fgets (in /usr/lib/system/libsystem_c.dylib)
==57220==    by 0x10003142E: jq_util_input_read_more (util.c:305)
==57220==    by 0x100030F67: jq_util_input_next_input (util.c:411)
==57220==    by 0x1000029D7: main (main.c:690)
==57220== 
true
==57220== 24 bytes in 1 blocks are possibly lost in loss record 4 of 30
....
==57220== 1,888 bytes in 59 blocks are definitely lost in loss record 30 of 30
```

Can you tell whether this is a jq issue, or a valgrind-on-a-Mac issue?

itchyny:
```yacc
'{' MkDict '}' {
  block o = gen_const_object($2);
  if (o.first != NULL)
    $$ = o;
  else
    $$ = BLOCK(gen_subexp(gen_const(jv_object())), $2, gen_op_simple(POP));
} |
```
`o` leaks...?

emanuele6:
No, it's not that. `block_free(o)` is a no-op if `!o.first`.
What gets leaked is the input of `halt_error/1`, I tried to figure out where the issue was earlier, but I didn't have much luck.
I will try to figure it out later when I have time.

itchyny:
https://github.com/jqlang/jq/blob/07ef97c414ef0370683a30a392223610912025fb/src/main.c#L226
`jv_copy` here is not good

.github/workflows/valgrind.yml
```diff
@@ -0,0 +1,51 @@
+name: valgrind
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "jq-*"
```
itchyny: We don't need to trigger on tags; all commits in master branch is already tested above.

emanuele6: I also thought it was useless, but I copied it from `ci.yml` because I thought maybe it was used for something

itchyny: Running CI on tags is important. When we create a release tag, the CI workflow runs twice for the branch and tag of the same commit. But since `github.ref` is different (`refs/tags/jq-1.X` vs `refs/heads/master`), release (and docker) job(s) only run on the tag.

emanuele6: Yeah, I just read `ci.yml` and noticed that it uses that to upload a release after a tag that starts with `jq-` is created; thank you

itchyny (approved):
LGTM!

nicowilliams:
Thank you!

.github/workflows/ci.yml
emanuele6: @pkoppstein `"Conditional jump or move depends on uninitialised value(s)"` for a libc function like `fgets()`, likely just means that you are running a static build of `jq` with `valgrind`, and `valgrind` doesn't work very well with static builds in general. I get the same kind of errors on GNU/Linux for a static build.

PR #2715, opened 2023-07-14
Fix memory leak on halt_error/1

itchyny:
I discovered where memory leak is occurring on using `halt_error/1`. I confirmed `make check` with valgrind enabled passes.
Ref: https://github.com/jqlang/jq/pull/2714#issuecomment-1635332661.


emanuele6:
Oh, yep, that's it! Thank you!

Issue #2716, opened 2023-07-14
support pick(last)


pkoppstein:
Using a version of jq with pick/1:
```
$ jq -nc '[[10,20],30] | pick(first) '
[[10,20]]

./jq -nc '[[10,20],30] | pick(.[length-1])'
[null,30]

$ jq -nc '[[10,20],30] | pick(last) '
jq: error (at <unknown>): Out of bounds negative array index

$ jq -nc '[[10,20],30] | pick(.[-1]) '
jq: error (at <unknown>): Out of bounds negative array index
```

So the failure of `pick(last)` could be resolved by redefining `def last: .[length-1];`
but is there a better solution?  It's not clear to me why pick(.[-1]) should fail.

itchyny:
Since `path(last)` yields `[-1]`, `null | setpath([-1]; 30)` fails with the negative index error.

pkoppstein:
Yes, but redefining last/1 as `.[length-1]` would resolve the issue, as suggested by:
```
$ jq -M -nc '[[10,20],30] | pick(.[length - 1])'
[null,30]
$ jq -M -nc '[[10,20],30] | pick(.[-1])'
jq: error (at <unknown>): Out of bounds negative array index
```

So I think the question is: is there a better way to resolve the issue?

itchyny:
Oops, I misread the last sentence of your comment. Yes, that looks like a reasonable change.

nicowilliams:
> So I think the question is: is there a better way to resolve the issue?

I think making ~`jv_set()`~ `jv_array_set()` take negative indices would be the better fix.

EDIT: Well, it tries to:

```C
jv jv_array_set(jv j, int idx, jv val) {
  assert(JVP_HAS_KIND(j, JV_KIND_ARRAY));

  if (idx < 0)   // <---- here
    idx = jvp_array_length(j) + idx;
  if (idx < 0) {
    jv_free(j);
    jv_free(val);
    return jv_invalid_with_msg(jv_string("Out of bounds negative array index"));
  }
  // copy/free of val,j coalesced
  jv* slot = jvp_array_write(&j, idx);
  jv_free(*slot);
  *slot = val;
  return j;
}
```

nicowilliams:
```
: ; ./jq -cn '
def dbg(f): (f|debug|empty),.;
def pick(pathexps):
    . as $in
  | reduce (path(pathexps)|dbg("path: \(.)")) as $a (
        null;
[...]

PR #2717, opened 2023-07-14
support pick(last); add to jq.test

pkoppstein:
Modify last/0 to allow e.g. pick(last) (q.v. #2716)

Include some tests for input, inputs, debug, debug(), pick()

Also some tidying of spacing in builtin.jq


tests/jq.test
```diff
@@ -1814,3 +1814,35 @@ true
 tojson | fromjson
 {"a":nan}
 {"a":null}
+
```
emanuele6, line 1843: Can you organize tests in sections like the rest of `jq.test`?

tests/jq.test
```diff
@@ -1814,3 +1814,35 @@ true
...
+[[10,20],30]
+[null,30]
+
+[inputs]
+null
+[]
+
+[input?]
+null
+[]
```
emanuele6: Calling `input/1` in `--run-tests` mode causes a segmentation fault which we should probably fix, but I am not sure how you expect those tests to work

tests/jq.test
pkoppstein: Not sure what you mean.  Do you mean: create new sections preceded by comments?

tests/jq.test
pkoppstein: I ran `--run-tests` without incident on a Mac.  So maybe there is a problem that needs to be fixed on some platforms?

UPDATE: It looks like you've found the source of the problem, but I'm not sure what protocol is appropriate here.  Should I remove the `input` test so this can be merged before your fix, or should I wait for your fix to be merged?  Or ....????

Regarding @nicowilliams's proposed solution to the underlying issue, it seems to me that since that change carries some risk whereas tweaking the def of last/0 does not, it would be appropriate to merge this PR in the near term (hopefully before it becomes "stale"), with a view to investigating the alternatives as time allows.

emanuele6: @pkoppstein The issue is that the struct field that specifies the input used by `input/1` is unitialised when running `--run-tests`. It could be that your system's implementation of `malloc()` returns values initialised to `0`, or something else, and those uninitaliased fields happened to be set to `0`, so it worked.
On GNU libc systems, `malloc()` typically returns memory initialised with `0x61` bytes for debug builds to help catch those problems, so the `jq->input_cb` field was set to `0x6161616161616161`, and, since it is not `NULL`, `input/1` tried to call it, triggering a segfault.
Anyway, anything could have happened.

tests/jq.test
emanuele6: AFAICT the format we are using for tests in `jq.tests` is:

`# section title`, 2 blank lines, and then a sequence of tests separated by a blank line, with optionally a subsection title.

```jq
# section1

# subsection 1.1

1
1
1

2
2
2

# subsection 1.2

3
3
3


# section 2

4
4
4
```

It is not consistent through the entire file, some earlier tests use
```jq
#
# section 1
#

# subsection 1.1

1
1
1
```

or
```jq
# section 1
1
1
1

2
2
2

# section 2
3
3
3
```

But generally thoughout the file, tests seem to be divided in sections that describe what they are testing, or what category of things they are testing, they are not just appended to the end without a title.

Also, since you are dividing those new tests with only one space from the previous ones, it looks like the are grouped with the tests related to parsing `nan` in JSON stings/inputs.

```jq
# nan is parsed as a valid NaN value from JSON

fromjson | isnan
"nan"
true

tojson | fromjson
{"a":nan}
{"a":null}

pick(.a.b.c)
null
{"a":{"b":{"c":null}}}

pick(first)
[1,2]
[1]

...
```

pkoppstein:
@nicowilliams @emanuele6 - I've merged this with https://github.com/jqlang/jq/commit/e79335e3a587d77954face2137fc3d4b5482129e
and run all the *.test test, so I believe this PR is ready for re-review.

nicowilliams:
I force pushed a rebase and re-arranging of the commits.  I dropped all the restyling of `src/builtin.jq` -- that should be done in a separate PR (do we have a style guide for jq code?  we probably should).

itchyny:
This PR is unable to merge due to too many unrelated coding style fixes.

pkoppstein:
@itchyny - nicowilliams wrote:

> I force pushed a rebase and re-arranging of the commits. I dropped all the restyling of src/builtin.jq 

~~Can't you get the necessary commits from what he's done?~~

I've closed this PR i.f.o. https://github.com/jqlang/jq/pull/2779

Thanks.

emanuele6:
@pkoppstein
You have restored the style changes @nicowilliams removed with a merge commit, an hour after he removed them. If that was accidental, I can revert that merge commit for you if you want.

pkoppstein:
@emanuele6 - Thanks for your offer, but in the meantime, I created another PR (https://github.com/jqlang/jq/pull/2779),
which is based on the most recent jqlang, so I'll close this PR, and hope for the best.

nicowilliams:
> @itchyny - nicowilliams wrote:
> 
> > I force pushed a rebase and re-arranging of the commits. I dropped all the restyling of src/builtin.jq
> 
> Can't you get the necessary commits from what he's done?

The problem is that after I force-pushed you then did a `git merge` and pushed again.

I recommend that you break out of the merge workflow.  Merging is a crutch you don't need.  Every time you want to say `git pull --merge`, or `git pull`, or `git merge`, you should imagine a kitten being mauled by a dog because of you doing that, then remember to try `git pull --rebase`, or `git fetch origin` then `git rebase origin/master` instead so that the kitten doesn't die.

Anyways, I get permission denied trying to push to this PR now.  What you should do:

```
$ # review what I pushed:
$ git log --patch --reverse $(git merge-base origin/master 8f5d3aee13128c169ace637e06ce3fd4405d1960)..8f5d3aee13128c169ace637e06ce3fd4405d1960
$ # if you agree to what I pushed, then:
$ git checkout tests-jq-test
$ git branch save-this-branch HEAD # so you don't lose your work
$ git reset --hard 8f5d3aee13128c169ace637e06ce3fd4405d1960 # this is what I force pushed
$ git push -f $your_repo_name_here tests-jq-test
```

then we can merge this.

pkoppstein:
@nicowilliams wrote:
> Merging is a crutch you don't need.

These days, I'm trying to steer clear of the git command line in favor of GitHub Desktop.  There's an option there to "Update from master", which I've sometimes used on a branch. Is that in effect what you're advising against?

nicowilliams:
> These days, I'm trying to steer clear of the git command line in favor of GitHub Desktop.  There's an option there to "Update from master", which I've sometimes used on a branch. Is that in effect what you're advising against?     

Evidently, yes.  If they have a "rebase to master" option please use that.

PR #2719, opened 2023-07-15
Initialise jq.input_cb{,_data} to NULL in jq_init()

emanuele6:
To avoid causing segmentation faults when `input/1` is called in a `jq_state` on which `jq_set_input_cb()` has not been called; e.g. the one used by `jq --run-tests`.

That segfault could also be fixed in run_jq_tests() by calling:
```c
jq_set_input_cb(jq, NULL, NULL);
```
But I think it makes sense to just make `jq_init()` initialise those values to NULL.

Ref: https://github.com/jqlang/jq/pull/2717#discussion_r1264338841

### DIFF
--- a/src/execute.c
+++ b/src/execute.c
@@ -1012,6 +1012,9 @@ jq_state *jq_init(void) {
   jq->exit_code = jv_invalid();
   jq->error_message = jv_invalid();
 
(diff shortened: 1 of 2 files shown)


nicowilliams:
Thanks!

PR #2720, opened 2023-07-16
Make stream parser error when a non-empty object or array is used as key

### LINKED ISSUE #2463: jq is partially laxer in what it parses when used with `--stream` flag
**Describe the bug**
When jq is used to validate any json input of arbitrary size to ensure there are no syntax errors inside, we would like to favor `--stream` flag to enable validation of files the internal representation of which exceeds our RAM or any imposed limits.

**To Reproduce**
```
% cat /tmp/test-jq.sh 
#!/bin/sh

check_JSON_validity() {
    json="$1"; shift
    printf '%s' "$json" | jq "$@" > /dev/null 2>&1
    if [ $? -eq 0 ]; then
        printf 'valid json'
    else
        printf 'jq returned %d' $?
    fi
}

jq --version || { >&2 echo "jq is not installed."; exit 1; }

compare_validity_checks() {
    INPUT="$1"
    if [ "$(check_JSON_validity "$INPUT")" = "$(check_JSON_validity "$INPUT" --stream)" ]; then
        echo "OK, that version of jq consistently validates input no matter if --stream mode is on or off."
        return 0
    else
        echo "NOT OK, that version of jq inconsistently chokes on invalid input depending on --stream flag."  
        return 1
    fi
}

compare_validity_checks '{["invalid", "input"]}'
compare_validity_checks '{"a":["valid", "input"]}'
compare_validity_checks '{"a":b["valid", "input"]}'
me@linux ~ % sh /tmp/test-jq.sh
jq-1.6
NOT OK, that version of jq inconsistently chokes on invalid input depending on --stream flag.
OK, that version of jq consistently validates input no matter if --stream mode is on or off.
OK, that version of jq consistently validates input no matter if --stream mode is on or off.
```

**Expected behavior**

`jq` should not be laxer concerning its interpretation of input in stream mode than it is in normal mode. If the laxness is immanently due to how stream mode works, you might rather consider an `--early-free-memory SELECTORS` option that I could set to '.' for mere validation purposes?

 - Ubuntu Linux 20.04
 - jq 1.6

**Additional context**
Long time preservation endeavours will also cover JSON files while they become more and more prominent in data processing across all branches and sciences. We need to validate all that data before we can archive it without risking to run out of memory digesting large datasets.

emanuele6:
Fixes #2463

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -246,6 +246,17 @@ static pfunc stream_token(struct jv_parser* p, char ch) {
   case '[':
     if (jv_is_valid(p->next))
       return "Expected a separator between values";
+    if (p->last_seen == JV_LAST_OPEN_OBJECT)
+      // Looks like {["foo"]}
+      return "Expected string key after '{', not '['";
+    if (p->last_seen == JV_LAST_COMMA) {
+      last = jv_array_get(jv_copy(p->path), p->stacklen - 1);
+      k = jv_get_kind(last);
+      jv_free(last);
+      if (k != JV_KIND_NUMBER)
+        // Looks like {"x":"y",["foo"]}
+        return "Expected string key after ',' in object, not '['";
+    }
     p->path = jv_array_append(p->path, jv_number(0)); // push
     p->last_seen = JV_LAST_OPEN_ARRAY;
     p->stacklen++;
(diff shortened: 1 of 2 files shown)


nicowilliams:
Thank you!

PR #2723, opened 2023-07-17
Add oniguruma workflow to test regex options

itchyny:
This PR introduces a new oniguruma Actions workflow to cover other options for `--with-oniguruma`. User can link with pre-installed library, or disable the option entirely. Automating these test cases prevents future regression of `make check` fails when built without oniguruma (e.x. #2292).


emanuele6 (approved):
LGTM!

nicowilliams:
Thanks!

PR #2727, opened 2023-07-18
Fix memory leaks in --with-oniguruma=no implementation of f_match

emanuele6:
Found by LeakSanitizer running `./jq -n '"" | test("")'` in a `--with-oniguruma=no --enable-ubsan --enable-asan` build.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1022,6 +1022,10 @@ static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
 }
 #else /* !HAVE_LIBONIG */
 static jv f_match(jq_state *jq, jv input, jv regex, jv modifiers, jv testmode) {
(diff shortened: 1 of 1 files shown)


nicowilliams:
There's also leaks in error cases in `f_match()` w/ Oniguruma, like at line 839 where we leak `testmode`.

emanuele6:
@nicowilliams
No, `jv_equal(testmode, jv_true())` at line 828 frees `testmode` (and `jv_true()`).
<s>Anyway a leak there would have been detected by the `valgrind` CI. ;)</s>(Never mind, I guess we don't have a test case for that code path.)

nicowilliams:
Thanks!

itchyny:
Do we enable valgrind in the oniguruma workflow? I'm in favor of that.

PR #2729, opened 2023-07-19
Avoid getpwuid for static linking

itchyny:
We use `getpwuid` for getting the home directory (when `$HOME` is unset), but this prevents the library to be fully static, and the compiler actually emits warnings on the usage; `warning: Using 'getpwuid' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking`. IMO checking `$HOME` is enough to expand `~/` in the library search paths. For reference, [`os.UserHomeDir()` in Go](https://github.com/golang/go/blob/go1.20/src/os/file.go#L489) is something like this (BTW this implementation let us know how we should handle home on Plan9 but nobody has reported on this topic so far).

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -28,9 +28,6 @@ extern "C"
 void *alloca (size_t);
 # endif
 #endif
-#ifndef WIN32
-#include <pwd.h>
-#endif
 
 #ifdef WIN32
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


nicowilliams:
Maybe we should only do this if building a statically-linked jq?

itchyny:
Not sure how to detect that, but even if that's possible, I'm opposed to leave these corner case codes, uncovered by the tests.

nicowilliams:
Thanks!

PR #2730, opened 2023-07-19
Enable valgrind in oniguruma workflow

itchyny:
This PR installs valgrind in the oniguruma workflow. This prevents future memory leakage when oniguruma is disbaled (e.x. #2727).


emanuele6 (approved):
LGTM!

PR #2733, opened 2023-07-19
Enable compilation to WebAssembly

robertaboukhalil:
This PR adds the ability to compile jq to WebAssembly using Emscripten:

I updated `main.c` to reset `options` to `0` if you're using Emscripten. The issue is that `options` is a global variable set to 0 only once, which makes sense in binary world where we only call `main()` once. But when running `jq.wasm` in the browser, being able to call `main()` multiple times in a row is useful so that you don't reinitialize the WebAssembly module every time. One example is that this has been really useful for my [jq playground](https://github.com/robertaboukhalil/jqkungfu) 🙂

Without this patch, `options` is not reset to 0, so the next time you call `main()`, it uses the previous value of options, which results in an error.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -146,7 +146,6 @@ enum {
   /* debugging only */
   DUMP_DISASM           = 65536,
 };
-static int options = 0;
 
 enum {
     JQ_OK              =  0,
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -292,6 +292,12 @@ int main(int argc, char* argv[]) {
   _setmode(fileno(stderr), _O_TEXT | _O_U8TEXT);
 #endif
 
+#ifdef __EMSCRIPTEN__
+  /* When compiling to WebAssembly with Emscripten, reset options to ensure we can
+     call main() multiple times without reinitializing the WebAssembly module. */
+  options = 0;
```
nicowilliams: Can't we make `options` local to `main()`?

nicowilliams: We could make `options` local to `main()` by passing it to `process()`.

robertaboukhalil: Great idea, it looks cleaner now, thanks!

nicowilliams:
Thanks!

PR #2737, opened 2023-07-20
manual.yml: some clarifications plus fixes for markdown bloopers

pkoppstein:
E.g. 

change:  The simplest *useful* filter is `.foo`.
to:           The simplest *useful* filter has the form `.foo`.


itchyny (approved):
LGTM

PR #2738, opened 2023-07-20
Fix destroying of version when building from tarball

colinleroy:
This prevents destroying of `src/version.h` when running `make clean` after `./configure --disable-maintainer-mode`.
It also fixes `scripts/version` to fetch version from `src/version.h` when we are not in a git repository.

Hope this helps,
Colin

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -120,6 +120,7 @@ src/builtin.o: src/builtin.inc
 if MAINTAINER_MODE
 CLEANFILES = src/version.h .remake-version-h src/builtin.inc
 else
+DISTCLEANFILES = src/version.h
 CLEANFILES = .remake-version-h src/builtin.inc
 endif
 bin_PROGRAMS = jq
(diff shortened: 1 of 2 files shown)


itchyny:
Version is embedded in configure in the tarball, so fetching version from src/version.h to write to src/version.h is incorrect.

itchyny:
Since version.h is a part of BUILT_SOURCES (just like builtin.inc) even in non-maintainer mode, I'm afraid the file shouldn't be included in the tarball. And generate_ver works well even if it is not in git repository.

colinleroy:
> Since version.h is a part of BUILT_SOURCES (just like builtin.inc) even in non-maintainer mode, I'm afraid the file shouldn't be included in the tarball. And generate_ver works well even if it is not in git repository.

Yes, but after a make clean, src/version.h disappears, and rebuilding recreates it as
```
#define JQ_VERSION ""
```

itchyny:
I can't reproduce that. `./configure && make clean && make src/version.h` restores the version.

colinleroy:
I'll do more tests and close/update this PR, it's possible I did something wrong. Thanks!

colinleroy:
Forget it, that's my bad. Building a debian package triggers an autoreconf, this regenerates configure and that's what is going wrong.

PR #2744, opened 2023-07-22
Improve manual in various ways (inputs, sort_by, foreach sections, etc.)

### LINKED ISSUE #1047: Manual: please explicitly explain the .foo[] syntax.
I have looked through the manual, and it did not look to me like you are ever explicitly explaining the .foo[] syntax: it just appears out of the blue in the section on **Array Construction** as part of the example `[.items[].name]`.  By analogy with `.foo.bar` being syntactic sugar for `.foo | .bar` (which you do explain in section **`.foo`, `.foo.bar`**), I guessed that the syntax `.foo[]` must be syntactic sugar for `.foo | .[]`; I have done a bit of experimentation, and so far, I haven't been able to come up with an example that refutes my guess -- but the point of this issue is that I shouldn't have had to guess.

One additional point of confusion for me is that, if I pedantically apply the _`.foo.bar` desugars to `.foo | .bar`_ line of thinking, I would expect to be able to write `.foo.[]` and have it desugar to `.foo | .[]` -- but that didn't work for me in my experiments: only `.foo[]` syntax was accepted.

If my guess is indeed what you intended (I haven't looked at your parser yet), then I think a logical place to add this information is in the section that describes **`.[]`**.  I think a single sentence would do. Perhaps something like

> A filter of the form `.foo[]` is equivalent to `.foo|.[]`

### LINKED ISSUE #1148: foreach optional argument
The documentation ignores the second usage of `foreach`, and only documents the full usage:

```
foreach EXP as $var (INIT; UPDATE; EXTRACT)
```

In fact the `EXTRACT` argument is optional, and the two following forms are equivalent:

```
foreach EXP as $var (INIT; UPDATE)
foreach EXP as $var (INIT; UPDATE; .)
```

### LINKED ISSUE #1422: Docs uses plain strings as examples, not JSON
I try to use jq for all my JSON manipulations now but docs are not helpful. All more advanced features involve plain strings and it's not always useful, example:

https://stedolan.github.io/jq/manual/#index(s),rindex(s)

In this example I would like to see how to use index function to get index of element in big array to use it in next filters. Instead I see very basic index usage which is not too helpful.

This makes jq hard to introduce to anyone, besides basic usage to extract part of jsons.

### LINKED ISSUE #2036: Documentation is not correct about behaviour of recurse builtin
**Describe the bug**
jq-1.6 documentation says,

>recurse(f) is identical to recurse(f; . != null) and can be used without concerns about recursion depth.

however, from jq-1.6, the filter for null was abolished, which is good and consistent, I believe.

**To Reproduce**

Just see the documentations and search for definition for the one parameter ```recurse``` function.

- https://stedolan.github.io/jq/manual/v1.6/#Builtinoperatorsandfunctions
- https://stedolan.github.io/jq/manual/#Builtinoperatorsandfunctions

**Expected behavior**

The description should be like following

> ```recurse(f)``` is identical to ```recurse(f; true)``` and can be used without concerns about recursion depth.

since ```recurse(f)``` is defined ```def recurse(f): def r: ., (f | r); r;``` in ```builtin.jq```

 - ALL
 - jq version : 1.6 or leter

**Additional context**
None

### LINKED ISSUE #2103: Documentation sort or sort_by
https://github.com/stedolan/jq/blob/ccc79e592cfe1172db5f2def5a24c2f7cfd418bf/docs/content/manual/manual.yml#L1317

Should this be
`sort_by` may be used to sort by a particular field ?

If it's possible to do this with just sort, can somebody give an example?

Thanks very much

### LINKED ISSUE #2169: foreach: doc recommendation
I found the `foreach` example a little hard to understand. Can we have something more Hello World-ish?, eg
https://jqplay.org/s/4GNwRCpQLX
```
jq 'foreach .[] as $item (0; . + 1; "\(.) - \($item)")'

input: ["apple", "banana", "cherry"]

output:
"1 - apple"
"2 - banana"
"3 - cherry"
```

Also, i'm curious why it starts at 1, not 0 -- this seems different than a foreach loop, eg, in JavaScript or Python. Would you be able to explain (here and/or in the docs)? Is this a bug?

(speaking of JavaScript, the "s" is not capitalized in the docs)

Thanks!

### LINKED ISSUE #2231: `error` ignores null input
**Describe the bug**
Both `error/0` and `error/1` do nothing for `null` (not empty!) input

**To Reproduce**

```
$ jq -n error
$ echo $?
0

$ jq -n 'error(null)'
$ echo $?
0
$
```

**Expected behavior**

Errors should never be ignored, should write stderr message and cause non-zero exit code

```
$ jq -n error
jq: error (at <unknown>) (not a string): null
$ echo $?
5
```

```
$ jq --version
jq-1.6
$ uname
Darwin
$ sw_vers -productVersion
11.0.1
```

### LINKED ISSUE #2412: recurse(f) is not identical to recurse(f; .!=null)
**Describe the bug**
The manual states:

> recurse(f) is identical to recurse(f; . != null) and can  be  used  without  concern about recursion depth.

That is not actually true: `recurse(f)` only stops when there is no more input to process, not when `.` becomes `null`; it is identical to `recurse(f; true)`.

**To Reproduce**
```bash
bash-5.1$ jq -cn '{foo:{bar:2,foo:{x:3}}} | recurse(.foo; . != null)'
{"foo":{"bar":2,"foo":{"x":3}}}
{"bar":2,"foo":{"x":3}}
{"x":3}

bash-5.1$ jq -cn '{foo:{bar:2,foo:{x:3}}} | recurse(.foo)'  # infinite loop
{"foo":{"bar":2,"foo":{"x":3}}}
{"bar":2,"foo":{"x":3}}
{"x":3}
null
null
null
null
^C
```

**Expected behavior**
The current behaviour for `recurse(f)` is fine as is in my opinion, but the manual is wrong.

 - OS and Version: Arch Linux
 - jq version: 1.6

### LINKED ISSUE #2467: Docs: `sort_by()` doesn't mention sorting by multiple keys of an object

itchyny:
- Add error/0 and mentions null input behavior (close #2231)
- Explain value iterator suffix syntax .foo[] (close #1047)
- Mention array slicing is also zero-based (close #2094)
- Add examples of input and inputs filters (close #2216, close #2470)
- Improve sort_by about multiple values (close #2103, close #2467, close #2474)
- Improve foreach section and simplify examples (close #1148, close #2169)
- Fix recurse/1 document on how it is identical using recurse/2 (close #2036, close #2412)
- Add non-string examples of index/1, rindex/1 (close #1422)
- Simplify the example of truncate_stream/1 (close #1736)

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -79,8 +79,8 @@ sections:
       jq filters run on a stream of JSON data. The input to jq is
       parsed as a sequence of whitespace-separated JSON values which
       are passed through the provided filter one at a time. The
-      output(s) of the filter are written to standard out, again as a
-      sequence of whitespace-separated JSON data.
+      output(s) of the filter are written to standard output, as a
+      sequence of newline-separated JSON data.
 
       Note: it is important to mind the shell's quoting rules.  As a
       general rule it's best to always quote (with single-quote
@@ -132,7 +132,7 @@ sections:
 
       * `--stream-errors`:
 
-        Like `--stream`, but invalid JSON inputs yield array calues
+        Like `--stream`, but invalid JSON inputs yield array values
         where the first element is the error and the second is a path.
         For example, `["a",n]` produces ["Invalid literal at line 1,
         column 9",[1]]`.
@@ -334,7 +334,7 @@ sections:
           input, jq processing can sometimes make it appear as though
           it does.  For example, using the current implementation of
           jq, we would see that the expression:
-          
+
               1E1234567890 | .
 
           produces `1.7976931348623157e+308` on at least one platform.
@@ -372,7 +372,7 @@ sections:
             output: ['"Hello, world!"']
 
           - program: '.'
-            input:  '0.12345678901234567890123456789'
+            input: '0.12345678901234567890123456789'
             output: ['0.12345678901234567890123456789']
 
           - program: '[., tojson]'
@@ -398,7 +398,7 @@ sections:
           JSON object (aka dictionary or hash) as input, `.foo` produces
           the value at the key "foo" if the key is present, or null otherwise.
 
-          A filter of the form `.foo.bar` is equivalent to `.foo|.bar`.
+          A filter of the form `.foo.bar` is equivalent to `.foo | .bar`.
 
           The `.foo` syntax only works for simple, identifier-like keys, that
           is, keys that are all made of alphanumeric characters and
@@ -484,6 +484,7 @@ sections:
           Either index may be negative (in which case it counts
           backwards from the end of the array), or omitted (in which
           case it refers to the start or end of the array).
+          Indices are zero-based.
 
         examples:
           - program: '.[2:4]'
@@ -509,7 +510,8 @@ sections:
           entirely, it will return *all* of the elements of an
           array. Running `.[]` with the input `[1,2,3]` will produce the
           numbers as three separate results, rather than as a single
-          array.
+          array. A filter of the form `.foo[]` is equivalent to
+          `.foo | .[]`.
 
           You can also use this on an object, and it will return all
           the values of the object.
@@ -525,6 +527,10 @@ sections:
             input: '[]'
             output: []
 
+          - program: '.foo[]'
+            input: '{"foo":[1,2,3]}'
+            output: ['1','2','3']
+
           - program: '.[]'
             input: '{"a": 1, "b": 1}'
             output: ['1', '1']
@@ -533,7 +539,8 @@ sections:
         body: |
 
           Like `.[]`, but no errors will be output if . is not an array
-          or object.
+          or object. A filter of the form `.foo[]?` is equivalent to
+          `.foo | .[]?`.
 
       - title: "Comma: `,`"
         body: |
@@ -728,15 +735,15 @@ sections:
           Recursively descends `.`, producing every value.  This is the
           same as the zero-argument `recurse` builtin (see below).  This
           is intended to resemble the XPath `//` operator.  Note that
-          `..a` does not work; use `..|.a` instead.  In the example
-          below we use `..|.a?` to find all the values of object keys
+          `..a` does not work; use `.. | .a` instead.  In the example
+          below we use `.. | .a?` to find all the values of object keys
           "a" in any object found "below" `.`.
 
           This is particularly useful in conjunction with `path(EXP)`
           (also see below) and the `?` operator.
 
         examples:
-          - program: '..|.a?'
+          - program: '.. | .a?'
             input: '[[{"a":1}]]'
             output: ['1']
 
@@ -1190,12 +1197,25 @@ sections:
             input: 'null'
             output: ['[1,2,3]']
 
-      - title: "`error(message)`"
+      - title: "`error`, `error(message)`"
         body: |
 
-          Produces an error, just like `.a` applied to values other than
-          null and objects would, but with the given message as the
-          error's value.  Errors can be caught with try/catch; see below.
+          Produces an error with the input value, or with the message
+          given as the argument. Errors can be caught with try/catch;
+          see below.
+
+          When the error value is `null`, it produces nothing and works
+          just like `empty`. So `[null | error]` and `[error(null)]` both
+          emit `[]`.
+
+        examples:
+          - program: 'try error catch .'
+            input: '"error message"'
+            output: ['"error message"']
+
+          - program: 'try error("invalid value: \(.)") catch .'
+            input: '42'
+            output: ['"invalid value: 42"']
 
       - title: "`halt`"
         body: |
@@ -1481,19 +1501,25 @@ sections:
           sorted order), and if their keys are equal then the values
           are compared key by key.
 
-          `sort` may be used to sort by a particular field of an
-          object, or by applying any jq filter.
-
-          `sort_by(f)` compares two elements by comparing the result of
-          `f` on each element.
+          `sort_by` may be used to sort by a particular field of an
+          object, or by applying any jq filter. `sort_by(f)` compares
+          two elements by comparing the result of `f` on each element.
+          When `f` produces multiple values, it firstly compares the
+          first values, and the second values if the first values are
+          equal, and so on.
 
         examples:
           - program: 'sort'
             input: '[8,3,null,6]'
             output: ['[null,3,6,8]']
+
           - program: 'sort_by(.foo)'
-            input: '[{"foo":4, "bar":10}, {"foo":3, "bar":100}, {"foo":2, "bar":1}]'
-            output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":100}, {"foo":4, "bar":10}]']
+            input: '[{"foo":4, "bar":10}, {"foo":3, "bar":10}, {"foo":2, "bar":1}]'
+            output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":10}, {"foo":4, "bar":10}]']
+
+          - program: 'sort_by(.foo, .bar)'
+            input: '[{"foo":4, "bar":10}, {"foo":3, "bar":20}, {"foo":2, "bar":1}, {"foo":3, "bar":10}]'
+            output: ['[{"foo":2, "bar":1}, {"foo":3, "bar":10}, {"foo":3, "bar":20}, {"foo":4, "bar":10}]']
 
       - title: "`group_by(path_expression)`"
         body: |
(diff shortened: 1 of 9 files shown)


docs/content/manual/manual.yml
```diff
@@ -2924,32 +2944,72 @@ sections:
...
           The form is `foreach EXP as $var (INIT; UPDATE; EXTRACT)`.
-          Like `reduce`, `INIT` is evaluated once to produce a state
-          value, then each output of `EXP` is bound to `$var`, `UPDATE`
-          is evaluated for each output of `EXP` with the current state
-          and with `$var` visible.  Each value output by `UPDATE`
-          replaces the previous state.  Finally, `EXTRACT` is evaluated
-          for each new state to extract an output of `foreach`.
+          As an example, we'll pass `[1,2,3]` to this expression:
+
+              foreach .[] as $item (0; . + $item; . * 2)
```
wader: Possibly show "expansion" of foreach similar to for reduce?

docs/content/manual/v1.5/manual.yml
```diff
@@ -398,6 +398,10 @@ sections:
             input: '[]'
             output: []
 
+          - program: '.foo[]'
+            input: '{"foo":[1,2,3]}'
+            output: ['1','2','3']
```
wader, line 403: Should be?
```yml
output:
  - 1
  - 2
  - 3
```

wader: Sorry now i see, this yaml array?

wader (approved):
Nice work

docs/content/manual/manual.yml
itchyny: Curious what it looks like. Following query is the first one I come up with.
```jq
0 | (1 as $item | . + $item) | (. * 2), (
    (2 as $item | . + $item) | (. * 2), (
    (3 as $item | . + $item) | (. * 2)))
```

wader: Looks correct 👍  i remember seeing a comment in the jaq code explaining this in a similar way https://github.com/01mf02/jaq/blob/bac33c3a1747ea390de368ebc64e9175b5218ce5/jaq-core/src/filter.rs#L40-L62

wader: But think i it's probably easier to follow with a concrete example like for reduce in the jq docs?

docs/content/manual/manual.yml
```diff
@@ -474,6 +474,7 @@ sections:
           Either index may be negative (in which case it counts
           backwards from the end of the array), or omitted (in which
           case it refers to the start or end of the array).
+          The indices are zero-based.
```
nicowilliams: `s/The i/I/`?

docs/content/manual/manual.yml
```diff
@@ -1180,12 +1187,25 @@ sections:
...
         body: |
 
-          Produces an error, just like `.a` applied to values other than
-          null and objects would, but with the given message as the
-          error's value.  Errors can be caught with try/catch; see below.
+          Produces an error with the input value, or with the message
+          given as the argument. Errors can be caught with try/catch;
+          see below.
+
+          When the error value is `null`, it produces nothing and works
```
nicowilliams, line 1207: Funny, I never knew (or I'd forgotten) this.  And it seems useful, so if it's an accident, it's a useful accident.

docs/content/manual/manual.yml
```diff
@@ -2924,32 +2944,72 @@ sections:
...
-          for each new state to extract an output of `foreach`.
+          As an example, we'll pass `[1,2,3]` to this expression:
+
+              foreach .[] as $item (0; . + $item; . * 2)
 
-          This is mostly useful only for constructing `reduce`- and
-          `limit`-like functions.  But it is much more general, as it
-          allows for partial reductions (see the example below).
+          Like the `reduce` syntax, `. + $item` is run for each result
+          that `.[]` produces, but `. * 2` is run for each intermediate
```
nicowilliams: Maybe this example isn't so useful because it's the same as `foreach .[] as $item (0; (. + $item) * 2; .)`.  A more useful example would be one where the `EXTRACT` expression has a conditional.

docs/content/manual/manual.yml
```diff
@@ -2924,32 +2944,72 @@ sections:
...
+          When `EXTRACT` is omitted, the identity filter is used.
 
         examples:
-          - program: '[foreach .[] as $item
-                        ([[],[]];
-                        if $item == null then [[],.[0]] else [(.[0] + [$item]),[]] end;
-                        if $item == null then .[1] else empty end)]'
-            input: '[1,2,3,4,null,"a","b",null]'
-            output: ['[[1,2,3,4],["a","b"]]']
+          - program: 'foreach .[] as $item (0; . + $item)'
```
nicowilliams, line 3016: Ditto.  We really need to show that `EXTRACT` can `empty` to skip some outputs.

nicowilliams: Also, I've come to think that I should just have added a `reduce` with the extra `EXTRACT` expression and not called it `foreach`.  My apologies to all for that.

docs/content/manual/manual.yml
```diff
@@ -79,8 +79,8 @@ sections:
       jq filters run on a stream of JSON data. The input to jq is
       parsed as a sequence of whitespace-separated JSON values which
       are passed through the provided filter one at a time. The
-      output(s) of the filter are written to standard out, again as a
-      sequence of whitespace-separated JSON data.
+      output(s) of the filter are written to standard output, as a
+      sequence of newline-separated JSON data.
```
nicowilliams, line 83: We don't have to mention carriage return on Windows, do we?  (No, we probably shouldn't.)

pkoppstein:
Typo:
```
$ grep -n calues docs/content/manual/manual.yml
135:        Like `--stream`, but invalid JSON inputs yield array calues
```

docs/content/manual/manual.yml
itchyny: They are not the same, but I updated the example. I'm afraid the only advantage to use `EXTRACT` instead of piping the filter after foreach expression is that `$item` is available there.

docs/content/manual/manual.yml
pkoppstein: @nicowilliams - Don't forget that a fundamental difference between `reduce` and `foreach` (as implemented) is that executing a `break` within a `reduce` makes the `reduce` emit empty, e.g.
```
$ ./jq -n 'label $out | reduce (1, 2, false, 3) as $item (null; if .==false then ., break $out else $item end)'
$
```
manual.yml is unclear about this.  Perhaps you could clarify the behavior that should be documented.

[EDIT: irrelevant non-example removed]

docs/content/manual/manual.yml
nicowilliams: Ah, d'oh.  Can you tell I'm rusty?

docs/content/manual/manual.yml
nicowilliams: @pkoppstein you never execute the `break` in the second case because you never set the reduction state to `$item`.

itchyny:
Ready to review but let me confirm later again and control the merge timing.

nicowilliams:
LGTM!

docs/content/manual/manual.yml
pkoppstein: Sorry, I was in a hurry and distracted.  Anyway, the point remains: the documentation does not make it clear whether reduce's actual behavior (empty on `break`) is intended, so unless you think the actual behavior is a bug, I'd like to make it explicit.

(@itchyny - please advise whether for gojq you were just conforming to jq's actual behavior, and whether you think the jq documentation should be clarified.)

itchyny: I don't see any difference of `break` between `reduce` and `foreach`. This is a topic of `break`, and the behavior of using with `reduce` is already documented in the `Breaking out of control structures` section.
```sh
# foreach emits 0, 1, 3 and then breaks on $item == 3
jq -n 'label $out | foreach range(5) as $item (null; if $item > 2 then break $out else debug + $item end)'

# reduce breaks on $item == 3
jq -n 'label $out | reduce range(5) as $item (null; if $item > 3 then break $out else debug + $item end)'
```

pkoppstein: @itchyny - The section you refer to in manual.yml indicates that `break` can be used to break out of the `reduce`, but does not explicitly say what the result will be.  One might expect the result to be the memoized value at that point, since `reduce` otherwise emits that value.  

(Certainly, one might **want** there to be a way to use `reduce` in such a way, e.g. to short-circuit the computation of the max of a stream, where it is known beforehand that the items in the stream can never be larger than a certain value that may or may not appear in the stream.)

itchyny:
Thank you all!!!

PR #2747, opened 2023-07-22
Improve --help to mention all the command options

itchyny:
Listing all the command line options is useful in various ways; not just for for completion generators, but for offline environment. Ah, not only for offline environment, this improves our dev experience a lot; we can see the meaning of command options without searching for the manual. This PR resolves #2284, close #2504.

PR #2748, opened 2023-07-22
Remove undocumented and untested --stream-errors option

itchyny:
This PR drops `--stream-errors` option. This option is undocumented, and untested. Also buggy; `printf '[' | jq --stream-errors .` aborts with assertion error.

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -220,9 +220,8 @@ jv jv_dump_string(jv, int flags);
 char *jv_dump_string_trunc(jv x, char *outbuf, size_t bufsize);
 
 enum {
-  JV_PARSE_SEQ              = 1,
-  JV_PARSE_STREAMING        = 2,
-  JV_PARSE_STREAM_ERRORS    = 4,
+  JV_PARSE_SEQ       = 1,
+  JV_PARSE_STREAMING = 2,
 };
 
 jv jv_parse(const char* string);
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -72,7 +72,6 @@ static void parser_init(struct jv_parser* p, int flags) {
     p->path = jv_array();
   } else {
     p->path = jv_invalid();
-    p->flags &= ~(JV_PARSE_STREAM_ERRORS);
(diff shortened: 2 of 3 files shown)


nicowilliams:
EDIT: `printf '[' | ./jq -c --stream --stream-errors .` does not abort (it outputs `["Unfinished JSON term at EOF at line 1, column 1",[0]]`).  So another fix would be to make `--stream-errors` imply `--stream`.

It should be an easy fix:

```
(gdb) p jv_show(result,0)
["Unfinished JSON term at EOF at line 2, column 0",<invalid>]$1 = void
(gdb)
```

The problem is that we need to check for `.[1]` of `result` being invalid, and then do something about that: dump just `.[0]` of `result`, or drop the `.[1]` of `result.

Now, the question is: is `--stream-errors` useful?  I think it should be!

And this doesn't abort:

```
: ; printf '[' | ./jq --stream-errors '.[0]'
"Unfinished JSON term at EOF at line 1, column 1"
```

nicowilliams:
And:

```
: ; (./jq -cn '[{a:{b:["c"]}}]'; printf '['; ./jq -cn '["a",{b:"c"}]') | ./jq -c --stream --stream-error
s .
[[0,"a","b",0],"c"]
[[0,"a","b",0]]
[[0,"a","b"]]
[[0,"a"]]
[[0]]
[[0,0],"a"]
[[0,1,"b"],"c"]
[[0,1,"b"]]
[[0,1]]
["Unfinished JSON term at EOF at line 3, column 0",[0]]
```

nicowilliams:
One thing we should do is write a fuzzer for the streaming parser.  Also, we should document what is and what is not fuzzed so that users who need better security know what features to use / not use.

itchyny:
Okay, we need to fix the crash, and update the help (after #2747) and the manual.

nicowilliams:
> Okay, we need to fix the crash, and update the help (after #2747) and the manual.

I'll try to get to it tomorrow.  Maybe or maybe later tonight.

PR #2749, opened 2023-07-22
Add tests for --arg, --argjson, and similar options

itchyny:
I added tests for `--arg`, `--argjson`, `--slurpfile`, `--rawfile`, `--args`, `--jsonargs`, and `$ARGS`.


nicowilliams:
Thanks!

PR #2751, opened 2023-07-23
Fix stderr to output string with no decoration (fix #2063)

### LINKED ISSUE #2063: stderr doesn't print raw
**Describe the bug**
stderr doesn't print its input as a raw string but as an escaped json string.

**To Reproduce**
```sh
jq -n '"line1\nline2" | stderr | empty'
```
Outputs ```"line1\nline2"```

**Expected behavior**
As the manual states:
> **stderr**
> Prints its input in **raw** and compact mode to stderr with no additional decoration, not even a newline.

I expected the program to output be:
```
line1
line2
```

 - OS and Version: Redhat Enterprise Linux 6
 - jq version 1.6

itchyny:
As written in the manual, the `stderr` filter should output values in raw and compact mode with no additional decoration. This PR fixes #2063.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -232,6 +232,16 @@ grep "Expected string key after '{', not '\\['" $d/err > /dev/null
 echo '{"x":"y",["a","b"]}' | $JQ --stream > /dev/null 2> $d/err || true
 grep "Expected string key after ',' in object, not '\\['" $d/err > /dev/null
 
+# debug, stderr
+$VALGRIND $Q $JQ -n '"test", {} | debug, stderr' >/dev/null
+$JQ -n -c -j '"hello\nworld", null, [false, 0], {"foo":["bar"]}, "\n" | stderr' >$d/out 2>$d/err
+cat > $d/expected <<'EOF'
+hello
+worldnull[false,0]{"foo":["bar"]}
+EOF
+cmp $d/out $d/expected
+cmp $d/err $d/expected
+
 # --arg, --argjson, $ARGS.named
 $VALGRIND $JQ -n -c --arg foo 1 --argjson bar 2 '{$foo, $bar} | ., . == $ARGS.named' > $d/out
 printf '{"foo":"1","bar":2}\ntrue\n' > $d/expected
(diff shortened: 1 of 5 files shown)


tests/shtest
```diff
@@ -232,6 +232,16 @@ grep "Expected string key after '{', not '\\['" $d/err > /dev/null
 echo '{"x":"y",["a","b"]}' | $JQ --stream > /dev/null 2> $d/err || true
 grep "Expected string key after ',' in object, not '\\['" $d/err > /dev/null
 
+# debug, stderr
+$VALGRIND $Q $JQ -n '"test", {} | debug, stderr' >/dev/null
```
wader, line 236: Should output be compared somehow here?

wader: Unrelated thought: I wonder if i would be nicer if we replaced `$VALGRIND $Q $JQ` by having `jq` in path be a wrapper script that does `exec valgrind -q /path/to/real/jq/ ...`? or do we don't want to not run via valgrind in some cases when it's enabled?

itchyny: I put this line just for valgrind (and I actually found a leak in the first commit), and I heard Nico said we don't care about the output of debug/0 somewhere. I strongly hope that JQ takes care of valgrind automatically but there might some difficulties in testing the standard error output.

wader: Aha valgrind can messes up stdout/err? by "hope that JQ takes care" do you mean it would be nice if `$JQ` pointed to some path that runs jq with valgrind?

Yes think i read that comment from @nicowilliams also, but even if we make no guarantees about stableness maybe it can make sense to test debug output so that we would notice if it changes by accident? maybe we already do somewhere?

nicowilliams: `debug/0` should do something with its output.  We could have a command-line option to send that to some file, or `/dev/null`/`NUL`, and `debug/0` can decorate its output or not (but it helps that it does).

I don't really care about the mechanics of how `valgrind` or other memory debuggers get invoked, as long as a) such debuggers can be optional, b) they can be turned off (valgrind slows dev cycles).  If the current thing makes it easy to forget to use valgrind, sure, let's fix it.

nicowilliams: Also, if in `--run-tests` we set a `stderr` callback, we could have syntax in `.test` files for checking `stderr`'s output, though I don't think we need that given that a test here in `shtest` is just as well, and we only need `stderr` to test it.

nicowilliams:
Thanks!

PR #2758, opened 2023-07-24
builtin.jq: simpler and faster transpose

pkoppstein:
It turns out that using max/0 is faster than using a stream-oriented version of max

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -181,14 +181,8 @@ def combinations(n):
       | combinations;
 # transpose a possibly jagged matrix, quickly;
 # rows are padded with nulls so the result is always rectangular.
-def transpose:
-  if . == [] then []
-  else . as $in
-  | (map(length) | max) as $max
-  | length as $length
-  | reduce range(0; $max) as $j
-      ([]; . + [reduce range(0;$length) as $i ([]; . + [ $in[$i][$j] ] )] )
-  end;
+# Using map(length) turns out to be faster than using max(stream)
+def transpose: [range(0; map(length)|max) as $i | [.[][$i]]];
 def in(xs): . as $x | xs | has($x);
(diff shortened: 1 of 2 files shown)


itchyny:
Previously `[] | transpose` produces `[]` but after this PR it changes to an error.

pkoppstein:
@itchyny - For backward compatibility: https://github.com/jqlang/jq/pull/2758/commits/08da3a931a139d6eda5a7b1c7beabe70831f886f

jq.test now has a test case for `[] | transpose` #=> []

Thanks.

src/builtin.jq
```diff
@@ -181,14 +181,8 @@ def combinations(n):
...
 # rows are padded with nulls so the result is always rectangular.
-def transpose:
-  if . == [] then []
-  else . as $in
-  | (map(length) | max) as $max
-  | length as $length
-  | reduce range(0; $max) as $j
-      ([]; . + [reduce range(0;$length) as $i ([]; . + [ $in[$i][$j] ] )] )
-  end;
+# Using map(length) turns out to be faster than using max(stream)
```
itchyny: I'm still unsure about the stream version of max to be included or not. Users not following the history will be get confused by this comment; where is `max(stream)`?

pkoppstein: @itchyny - Here, max(stream) is simply short for any valid streaming version of max. That should be clear since no specific referent is mentioned. Feel free to change the wording, though.

itchyny: Let's just drop the sentence, we don't need such performance notes here.

pkoppstein: OK. Done in https://github.com/jqlang/jq/pull/2758/commits/a20c6d50425251c211bfa5c6608c7049d81bd331

itchyny (approved):
Thank you!

PR #2759, opened 2023-07-24
Make maintainer-mode default to disabled

wader:
This makes the standard build instructions a bit easier, just ./configure, and also requires less tools installed (bison).

Also i think few people probably want to generate the lexer and paser code.

### DIFF
--- a/.github/workflows/scanbuild.yml
+++ b/.github/workflows/scanbuild.yml
@@ -48,10 +48,7 @@ jobs:
                 CC: ${{ matrix.compiler }}
                 MAKEVARS: ${{ matrix.makevars }}
               run: |
-                autoreconf -i
-                rm src/lexer.c src/lexer.h
-                rm src/parser.c src/parser.h
-                ./configure --with-oniguruma=builtin YACC="$(which bison) -y" $COVERAGE
+                ./configure --with-oniguruma=builtin $COVERAGE
                 scan-build --keep-going make -j4
             - name: Test
               env:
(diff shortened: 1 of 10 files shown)


wader:
More of a proposal. I think this is more closer to what people expect. Maybe a second step could be to remove most uses of `--with-oniguruma=builtin` also?

.github/workflows/scanbuild.yml
```diff
@@ -48,10 +48,7 @@ jobs:
                 CC: ${{ matrix.compiler }}
                 MAKEVARS: ${{ matrix.makevars }}
               run: |
-                autoreconf -i
-                rm src/lexer.c src/lexer.h
-                rm src/parser.c src/parser.h
-                ./configure --with-oniguruma=builtin YACC="$(which bison) -y" $COVERAGE
+                ./configure --with-oniguruma=builtin $COVERAGE
```
wader, line 51: Was there a reason why scan-build would re-generate the lexer/parser code?

nicowilliams: I can't think of one.

nicowilliams:
LGTM!  Thanks!

.github/workflows/scanbuild.yml
```diff
@@ -48,10 +48,7 @@ jobs:
                 CC: ${{ matrix.compiler }}
                 MAKEVARS: ${{ matrix.makevars }}
               run: |
-                autoreconf -i
```
emanuele6, line 51: Oops

emanuele6: Should be fixed with https://github.com/jqlang/jq/commit/54fef09ac39d6fb003367d65709349c43078405e

nicowilliams: Thank you for fixing it @emanuele6!

PR #2761, opened 2023-07-24
Remove all #define _GNU_SOURCE in source files

emanuele6:
This is unnecessary since we are using AC_USE_SYSTEM_EXTENSIONS in configure.ac, which makes the compiler always run with -D_GNU_SOURCE=1 when building for GNU/Linux.

Also, the  #define _GNU_SOURCE  in src/util.c, and in src/inject_errors.c (--enable-inject-errors), ware not guarded by a #ifndef _GNU_SOURCE  so they were just triggering redefinition warnings.

---

This fixes the one warning that [#2695](https://github.com/jqlang/jq/pull/2695#issuecomment-1636968885) does not fix. :)

Also a similar warning only triggered in `--enable-inject-errors` builds.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1,6 +1,3 @@
-#ifndef _GNU_SOURCE
-# define _GNU_SOURCE
-#endif
 #ifndef __sun__
 # define _XOPEN_SOURCE
 # define _XOPEN_SOURCE_EXTENDED 1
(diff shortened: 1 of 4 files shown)


wader (approved):

itchyny (approved):

nicowilliams:
Thanks!

Issue #338, opened 2014-04-06
converting arrays to objects (iteration)


matter119:
I've been driving myself crazy trying to figure out the best method to do this, but I just can't seem to get it right. I have the following JSON data:

```
data: [[“building_A”,”floor_2”,[[“chair”,”1.0”],[“table”,”2.2”],[“desk”,”1.2”]],
          [“building_B”,”floor_1”,[[“chair”,”1.0”]]]
```

I would like to convert it to:

```
{ “index”: { “_id”:”building_A” }}
{ "building" : "building_A", "floor":"floor_2", "contents" : [ {"object":"chair","amount":1.0},{"object":"table","amount":2.2},{"object”:desk","amount":2.0}]}
{ “index”: { “_id”:”building_B” }}
{ "building" : "building_B", "floor":"floor_1", "contents" : [ {"object":"chair","amount":1.0}]}
```

But I cannot seem to get the contents array converted correctly. This comes close:

```
jq -c  '.data[] | { building: .[0], floor: .[1], contents: [{ object: .[2][][0], amount: .[2][][1]  }] }'
```

But the iteration is all wrong and I can seem to wrap my head around it.

Is there a forum where I can post this type of question?

Issue #337, opened 2014-04-05
Escape all the json except the values


unhammer:
I have an input file like

```
 { 
   "title": "My nice title", 
   "tagline": "Some tagline", 
  …
  }
```

and would like to get out

```
 <apertium-notrans>{ 
    "title": "</apertium-notrans>My nice title<apertium-notrans>",
    "tagline": "</apertium-notrans>Some tagline<apertium-notrans>", 
    …
   }</apertium-notrans>
```

– is something like this possible with jq? The manual seems most interested in just extracting or moving around values, but I would like to preserve all the surrounding structure as well, just escape it with these apertium-notrans tags.

PR #2764, opened 2023-07-24
Free expected value on invalid actual for test

wader:
valgrind (3.19.0) locally on macOS docker container seems to detect this leak but not CI.

```sh
$ cat modulemeta.test
modulemeta | .deps |= length
"c"
{"whatever":null,"deps":6}

$ valgrind --error-exitcode=1 --leak-check=full -q ./jq --run-tests < modulemeta.test
Test #1: 'modulemeta | .deps |= length' at line number 1
*** Insufficient results for test at line number 3: modulemeta | .deps |= length
0 of 1 tests passed (0 malformed, 0 skipped)
==14129== 482 (392 direct, 90 indirect) bytes in 1 blocks are definitely lost in loss record 7 of 8
==14129==    at 0x48407B4: malloc (vg_replace_malloc.c:381)
==14129==    by 0x13BC78: jv_mem_alloc (jv_alloc.c:122)
==14129==    by 0x1371C6: jvp_object_new (jv.c:1521)
==14129==    by 0x145840: parse_token (jv_parse.c:167)
==14129==    by 0x145840: scan (jv_parse.c:684)
==14129==    by 0x145840: jv_parser_next (jv_parse.c:793)
==14129==    by 0x1463CB: jv_parse_sized (jv_parse.c:865)
==14129==    by 0x136BD9: run_jq_tests (jq_test.c:183)
==14129==    by 0x1370FA: jq_testsuite (jq_test.c:34)
==14129==    by 0x13135C: main (main.c:582)
==14129==
```

### DIFF
--- a/src/jq_test.c
+++ b/src/jq_test.c
@@ -188,6 +188,7 @@ static void run_jq_tests(jv lib_dirs, int verbose, FILE *testdata, int skip, int
       }
       jv actual = jq_next(jq);
       if (!jv_is_valid(actual)) {
+        jv_free(expected);
         jv_free(actual);
         printf("*** Insufficient results for test at line number %u: %s\n", lineno, prog);
         pass = 0;


emanuele6:
Makes sense. If you want, you can add a regression test in tests/shtest with something like this:
```bash
./jq --run-tests <<'EOF'
.
1
foo
EOF
```

wader:
On another note: should we have tests that trigger test errors like this? or am i doing something wrong?

nicowilliams:
> On another note: should we have tests that trigger test errors like this? or am i doing something wrong?

You mean tests of the test system?  This leak is not a big deal because you're not supposed to have failing tests :)  But, yeah, I do think we probably ought to have a test in `tests/shtest` that `jq --run-tests` does indeed fail when tests fed to it fail, and run that under valgrind.  Good point!

nicowilliams:
Thanks!

wader:
> > On another note: should we have tests that trigger test errors like this? or am i doing something wrong?
> 
> You mean tests of the test system? This leak is not a big deal because you're not supposed to have failing tests :) But, yeah, I do think we probably ought to have a test in `tests/shtest` that `jq --run-tests` does indeed fail when tests fed to it fail, and run that under valgrind. Good point!

Sorry that sounded a bit weird, what i mean was this:
```sh
➜  jq git:(master) ✗ ./jq --run-tests < tests/jq.test 2>&1 | grep -C 2 Insufficient
Test #337: 'include "\(a)"; 0' at line number 1582
Test #338: 'modulemeta' at line number 1585
*** Insufficient results for test at line number 1587: modulemeta
Test #339: 'modulemeta | .deps |= length' at line number 1589
*** Insufficient results for test at line number 1591: modulemeta | .deps |= length
Test #340: 'import "syntaxerror" as e; .' at line number 1594
Test #341: '%::wat' at line number 1598
```

We current seem to have some tests that fail to be parsed, or is this how it's intended to be? but i agree leaks in the test code is not really a big deal, maybe nice to fix for less confusion while debugging something

nicowilliams:
> ➜ jq git:(master) ✗ ./jq --run-tests < tests/jq.test 2>&1 | grep -C 2 Insufficient Test #337: 'include "\(a)"; 0' at line number 1582 Test #338: 'modulemeta' at line number 1585 *** Insufficient results for test at line number 1587: modulemeta

The issue is that you're missing some `-L` options so tests of the module system can work.

wader:
Ah doh that makes sense

PR #2766, opened 2023-07-25
Reorder the command line options in the manual

itchyny:
I have always thought the order of command line options in the manual is very unfriendly. It lists `--seq` and `--stream` at the top, but they are used less often than `--null-input` and `--compact-output`. In #2747, I ordered the list carefully to make it easy  to understand, and in this PR, I reordered the manual to match the output of `--help`. Roughly in the order of input options, output options, advanced input/output options (`--stream`, `--seq`), module options, arg options, and other options.


docs/content/manual/manual.yml
```diff
@@ -104,74 +104,58 @@ sections:
       You can affect how jq reads and writes its input and output
       using some command-line options:
 
-      * `--version`/`-V`:
```
nicowilliams, line 107: You don't think `--version` should be first?

nicowilliams:
LGTM!

docs/content/manual/manual.yml
itchyny: I think the option (at least the long option) is obvious and don't think it should be first.

itchyny:
Thanks.

PR #2768, opened 2023-07-25
Remove deprecated --argfile option

itchyny:
The `--argfile` option has been documented as `Do not use.` since jq 1.5. It has been 8 years, so let's remove this.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -257,14 +257,6 @@ sections:
         available in the program and has a string whose contents are to the texts
         in the file named `bar`.
 
-      * `--argfile variable-name filename`:
-
-        Do not use.  Use `--slurpfile` instead.
-
-        (This option is like `--slurpfile`, but when the file has just
-        one text, then that is used, else an array of texts is used as
-        in `--slurpfile`.)
-
       * `--args`:
 
         Remaining arguments are positional string arguments.  These are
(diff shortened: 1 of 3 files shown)


wader:
Oh never used i think but assumed it was like `--raw-file` but is more like `--argjsonarrayfirst`?

wader (approved):

wader:
Is there some good compatibility argument for keeping it?

nicowilliams:
Thanks!

pkoppstein:
@wader asked:

> Is there some good compatibility argument for keeping it?

Yes, I believe so, mainly because (a) it's all over the internets (google counts 3,260 pages) and (b) there isn't yet a good substitute (e.g. --slurpfile always slurps).

Part of the reason it's been in limbo so long is that there has been some hope that jq's I/O functionality would be significantly overhauled, at which time `--argfile` would die a natural death.  

Anyway, in my opinion, it would have been better to leave it in for 1.7; in fact, maybe it's not too late to revert this particular commit???? @nicowilliams? @itchyny?

nicowilliams:
@pkoppstein it's been removed.  It will stay removed.  There was a deprecation warning for a long time.  People will adjust.

philpax:
For anyone else coming across this PR who needs to update their existing scripts, and are OK with subpar performance: you can replace `--argfile` with `--slurpfile` and index into the resulting array.

Here's an example:
```
jq -c '{ title, comments: $comments[.name] }' --argfile comments "comments.ndjson" "submissions.ndjson" > output.ndjson
```
should become
```
jq -c '{ title, comments: ($comments[0][.name]) }' --slurpfile comments "comments.ndjson" "submissions.ndjson" > output.ndjson
```

This may not be sufficient for your use case, or there may be better ways of doing it, but this got me back up and running.

PR #2771, opened 2023-07-25
Make sure to free default @text format on error

emanuele6:
It seems that bison doesn't call destructors for mid-rule action components on error, since it does not know their type.

A mid-rule action was used to allocate the "text" string used as format by string literals without a format, which would leak on error.
This patch replaces it with a new NoFormat component of type \<literal\>.
Now bison will call jv_free() on that string after a syntax error.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60836

```
$ ./jq '"'
jq: error: syntax error, unexpected end of file, expecting QQSTRING_TEXT or QQSTRING_INTERP_START or QQSTRING_END (Unix shell quoting issues?) at <top-level>, line 1:
"
jq: 1 compile error

=================================================================
==1495450==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 21 byte(s) in 1 object(s) allocated from:
    #0 0x7fc21aee1359 in __interceptor_malloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x557ccb6ff608 in jv_mem_alloc src/jv_alloc.c:141

SUMMARY: AddressSanitizer: 21 byte(s) leaked in 1 allocation(s).
```

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -366,9 +366,9 @@ enum yysymbol_kind_t
   YYSYMBOL_FuncDef = 79,                   /* FuncDef  */
   YYSYMBOL_Params = 80,                    /* Params  */
   YYSYMBOL_Param = 81,                     /* Param  */
-  YYSYMBOL_String = 82,                    /* String  */
-  YYSYMBOL_83_1 = 83,                      /* @1  */
-  YYSYMBOL_84_2 = 84,                      /* @2  */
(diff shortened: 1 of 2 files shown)


nicowilliams:
LGTM.  Waiting on builds.

PR #2774, opened 2023-07-25
Only care about RS when parsing with --seq

emanuele6:
RS should only only have special meaning when parsing json-seq.
Before this fix, RS was sometimes treated as a value terminator, causing weird results.

Bug discovered by OSS fuzz.
Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60835

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -610,11 +610,11 @@ static int stream_check_done(struct jv_parser* p, jv* out) {
   }
 }
 
-static int parse_check_truncation(struct jv_parser* p) {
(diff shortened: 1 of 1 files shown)


emanuele6:
Hmm, I tried opening this PR using `gh` (github's command line tool), and it pushed the branch to the jqlang repo instead of my fork; sorry about that.

nicowilliams:
> Hmm, I tried opening this PR using `gh` (github's command line tool), and it pushed the branch to the jqlang repo instead of my fork; sorry about that.

That's not really a problem.  I've done that on occasion.

nicowilliams:
Thanks!

PR #2775, opened 2023-07-26
Enable gamma, drem on macOS

itchyny:
This is very related to what I do in #2756, but current implementation disallows gamma and drem on macOS. I fixed the identifier definitions to enable gamma and drem on macOS. I also fixed some macro for exp10 and significand to make them look similar.


nicowilliams:
Thanks!

PR #2776, opened 2023-07-26
builtin.jq: whilst

pkoppstein:
whilst is like while but more conventional.


src/builtin.jq
```diff
@@ -139,13 +139,18 @@ def gsub($re; s): sub($re; s; "g");
 ########################################################################
 # generic iterator/generator
 def while(cond; update):
-     def _while:
-         if cond then ., (update | _while) else empty end;
-     _while;
+  def _while:
```
nicowilliams, line 142: We need to define a code style for jq.  I've become very happy with 4-space indentation except for 2 spaces for `|`, like this:

```jq
    .
  | stuff
```

I think until we decide on a style for jq code, maybe it'd be best to not restyle existing jq code.

itchyny:
Any reference to convince this naming is adopted somewhere else?

src/builtin.jq
pkoppstein: I was just trying to ensure while/whilst/until were consistently formatted, admittedly in the 2-space style -- two spaces saves space-time, no?  I thought that was important for builtin.jq ...

nicowilliams:
> whilst is like while but more conventional.

More conventional?  I think I made a mistake making `until/2` so different from `while/2`, but `whilst/2` isn't more like `until/2`, so I don't quite understand the motivation.  Can you show us a problem that's easier to solve with `whilst/2` than with `while/2`?

pkoppstein:
@itchyny wrote:
> Any reference to convince this naming is adopted somewhere else?

Who knows?  I couldn't think of a better name, and the functionality of the proposed `whilst` is very similar to the existing `while`, as the word "whilst" is similar to "while".  More importantly:
```
In some Midlands dialects, there has been a distinct difference - "whilst" meaning "during the time when" and "while" meaning "until, up to point when", e.g. "Look after this whilst I'm gone" or "Look after this while I come back". Sounds wrong to the purist, but in Rutland they would know what you meant.
```

In other words, this distinction which can be drawn in English parallels the distinction between the proposed and existing jq defs.

(I was toying with `emit_while` but that would probably cause even more dissatisfaction because it suggests a closer affinity to jq's `while` than it has.)

src/builtin.jq
nicowilliams: Changing style of surrounding code to be more consistent is something I like to do -though it can be irritating because it clutters diffs and complicates use of `git blame`-, but in this case I'm not sure what style we ought to adopt, and my own jq-coding style has change over time, so for `src/builtin.jq` I would prefer we don't gratuitously re-style until we decide on a style.  Conversely, I'm not asking you to adopt a style for jq code for now either -- if eventually we have to restyle all of `src/builtin.jq`, so be it.

(For C code in jq the style is fairly consistent already, and very much in @stedolan's C style.)

pkoppstein:
@nicowilliams wrote:

> Can you show us a problem ...

Believe me, whilst/2 was born of necessity.  On the web: 
* https://rosettacode.org/wiki/Abelian_sandpile_model/Identity#jq
* https://rosettacode.org/wiki/Numbers_k_such_that_the_last_letter_of_k_is_the_same_as_the_first_letter_of_k%2B1#jq

By saying "whilst is like while but more conventional", I meant that this new def is somewhat similar to jq's existing `while` but is closer to traditional while loops such as C's:

https://www.tutorialspoint.com/cprogramming/c_while_loop.htm

Perhaps those familiar with such "while" loops would prefer that the "while" in builtin.jq behaved like that, but that's water under the bridge.

wader:
> We need to define a code style for jq. I've become very happy with 4-space indentation except for 2 spaces for |, like this:

For larger jq source files i've settle on a haskell/elm-ish style, 2 space indent and then explicit parentheses if some pipelines ends up multiline to make them align nicely and looks similar to array and objects literals. Also for some reason started using underscore to prefix for local functions. Can have a look at jqjq if you want an example.

Also wish there was a jqfmt (and vet/lint) :)

src/builtin.jq
pkoppstein: @nicowilliams - In case you haven't noticed, @itchyny's builtin.jq in gojq uses the two-space style :-)

https://github.com/itchyny/gojq/blob/main/builtin.jq

itchyny:
Honestly feels kind of a literature that non-native speakers cannot tell the difference (or even worse cannot spell correctly).

src/builtin.jq
nicowilliams: @pkoppstein no, I've not noticed.  It's on my list to eventually go take a look at `fq`, `jqjq`, `gojq`, `rq`, and others, but my time is limited.  What they do for style is interesting, but what they do for semantics is much more interesting as we want jq-alikes to interoperate.  As for style, like I said, my own jq style has evolved and I think we should discuss that separately and not gratuitously restyle `src/builtin.jq` _for the time being_ even though I otherwise would prefer that we do restyle as we go.

nicowilliams:
> Honestly feels kind of a literature that non-native speakers cannot tell the difference (or even worse cannot spell correctly).

The main problem is that I get no real sense of the semantic difference between `while/2` and `whilst/2` from the two functions' names.  English is my third language, and that might be why, but at any rate, I'm asking @pkoppstein to explain how this new builtin is useful.

pkoppstein:
@nicowilliams wrote:
> explain how this new builtin is useful.

References to code examples were already provided - see "rosettacode" refs above.
But the important point is that the utility is similar to C's `while`.

As for naming - I don't see a huge problem, as manual.yml makes the distinction quite clear.  

As for mixing them up - well, we already have that problem because jq's `while` has a different logic than C's.  

If you don't like `whilst`, I doubt you'll like any of the other alternatives I've considered - whilst_do; while_recurse; if_recurse; cond; ...

[@itchyny - Since the name `while` has already been taken, what alternatives would you suggest?]

pkoppstein:
PR https://github.com/jqlang/jq/pull/2767 adds `abs` so this ER could be closed but for one point - the status of `$number|length`.

Until recently (that is, prior to 21 June 2023), v1.6 of manual.yml was silent about `$number|length` (*), and so
if one had wanted to, the v1.7 version could easily have continued to be silent.  

Since it would still be possible to deprecate or discourage `length` on numbers, for now I'll leave this issue open for further comment.

It might also be worth reflecting on the wisdom of being cautious about **retroactively adding specification detail to prior versions of manual.yml**, at least when there's the probability that its omission was intentional (the probability here being related to doubts about the superabundance of polymorphism).

---
(*)

dd5ce98c docs/content/manual/v1.6/manual.yml   (Alex Jurkiewicz        2023-06-21 19:21:39 +1000  750)           - The length of a **number** is its absolute value.

PR #2777, opened 2023-07-27
builtin.jq: lpad, nwise

pkoppstein:
lpad: left pad
Retain _nwise for backward compatibility.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -1975,6 +1975,38 @@ sections:
             input: 'null'
             output: ['"less"']
 
+      - title: "`lpad($len; $fill)`, `lpad($len)`"
+        body: |
+
+          These functions can be used to pad the input string. `$len` should be a non-negative
+          integer, and `$fill` should be a non-empty string.
+
+          If the input string has length less than `$len`, then both these functions
+          will pad it using the characters in the `$fill` string to produce
+          a string of length `$len`. Typically, `$fill` is given as a string of length 1.
+
+          `lpad($len)` is equivalent to `lpad($len; " ")`.
+
+        examples:
+          - program: 'map(lpad(2))'
+            input: '["", "a"]'
+            output: ['["  ", " a"]']
+
+          - program: 'map(lpad(4; "- "))'
+            input: '["", "a"]'
+            output: ['["- - ","- -a"]']
+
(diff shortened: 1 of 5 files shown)


nicowilliams:
> Retain _nwise for backward compatibility.

No need.  If it starts with an `_` then it's jq-private.

pkoppstein:
>  If it starts with an _ then it's jq-private.

True, but so far as I know, there's hardly any cost to retaining it for now, whereas removing it for a point release would probably just be aggravating to anyone like me who has many scripts that use it.  

So my thought was: jettison it in 2.0 but in the meantime, isn't it nice to know there's some extra baggage that can be thrown overboard if that becomes necessary?

wader:
Should we open an issue discussing possible additional convenience builtins for post 1.7? but i think we should be careful to not clutter the namespace too much. `nwise` looks like a good candidiate, i have `chunk` in fq that works similar, would gladly replace it. See https://github.com/wader/fq/blob/master/doc/usage.md#function-added-in-fq for some other additional ones i've come to use a lot.

nicowilliams:
> Should we open an issue discussing possible additional convenience builtins for post 1.7? but i think we should be careful to not clutter the namespace too much. `nwise` looks like a good candidiate, i have `chunk` in fq that works similar, would gladly replace it. See https://github.com/wader/fq/blob/master/doc/usage.md#function-added-in-fq for some other additional ones i've come to use a lot.

Yes, I think maybe we should adopt a change freeze so we can focus on making a release.  This weekend I'll put time into re-generating the release key (or changing its name) and then making GitHub Actions workflows sign artifacts.  Why not right now?  Because I need a solid chunk of time to figure out the details.  I find working with CI/CDs to be very time consuming.  Working on features as time allows is much easier by comparison.  I think release mechanics and misc bugs (and any other bugs found by fuzzing in the next few days) should be the only things we should focus on until we get a release done.

pkoppstein:
Since one of the "checks" was (mysteriously) unsuccessful, and since this PR is already quite old,
and since I also wanted to add a stream-oriented nwise/2, I'm closing it in favor of https://github.com/jqlang/jq/pull/2786

In this superseding PR, I've dropped both _nwise/1 and _nwise/2 (in accordance with remarks made by @nicowilliams),
but I'd be happy to add either or both of them back.

PR #2780, opened 2023-07-27
Hardcode 0/0 => NaN instead of letting C divide 0 by 0

emanuele6:
In C, division by 0 is unspecified in all cases, so we can't expect a C compiler to always generate code that makes a / b yield NaN if both a and b are 0.

Fixup from 886a9b18b10e4e31b8c746d1f243043d2b7ea234

---
Ref: https://github.com/jqlang/jq/pull/2253#discussion_r1276708144

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -383,8 +383,13 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
 static jv f_divide(jq_state *jq, jv input, jv a, jv b) {
   jv_free(input);
   if (jv_get_kind(a) == JV_KIND_NUMBER && jv_get_kind(b) == JV_KIND_NUMBER) {
-    if (jv_number_value(b) == 0.0 && jv_number_value(a) != 0.0)
-      return type_error2(a, b, "cannot be divided because the divisor is zero");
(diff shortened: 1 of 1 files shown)


emanuele6:
I am not sure if this is the right fix, but it at least removes the undefined behaviour keeping the existing intended behaviour.

Do we really want to make `0/0` return NaN instead of throwing an division by 0 error? Why are we making `0/0` return Nan, but `1/0` throw an error instead of returning positive infinity? Do we want to make the arithmetic operators IEEE 754 compliant?

emanuele6:
> Do we really want to make 0/0 return NaN instead of throwing an division by 0 error? Why are we making 0/0 return Nan, but 1/0 throw an error instead of returning positive infinity?

I guess @itchyny made `f_divide` return NaN for `a==0.0 && b==0.0` because constant-folded `0/0` would yield NaN, while `0 | ./0` would throw an error, and wanted to make things more consistent.

Constant-folding considers an error any expression that returns infinity (positive or negative), while `f_divide` used to only consider expressions with `b==0.0` errors; now it considers an error any expression with `b==0.0&&a!=0.0` which is still not exactly the same as constant-folding, but it's closer.

Another reason why we should centralise how we do division.

PR #2782, opened 2023-07-28
Fix memory leak for  { $foo: bar }

emanuele6:
{ BINDING: ExpD } wasn't freeing BINDING.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60865

---

OSS Fuzz insists we fix this asap.
file


emanuele6:
I added a test that uses `{ BINDING }` and `{ BINDING: foo }`

nicowilliams:
Thanks!

PR #2783, opened 2023-07-28
Make sure jv_string_value($1) is used before $1 is freed

emanuele6:
While reviewing `src/parser.y`, I noticed that the code in the rules for the `as { $x }` and `as { $x: foo }` destructing bindings relied on the rightmost argument expression of a function call being evaluated before the leftmost argument expression.
This is currently not causing problems, but it is not guaranteed that the expression of the last argument is evaluated first.
This patch uses a temporary variable to ensure it is evaluated first, so we don't accidentally use `$1` after free.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -977,11 +977,11 @@ static const yytype_int16 yyrline[] =
      740,   743,   746,   749,   752,   755,   758,   761,   783,   787,
      791,   794,   806,   811,   812,   813,   814,   817,   820,   825,
      830,   833,   838,   841,   846,   850,   853,   858,   861,   866,
-     869,   874,   877,   880,   883,   886,   889,   897,   903,   906,
-     909,   912,   915,   918,   921,   924,   927,   930,   933,   936,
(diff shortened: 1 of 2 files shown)


emanuele6:
<s>`{ $x: foo }` and</s> `{ $x }` also have this problem in object construction, I will fix those too

emanuele6:
There were also a few other places using `jv_copy()` that have the same problem

emanuele6:
I am not sure why this didn't trigger any use after free errors. Some of those calls rely on the first argument being evaluated first, and others rely on the last argument being evaluated first. Maybe we were lucky and the refcount always happens to be greater than 1 in those places.

emanuele6:
Oh, never mind. I just checked the definition of `gen_const`, and noticed that it does not free its argument, it only gets ownership of that reference (like `std::move` in C++), so, it is fine to call `jv_copy(foo)`, or `jv_string_value(foo)` after calling `gen_const(foo)`, since `foo` still points to that reference, that was not freed.
Sorry about that, closing...

PR #2786, opened 2023-07-29
builtin.jq: lpad, nwise/1, nwise/2

pkoppstein:
lpad: left pad
nwise/1 replaces _nwise/1 and works on arrays and strings. nwise($n; stream) is a stream-oriented version of nwise/1, and places no limits on what may appear in the stream. It has $n as its first argument mainly in deference to limit/2.

_nwise/2 is dropped.

### DIFF
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1860,6 +1860,10 @@ nwise(2)
 null
 []
 
+try nwise(0) catch .
+[1]
+"nwise: argument should be non-negative"
+
 map(lpad(4))
 ["","a"]
 ["    ","   a"]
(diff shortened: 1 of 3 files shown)


tests/jq.test
```diff
@@ -1848,6 +1848,26 @@ map(abs)
 [0.1,1000000000000000002]
 [1e-1, 1000000000000000002]
 
+nwise(2)
+null
+null
+
+nwise(2)
+[]
+[]
```
wader, line 1857: Add a negative $n test case as it seems to get stuck atm, should throw error?

pkoppstein: Good points.  QED. 

Not sure why the "Building man page...." test fails, as I don't have that problem on a Mac.  Any ideas?

itchyny:
Let's discuss on the design of padding function(s) in #2033. Especially on how this is less useful on printing strings with multi-width characters in table-like format in terminals.

PR #2790, opened 2023-07-29
Require a main program (fix #2785)

### LINKED ISSUE #2785: Using `-f` with a file that has no main program can assume the main program is `.`, or produce an incorrect error
See #2353.

Reproducer:

```
$ printf 'def a: 0;\n' > m.jq
$ jq -n -f m.jq
null
$ printf 'def a: 0;\ndef b: a;' > m.jq
$ jq -n -f m.jq
jq: error: a/0 is not defined at <top-level>, line 2:
def b: a;
jq: 1 compile error
$ 
```

What's happening is that if there's no main program in the file specified with `-f` then jq assumes `.`, but probably does not bother fully-linking the jq code in that file, then it detects that there are unbound symbols and produces an error.

What should happen: it should always be an error to use `-f` with a jq program file that has no main program.

nicowilliams:


nicowilliams:
I wouldn't go so far as to say that this has to be in 1.7.

src/linker.c
```diff
@@ -407,6 +407,12 @@ int load_program(jq_state *jq, struct locfile* src, block *out_block) {
   if (nerrors)
     return nerrors;
 
+  if (!block_has_main(program)) {
+    jq_report_error(jq, jv_string_fmt("jq: error: Top-level program given (try \".\")\n"));
```
emanuele6: `s/jv_string_fmt/jv_string/`

src/linker.c
```diff
@@ -407,6 +407,12 @@ int load_program(jq_state *jq, struct locfile* src, block *out_block) {
   if (nerrors)
     return nerrors;
 
+  if (!block_has_main(program)) {
+    jq_report_error(jq, jv_string("jq: error: Top-level program not given (try \".\")\n"));
```
itchyny: I don't think we need a newline at the end because err_cb prints with it.

itchyny (approved):

nicowilliams:
Pushing w/o waiting for checks to save builder minutes.

PR #2793, opened 2023-07-30
Fix 2189

nicowilliams:


docs/content/manual/manual.yml
```diff
@@ -568,13 +572,13 @@ sections:
...
-          pretty much the same as the Unix shell's pipe, if you're used to
-          that.
+          similar to the Unix shell's pipe, if you're used to that.
 
           If the one on the left produces multiple results, the one on
           the right will be run for each of those results. So, the
           expression `.[] | .foo` retrieves the "foo" field of each
-          element of the input array.
+          element of the input array.  This is a cartesian product,
+          which can be surprising.
```
itchyny, line 581: Maybe we need an example to clarify what this sentence means? Like `1, 2 | ., -.`. If we mention cartesian product, we may need it in the binary operator section; `(1, 2) * (3, 4)`.

emanuele6 (approved):
LGTM
Can you regenerate `jq.1.prebuilt` before merging this? (It still says `July 2023` instead of `August 2023`)

PR #2795, opened 2023-07-31
Simple and efficient implementation of walk/1

### LINKED ISSUE #2584: Deleting objects using walk has side effects

### LINKED ISSUE #2611: ER: simpler, faster `walk/1`

pkoppstein:
Resolves #2584; also resolves #2611
and supersedes https://github.com/jqlang/jq/pull/2655

Note that according to the revised implementation:

`{x:0} | walk(.,1)` is equivalent to `{x:0} | walk(.), walk(1)`

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -246,13 +246,14 @@ def bsearch($target):
 
 # Apply f to composite entities recursively, and to atoms
 def walk(f):
-  . as $in
-  | if type == "object" then
-      reduce keys_unsorted[] as $key
-        ( {}; . + { ($key):  ($in[$key] | walk(f)) } ) | f
-  elif type == "array" then map( walk(f) ) | f
-  else f
-  end;
+  def w:
+    if type == "object"
+    then map_values(w)
+    elif type == "array" then map(w)
+    else .
+    end
+    | f;
+  w;
 
 # pathexps could be a stream of dot-paths
 def pick(pathexps):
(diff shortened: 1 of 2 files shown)


itchyny (approved):
LGTM

pkoppstein:
@itchyny -  Thanks.  Sorry about the rebase problem.  I just used the "Rebase" button on GitHub Desktop, followed by "Push".

PR #2797, opened 2023-07-31
Fix constant folding of division and reminder with zero divisor

itchyny:
Previously constant folding of zero division (e.x. 1/0) produces a compile error. This was incorrectly implemented by checking if the division result is infinite, so produces wrong results compared to the query where no constant folding is processed (e.x. 1e308/0.1). This patch delays the operation when the divisor is zero. This makes the results more consistent, but changes the exit code on zero division from 3 to 5. Also 0/0 now produces the zero division error, not NaN.

This patch also fixes the modulo operation. Previously constant folding logic does not take care of the % operator, but now it folds if the both dividend and divisor are safe numbers to cast to the integer type, and the divisor is not zero. This patch also fixes some code that relies on undefined cast behaviors in C. The modulo operation produces NaN if either the dividend or divisor is NaN.

Closes #2252, also closes #2780.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -383,7 +383,7 @@ static jv f_multiply(jq_state *jq, jv input, jv a, jv b) {
 static jv f_divide(jq_state *jq, jv input, jv a, jv b) {
   jv_free(input);
   if (jv_get_kind(a) == JV_KIND_NUMBER && jv_get_kind(b) == JV_KIND_NUMBER) {
-    if (jv_number_value(b) == 0.0 && jv_number_value(a) != 0.0)
+    if (jv_number_value(b) == 0.0)
       return type_error2(a, b, "cannot be divided because the divisor is zero");
     jv r = jv_number(jv_number_value(a) / jv_number_value(b));
     jv_free(a);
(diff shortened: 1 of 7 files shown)


src/builtin.c
```diff
@@ -396,21 +396,30 @@ static jv f_divide(jq_state *jq, jv input, jv a, jv b) {
   }
 }
 
+#define dtoi(n) ((n) < INTMAX_MIN ? INTMAX_MIN : -(n) < INTMAX_MIN ? INTMAX_MAX : (intmax_t)(n))
```
nicowilliams, line 399: Oh, `intmax_t`...  `intmax_t` has to be used very carefully, and... you are using it carefully.  Good!

nicowilliams:
Looks good to me!

wader (approved):
LGTM but i'm not very familiar with div/mod int safety

nicowilliams:
Thanks!

emanuele6:
This should have fixed https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60848

`0/0` used to trigger an UBSAN warning when it was constant folded.

emanuele6:
I think you changed the name of your `is_unsafe_double` macro to `is_unsafe_to_int_cast` in `src/parser.y`, but you forgot to regenerate `src/parser.c` after that

nicowilliams:
> I think you changed the name of your `is_unsafe_double` macro to `is_unsafe_to_int_cast` in `src/parser.y`, but you forgot to regenerate `src/parser.c` after that

Good catch!

I just pushed a re-gen in 90a6b2d9.

PR #2798, opened 2023-07-31
Add new release public key

nicowilliams:


wader (approved):
LGMT but don't know if there is any convension. Googled a bit only found https://github.com/yt-dlp/yt-dlp/blob/master/public.key

PR #2800, opened 2023-07-31
Declare  cfunction.fptr  as jv (*)() to avoid having to cast everywhere

emanuele6:
You only need to specify the return type in a function pointer declaration in C.

If you use `()` in the declaration, the function pointer can be called with any arguments, and the type of the arguments is decided for each function call based on the types of the arguments used for the call. (To declare a function pointer for a function with no arguments, you use `(void)`.)

Since all the cfunction structs have a  fptr  that points to a functions that return `jv`, not `void`, we can we can just declare cfunction.fptr as `jv (*)()` and avoid having those annoying and technically not C-standard compliant casts everywhere.


nicowilliams:
~Should we revert f733a153?~  EDIT: Ah, you did.

nicowilliams:
Thanks!

itchyny:
I'm facing different warnings with clang after this fix.
```c
src/execute.c:924:35: warning: passing arguments to a function without a prototype is deprecated in all versions of C and
      is not supported in C2x [-Wdeprecated-non-prototype]
      case 1: top = function->fptr(jq, in[0]); break;
                                  ^
src/execute.c:925:35: warning: passing arguments to a function without a prototype is deprecated in all versions of C and
      is not supported in C2x [-Wdeprecated-non-prototype]
      case 2: top = function->fptr(jq, in[0], in[1]); break;
                                  ^
src/execute.c:926:35: warning: passing arguments to a function without a prototype is deprecated in all versions of C and
      is not supported in C2x [-Wdeprecated-non-prototype]
      case 3: top = function->fptr(jq, in[0], in[1], in[2]); break;
                                  ^
src/execute.c:927:35: warning: passing arguments to a function without a prototype is deprecated in all versions of C and
      is not supported in C2x [-Wdeprecated-non-prototype]
      case 4: top = function->fptr(jq, in[0], in[1], in[2], in[3]); break;
                                  ^
src/execute.c:928:35: warning: passing arguments to a function without a prototype is deprecated in all versions of C and
      is not supported in C2x [-Wdeprecated-non-prototype]
      case 5: top = function->fptr(jq, in[0], in[1], in[2], in[3], in[4]); break;
                                  ^
```

emanuele6:
Hmm, I knew `C23` removes support for `K&R` function declarations, but I didn't know it also removed support for calling K&R function pointers.
Here `clang` is complaining that `K&R` function pointer calls are "deprecated" in ANSI/Standard C; I guess we can cast those pointers before calling to make it happy.

PR #2801, opened 2023-07-31
Enable stack protection (CI release executables)

### LINKED ISSUE #1514: Enable stack protection in released binary
I've recently launched AWS Inspector scan across our systems, and one of the findings was:

> **Finding:** The following executable files on instance i-xxxxxx do not support stack cookies: /usr/bin/jq.  
**Severity:** Medium  
**Description:** This rule detects the presence of third-party software that is compiled without support for stack cookies. Stack cookies increase system security by defending against stack-based buffer overflow and other memory corruption attacks.  
**Recommendation:** It is recommended that you uninstall this software from your assessment target if you are not using it, or contact the vendor to get an updated version of this software with stack cookies enabled.

I've then checked latest available release with [hardening-includes](https://packages.debian.org/wheezy/hardening-includes) package:

```
# ./jq-linux64 --version
jq-1.5
# sha256sum ./jq-linux64
c6b3a7d7d3e7b70c6f51b706a3b90bd01833846c54d32ca32f0027f00226ff6d  ./jq-linux64
# hardening-check ./jq-linux64
./jq-linux64:
 Position Independent Executable: no, normal executable!
 Stack protected: no, not found!
 Fortify Source functions: unknown, no protectable libc functions used
 Read-only relocations: yes
 Immediate binding: no, not found!
```

Then I've downloaded corresponding source package `jq-1.5.tar.gz` and compiled it with just `./configure` and `make`. That gave me this result:

```
# ./jq --version
jq-1.5-1-g940132e-dirty
# hardening-check ./jq
./jq:
 Position Independent Executable: no, normal executable!
 Stack protected: yes
 Fortify Source functions: yes (some protected functions found)
 Read-only relocations: yes
 Immediate binding: no, not found!
```

After deploying this binary to my hosts, subsequent AWS Inspector scans raised no security issues about `jq`.

Please consider enabling security features for released binaries by default.

nicowilliams:
Resolves #1514.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,8 @@ jobs:
             --disable-valgrind \
             --with-oniguruma=builtin \
             --enable-static \
-            --enable-all-static
+            --enable-all-static \
+            CFLAGS="-O2 -pthread -fstack-protector-all"
           make -j"$(nproc)"
           file ./jq
           cp ./jq jq-${{ env.SUFFIX }}
@@ -157,7 +158,8 @@ jobs:
             --disable-valgrind \
             --with-oniguruma=builtin \
             --enable-static \
-            --enable-all-static
+            --enable-all-static \
+            CFLAGS="-O2 -pthread -fstack-protector-all"
           make -j"$(nproc)"
           strip ./jq
           file ./jq
@@ -234,7 +236,8 @@ jobs:
             --with-oniguruma=builtin \
             --disable-shared \
             --enable-static \
-            --enable-all-static
+            --enable-all-static \
+            CFLAGS="-O2 -pthread -fstack-protector-all"
           make -j$(nproc)
           file ./jq.exe
           cp ./jq.exe jq-${{ env.SUFFIX }}.exe


.github/workflows/ci.yml
```diff
@@ -94,6 +94,7 @@ jobs:
             --with-oniguruma=builtin \
             --enable-static \
             --enable-all-static
```
itchyny: Needs line continuation.

nicowilliams:
`-mshstk` didn't work, though I'm sure it would work if we only use clang.  I suppose we could switch to clang for this.  Also, I should enable this for OS X and for Windows too.

.github/workflows/ci.yml
```diff
@@ -158,7 +159,8 @@ jobs:
             --disable-valgrind \
             --with-oniguruma=builtin \
             --enable-static \
-            --enable-all-static
+            --enable-all-static \
+            CFLAGS="-O2 -pthread -fstack-protector-all"
```
wader, line 162: Doh i somehow assumed `-fsanitize=safe-stack` was clang's versions of `-fstack-protector-all` but seems to be two different stack protection techniques.

I still haven't managed to get `-fsanitize=safe-stack` to work for darwin x86, but do seems to work on linux 🤷

wader:
I think this is good enough for 1.7. It would be great post 1.7 to provide these things using configure options somehow.

wader (approved):

itchyny:
Did you check the executables built on CI to make sure hardening-check?

nicowilliams:
> Did you check the executables built on CI to make sure hardening-check?

Not yet.  Though, hmm, how would I?  Use a debugger to try to force a buffer overflow?

wader:
> > Did you check the executables built on CI to make sure hardening-check?
> 
> Not yet. Though, hmm, how would I? Use a debugger to try to force a buffer overflow?

Seems like [hardened-check](https://manpages.ubuntu.com/manpages/trusty/man1/hardening-check.1.html) should be able to detect it heuristically for ELF at least:
> Stack Protected<br>
>               This indicates that there is evidence that the ELF was compiled with the [gcc]>(https://manpages.ubuntu.com/manpages/trusty/man1/gcc.1.html)(1)
>               option -fstack-protector (e.g. uses __stack_chk_fail). The program will be
>               resistant to having its stack overflowed.<br>           
>               When an executable was built without any character arrays being allocated on the
>               stack, this check will lead to false alarms (since there is no use of
>               __stack_chk_fail), even though it was compiled with the correct options.

wader:
@nicowilliams noticed a fixup commit slipped in

nicowilliams:
image

What's stripping the executable?

wader:
Think this https://github.com/jqlang/jq/blob/master/.github/workflows/ci.yml#L74 probably?

nicowilliams:
> Think this https://github.com/jqlang/jq/blob/master/.github/workflows/ci.yml#L74 probably?

Ay, yes.  Well, I suppose I could build jq twice, once w/o stripping, to check if it has stack protection.  Or maybe just punt.  I'm doing a build without `-s` and if that works I'll put it back and drop the `nm` check and call it a day.

nicowilliams:
Without stripping we see:

image

So I'm going to undo the `nm` check and restore stripping.

wader (approved):
👍 I think we can makes this nicer after 1.7

nicowilliams:
I wonder if we could have a command-line option to indicate whether jq was built with any stack protection features.  Or maybe we could have a command-line option that shows the build options used to build jq, just like many programs have.

wader:
> I wonder if we could have a command-line option to indicate whether jq was built with any stack protection features. Or maybe we could have a command-line option that shows the build options used to build jq, just like many programs have.

Yeap that feels like quite neat solution. I poked around a bit with harden-checker and noticed that with static builds the `__stack_chk_fail` symbol seem to be included even when `-fstack-protector` is not used (so it part of glibc?), feels a bit shaky, so your proposal seems safer.

nicowilliams:
> Yeap that feels like quite neat solution. I poked around a bit with harden-checker and noticed that with static builds the `__stack_chk_fail` symbol seem to be included even when `-fstack-protector` is not used (so it part of glibc?), feels a bit shaky, so your proposal seems safer.

With `--run-tests` we could spawn a thread to run a bit of vulnerable code and test it.  Not sure that's a great idea or a terrible idea.  Anyways, I opened #2817 to have jq print it's build configuration -- it captures only `./configure options`, but maybe we could add to it a way to record the versions of all the relevant tools (autotools, compiler, bison, Oniguruma, etc.), and we could always do that later.

PR #2804, opened 2023-08-01
Fix rounding small (but not too small) numbers to zero on calculation (fix #2685)

### LINKED ISSUE #2685: Small numbers are rounded to zero on calculation

itchyny:
The rounding happens on `decNumberReduce` which reduces the number literal to another decNumber with smaller size to fit to the char buffer we prepare. ~Using `decimal64` from the library, we can avoid reducing.~ Adopted @leonid-s-usov's patch. This PR fixes #2685.


Makefile.am
leonid-s-usov, line 1: maybe we should discard changes to the file completely? I believe the only change is that you've added jv.h to LIBJQ_INCS but is it required there?

nicowilliams: Yeah, why this change?

leonid-s-usov: it comes from @itchyny's initial commit. I just removed the new filed from the lists, but haven't reverted the change back

itchyny: Ah, okay. I'll drop this.

nicowilliams:
Thanks!

nicowilliams:
@itchyny this crashes with `./jq -cn '5E480463400+9E-6000000099'`.

PR #2805, opened 2023-08-01
Simplify  String  rule in parser

emanuele6:
Use a `StringStart` component that is either `FORMAT QQSTRING_START` or `QQSTRING_START` instead of having two similar rules for `String`.

This is simpler and avoids having to use an untyped mid-rule action component to copy `FORMAT` at the top of the stack before `QQString`, and having to use `jv_free($<literal>3)` instead of `jv_free($1)` just to make bison not complain about the "unused" mid-rule component.


itchyny:
My first thought of skimming the rules is that including the format prefix in String is an unintended accident. `.@html`, `{@html:.}` are invalid but `.@html""`, `{@html"":.}` are valid. I think better rule is `Term := String | Format | Format String | etc`, which disallows unexpected formatted string usages around the syntax.

emanuele6:
Why should you only be allowed to use a different format for a string literal in an expression, and not when using a string literal in a object constructor as key, or as `."string"`, or as path for `import`/`include`? What is the point of forcing the user to use `"\(@foo "blah")"` instead of just `@foo "blah"` if they want to do that?
In python, and javascript, you can use a `f''` or <code>``</code> string interpolation wherever you can use a string literal.

nicowilliams:
I like this.  It's easier to understand.

nicowilliams:
LGTM.  I really like this simplification.  I'm not seeing what it might break either.

> My first thought of skimming the rules is that including the format prefix in String is an unintended accident. `.@html`, `{@html:.}` are invalid but `.@html""`, `{@html"":.}` are valid. I think better rule is Term := String | Format | Format String | etc, which disallows unexpected formatted string usages around the syntax.

`.@html`, `{@html:.}` remain invalid with this patch, and `.@html""`, `{@html"":.}` remain valid with this patch.  Today I learned that the format specifier is an integral part of the string literal, and... I think it makes sense.

So I think this is ready, unless there's any other objections.  Changing nothing in the language while simplifying the parser is a win.

itchyny (approved):
Okay. Looks good.

nicowilliams:
Thanks!

PR #2806, opened 2023-08-01
Fix overflow on integer casting on modulo operator

itchyny:
I noticed that `pow(2, 63)` is unsafe to cast to int. I was carefully flipped the sign, but failed to notice the boundary bug. Ref: #2797

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -396,7 +396,7 @@ static jv f_divide(jq_state *jq, jv input, jv a, jv b) {
   }
 }
 
-#define dtoi(n) ((n) < INTMAX_MIN ? INTMAX_MIN : -(n) < INTMAX_MIN ? INTMAX_MAX : (intmax_t)(n))
+#define dtoi(n) ((n) < INTMAX_MIN ? INTMAX_MIN : -(n) <= INTMAX_MIN ? INTMAX_MAX : (intmax_t)(n))
 static jv f_mod(jq_state *jq, jv input, jv a, jv b) {
   jv_free(input);
   if (jv_get_kind(a) == JV_KIND_NUMBER && jv_get_kind(b) == JV_KIND_NUMBER) {
--- a/src/parser.y
+++ b/src/parser.y
@@ -235,7 +235,7 @@ static block constant_fold(block a, block b, int op) {
       res = jv_number(na / nb);
       break;
     case '%':
-#define is_unsafe_to_int_cast(n) (isnan(n) || (n) < INTMAX_MIN || -(n) < INTMAX_MIN)
+#define is_unsafe_to_int_cast(n) (isnan(n) || (n) < INTMAX_MIN || -(n) <= INTMAX_MIN)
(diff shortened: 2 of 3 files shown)


nicowilliams:
Did that raise a warning somewhere?

src/builtin.c
```diff
@@ -396,7 +396,7 @@ static jv f_divide(jq_state *jq, jv input, jv a, jv b) {
   }
 }
 
-#define dtoi(n) ((n) < INTMAX_MIN ? INTMAX_MIN : -(n) < INTMAX_MIN ? INTMAX_MAX : (intmax_t)(n))
+#define dtoi(n) ((n) < INTMAX_MIN ? INTMAX_MIN : -(n) <= INTMAX_MIN ? INTMAX_MAX : (intmax_t)(n))
```
nicowilliams, line 399: I thought about this when reviewing the original, but `n` is a double, so I don't think there's a problem here because `INTMAX_MIN` should be promoted to double.

src/parser.y
```diff
@@ -235,7 +235,7 @@ static block constant_fold(block a, block b, int op) {
       res = jv_number(na / nb);
       break;
     case '%':
-#define is_unsafe_to_int_cast(n) (isnan(n) || (n) < INTMAX_MIN || -(n) < INTMAX_MIN)
+#define is_unsafe_to_int_cast(n) (isnan(n) || (n) < INTMAX_MIN || -(n) <= INTMAX_MIN)
```
nicowilliams, line 238: Ditto.

src/builtin.c
itchyny: But `n = 9223372036854775808.0 = pow(2, 63)` exceeds `INTMAX_MAX = 9223372036854775807`, thus unsafe to cast, and `-n < INTMAX_MIN = -9223372036854775808` (exactly in double) is falsy. Assuming `INTX_MAX == -1-INTX_MIN`, casting a double `n` overflows when `n >= INTX_MAX+1` i.e. `n >= -INTX_MIN` i.e. `-n <= INTX_MIN` (the last one is only safe comparison in double).

nicowilliams: But `INTMAX_MAX` here gets promoted to a `double`, so it should be safe.

nicowilliams: Did you get a warning with UBSAN?

itchyny: No, I just noticed while renaming the macro (you were faster though). I originally referred to https://stackoverflow.com/a/60801229/1363488.

nicowilliams: Ok, so we can't represent `INTMAX_MIN` and `INTMAX_MAX` exactly as doubles, but then changing the comparison from `<` to `<=` won't make much difference.  We're relying on the cast from double to `intmax_t` to be safe.

https://stackoverflow.com/questions/51104995/can-a-conversion-from-double-to-int-be-written-in-portable-c

If we want to be extra careful we should probably have global read-only `double` constants that represent the smallest and largest integer that can be represented exactly in `double`, but being careful to allow `double` to not be an IEEE 754 64-bit double.

PR #2814, opened 2023-08-01
Improve README

owenthereal:
Improve README to be more structural and more readable. A reander is available [here](https://github.com/jqlang/jq/blob/26e0e9384c4a67c02daf9deeccd6ebda69c3f5f0/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Build a statically linked version:
 make LDFLAGS=-all-static
 ```
 
-To build from a released tarball (available on the release page), skip the `autoreconf` step, and flex or bison won't be needed.
+If you're not using the latest git version but instead building a released tarball (available on the release page), skip the `autoreconf` step, and flex or bison won't be needed.
 
 ##### Cross-Compilation
 
--- a/README.md
+++ b/README.md
@@ -1,13 +1,8 @@
-jq
-==
+# jq
 
-## Introduction
+`jq` is a lightweight and flexible command-line JSON processor akin to`sed`,`awk`,`grep`, and friends for JSON data. It's written in portable C and has zero runtime dependencies, allowing you to easily slice, filter, map, and transform structured data.
 
-**jq** is a lightweight and flexible command-line JSON processor akin to `sed`, `awk`, `grep`, and friends for JSON data. It's written in portable C and has zero runtime dependencies, allowing you to easily slice, filter, map, and transform structured data.
-
-## Getting Started
-
-### Documentation
+## Documentation
 
 - **Official Documentation**: [jqlang.github.io/jq](https://jqlang.github.io/jq)
 - **Try jq Online**: [jqplay.org](https://jqplay.org)
--- a/README.md
+++ b/README.md
@@ -1,76 +1,65 @@
 jq
 ==
 
-jq is a lightweight and flexible command-line JSON processor.
+## Introduction
 
-If you want to learn to use jq, read the documentation at
-[https://jqlang.github.io/jq](https://jqlang.github.io/jq).  This
-documentation is generated from the docs/ folder of this repository.
-You can also try it online at [jqplay.org](https://jqplay.org).
+**jq** is a lightweight and flexible command-line JSON processor akin to `sed`, `awk`, `grep`, and friends for JSON data. It's written in portable C and has zero runtime dependencies, allowing you to easily slice, filter, map, and transform structured data.
 
-If you want to hack on jq, feel free, but be warned that its internals
-are not well-documented at the moment. Bring a hard hat and a
-shovel.  Also, read the wiki: https://github.com/jqlang/jq/wiki, where
-you will find cookbooks, discussion of advanced topics, internals,
-release engineering, and more.
+## Getting Started
 
-Source tarball and built executable releases can be found on the
-homepage and on the github release page, https://github.com/jqlang/jq/releases.
-Docker image is available at https://github.com/jqlang/jq/pkgs/container/jq.
+### Documentation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,76 +1,65 @@
 jq
 ==
 
-jq is a lightweight and flexible command-line JSON processor.
+# Introduction
```
itchyny:
```suggestion
## Introduction
```
The only top level header is jq

owenthereal: Updated header level with `--amend`

README.md
```diff
@@ -1,76 +1,65 @@
...
 
-    # git clean ...
-    autoreconf -i
-    ./configure
-    make distclean
-    scripts/crosscompile <name-of-build> <configure-options>
+```console
+git submodule update --init # if building from git to get oniguruma
+autoreconf -i               # if building from git
+./configure --with-oniguruma=builtin
```
itchyny, line 34: I always feel `--with-oniguruma=builtin` in instruction is verbose because it detects automatically but do we still need this to prevent the fallback to pre-installed libonig?

README.md
```diff
@@ -1,76 +1,65 @@
 jq
 ==
 
-jq is a lightweight and flexible command-line JSON processor.
+## Introduction
```
nicowilliams: Too many headings?  "Introduction" is not a needed heading given that we already have a `jq` heading at line 1.

nicowilliams: image

See what I mean by too many headings?  It's a bit of a jumble of different size fonts because there's so little text between each heading.

Easy fix: cut out a lot of the headings :)

README.md
```diff
@@ -1,76 +1,65 @@
...
-documentation is generated from the docs/ folder of this repository.
-You can also try it online at [jqplay.org](https://jqplay.org).
+**jq** is a lightweight and flexible command-line JSON processor akin to `sed`, `awk`, `grep`, and friends for JSON data. It's written in portable C and has zero runtime dependencies, allowing you to easily slice, filter, map, and transform structured data.
 
-If you want to hack on jq, feel free, but be warned that its internals
-are not well-documented at the moment. Bring a hard hat and a
-shovel.  Also, read the wiki: https://github.com/jqlang/jq/wiki, where
-you will find cookbooks, discussion of advanced topics, internals,
-release engineering, and more.
+## Getting Started
```
nicowilliams: Remove this heading too.

README.md
```diff
@@ -1,76 +1,65 @@
...
+### Docker Image
 
-After make finishes, you'll be able to use `./jq`.  You can also
-install it using:
+Pull the [jq image](https://github.com/jqlang/jq/pkgs/container/jq) to start quickly with Docker.
 
-    sudo make install
+### Building from source
 
-If you're not using the latest git version but instead building a
```
nicowilliams, line 44: I feel like we're losing a lot of detail -- detail that probably doesn't belong in this README, but it belongs somewhere.

README.md
```diff
@@ -1,76 +1,65 @@
...
-released tarball (available on the website), then you won't need to
-run `autoreconf` (and shouldn't), and you won't need flex or bison.
+#### Dependencies
 
-To cross-compile for OS X and Windows, see docs/Rakefile's build task
```
nicowilliams, line 48: More detail we probably shouldn't lose completely.

README.md
```diff
@@ -1,76 +1,65 @@
...
-    git submodule update --init # if building from git to get oniguruma
-    autoreconf -i               # if building from git
-    emconfigure ./configure --with-oniguruma=builtin
-    emmake make EXEEXT=.js CFLAGS="-O2" LDFLAGS="-s EXPORTED_RUNTIME_METHODS=['callMain']"
+To build from a released tarball (available on the release page), skip the `autoreconf` step, and flex or bison won't be needed.
 
+##### Cross-Compilation
 
-# Community
+For details on cross-compliation, check out the [GitHub Actions file](.github/workflows/ci.yml) and the [cross-compliation wiki page](https://github.com/jqlang/jq/wiki/Cross-compilation).
```
nicowilliams, line 50: Oh, I see, right, you're linking to the wiki pages.

README.md
owenthereal: Removed the unecessary headings :). A new render is [here](https://github.com/jqlang/jq/blob/26e0e9384c4a67c02daf9deeccd6ebda69c3f5f0/README.md)

README.md
owenthereal: Done

README.md
owenthereal: I put back this `if you're not ...` line to avoid confusions, the detail wasn't lost, it was made more concise.

README.md
owenthereal: Yes, the cross-compilation details were moved to the wiki page. I'm trying to make the README more terse and include links for more details.

README.md
owenthereal: Indeed, the `--with-oniguruma=builtin` is in the [default `./configure` flags](https://github.com/jqlang/jq/blob/fd0d475694bdaf5261340aff9a6a4437b04124dc/Makefile.am#L228): . Perhaps let's make the [help](https://github.com/jqlang/jq/blob/fd0d475694bdaf5261340aff9a6a4437b04124dc/configure.ac#L248-L250) more clear `builtin` is the default.

nicowilliams: I guess this is going to be more reliable for users, so it's probably best to leave it.  Users with a slight clue will think about this and maybe exclude this option?

nicowilliams:
LGTM.

README.md
itchyny: Okay, I understand.

itchyny (approved):

PR #2817, opened 2023-08-02
Add --config option printing ./configure options used

### LINKED ISSUE #2816: Add command-line option to show build options used to build jq
It would be nice to have a command-line option by which to show the `./configure` build options used to build jq.  Or perhaps a digested version of that, like if Oniguruma was used from the sub-module, whether stack protection features were enabled (and which), etc.

nicowilliams:
```
./jq --config
'--prefix=/tmp/jq' '--with-oniguruma=builtin' '--enable-maintainer-mode' '--srcdir=..' 'CFLAGS=-pthread -fstack-protector-all'
```

We could add lines recording compiler name and version, flex/bison version, autotools versions, etc.

Closes #2816.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,12 +33,12 @@ ACLOCAL_AMFLAGS = -I config/m4
 # header file creation so we'll use good old make
 if MAINTAINER_MODE
 BUILT_SOURCES = src/lexer.h src/lexer.c src/parser.h src/parser.c 	\
-		src/builtin.inc src/version.h
+		src/builtin.inc src/config_opts.inc src/version.h
 src/lexer.c: src/lexer.l
 	$(AM_V_LEX) flex -o src/lexer.c --header-file=src/lexer.h $<
 src/lexer.h: src/lexer.c
 else
-BUILT_SOURCES = src/builtin.inc src/version.h
+BUILT_SOURCES = src/builtin.inc src/config_opts.inc src/version.h
 .y.c:
 	$(AM_V_YACC) echo "NOT building parser.c!"
 .l.c:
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -580,6 +581,11 @@ int main(int argc, char* argv[]) {
         ret = JQ_OK;
         goto out;
       }
+      if (isoption(argv[i], 'V', "config", &short_opts)) {
```
itchyny: Note that `-V` is for `--version`. I don't think we will normally use this option, so we don't need a short option.

nicowilliams: Oops, yes.

itchyny:
This resolves #2816 but assigned milestones are different.

nicowilliams:
Milestones now line up.

Makefile.am
```diff
@@ -109,14 +109,20 @@ generate_ver = ver="`{ $(srcdir)/scripts/version || echo '$(VERSION)' ; } | sed
 src/version.h: .remake-version-h
 	mkdir -p src
 	$(AM_V_GEN) $(generate_ver); echo "$$ver" > $@
-src/main.c: src/version.h
+src/config_opts.inc:
+	mkdir -p src
+	$(AM_V_GEN) if test -x ./config.status; then \
+	  ./config.status --config; \
+	else echo "(unknown)"; \
+	fi | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g' -e 's/^/"/' -e 's/$$/\\n"/' -e 's/^/#define JQ_CONFIG /' > $@
```
wader: `--build-configuration` prints an extra new line for me, maybe change `'s/$$/\\n"/` to `'s/$$/"/`?

wader (commented):
Except my comment i think it looks good 👍

btw for fq i expose some or these things using a function that return an structure with various information that is then used by the cli code written in jq code, kind of nice. Maybe could be interesting in the future? maybe not cli in jq but expose things as a function or $global

Makefile.am
nicowilliams: The format is uncommitted.  It can have noise :)

nicowilliams:
Now with `$JQ_BUILD_CONFIGURATION`.

wader (approved):

PR #2818, opened 2023-08-03
Fix crash on numeric comparison (ref #2804)

itchyny:
This PR fixes a crash introduced in #2804. Comparison of numbers with huge exponent difference causes crash in `decUnitCompare`.


leonid-s-usov (approved):
Looks good. So it seems this must have been failing also before our rounding fix, cause this setting was copied as is

emanuele6:
Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61050
`5E480463400+9E-6000000099` seg faults during compilation (constant folding).

itchyny:
Thank you.

nicowilliams:
Thank you!

leonid-s-usov:
I was puzzled by this issue and made some additional research when I got to this. It showed two issues
1. My original implementation was buggy. When initializing the default decContext, I set the `emin` to `DEC_MIN_EMAX`, which is 0, instead of `DEC_MIN_EMIN`, which is `-999999999`.
2. We should probably do a better job at parsing the `ctx.status` post-conversion, because in the case of `5E-5000000000` it's an underflow, and a round, and a clamp, and something else - so, a bouquet of error flags that we do nothing about.

The problem is injected into the runtime in `jvp_literal_number_new` at ```  decNumberFromString(&n->num_decimal, literal, ctx);```. This is where the dec context that we supply affects the resulting number, and the issue arizes when the parsed string represents a number of precision that doesn't fit. 

This fix incidentally does well at setting the exp magnitude post conversion from string to 1 less than the one that'd crash. Another change that behaves the same way would be to reduce the `emin` by one (+1 since it's negative), which is also [suggested in the documentation](https://speleotrove.com/decimal/dncont.html): 
> If the adjusted exponent for a result or conversion would be smaller than emin then the result is subnormal. If the result is also inexact, an underflow results. The exponent of the smallest possible number (closest to zero) will be emin–digits+1.[[1]](https://speleotrove.com/decimal/dncont.html#footnote.1)  emin is usually set to –emax or to –(emax–1). 

However, this fix is the same as the emin+1 because the expression that calculates the problematic value is in `decFinalize`:
```
  Int tinyexp=set->emin-dn->digits+1;   // precalculate subnormal boundary
```
one can see that it references both `digits` and `emin`, so the result is equally affected.

My biggest concern, however, is that it seems we have insufficient testing around number conversion, especially when they are over/under flown.

PR #2819, opened 2023-08-03
Constant fold all kinds of constants

emanuele6:
Also use `jv_equal()` to constant fold `==` and `!=` instead of `jv_cmp()` to match the implementation of `_equal` and `_notequal`.


src/parser.y
```diff
@@ -202,63 +203,33 @@ static block constant_fold(block a, block b, int op) {
...
+  case NEQ: res = binop_notequal(jv_a, jv_b); break;
+  case '<': res = binop_less(jv_a, jv_b); break;
+  case '>': res = binop_greater(jv_a, jv_b); break;
+  case LESSEQ: res = binop_lesseq(jv_a, jv_b); break;
+  case GREATEREQ: res = binop_greatereq(jv_a, jv_b); break;
   }
 
-  if (jv_get_kind(res) == JV_KIND_INVALID)
+  if (!jv_is_valid(res)) {
+    jv_free(res);
```
emanuele6: Is there a way to generate a constant error block? That would be cool. =)

emanuele6:
I noticed that string comparison, string division, string-number multiplication, etc were not being constant folded, so I decided to finally export the binary operators' implementations from `builtin.c`, and use those to implement constant folding so now all type combinations are supported and the constant folding behaviour is exactly the same as runtime.

src/parser.y
emanuele6: I added a way to do it!

src/opcode_list.h
```diff
@@ -47,3 +47,5 @@ OP(GENLABEL, NONE, 0, 1)
 
 OP(DESTRUCTURE_ALT, BRANCH, 0, 0)
 OP(STOREVN, VARIABLE, 1, 0)
+
+OP(ERROR, CONSTANT, 1, 0)
```
emanuele6: Is `1, 0` correct here? I think `stack_in == 1` is correct, but the second value (`stack_out`) is not used for anything so I am not sure what it should be.

emanuele6:
I renamed the opcode to `ERRORK`. It seems that on windows, `ERROR` is already used for something else.

emanuele6:
I made a mistake when renaming `ERROR`, now it should work

nicowilliams:
Well done!

nicowilliams:
Thanks!

PR #2823, opened 2023-08-04
Let error(null) throw null

emanuele6:
This patch removes the weird behaviour of `jv_invalid_with_msg(jv_null())` that returns `jv_invalid()` (i.e. empty), instead of a boxed `jv_null()`.

The previous behaviour of `null|error` was obviously unintentional, and allowing `jv_invalid_with_msg()` to return values on which you can't call `jv_invalid_get_msg()` is only error prone.

---

I don't see a reason to keep and document this behaviour, it is obviously a bug, it is not useful, and it was not documented in jq 1.6.

Either way, `jv_invalid_with_msg(jv_null())` should be fixed to not return `jv_invalid()` in my opinion.
The "`null|error` is equivalent to `empty`" behaviour can be implemented in `f_error` if we really want to keep it.

```patch
diff --git a/src/builtin.c b/src/builtin.c
index 0d0ac23..e84bef9 100644
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1139,6 +1139,8 @@ static jv f_nan(jq_state *jq, jv input) {
 }
 
 static jv f_error(jq_state *jq, jv input) {
+  if (jv_get_kind(input) == JV_KIND_NULL)
+    return jv_invalid();
   return jv_invalid_with_msg(input);
 }
 
```

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -1231,10 +1231,6 @@ sections:
           given as the argument. Errors can be caught with try/catch;
           see below.
 
-          When the error value is `null`, it produces nothing and works
-          just like `empty`. So `[null | error]` and `[error(null)]` both
-          emit `[]`.
-
         examples:
           - program: 'try error catch .'
             input: '"error message"'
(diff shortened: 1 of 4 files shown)


nicowilliams:
Thanks!

PR #2827, opened 2023-08-06
Don't wrap negative indices twice (fix #2826)

nicowilliams:
Resolves #2826


tests/jq.test
```diff
@@ -2004,3 +2004,9 @@ walk(1)
 walk(select(IN({}, []) | not))
 {"a":1,"b":[]}
 {"a":1}
+
+# Issue #2826
+.[-5]
```
itchyny, line 2009: We can merge the test at L251 with the following query; `[.[range(-4; 4)]]`.

nicowilliams:
But now:

```
./jq -cn '[0, 1, 2]|path(.[-5])'
[-2]
```

itchyny:
Can we just drop support of index wrapping for the coming release and consider how pick should work on arrays later?

src/execute.c
```diff
@@ -695,18 +695,23 @@ jv jq_next(jq_state *jq) {
...
       }
       jv v = jv_get(t, jv_copy(k));
       if (jv_is_valid(v)) {
-        path_append(jq, k, jv_copy(v));
+        path_append(jq, norm_k, jv_copy(v));
         stack_push(jq, v);
+        jv_free(k);
       } else {
+        jv_free(norm_k);
         jv_free(k);
```
emanuele6: This is calling `jv_free(k)` in both branches; also you can just use `k` for the `jv_get()` at line 708.

nicowilliams:
> Can we just drop support of index wrapping for the coming release and consider how pick should work on arrays later?

The current version of this PR does not break `pick/1`.

Issue #2831, opened 2023-08-08
jq-windows-amd64.exe from 1.7rc1 doesn't work in a clean Windows environment


generalmimon:
**Describe the bug**
The `jq-windows-amd64.exe` executable from the [jq 1.7rc1](https://github.com/jqlang/jq/releases/tag/jq-1.7rc1) release cannot be run on a clean Windows environment from `cmd.exe` (for example) without having to install the MSYS2 toolchain and make it available to the jq binary, it seems. This is unlike previous jq releases, which work on Windows out of the box just fine - I checked `jq-win64.exe` from [jq 1.6](https://github.com/jqlang/jq/releases/tag/jq-1.6) and [jq 1.5](https://github.com/jqlang/jq/releases/tag/jq-1.5).

**To Reproduce**

First of all, hopefully don't have `msys-2.0.dll` in the search path somehow, in which case you won't be able to reproduce the issue. For example if I run `cmd` from within Git for Windows (available from https://git-scm.com/download/win; there's also a "Portable" edition there that doesn't require installation), the `jq-windows-amd64.exe` binary works there because the Git for Windows shell modifies the environment so that the `msys-2.0.dll` becomes available within the session and I can use `where msys-2.0.dll` (which is like `which`, only for `cmd.exe`) to verify that:

   ![image](https://github.com/jqlang/jq/assets/47499687/a10c0521-b2e6-4d51-b789-b2ffd3726165)

---

1. Download https://github.com/jqlang/jq/releases/download/jq-1.7rc1/jq-windows-amd64.exe
2. Open `cmd.exe`, for example by pressing the <kbd>Win</kbd> key to bring up the Start Menu, typing `cmd` and accepting the first suggestion.
3. `cd` to the directory where the `jq-windows-amd64.exe` binary was downloaded.
4. Type `.\jq-windows-amd64 --version` and press <kbd>Enter</kbd>.

See the screenshot below just after following these steps:

![2023-08-08_jq-1 7rc1-msys2-dll-not-found](https://github.com/jqlang/jq/assets/47499687/4ac2eb76-2fb9-402f-858c-51f39a61f0d2)

[...]

PR #2835, opened 2023-08-10
Make sure to init key and dbl_key also on win32

wader:
Hopefully fixes page fault for mingw build

Related to #2831


nicowilliams:
Thanks!

PR #2837, opened 2023-08-12
Expose a module's function names in `modulemeta`

mrwilson:
This PR adds a new field, `defs`, to the output of `modulemeta`.

This field contains the names and arities of functions defined within that module. For example

```jq
module {
  name: "example",
  version: "0.0.0"
};

def foo(): 1;
def bar($baz): $baz + 1;
```

```sh
$ jq -L. --null-input '"example" | modulemeta | .defs[]'

"foo/0"
"bar/1"
```

The intent of this change is to make it easier for tools / processes built on top of `jq` to introspect about function definitions within the modules.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -3633,7 +3633,8 @@ sections:
 
           Takes a module name as input and outputs the module's metadata
           as an object, with the module's imports (including metadata)
-          as an array value for the `deps` key.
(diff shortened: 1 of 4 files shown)


nicowilliams:
Do we want to also list the module's `$bindings`, if any?

nicowilliams:
Thanks!

PR #2838, opened 2023-08-12
Add builtin function `compact`

owenthereal:
`compact` removes nulls from an object.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -1119,7 +1119,7 @@ sections:
       - title: "`compact`"
         body: |
 
-          The builtin function `compact` removes nulls from an object.
+          The builtin function `compact` removes nulls from an object or an array recursively.
 
         examples:
           - program: "compact"
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -1116,6 +1116,19 @@ sections:
             input: '["foo", "bar", "baz"]'
             output: ['["foo"]']
 
+      - title: "`compact`"
+        body: |
+
+          The builtin function `compact` removes nulls from an object.
+
+        examples:
+          - program: "compact"
+            input: '{"foo": null, "bar": "baz"}'
+            output: ['{"bar": "baz"}']
+          - program: "compact"
+            input: '["foo", "bar", null]'
+            output: ['["foo", "bar"]']
+
       - title: "`getpath(PATHS)`"
         body: |
 
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -256,7 +256,7 @@ def walk(f):
   w;
 
 # compact removes nulls from an array or an object
-def compact: del(..|nulls);
+def compact: del(.. | nulls);
 
 # pathexps could be a stream of dot-paths
 def pick(pathexps):
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -255,6 +255,9 @@ def walk(f):
     | f;
   w;
 
+# compact removes nulls from an array or an object
+def compact: del(..|nulls);
+
 # pathexps could be a stream of dot-paths
 def pick(pathexps):
   . as $in
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -1124,6 +1124,25 @@ jq \'del(\.[1, 2])\'
 .
 .IP "" 0
 .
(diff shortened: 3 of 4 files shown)


docs/content/manual/manual.yml
```diff
@@ -1109,6 +1109,19 @@ sections:
             input: '["foo", "bar", "baz"]'
             output: ['["foo"]']
 
+      - title: "`compact`"
+        body: |
+
+          The builtin function `compact` removes nulls from an object.
```
wader: from object and array? maybe mention that it is recursive?

src/builtin.jq
```diff
@@ -255,6 +255,9 @@ def walk(f):
     | f;
   w;
 
+# compact removes nulls from an array or an object
+def compact: del(..|nulls);
```
wader: style seem to be whitespace around operators

wader:
Implementation LGTM but i guess we have to decide if it goes into 1.7 or not? as i understood around the discussion about `pick` we wanted a compact but for arrays only but i could have misunderstood?

nicowilliams:
You have to also include a regen of `jq.1.prebuilt` in order for the checks to pass.

docs/content/manual/manual.yml
owenthereal: Addressed in https://github.com/jqlang/jq/pull/2838/commits/7052fa1d867aceacac529248c4165be24823361d

src/builtin.jq
owenthereal: Addressed in https://github.com/jqlang/jq/pull/2838/commits/58a5d09dbdd2b89a1eeb15468c48b32f2f5f4095

owenthereal:
@wader:

> Implementation LGTM but i guess we have to decide if it goes into 1.7 or not? as i understood around the discussion about pick we wanted a compact but for arrays only but i could have misunderstood?

I personally think that it's fine to go into 1.7 unless we think 1.7 should have zero new feature. `compact` could be the building blocks for other stuff as well besides improving the existing `pick` implementation. 

@nicowilliams:

> You have to also include a regen of jq.1.prebuilt in order for the checks to pass.

Fixed in https://github.com/jqlang/jq/pull/2838/commits/04a0dc5cf662afc5a88c6f88dbd1dcbacf70befc

This PR is ready for review again 😄

pkoppstein:
@owenthereal wrote:
> compact could be the building blocks for other stuff as well besides improving the existing pick implementation.

I am not sure whether the above refers to the behavior of `pick` on arrays or objects, but in either case, I would like to point out that the way `pick` inserts nulls in both cases is both intended and (at least in part) documented.  

Consider in particular the fact that `pick` will add "null" values to object inputs, e.g. `{} | pick(.x)` produces `{"x": null}`.  

This behavior fits in properly with the jq scheme of things, as evidenced for example by the way it flows from the simplicity of the def of `pick`.  In this connection, it should also be remembered that, at least from the point of view of jq's `.x` selector, the absence of a key named "x" is equivalent to a null-valued key named "x".  Thus `pick(.x)` can be understood as an imperative: evaluate `.x` and ensure the result is reflected in the output. 

On the other hand, it might make sense for `keep/1` (as being developed by @nicowilliams) to avoid inserting nulls into arrays and objects.  Should `{} | keep(.x)` raise an exception?

pkoppstein:
@wader wrote:
> we have to decide if it goes into 1.7 or not?  

My thinking is that it shouldn't, partly because we're already at jq 1.7rc2, but mainly because there hasn't been much discussion about it.  Should we have guidelines as to when new trivial builtins should be added?  In any case, it is very specific: in my experience, different users typically want to delete various combinations of distinguished values besides `null`, notably `[]`, `""`, and/or other strings signifying the absence of a value.

emanuele6:
Personally, I think just `del(.. | nulls)` is easier to understand; do we want to merge this alias anyway?

nicowilliams:
> Personally, I think just del(.. | nulls) is easier to understand; do we want to merge this alias anyway?

I tend to agree.  I'm not sure I'd know what `compact` would do without first looking at the docs.  I mean, that's often the case with jq builtins, but maybe we should try not to add more of these?  The changes look ok besides that.

itchyny:
Based on the comments from other maintainers, unlikely to be accepted as a builtin, closing.

PR #2842, opened 2023-08-13
Cast function pointers without prototype before calling them

emanuele6:
clang complained that this is deprecated in all versions of standard C, and unsupported in C2x.

---
Ref: https://github.com/jqlang/jq/pull/2800#issuecomment-1676320945


itchyny (approved):
LGTM!

PR #2843, opened 2023-08-13
Fixed `tm::is_dst` overriding `my_mktime`s efforts to be UTC

### LINKED ISSUE #2490: fromdate produces an Epoch timestamp that is one hour off
**Describe the bug**

Parsing an ISO8601 timestamp yields a timestamp that is off by 3600 seconds.

**To Reproduce**

```
echo 1665511200 | jq todate | jq fromdate
1665514800
```

**Expected behavior**

Expected the round trip from `todate` (which produces an ISO8601) to `fromdate` would yield the same result.

 - OS and Version: macOS Monterey 12.6, also reproduced on Linux (Fedora 34)
 - jq version 1.6

**Additional context**

🤷

MCLooyverse:
This (likely) fixes #2490 by insuring the broken-down time given to `my_mktime` is never interpreted in DST.

I'm not certain of the consequences of mutating `tm` right now.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1355,7 +1355,7 @@ static time_t my_mktime(struct tm *tm) {
   tz = (tz = getenv("TZ")) != NULL ? strdup(tz) : NULL;
   if (tz != NULL)
     setenv("TZ", "", 1);
-	tm->is_dst = 0;
+  tm->is_dst = 0;
   time_t t = mktime(tm);
   if (tz != NULL)
     setenv("TZ", tz, 1);
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1355,6 +1355,7 @@ static time_t my_mktime(struct tm *tm) {
   tz = (tz = getenv("TZ")) != NULL ? strdup(tz) : NULL;
   if (tz != NULL)
     setenv("TZ", "", 1);
+	tm->is_dst = 0;
   time_t t = mktime(tm);
   if (tz != NULL)
     setenv("TZ", tz, 1);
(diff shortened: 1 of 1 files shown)


src/builtin.c
```diff
@@ -1355,6 +1355,7 @@ static time_t my_mktime(struct tm *tm) {
   tz = (tz = getenv("TZ")) != NULL ? strdup(tz) : NULL;
   if (tz != NULL)
     setenv("TZ", "", 1);
+  tm->is_dst = 0;
```
nicowilliams, line 1358: Ah, so we're not fully initializing `struct tm`s...  Do we also need to set `tm_zone`?

nicowilliams: Would it not be better to initialize this in the callers?

Should we add `is_dst` to the jq broken-down time representation?

MCLooyverse: False alarm!  The error I'm fixing is *already* fixed on origin/master, just not in the v1.6 release I had.

> Ah, so we're not fully initializing `struct tm`s... Do we also need to set `tm_zone`?

Actually, in `jv2tm`, a comment explicitly mentions that `tm->tm_isdst` gets zeroed by a `memset`.  I don't know why my code compiles, since my `is_dst` *should* be a typo for `tm_isdst`.

nicowilliams: So should we close this then?  There's nothing left to do?

MCLooyverse: Yes.  Also, the reason my code doesn't fail is because that last-resort path of `my_mktime` isn't even compiled on my machine.  Somehow, when I did `gcc -E` to discover what my machine uses, I got the wrong answer ¯\_(ツ)_/¯.

PR #2845, opened 2023-08-14
Better checking for `$v` not being a path expression (Fix #2844)

### LINKED ISSUE #2844: `jq -cn '(.[3]|debug) as $a | $a = 5'` doesn't report an error, but `$a` is not a path expression
Reported by @ncfavier on IRC.

We should consider having `LOADV`/`LOADVN` raise an error when `jq->subexp_nest != 0`.  Except that `jq->subexp_nest == 0` is not sufficient indication that the current instruction is executing inside `path/1`, but we can change the initial value of `jq->subexp_nest` to be `1` instead of `0`, that way it's only ever `0` when running inside `path/1`.

nicowilliams:
This PR makes `$binding`s an error when in a path expression running inside `path/1`.

E.g., `jq -cn '. as $a | $a'` outputs `null` as expected, while `jq -cn '. as $a | $a = 5'` now raises an error (it used to output `5`, though the `$a` binding wasn't actually updated, as `jq -cn '. as $a | $a = 5 | $a'` used to output `null`).

### DIFF
--- a/src/bytecode.c
+++ b/src/bytecode.c
@@ -4,6 +4,7 @@
 
 #include "bytecode.h"
 #include "jv_alloc.h"
+#include "locfile.h"
 
 // flags, length
 #define NONE 0, 1
@@ -92,6 +93,20 @@ static struct bytecode* getlevel(struct bytecode* bc, int level) {
   return bc;
 }
 
+jv get_location(struct bytecode *bc, uint16_t *codeptr) {
+  char n[20];
+  int i = codeptr - bc->code;
+
+  do {
+    if (bc->location_offsets[i] != -1)
+      break;
+  } while (i--);
+  if (i < 0 || bc->location_offsets[i] == -1)
+    return jv_copy(bc->fname);
+  (void) snprintf(n, sizeof(n), "%d", bc->location_offsets[i]);
+  return JV_STRING(jv_copy(bc->fname), jv_string(":"), jv_string(n));
+}
+
 void dump_operation(struct bytecode* bc, uint16_t* codeptr) {
   int pc = codeptr - bc->code;
   printf("%04d ", pc);
@@ -146,6 +161,13 @@ void dump_operation(struct bytecode* bc, uint16_t* codeptr) {
   }
 }
 
+struct bytecode *bytecode_alloc(struct locfile *lf) {
+  struct bytecode* bc = jv_mem_calloc(1, sizeof(struct bytecode));
+  bc->fname = jv_copy(lf->fname);
+  bc->program = jv_string_sized(lf->data, lf->length);
+  return bc;
+}
+
 void bytecode_free(struct bytecode* bc) {
   if (!bc)
     return;
@@ -157,5 +179,8 @@ void bytecode_free(struct bytecode* bc) {
     symbol_table_free(bc->globals);
   jv_mem_free(bc->subfunctions);
   jv_free(bc->debuginfo);
+  jv_free(bc->fname);
+  jv_free(bc->program);
+  jv_mem_free(bc->location_offsets);
   jv_mem_free(bc);
 }
--- a/src/bytecode.c
+++ b/src/bytecode.c
@@ -105,7 +105,8 @@ jv get_location(struct bytecode *bc, uint16_t *codeptr) {
   if (offset == -1)
     return jv_copy(bc->fname);
   const char *prog = jv_string_value(bc->program);
-  for (i = 0; i < offset; i++) {
+  int len = jv_string_length_bytes(jv_copy(bc->program));
+  for (i = 0; i < offset && i < len; i++) {
     if (prog[i] == '\n') {
       lineno++;
       column = 0;
--- a/src/bytecode.c
+++ b/src/bytecode.c
@@ -94,17 +94,27 @@ static struct bytecode* getlevel(struct bytecode* bc, int level) {
 }
 
 jv get_location(struct bytecode *bc, uint16_t *codeptr) {
-  char n[20];
-  int i = codeptr - bc->code;
+  char buf[41];
+  int32_t offset = -1;
+  int32_t i = codeptr - bc->code;
+  int lineno = 0, column = 0;
 
   do {
-    if (bc->location_offsets[i] != -1)
-      break;
-  } while (i--);
-  if (i < 0 || bc->location_offsets[i] == -1)
+    offset = bc->location_offsets[i];
+  } while (offset == -1 && i--);
+  if (offset == -1)
     return jv_copy(bc->fname);
-  (void) snprintf(n, sizeof(n), "%d", bc->location_offsets[i]);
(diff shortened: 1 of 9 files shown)


nicowilliams:
This may be leaving us with dead code in `PATH_END`, `INDEX`/`INDEX_OPT`, and `EACH`/`EACH_OPT` where we do a heuristic check of whether we're producing an invalid path expression.  Perhaps `path_intact()` is dead code now.  We should think about what other instructions make an expression not-a-path-expression, such as:

 - `APPEND`
 - `INSERT`
 - `MODULEMETA` (EDIT: no, this one is a pseudo-instruction not used at run-time)
 - `DEPS` (EDIT: no, this one is a pseudo-instruction not used at run-time)
 - etc.

As it happens we now get:

```
: ; ./jq -c '[range(.)] = .' <<< 1
jq: error (at <stdin>:1): Invalid path expression ($binding is not a path expression)
```

which is... correct but confusing because there is no `$varname` in this example.

Compare to 1.6:

```
jq -c '[range(.)] = .' <<< 1
jq: error (at <stdin>:1): Invalid path expression with result [0]
```

Neither case yields a terribly useful error message.

Note that we've lost the names of `$binding`s at run-time, so we can't make the error very nice.  Ay.

Another thing we've lost is all source code location information.  It'd be nice to retain it, at least in some debug mode, then we could say `Invalid path expression at <filename>:<line-number>:<column>`.  Basically we should add something to `struct bytecode` that we can use to map instruction addresses to source code locations, maybe: a) the name of the source file, b) an array of `codelen` line and column numbers.

At the very least the error message should be a bit more useful, and maybe the manual should make it easier to find the description of what is a path expression by having an example that yields this error message.

nicowilliams:
And there's the valgrind issues to fix.

nicowilliams:
Now with a WIP commit to add run-time program location information!  This is incomplete because currently `jq_compile_args()` doesn't get the name of the file containing the program to compile(!), so the filename for the main program is always reported as `<top-level>`, which... is not very user-friendly.

Still, this is going to make error reporting much much nicer.  It will also make it easier to write a source debugger mode for jq at some point.

nicowilliams:
And now we have:

```
: ; cat /tmp/a.jq
. as $a | $a = 5
: ; ./jq -cnf /tmp/a.jq
jq: error (at <unknown>): Invalid path expression ($binding is not a path expression) at /tmp/a.jq:0
```

So that's a bit confusing.  We want to have two filenames and locations: one for the _program_ and one for the _input_, so this needs to be crafted a bit better.

Also, the number associated with the source code filename is currently an _offset_ -- we should have `get_location()` report a line number and column number instead!

nicowilliams:
I think we need to add more `gen_location()` calls in `src/parser.y` in order to have access to better location information at runtime to go with the changes here.

nicowilliams:
```
: ; cat /tmp/a.jq
. |
    {(.): (.)} = 0
: ; ./jq -cnf /tmp/a.jq
jq: error (input at <unknown>): Invalid path expression (objects are not path expressions) at /tmp/a.jq:1:4
```

That's starting to look better.

src/bytecode.c
```diff
@@ -92,6 +93,31 @@ static struct bytecode* getlevel(struct bytecode* bc, int level) {
...
+  for (i = 0; i < offset && i < len; i++) {
+    if (prog[i] == '\n') {
+      lineno++;
+      column = 0;
+    } else {
+      column++;
+    }
+  }
+  (void) snprintf(buf, sizeof(buf), "%d:%d", lineno, column);
+  return JV_STRING(jv_copy(bc->fname), jv_string(":"), jv_string(buf));
```
emanuele6, line 118: Use `jv_string_fmt` so we don't have to care about possible overflows here

src/compile.c
```diff
@@ -1310,13 +1310,15 @@ static int compile(struct bytecode* bc, block b, struct locfile* lf, jv args, jv
   }
   uint16_t* code = jv_mem_calloc(sizeof(uint16_t), bc->codelen);
   bc->code = code;
+  bc->location_offsets = jv_mem_calloc(sizeof(uint32_t), bc->codelen);
```
emanuele6, line 1313: calloc arguments are backwards

src/execute.c
```diff
@@ -330,10 +344,14 @@ void jq_report_error(jq_state *jq, jv value) {
   jq->err_cb(jq->err_cb_data, value);
 }
 
-static void set_error(jq_state *jq, jv value) {
+static void set_error(jq_state *jq, uint16_t *pc, jv value) {
   // Record so try/catch can find it.
   jv_free(jq->error);
-  jq->error = value;
+  char errbuf[15];
+  value = jv_invalid_get_msg(value);
```
emanuele6, line 351: `assert(!jv_is_valid(value))`?

src/execute.c
```diff
@@ -330,10 +344,14 @@ void jq_report_error(jq_state *jq, jv value) {
...
+  if (jv_get_kind(value) != JV_KIND_STRING)
+    value = jv_string(jv_dump_string_trunc(value, errbuf, sizeof(errbuf)));
+  jq->error = jv_invalid_with_msg(JV_STRING(value, jv_string(" at "), get_location(frame_current(jq)->bc, pc)));
```
emanuele6, line 354: use `jv_string_fmt`

src/execute.c
```diff
@@ -522,9 +564,19 @@ jv jq_next(jq_state *jq) {
         jv_free(max);
         goto do_backtrack;
       } 
+#if 0
```
emanuele6, line 567: adding commented out code?

src/jv.h
```diff
@@ -131,6 +131,32 @@ jv jv_string_split(jv j, jv sep);
 jv jv_string_explode(jv j);
 jv jv_string_implode(jv j);
 
+#define JV_STRING_1(e) (jv_string_concat(jv_string(""),e))
```
emanuele6, line 134: unnecessary, `jv_string_fmt` is cleaner

emanuele6 (changes requested):

src/compile.c
nicowilliams: Heheh, well, they are multiplied, and multiplication is commutative, but sure.  (Notice I was keeping the style in the same file, specifically the `jv_mem_calloc()` call two lines above.)

src/execute.c
nicowilliams: We can check this by code inspection and save a branch in debug mode.

src/execute.c
nicowilliams: Oops

src/compile.c
emanuele6: https://github.com/jqlang/jq/pull/2978 https://github.com/jqlang/jq/commit/527c5abab17c1f8f717e00c3938e0361b4f33ae6

nicowilliams:
I need to get back to this...

PR #2849, opened 2023-08-16
Fix overflow on numeric comparison

itchyny:
Although #2839 fixed the overflow of exponent subtraction, there still is possibility of overflow in the `D2U` macro. This patch fixes the overflow in the `D2U` macro, and also truncates the maximum digits to `DEC_MAX_DIGITS` as suggested by https://github.com/jqlang/jq/pull/2839#discussion_r1292774969. @nicowilliams reported the boundary test fails on some machine (why didn't CI catch this?).


nicowilliams:
> why didn't CI catch this?

FYI this is x64:

```
: ; file jq
jq: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=ea19a42df34bbd45de6d686cdde59b5fe573570f, for GNU/Linux 3.2.0, with debug_info, not stripped
```

leonid-s-usov (approved):

itchyny:
Thank you!

PR #2854, opened 2023-08-21
 Add continuation character in string literals 

axiles:
This allows to have long string literals without having to introduce
newline characters inside it.

Example:

"Hello\ 
 World"

(space before `World`)

is the same as `"Hello World"`.

Also some tests were added.

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -463,6 +463,7 @@ static pfunc found_string(struct jv_parser* p) {
       case 't': *out++ = '\t'; break;
       case 'n': *out++ = '\n'; break;
       case 'r': *out++ = '\r'; break;
+      case '\n': break;
 
       case 'u':
         /* ahh, the complicated case */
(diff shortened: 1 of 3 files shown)


nicowilliams:
This is not standard JSON.  I think we want to have a discussion about JSON extensions in general.

As for the jq language, we already have a way to continue strings: just add them with `+`:

```jq
"this string is " +
"continued"
```

nicowilliams:
See #2855.

axiles:
I understand that there could be a discussion regarding the jq language. However, I have just realized that this PR, in its current state, also allows this feature in the input which was NOT my intention.

Therefore, before any further discussion I need to:
- modify this PR such that this feature (backslash + newline)  is accepted in jq but not in the input (like the comments or the newline character in the literal)
- add some tests to check that this such quotes are indeed rejected when present in the input.

For these reasons, I have converted this PR to draft.

emanuele6:
This PR is stale; and marked as a draft.
Also `jq` constant folds `"foo" + "bar"` to `"foobar"` when compiling the `jq` code, so you can just use `+`.
Closing...

PR #2857, opened 2023-08-24
builtin.jq: Add `assert/2`

mrwilson:
This PR adds a new builtin called `assert(test; $message)`.

The `assert(test; $message)` function applies the test to the input entity.
If the test passes (returns true) then the output is the original input.
If the test fails then JQ throws an error containing the message passed in.
          
This can be used to short-circuit pipelines in scripts such as when you don't want to proceed if there's a missing key, or the input is an empty array.

The motivation for this was when I was writing some scripts around some complex object structures and I wanted to break out of the script entirely if specific fields were missing. Rather than test for this in a wrapper shell script, I wanted to generalise the approach to "apply a test, continue if test passes, error if it doesn't"

Normal path

```jq
// Input
[{a: 4, b: 2}, {a: 7, b: 6}] | assert(map(.b) | add > 5; "your message here")

// Output
[
  {
    "a": 4,
    "b": 2
  },
  {
    "a": 7,
    "b": 6
  }
]
```

Error path

```jq
// Input
[{a: 4, b: 2}, {a: 7, b: 6}] | assert(map(.b) | add > 10; "your message here")

// Output
jq: error (at XXX): assertion failed: your message here
```

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -264,6 +264,11 @@ def pick(pathexps):
 # ensure the output of debug(m1,m2) is kept together:
 def debug(msgs): (msgs | debug | empty), .;
 
+# assert property with message
+# success passes the original input through
+# failure raises an error with the original error message
+def assert(test; $message): if test then . else error("assertion error: \($message)") end;
+
 # SQL-ish operators here:
 def INDEX(stream; idx_expr):
   reduce stream as $row ({}; .[$row|idx_expr|tostring] = $row);
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -1974,6 +1974,29 @@ sections:
             output:
               - '[{"a":{"b":2}}]'
 
+      - title: "`assert(test; $message)`"
+        body: |
+          
+          The `assert(test; $message)` function applies the test to the
+          input entity. If the test passes (returns true) then the output
+          is the original input. If the test fails then JQ throws an error
+          containing the message passed in.
+          
+          This can be used to short-circuit pipelines in scripts such as when
+          you want to stop if there is a missing key, or the input is an empty
+          array.
+
+        examples:
+          - program: 'assert(length > 0; "must not be empty")'
+            input: '[1,2,3,4]'
+            output:
+              - '[1,2,3,4]'
+
+          - program: 'try assert(values | add >= 10 ; "values must total at least 10") catch .'
+            input: '{"a": 1, "b": 2 }'
+            output:
+              - '"assertion error: values must total at least 10"'
+
       - title: "`$JQ_BUILD_CONFIGURATION`"
         body: |
 
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -2152,6 +2152,28 @@ jq \'walk( if type == "object" then with_entries( \.key |= sub( "^_+"; "") ) els
 .
 .IP "" 0
 .
+.SS "assert(test; $message)"
+The \fBassert(test; $message)\fR function applies the test to the input entity\. If the test passes (returns true) then the output is the original input\. If the test fails then JQ throws an error containing the message passed in\.
+.
+.P
+This can be used to short\-circuit pipelines in scripts such as when you want to stop if there is a missing key, or the input is an empty array\.
+.
+.IP "" 4
+.
+.nf
+
+jq \'assert(length > 0; "must not be empty")\'
+   [1,2,3,4]
+=> [1,2,3,4]
+
+jq \'try assert(values | add >= 10 ; "values must total at least 10") catch \.\'
+   {"a": 1, "b": 2 }
+=> "assertion error: values must total at least 10"
+.
+.fi
+.
+.IP "" 0
+.
 .SS "$JQ_BUILD_CONFIGURATION"
 This builtin binding shows the jq executable\'s build configuration\. Its value has no particular format, but it can be expected to be at least the \fB\./configure\fR command\-line arguments, and may be enriched in the future to include the version strings for the build tooling used\.
 .
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -2024,3 +2024,11 @@ walk(1)
 walk(select(IN({}, []) | not))
 {"a":1,"b":[]}
 {"a":1}
+
+assert(length > 0; "empty array")
+[1]
+[1]
+
+try assert(length == 0; "non-empty array") catch .
+[1]
+"assertion error: non-empty array"
--- a/tests/man.test
+++ b/tests/man.test
@@ -658,6 +658,14 @@ walk(if type == "array" then sort else . end)
 [[4, 1, 7], [8, 5, 2], [3, 6, 9]]
 [[1,4,7],[2,5,8],[3,6,9]]
 
+assert(length > 0; "must not be empty")
+[1,2,3,4]
+[1,2,3,4]
+
+try assert(values | add >= 10 ; "values must total at least 10") catch .
+{"a": 1, "b": 2 }
+"assertion error: values must total at least 10"
+
 $ENV.PAGER
 null
 "less"


wader:
This would be really useful if it was possible to have a assert/1 that included the expression and input in the message. But that would require some special handling i guess like compiler rewrite assert/1 to assert/2 with a generated message or some magic

mrwilson:
> This would be really useful if it was possible to have a assert/1 that included the expression and input in the message. But that would require some special handling i guess like compiler rewrite assert/1 to assert/2 with a generated message or some magic

That would definitely be a useful capability, but beyond my current understanding of the codebase I think. 

Is there a way within the API to string-ify the operations in a particular block to get the actual expression rather than the evaluated expression?

If there was, then I could see `assert/1` being delegated to `assert/2` with a NULL second parameter, and during the compilation step synthesising a string value from the expression in the first parameter.

nicowilliams:
Making a magical `assert/1`/`assert/2` that stringifies the expression is a great idea.  Maybe we'd want a magical function which works at the compiler level and which, given a lambda, chases its `block` representation to find the body and its start and end `location`s, then have it yield a `LOADK` that has the string of that, then `assert`-like functions  can be jq-coded and call this magical function to get a string representation of the lambda.

wader:
Yeap that sounds like a nice implementation that also could be useful for other things. Would it also make sense to include the input? possibly in some truncated form.

So something like:
```jq
def assert(f): assert(f; "\(_block_source_string(f)) failed with input \(.)")
```

nicowilliams:
> Yeap that sounds like a nice implementation that also could be useful for other things. Would it also make sense to include the input? possibly in some truncated form.

No magic is needed for that.  `def assert(prog; $msg): . as $input | try prog catch ... | ...;`.

mrwilson:
I'm happy to try and put a change together for what's been discussed in the thread, if you're able to give me a few pointers where to start.

Is this something that would be largely implemented in the compiler code, the execute code, the parser, or a combination of all these things?

nicowilliams:
> I'm happy to try and put a change together for what's been discussed in the thread, if you're able to give me a few pointers where to start.

That's very ambitious.  Good!  As it happens I just started looking so I can give you a few pointers below.

You'll want to make use of the jq internals pages linked from the sidebar of the [jq wiki](https://github.com/jqlang/jq/wiki).

> Is this something that would be largely implemented in the compiler code, the execute code, the parser, or a combination of all these things?

I think a combination of more `gen_location()` calls in `src/parser.y` so that we have more location information in the `block` representation of parsed and linked programs, a new opcode, a byte-coded jq function in `src/builtin.c`, and code in `compile()` in `src/compile.c` that looks for calls to functions that have only one instruction being that opcode and replaces those calls with a `LOADK` with an `imm.constant` that has the program from the start referenced by the the `first` instruction of the `CLOSURE_CREATE` to the `last`.

However, there seems to be a difficulty with this approach, which is that by the time we find the `CALL_JQ` instructions to replace with `LOADK`s, the closure we're trying to get the program for lacks the information we need.

Let's look at what happens when we try to compile a program like `def f(fe): fe; def g(ge): f(ge); g("a")`.  Let's imagine that `f/1` is our magic function, so we want to find calls to it and replace them with a constant that is the body of its argument closure.  But which closure body should we capture? `ge` or `"a"`?  If `ge` then this may not be magical enough -- maybe we want to say that when the body of the closure is just a call to another closure then we want to chase that all the way to the first place where it's not just a call to another closure.  We could resolve this question by having two magical functions, one that would capture `ge` and one that would capture `"a"` in this example.

Anyways, here's a `gdb` session that shows what I've in mind, and the difficulty that arises.  A couple of things you should know: first, this is happening during the compilation of `block` jq program representation to bytecode, and this is _after_ parsing _and_ linking.

```
: ; gdb --args ./jq --debug-dump-disasm --debug-trace -cn 'def f(fe): fe; def g(ge): f(ge); g("a")'
...
(gdb) b compile if b.first.op == CALL_JQ &&  b.first.arglist.first.op == CLOSURE_CREATE
...
(gdb) r
Starting program: /home/nico/ws/jq/build/jq --debug-dump-disasm --debug-trace -cn 'def f(fe): fe; def g(ge): f(ge); g("a")'

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Error in testing breakpoint condition:
Cannot access memory at address 0x10

Breakpoint 3, compile (bc=0x555555703d50, b=..., lf=0x55555565f3a0, args=..., env=0x7fffffffe610) at ../src/compile.c:1237
1237      int errors = 0;
(gdb)
(gdb) p b.first[0]
$24 = {next = 0x0, prev = 0x0, op = CALL_JQ, imm = {intval = 0, target = 0x0, constant = {kind_flags = 0 '\000', pad_ = 0 '\000', offset = 0, size = 0, u = {
        ptr = 0x0, number = 0}}, cfunc = 0x0}, locfile = 0x55555565f3a0, source = {start = 11, end = 13}, bound_by = 0x55555565f680,
  symbol = 0x55555565f800 "fe", any_unbound = 0, referenced = 0, nformals = -1, nactuals = 0, subfn = {first = 0x0, last = 0x0}, arglist = {first = 0x0,
    last = 0x0}, compiled = 0x0, bytecode_pos = -1}
(gdb)
(gdb) # We do have location information here for the body of `f/1`:
(gdb) p b.first.locfile[0]
$28 = {fname = {kind_flags = 133 '\205', pad_ = 0 '\000', offset = 0, size = 0, u = {ptr = 0x55555565f3f0, number = 4.6355705923692457e-310}},
  data = 0x55555565f420 "def f(fe): fe; def g(ge): f(ge); g(\"a\")", length = 39, linemap = 0x55555565f450, nlines = 1, error = 0x0, jq = 0x55555565d550,
  refct = 5}
(gdb)
(gdb) # So we're compiling the body of `f/1` here.  One can tell
(gdb) # because we're `CALL_JQ`ing `b.first.symbol` is `"fe"`,
(gdb) # which in the jq program we're compiling happens only
(gdb) # in the body of `f/1`.
(gdb) #
(gdb) # Notice that we don't have anything interesting here:
(gdb) #  - no subfn
(gdb) #  - no arglist
(gdb) #
(gdb) # We do have a bound_by, so let's look:
(gdb) p b.first.bound_by
$25 = (struct inst *) 0x55555565f680
(gdb) p b.first.bound_by[0]
$26 = {next = 0x0, prev = 0x0, op = CLOSURE_PARAM, imm = {intval = 0, target = 0x0, constant = {kind_flags = 0 '\000', pad_ = 0 '\000', offset = 0,
      size = 0, u = {ptr = 0x0, number = 0}}, cfunc = 0x0}, locfile = 0x0, source = {start = -1, end = -1}, bound_by = 0x55555565f680,
  symbol = 0x55555565f730 "fe", any_unbound = 0, referenced = 1, nformals = 0, nactuals = -1, subfn = {first = 0x0, last = 0x0}, arglist = {first = 0x0,
    last = 0x0}, compiled = 0x555555703d50, bytecode_pos = -1}
(gdb) # And notice that here too we have nothing interesting because
(gdb) # it's already been compiled.
(gdb) # We do... have the locfile, though it's not filled in, but
(gdb) # that might not be enough because we might need to chase this
(gdb) # closure to find its actual body.
(gdb)
```

So some work has to be done to make it possible to get to the body of the closure.

pkoppstein:
An important feature of many assertion facilities that is currently
missing from the current proposal is the ability to turn assertion checking
on or off.  The ability to turn them off is important, for example
because (in the words of Tony Hoare) it facilitates the use of
assertions "for program documentation, to assist later developers to
evolve the product to meet new market needs"
(http://www.cs.cornell.edu/courses/CS2110/2014sp/assignments/a0/assertearlyassertoften.pdf)

jq provides two convenient mechanisms for such toggling:

1) global $-variables (set on the command-line)
2) access to environment variables.

An assertion checking module (assert.jq) that uses both techniques is available at
https://gist.github.com/pkoppstein/9b0f49a213a5b8083bab094ede06652d

This module also allows one to choose between `error`, `stderr`, and `debug`
for the handling of assertion violations.  The use of `stderr` or `debug`
allows program execution to continue after an assertion violation has
been detected.  The form `assert(PREDICATE; $__loc__)` is supported for
pinpointing the occurrence of an assertion violation.

Please also note that any enhancements of jq that support retrieving a string
representation of a jq expression could be easily incorporated into `assert.jq`.

With one very small change, the module can also be used with gojq, the Go implementation of jq.

nicowilliams:
> An important feature of many assertion facilities that is currently
> missing from the current proposal is the ability to turn assertion checking
> on or off.

As it happens I recently wrote this for debugging:

```jq
    # Enable debug output with jq --argjson debug "$level" ...
    def debug1($msg): ($msg | select($ARGS.named.debug > 0) | debug | empty), .;
    def debug2($msg): ($msg | select($ARGS.named.debug > 1) | debug | empty), .;
    def debug3($msg): ($msg | select($ARGS.named.debug > 2) | debug | empty), .;
```

and I'm tempted to have this as builtins _and_ to have a `-d` command-line option that can be used instead of having to use `--argjson debug "$level"`.

I used this for debugging my `keep/1` inspired by your `pick/1`.

Looking at your gist I see you about the same idea.

nicowilliams:
At the parser and compiler layers we use the C type [`block`](https://github.com/jqlang/jq/wiki/Internals%3A-block-representation) to represent jq programs and fragments.

pkoppstein:
@nicowilliams wrote:
> def debug1($msg): ($msg | select($ARGS.named.debug > 0 | debug | empty), .;

This of course will incur the (potentially non-negligible) cost of computing `$msg` even if $debug == 0.

By contrast, the `assert.jq` module previously mentioned goes to some lengths to ensure that if assertion-checking is turned off, the run-time costs associated with assertions will always be negligible.  As in C or Java, etc.

If there is any hope of some variation of assert.jq being included in builtin.jq, please let me know and I'll prepare a PR accordingly.

src/builtin.jq
```diff
@@ -264,6 +264,11 @@ def pick(pathexps):
 # ensure the output of debug(m1,m2) is kept together:
 def debug(msgs): (msgs | debug | empty), .;
 
+# assert property with message
+# success passes the original input through
+# failure raises an error with the original error message
+def assert(test; $message): if test then . else error("assertion error: \($message)") end;
```
tianon, line 270: If this drops the `$` on `$message`, it'll only evaluate the expression if there's an error, right?  That doesn't matter for simple messages, but might be relevant for more complex ones.

```suggestion
def assert(test; message): if test then . else error("assertion error: \(message)") end;
```

PR #2858, opened 2023-08-25
Check `nomem_handler->handler` before calling it

nicowilliams:
Traditionally we `abort()` on `ENOMEM`.  And in `master` we've been segfaulting on `ENOMEM` by trying to call a function at address `0x0`.  With this PR we'll go back to calling `abort()`.  But maybe we should just exit with an error and not `SIGABRT`, and therefore too without dumping core.  Thoughts?

This was reported by @chameleon10712.

### DIFF
--- a/src/jv_alloc.c
+++ b/src/jv_alloc.c
@@ -111,7 +111,7 @@ static void memory_exhausted() {
   tsd_init_nomem_handler();
 
   nomem_handler = pthread_getspecific(nomem_handler_key);
-  if (nomem_handler)
+  if (nomem_handler && nomem_handler->handler)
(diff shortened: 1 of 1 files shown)


wader (approved):
As its easy for a user to trigger huge allocatios i guess it make sense to not abort in that case. Would it make sense to print something to stderr? thinking possible silent exit might be confusing?

nicowilliams:
> As its easy for a user to trigger huge allocatios i guess it make sense to not abort in that case. Would it make sense to print something to stderr? thinking possible silent exit might be confusing?

We do print to `stderr`.  The question is only whether we should `abort()` or `exit()` (or `_exit()`).

nicowilliams:
I we want to not `abort()` we can always make it so in a follow-up PR.

PR #2860, opened 2023-08-26
Make jq_get_lib_dirs return an empty array if JQ_LIBRARY_PATH is not set

emanuele6:
For the jq_state used by the jq utility, the JQ_LIBRARY_PATH attribute
will always be set, but, in general, it is possible that it might not be.

If it is not set, jq_get_lib_dirs() will return jv_invalid().

That is not good, because some code in linker.c expects it to always
returns an array.

This patch makes jq_get_lib_dirs() return an empty array if
JQ_LIBRARY_PATH is not set to prevent problems.

This issue made OSS fuzz trigger failed assertions every time it tried
to compile a script that uses "include".

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61796

### DIFF
--- a/src/execute.c
+++ b/src/execute.c
@@ -1267,7 +1267,8 @@ jv jq_get_prog_origin(jq_state *jq) {
 }
 
 jv jq_get_lib_dirs(jq_state *jq) {
-  return jq_get_attr(jq, jv_string("JQ_LIBRARY_PATH"));
+  jv lib_dirs = jq_get_attr(jq, jv_string("JQ_LIBRARY_PATH"));
+  return jv_is_valid(lib_dirs) ? lib_dirs : jv_array();
 }
 
 void jq_set_attrs(jq_state *jq, jv attrs) {


src/execute.c
```diff
@@ -1267,7 +1267,8 @@ jv jq_get_prog_origin(jq_state *jq) {
 }
 
 jv jq_get_lib_dirs(jq_state *jq) {
-  return jq_get_attr(jq, jv_string("JQ_LIBRARY_PATH"));
+  jv lib_dirs = jq_get_attr(jq, jv_string("JQ_LIBRARY_PATH"));
+  return jv_is_valid(lib_dirs) ? lib_dirs : jv_array();
```
nicowilliams, line 1271: If `lib_dirs` is not valid it might need to be freed nonetheless if it has a message.

emanuele6: @nicowilliams No. That is not possible.

nicowilliams: Oh, because `jq_get_attr()` never does that.  Got it.

nicowilliams:
Thanks!

Issue #2260, opened 2021-02-01
Improved list destructuring by adding a 'rest' operator


reegnz:
**Describe the request**

I am trying to work with list destructuring. I want to be able to destructure array-like types (eg. array, string) into 'head' and 'tail' parts.

This is what I've been doing so far:
```sh
def head: .[0:1] | select(length >0) | if type == "array" then .[] else . end;
def tail: .[1:];
```

```sh
{ head: head, tail: tail }
```

This works for destructuring both strings and arrays.

**Expected behavior**

* Either include above two functions into the builtin.jq or
* Support destructuring with  a 'rest' operator (similar to how [javascript destructuring can use the rest operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) or how [python3 is using the starred expression](https://www.python.org/dev/peps/pep-3132/) to do it.

The operator one I would imagine to look like something in javascript, like `...`:

```sh
echo "[1,2,3,4,5]" | jq '. as [$head, ...$tail] | {$head, $tail}
[
  1,
  [
    2,
    3,
    4,
    5
  ]
]
```

**Environment (please complete the following information):**
 - OS and Version: macOS
 - jq version 1.6

vintnes:
I got half-way through writing tests for a new operator (generally frowned upon) before I realized you can already do this relatively concisely (and very flexibly) with slices:

    [.[0], .[1:-1], .[-1]] as [$head, $mid, $tail] | $head, $mid, $tail
    [0, 1, 2, 3, 4]
    0
    [1,2,3]
    4

A greedy operator (let's assume `*var`) begs the question of ambiguity.  
What should the output be for something like this?

    [0, 1, 2, 3, 4] as [*head, *tail] | $head, $tail

PR #2882, opened 2023-09-07
Add head and tail builtins

reegnz:
head:
* returns a single value or empty if input is a list
* returns a string or empty if input is a string
* fails if input is neither list or string

tail:
* returns a list if input is a list
* returns a string if input is a string
* fails if input is neither list or string

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -169,6 +169,8 @@ def nth($n; g):
   else label $out | foreach g as $item ($n + 1; . - 1; if . <= 0 then $item, break $out else empty end) end;
 def first: .[0];
 def last: .[-1];
+def head: select(length >0) | .[:1] | if type == "array" then .[] end;
+def tail: .[1:];
 def nth($n): .[$n];
 def combinations:
     if length == 0 then [] else
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1994,6 +1994,45 @@ any(keys[]|tostring?;true)
 {"a":"1","b":"2","c":"3"}
 true
 
+[head, tail]
+[ 1, 2, 3 ]
+[ 1, [ 2, 3 ] ]
+
+[head, tail]
+[ 1 ]
+[ 1, [] ]
+
+[head, tail]
+[]
+[ [] ]
+
+[head, tail]
+"abc"
+[ "a", "bc" ]
+
+[head, tail]
+[ "a" ]
+[ "a", "" ]
+
+[head] + tail
+[ 1, 2, 3 ]
+[ 1, 2, 3 ]
+
+[head] + tail
+[ 1 ]
+[ 1 ]
+
+[head] + tail
+[]
+[]
+
+head + tail
+"abc"
+"abc"
+
+head + tail
+"a"
+"a"
 
 # explode/implode
 # test replacement character (65533) for outside codepoint range and 0xd800 (55296) - 0xdfff (57343) utf16 surrogate pair range


reegnz:
These primitives helped me in implementing some convenience functions on top of them: https://github.com/reegnz/dotfiles/tree/master/jq

Most notably converting to/from camelcase.

Capitalizing strings is also easier to implement with such primitive filters exposed by jq.

reegnz:
Alternatively, one could use these as heads and tails (more cryptic, but gets the job done:

For lists:
```yaml
[ .[:1][], .[1:]]
```

For strings:
```yaml
[.[:1], .[1:]]
```

The unwrapped head for lists is inspired by other languages usually called destructuring, eg.:
https://elixir-lang.org/getting-started/pattern-matching.html
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
https://www.typescriptlang.org/docs/handbook/variable-declarations.html#destructuring

tests/jq.test
```diff
@@ -1994,6 +1994,45 @@ any(keys[]|tostring?;true)
...
+[head, tail]
+[]
+[ [] ]
+
+[head, tail]
+"abc"
+[ "a", "bc" ]
+
+[head, tail]
+[ "a" ]
```
jadutter, line 2014: This test is failing
```
Expected ["a",""], but got ["a",[]] 
```

src/builtin.jq
```diff
@@ -169,6 +169,8 @@ def nth($n; g):
   else label $out | foreach g as $item ($n + 1; . - 1; if . <= 0 then $item, break $out else empty end) end;
 def first: .[0];
 def last: .[-1];
+def head: select(length >0) | .[:1] | if type == "array" then .[] end;
```
jadutter, line 172: Based on the tests, it seems more accurate to think of these functions as 
`firstItem` and `remainingItems`. 
Calling them `head` and `tail` seems misleading because it implies they would work like the equivalent shell commands that read the start and end of a file for n lines. 
Instead, `head` always returns one item, while `tail` returns all except the first item.

I'm not confident these functions (in their current state) would be useful as builtins for the community as a whole. They seem like more of a niche thing that could exist in an independent [module](https://jqlang.github.io/jq/manual/#modules) as seems to be the [current case](https://github.com/jqlang/jq/pull/2882#issuecomment-1710486600).

tests/jq.test
```diff
@@ -1994,6 +1994,45 @@ any(keys[]|tostring?;true)
...
 
+[head, tail]
+[ 1, 2, 3 ]
+[ 1, [ 2, 3 ] ]
+
+[head, tail]
+[ 1 ]
+[ 1, [] ]
+
+[head, tail]
```
jadutter, line 2005: Given it's possible for `head` to return `empty` values, it's probably better to write tests that check these functions independently. Otherwise, in tests like this, it's not obvious which function returned an `empty` value.

src/builtin.jq
wader: As @reegnz  references in https://github.com/jqlang/jq/pull/2882#issuecomment-1710554500 the head and tail terminology comes from pattern matching in function languages.

pkoppstein:
Sorry to be so negative, but I don't see the point of these additions, and indeed based on my present understanding, they seem to just add bloat at best and confusion at worst (confusion, because of the history of `car` and `cdr`; see also @jadutter's remarks). 

Perhaps I'm missing something?  Please feel free to enlighten me :-)

reegnz:
I don't mind the specific terminology used. Can be `head` and `tail`, or `first` and `rest` or even `firstItem` vs `remainingItems` (although this one would diverge most from existing naming in other languages).

Calling it bloat sounds funny to me. It's true, most functional languages are complex because they have stdlib-s with higher order abstractions shipping with them. Wouldn't call that bloat though, that's a pretty big stretch.

Just please give this the benefit of the doubt instead of straight up calling it bloat and dismissing it. People do make use of taking the first item and remaining items of a list.

An alternative to filters is to make destructuring a language-level feature, like in javascript and clojure.

reegnz:
BTW original issue I've opened was #2260 this was just a quick run at the problem, I don't mind how it's solved.

itchyny:
These will unlikely be included as builtins, closing.

PR #2883, opened 2023-09-07
Add rename_keys builtin

reegnz:
map_values is already present, it would be natural to also have a
map_keys builtin.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -32,6 +32,7 @@ def _modify(paths; update):
         )
     ) | . as $dot | $dot[0] | delpaths($dot[1]);
 def map_values(f): .[] |= f;
+def map_keys(f): walk(if type == "object" then with_entries(.key | f) else . end);
 
 # recurse
 def recurse(f): def r: ., (f | r); r;
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -255,7 +255,7 @@ def walk(f):
     | f;
   w;
 
-def map_keys(f): walk(if type == "object" then with_entries(.key |= f) else . end);
+def rename_keys(f): walk(if type == "object" then with_entries(.key |= f) else . end);
 
 # pathexps could be a stream of dot-paths
 def pick(pathexps):
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -32,7 +32,6 @@ def _modify(paths; update):
         )
     ) | . as $dot | $dot[0] | delpaths($dot[1]);
 def map_values(f): .[] |= f;
-def map_keys(f): walk(if type == "object" then with_entries(.key |= f) else . end);
 
 # recurse
 def recurse(f): def r: ., (f | r); r;
@@ -255,6 +254,8 @@ def walk(f):
     end
     | f;
   w;
+
+def map_keys(f): walk(if type == "object" then with_entries(.key |= f) else . end);
 
 # pathexps could be a stream of dot-paths
 def pick(pathexps):
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -32,7 +32,7 @@ def _modify(paths; update):
         )
     ) | . as $dot | $dot[0] | delpaths($dot[1]);
 def map_values(f): .[] |= f;
-def map_keys(f): walk(if type == "object" then with_entries(.key | f) else . end);
+def map_keys(f): walk(if type == "object" then with_entries(.key |= f) else . end);
 
 # recurse
 def recurse(f): def r: ., (f | r); r;
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -637,6 +637,10 @@ map_values(.+1)
 [0,1,2]
 [1,2,3]
 
+map_keys(split("_")|map((.[:1] | ascii_upcase) +.[1:])|join(""))
+{"test_alpha": 1, "test_beta": {"test_gamma": 3}}
+{"TestAlpha": 1, "TestBeta": {"TestGamma": 3}}
+
 #
 # User-defined functions
 # Oh god.
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -637,10 +637,6 @@ map_values(.+1)
 [0,1,2]
 [1,2,3]
 
-map_keys(split("_")|map((.[:1] | ascii_upcase) + .[1:])|join(""))
-{"test_alpha": 1, "test_beta": {"test_gamma": 3}}
-{"TestAlpha": 1, "TestBeta": {"TestGamma": 3}}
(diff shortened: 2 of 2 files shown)


reegnz:
I've used this mainly to convert between camel-case and snake-case and similar use-cases:
https://github.com/reegnz/dotfiles/tree/master/jq

I thought the community might benefit if it's made a jq builtin.

itchyny:
Users cannot control the target object so walking in this builtin is not natural.

reegnz:
@itchyny not sure I get what you mean. Could you give an example?
I'm trying to figure it out on my own, but map_values also doesn't allow for controlling the target-object. (Maybe I'm not understanding what you mean by 'target' and 'control').

The point of the function is to map the keys of objects. The term "keys" don't make sense to me if the target is not "object". If the target is "object" that means it has key/value pairs, where the key can be mapped to a different value.

Thanks for your help.

emanuele6:
`map_values` does not walk and update all the object in the input, and also works on arrays, so why would it be *natural* to have a `map_keys` that updates the keys of all the object deep inside the input?

Also it is very weird to use `walk(if type == "object" then ... else . end)` here. Why do you need to walk instead of just looping all the objects in the input and changing their keys?
```jq
def yourfunction(f): (.. | objects) |= with_entries(.key |= f);
```

reegnz:
> Why do you need to walk instead of just looping all the objects in the input and changing their keys?

Didn't know of a better way. Now I probably do. ;)

EDIT: @emanuele6 your proposal doesn't work when I want to map_keys of the top-level object.

reegnz:
> why would it be natural to have a map_keys that updates the keys of all the object deep inside the input?

@emanuele6  When you want to switch all keys from CamelCase to snake_case for example, or the reverse, you need to walk all keys and transform each and every one of them.

Not seeing why we argue about some use-case being 'natural' or not. I don't get what you mean by it being natural or not.

reegnz:
Am I right in assuming we'd want a map_values that also maps list values based on the input function? But that's just `map`.

reegnz:
So I guess the criticism of it not being natural is the name of the builtin? Eg. map_keys shouldn't be recursive?
In that case you can already do `with_entries`, my point was to have a builtin that is `with_entries` but recursive.

emanuele6:
I see, you need to walk because renaming the keys will change the paths, so keys need to be renamed in the correct order from the bottom up; `walk(objects |= with_entries(.key |= f))`.

> > why would it be natural to have a map_keys that updates the keys of all the object deep inside the input?

> @emanuele6 When you want to switch all keys from CamelCase to snake_case for example, or the reverse, you need to walk all keys and transform each and every one of them.

I do understand what the function wants to do.

> Not seeing why we argue about some use-case being 'natural' or not. I don't get what you mean by it being natural or not.

In the description of the PR you said "map_values is already present, it would be natural to also have a map_keys builtin.", but this function hardly has anything to do to with `map_values`, `map_values` only modifies the values at the current level.

emanuele6:
Anyway, one way to achieve the same thing without needing to loop in `walk` order could be:
```jq
def yourfunction(f): fromstream(tostream | (.[0][] | strings) |= f);
```

I am not sure if that is better though. (probably not just because you would have to update a lot more strings.)

reegnz:
Ah, thank you for that explanation. So that comes from my own definition then. :)
Now I see how that's a valid criticism.

What would be a good name for this then? Can it be considered a useful builtin?

reegnz:
BTW what is the concern of avoiding using walk for this? Is it an expensive operation? It defeats some optimizations?

emanuele6:
> BTW what is the concern of avoiding using walk for this? Is it an expensive operation? It defeats some optimizations?

Not particularly, I just thought it would not be necessary, but now I realise it is because the paths change if you rename keys.

> Can it be considered a useful builtin?

Maybe

> What would be a good name for this then?

`renamekeys`?

emanuele6:
Maybe there could also be a `map_keys` function preferably implemented as `with_entries(. + { key: (.key | f) })` instead of `.key |= f`, so you can do stuff like this (map a key, to multiple keys):

```bash
$ jq 'def map_keys(f): with_entries(. + { key: (.key | f) }); map_keys(., ascii_downcase)' <<< '{"FOO":1,"bar":2,"baZ":3}'
{
  "FOO": 1,
  "foo": 1,
  "bar": 2,
  "baZ": 3,
  "baz": 3
}
```

And `renamekeys` could use `map_keys`.

@reegnz
Anyway, note that some users want to be conservative with introducing new jq-coded builtins, because the way they are currently implemented, they are not pre-compiled, and `jq` will have to compile them at startup even if you don't use them, so it will make startup time slower.
Honestly I don't think we are the point that startup time is a concern; `jq -n .` runs fast; but that appears to be a concern for some, so let's wait on reviews, and see if other people are interested.

itchyny:
I see you want a recursive filter included as builtin but such filter has limited usage and unlikely to be included due to the concern of growing builtins filters.
Filter updating the top-level object is more flexible that user can choose the target object; when they want recursive version just do `walk(objects |= map_keys(f))`; or they might want to change keys of objects with specific conditions.

reegnz:
I can get behind `map_keys` not being recursive.
Either way, this is already quite concise that a built-in is not even that necessary:

```sh
jq 'walk(objects |= with_entries(.key |= ascii_upcase))'
```

Semantically I still think `map_keys(ascii_upcase)` would make sense though, as IMHO it carries intent much better than plainly using `with_entries(.key |= ascii_upcase)`.

But I understand there is a contention between two groups of users: one group wants to keep core jq lightweight, which is a perfectly valid option.
The other group wants higher order functions built in that are helping with less abstract issues (eg. converting between camelcase and snake-case). The latter group might be just me. 😅

Regarding startup time and some users not wanting to extend builtins because of that:
Maybe both groups needs could be met by having built-in modules, that are not loaded by default? Eg. a blessed stdlib built right into jq, so specific use-cases (like converting case) could still live as the canonical implementation within jq, but not loaded by default? (emphasis being built-in, so it comes with jq, no need to download modules with a non-existing package manager)

Either way I'm happy we got to a point where my own set of private functions were already made simpler.

Anyway, I'm very grateful for both of your input so far, thanks @itchyny  and @emanuele6. 🙇

nicowilliams:
> map_values is already present, it would be natural to also have a
> map_keys builtin

So why not call this `map_keys`?

itchyny:
Although `def map_keys(f): with_entries(.key |= f)` is a good definition and likely to be in the builtins IMO, but current naming and definition of the patch are not acceptable, so closing the PR.

PR #2904, opened 2023-09-18
Fix the default colors to use 39, the default foreground color

### LINKED ISSUE #2903: Default color of jq is indistinguishable under Solarized Light background
**Describe the bug**
Since jq 1.7, its default color is very hard to read when using Solarized Light.

**To Reproduce**

Tilix 1.9.5, with Solarized Light color scheme.

```
> echo '{"a": 1, "b": "b", "c": [1,2,3,4,5]}' | jq .
```

image

**Expected behavior**

Clean output like jq 1.6.

- OS and Version: Arch Linux
- jq version 1.7

**Additional context**

~~This may be caused by https://github.com/jqlang/jq/pull/2824.~~

taoky:
This reverts commit 07b18a33feb6ccab08be221657e19b82abf0d84e, and fix #2903.

According to Linux manpage `console_codes(4)`, "color 39" is not a undefined value:

```
39         set default foreground color (before Linux 3.16: set underscore off, set default foreground color)
```

I'm afraid that what in https://github.com/jqlang/jq/pull/2032 does not hold, and setting its value to 37 hurts users with light background, as "color 37" is:

```
37         set white foreground
```

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -3677,7 +3677,7 @@ sections:
         - color for object keys
 
       The default color scheme is the same as setting
-      `JQ_COLORS="0;90:0;37:0;37:0;37:0;32:1;37:1;37:1;34"`.
+      `JQ_COLORS="0;90:0;39:0;39:0;39:0;32:1;39:1;39:1;34"`.
 
       This is not a manual for VT100/ANSI escapes.  However, each of
       these color specifications should consist of two numbers separated
--- a/docs/content/manual/v1.6/manual.yml
+++ b/docs/content/manual/v1.6/manual.yml
@@ -3343,7 +3343,7 @@ sections:
         - color for objects
 
       The default color scheme is the same as setting
-      `"JQ_COLORS=1;30:0;37:0;37:0;37:0;32:1;37:1;37"`.
+      `"JQ_COLORS=1;30:0;39:0;39:0;39:0;32:1;39:1;39"`.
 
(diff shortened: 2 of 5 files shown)


taoky:
And according to ECMA standard:

https://www.ecma-international.org/wp-content/uploads/ECMA-48_2nd_edition_august_1979.pdf, page 41 (PDF page 49):

> 39 (reserved for future standardization)

https://www.ecma-international.org/wp-content/uploads/ECMA-48_3rd_edition_march_1984.pdf, page 46 (PDF page 53), and https://www.ecma-international.org/wp-content/uploads/ECMA-48_5th_edition_june_1991.pdf, page 62 (PDF page 76):

> 39 default display colour (implementation-defined)

nicowilliams:
Wouldn't we also need to update the main manual?

Otherwise LGTM.

docs/content/manual/v1.7/manual.yml
```diff
@@ -3677,7 +3677,7 @@ sections:
         - color for object keys
 
       The default color scheme is the same as setting
-      `JQ_COLORS="0;90:0;37:0;37:0;37:0;32:1;37:1;37:1;34"`.
+      `JQ_COLORS="0;90:0;39:0;39:0;39:0;32:1;39:1;39:1;34"`.
```
itchyny: Do not update this file. v1.7 ships with `0;37` as defaults.

itchyny (approved):

PR #2915, opened 2023-09-28
main.c: Remove unused EXIT_STATUS_EXACT option

emanuele6:
In process there is a suspicious `options |= EXIT_STATUS_EXACT` that is run when the jq script is terminated by `halt`, or `halt_error`.

That line of code acutally does nothing because options is a local argument variable, and is not passed as a pointer. It was probably meant
to be a `*options |= EXIT_STATUS_EXACT` with the options argument passed as a `int *`.

In any case, we do not want to run the code in main() that was supposed to run if `EXIT_STATUS_EXACT` is set (but didn't since it is never added to options); as far as I can tell, we only want to run that code when the --exit-status/-e option is passed.

So I removed `EXIT_STATUS_EXACT` completely, and the useless assignment, instead of fixing it since it was not used for anything else.

Useless assignment detected by clang-tidy.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -161,7 +161,6 @@ enum {
   RAW_NO_LF             = 1024,
   UNBUFFERED_OUTPUT     = 2048,
   EXIT_STATUS           = 4096,
-  EXIT_STATUS_EXACT     = 8192,
   SEQ                   = 16384,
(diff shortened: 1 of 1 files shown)


nicowilliams:
Thanks!

PR #2916, opened 2023-09-28
Actually use number correctly casted from double to int as index

emanuele6:
The code was using `(int)jv_number_value(k)` instead of `(int)didx`.

follow-up from 0e067ef93605493060392f0999be27694146fca4 (#2821)

Useless assignments to `didx` detected by clang-tidy.


emanuele6:
I forgot to create a PR for this patch earlier :/

nicowilliams:
Thanks!

PR #2926, opened 2023-10-08
Remove undefined behavior caught by LLVM 10 UBSAN

### LINKED ISSUE #2881: Currently-harmless UB in `stack_reset()` and `stack_reallocate()`
Reported [here](https://donotsta.re/notice/AZVP1zPjXVEuWb6Klk).

clang w/ `-fsanitize=undefied` reports:

```
../src/exec_stack.h:64:32: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/exec_stack.h:64:32 in
../src/exec_stack.h:83:36: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/exec_stack.h:83:36 in
```

These are cases doing pointer arithmetic with `NULL` pointers and then yield `NULL` pointers.  If the compiler were to try to optimize these functions away due to this, then that would cause jq to break immediately -- the tests certainly wouldn't pass, especially with `valgrind`.  Therefore it is safe to say that for all the tested builds of jq this bit of UB is not _currently_ a problem.

emanuele6:
Thanks to @Gaelan for the patch!
Taken from #2108.

Fixes #2881

### DIFF
--- a/src/exec_stack.h
+++ b/src/exec_stack.h
@@ -61,8 +61,10 @@ static void stack_init(struct stack* s) {
 
 static void stack_reset(struct stack* s) {
   assert(s->limit == 0 && "stack freed while not empty");
-  char* mem_start = s->mem_end - ( -s->bound + ALIGNMENT);
-  free(mem_start);
+  if(s->mem_end != NULL) {
+    char* mem_start = s->mem_end - ( -s->bound + ALIGNMENT);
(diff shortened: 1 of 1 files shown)


nicowilliams:
Thanks!

PR #2931, opened 2023-10-12
Generate links in man page

### LINKED ISSUE #2930: man page missing text
**Describe the bug**
Text is missing from the man page. It looks like anytime there is a link, the link text and following text to the end of the line is removed.

**To Reproduce**
For a specific example, look at the first two sentences of under the [REGULAR EXPRESSIONS section](https://man.archlinux.org/man/jq.1.en#REGULAR_EXPRESSIONS).
It reads:
> jq uses the
>
> Oniguruma supports several flavors of regular expression, so it is important to know that jq uses the

**Expected behavior**
It should contain the [full text](https://jqlang.github.io/jq/manual/#regular-expressions):
> jq uses the Oniguruma regular expression library, as do PHP, TextMate, Sublime Text, etc, so the description here will focus on jq specifics.
>
> Oniguruma supports several flavors of regular expression, so it is important to know that jq uses the "Perl NG" (Perl with named groups) flavor.

**Environment:**
- Archlinux
- jq version 1.7

**Additional context**
To rule out a packaging bug, I also built `jq` manually from the git source. I made sure I had the correct dependencies installed and followed the [documentation build instructions](https://github.com/jqlang/jq/blob/master/docs/README.md). `jq.1` is still missing the same text.

It is worth noting that the text is also missing from the [prebuilt man page](https://github.com/jqlang/jq/blob/master/jq.1.prebuilt#L2741).

emanuele6:
Some \[foo](bar) links were added to manual.yml without updating build_manpage.py to make it generate roff for "a" tags.

Fixes #2930

### DIFF
--- a/docs/build_manpage.py
+++ b/docs/build_manpage.py
@@ -86,6 +86,9 @@ class RoffWalker(object):
                     self.__write_cmd('.P')
                 self._write_element(root, ensure_newline=(parent_tag != 'li'))
 
+            elif root.tag == 'a':
+                self._write_element(root, ensure_newline=(parent_tag != 'li'))
+
             elif root.tag == 'ul':
                 if self._ul_is_special(root):
                     li = root[0]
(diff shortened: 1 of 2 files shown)


wader (approved):

nicowilliams:
Thanks!

PR #2932, opened 2023-10-13
Add last resort version string in scripts/version

nicowilliams:
Ugh.  We do already include some built sources (e.g., outputs of Bison and Flex, the prebuilt man page), so why not a bit more?  Well, I'm not going that far -- I'm not including `src/version.h` here, just hard-coding a last resort version string into `scripts/version`.

This PR is really about exploring what, if _anything_ we might do about these two issues:

1. GitHub _insists_ on including a tarball of the workspace of the repo checked out at the release tag when making releases.  That tarball does not include `.git`, so `scripts/version` can't figure out the current version being built.
2. Running `autoconf`/`autoreconf` in the dist tarballs we publish in our releases clobbers the `src/version.h` included in the tarball, and once again there is no `.git` so `scripts/version` can be used by `Makefile` to recover from this.

This PR adds a fallback at the price of having one more thing to update at release time.  It might be better instead to commit `src/version.h`, but that's also going to be a pain since we'd have to do that at every commit!

### DIFF
--- a/scripts/version
+++ b/scripts/version
@@ -2,7 +2,15 @@
 set -eu
 
 cd "$(dirname "$0")/../"
-test -d .git || exit 1
+if ! [ -d .git ]; then
+  # Building from default GitHub tarball, or building from dist after running
+  # autoreconf unnecessarily.  This is something we can fail to update, but
+  # it's been generating a lot of questions.
+  #
+  # Now we'll have to make `Release jq-<version>` commits that change this:
+  echo "jq-1.7-misconfigured"
+  exit 0
+fi
 
 if git describe --tags --match 'jq-*' >/dev/null 2>&1; then
   git describe --tags --match 'jq-*' --dirty | sed 's/^jq-//'


wader:
Seems like we're not alone with the confusing "Source code" download links https://github.com/orgs/community/discussions/6003. I subscribed to the thread to keep an eye, i guess we would like to disable them if it was possible?

Could an option to be fallback generate some version string that made it very clear something is not right?

emanuele6:
Alternatively you could make `make` keep the old `version.h` if `scripts/version` fails:
```diff
diff --git a/Makefile.am b/Makefile.am
index f7bdfc2..d6c15f9 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -103,12 +103,12 @@ endif
 # Remake the version.h header file if, and only if, the git ID has changed
 .PHONY: .FORCE
 .FORCE:
-generate_ver = ver="`{ $(srcdir)/scripts/version || echo '$(VERSION)' ; } | sed 's/.*/\#define JQ_VERSION \"&\"/'`"
+generate_ver = ver="\#define JQ_VERSION \"`$(srcdir)/scripts/version || { echo '$(VERSION)'; false ;}`\""
 .remake-version-h: .FORCE
-	@ $(generate_ver); test "x`cat src/version.h 2>/dev/null`" = "x$$ver" || touch .remake-version-h
+	@if $(generate_ver); then test "x`cat src/version.h 2>/dev/null`" = "x$$ver" || touch $@; fi
 src/version.h: .remake-version-h
 	mkdir -p src
-	$(AM_V_GEN) $(generate_ver); echo "$$ver" > $@
+	$(AM_V_GEN) if $(generate_ver); then echo "$$ver" > $@; else touch $@; fi
 src/config_opts.inc:
 	mkdir -p src
 	$(AM_V_GEN) if test -x ./config.status; then \

```

nicowilliams:
> Could an option to be fallback generate some version string that made it very clear something is not right?

Oh, that's an interesting idea.  I like it.

nicowilliams:
Another option is to have `.github/workflows/ci.yml` make a commit to set the version (thus no need to worry about forgetting to do it manually).

PR #2935, opened 2023-10-21
Fix possible uninitialised value dereference if jq_init() fails

emanuele6:
If `jq_init()` fails, `goto out` would try to free `input_state` which is uninitialised. I initialised `input_state` to `NULL` to fix the problem.

I also fixed `jq_util_input_init()` not handling OOM errors by returning `NULL`, and added code to make jq exit cleanly if it returns `NULL`. The codebase is filled with these kinds of problems, but this one was easy to fix, so might as well fix it now...

Ref: https://github.com/jqlang/jq/pull/2934#discussion_r1367795641

Reported-By: Klemens Nanni <kn@openbsd.org>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -310,6 +310,7 @@ int umain(int argc, char* argv[]) {
 int main(int argc, char* argv[]) {
 #endif
   jq_state *jq = NULL;
+  jq_util_input_state *input_state = NULL;
   int ret = JQ_OK_NO_OUTPUT;
   int compiled = 0;
   int parser_flags = 0;
@@ -336,7 +337,7 @@ int main(int argc, char* argv[]) {
 
   jq = jq_init();
   if (jq == NULL) {
-    perror("malloc");
+    perror("jq_init");
     ret = JQ_ERROR_SYSTEM;
     goto out;
   }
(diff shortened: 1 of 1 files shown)


src/util.c
```diff
@@ -226,10 +226,12 @@ jq_util_input_state *jq_util_input_init(jq_util_msg_cb err_cb, void *err_cb_data
     err_cb_data = stderr;
   }
   jq_util_input_state *new_state = jv_mem_calloc(1, sizeof(*new_state));
-  new_state->err_cb = err_cb;
-  new_state->err_cb_data = err_cb_data;
-  new_state->slurped = jv_invalid();
-  new_state->current_filename = jv_invalid();
+  if(new_state) {
```
klemensn: Space between `if` and `(`?

klemensn (approved):
LGTM

src/util.c
emanuele6: Hmm, yeah, sorry, I had my editor configured for `curl`-style

src/util.c
```diff
@@ -226,10 +226,12 @@ jq_util_input_state *jq_util_input_init(jq_util_msg_cb err_cb, void *err_cb_data
     err_cb_data = stderr;
   }
   jq_util_input_state *new_state = jv_mem_calloc(1, sizeof(*new_state));
-  new_state->err_cb = err_cb;
-  new_state->err_cb_data = err_cb_data;
-  new_state->slurped = jv_invalid();
-  new_state->current_filename = jv_invalid();
+  if (new_state) {
```
nicowilliams: `jv_mem_calloc()` `abort()`s if `ENOMEM`.

emanuele6: Oh, my bad.
Isn't that a `libjq` function? That is not very nice! D:

emanuele6: I removed that part of patch.

nicowilliams:
Thanks!

PR #2937, opened 2023-10-28
jvp_object_contains: remove unnecessary jv_copy

emanuele6:


emanuele6:
I just found a few other unnecessary `jv_copy` in src/builtin.c

src/builtin.c
```diff
@@ -58,24 +58,23 @@ BINOPS
...
 
 static jv type_error2(jv bad1, jv bad2, const char* msg) {
   char errbuf1[15],errbuf2[15];
+  const char *badkind1 = jv_kind_name(jv_get_kind(bad1));
+  const char *badkind2 = jv_kind_name(jv_get_kind(bad2));
   jv err = jv_invalid_with_msg(jv_string_fmt("%s (%s) and %s (%s) %s",
-                                             jv_kind_name(jv_get_kind(bad1)),
-                                             jv_dump_string_trunc(jv_copy(bad1), errbuf1, sizeof(errbuf1)),
-                                             jv_kind_name(jv_get_kind(bad2)),
-                                             jv_dump_string_trunc(jv_copy(bad2), errbuf2, sizeof(errbuf2)),
```
wader, line 75: For understanding: was thinking first just remove the `jv_copy` and `jv_free` below would be enough but then i remembered arg order is undefined i think?

wader (approved):

src/builtin.c
emanuele6: Yep, it is undefined. Also, iirc most implementations evaluate right-to-left.

src/builtin.c
```diff
@@ -58,24 +58,23 @@ BINOPS
 
 static jv type_error(jv bad, const char* msg) {
   char errbuf[15];
-  jv err = jv_invalid_with_msg(jv_string_fmt("%s (%s) %s",
-                                             jv_kind_name(jv_get_kind(bad)),
-                                             jv_dump_string_trunc(jv_copy(bad), errbuf, sizeof(errbuf)),
+  const char *badkind = jv_kind_name(jv_get_kind(bad));
```
nicowilliams, line 61: So we're just trying to make `type_error()` fast?  :)

nicowilliams:
Thanks!

src/builtin.c
emanuele6: I was reading the source code for `contains/1` and I found that unnecessary `jv_copy()`, then I looked for other unnecessary copies in `src/builtin.c` for fun, and I only found those that were easily avoidable. =)

nicowilliams: I rather liked having things like this as one C expression :) but you're right that unnecessary `jv_copy()`s are bad.

PR #2942, opened 2023-11-03
Comment bug fixes, and fully support Tcl-style multiline comments

emanuele6:
* bugfix: comments were incorrectly being terminated by CR; for example
 
      jq -n $'1 #foo\r'

  fails to compile because the CR character terminates the comment, and CR is not a valid character in jq syntax.

* improvement: comments fully support Tcl-style line continuation.  Previously this was only "supported" in `-f` scripts, whose first line starts with "#!", and second line starts with `#` and ends with `\`, only for the comment on the second line, only for one extra line.

* man: document comment syntax, which was previously undocumented.

* tests: add regression tests for the bugfix, and some tests for line continuation in comments.

### DIFF
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -229,7 +229,7 @@ sections:
       * `-f filename` / `--from-file filename`:
 
         Read filter from the file rather than from a command line, like
-        awk's -f option. You can also use '#' to make comments.
+        awk's -f option.
 
       * `-L directory`:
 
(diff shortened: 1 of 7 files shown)


docs/content/manual/manual.yml
```diff
@@ -3537,6 +3537,68 @@ sections:
...
+            if . == []
+              then inputs
+              else .[]
+            end |
+            . as $dot |
+            try
+              tonumber
+            catch
+              (@json "sum: Invalid number \($dot).\n" | halt_error(1))
+          ) as $n (0; . + $n)
```
emanuele6: Hmm, I should probably make it reject nan or it will output null instead of a number. =)

nicowilliams:
LGTM!

nicowilliams:
Thanks!

PR #2945, opened 2023-11-15
refactor: move bsearch function to C-code

### LINKED ISSUE #527: bsearch as C-coded builtin
As best I can tell, based on my timings on a Mac, a jq-coded binary search algorithm on a sorted array can't compete even with the C-coded linear search.  The jq version of bsearch that I used is appended.

Thus I would like to request a C-coded bsearch that would always terminate and that would return the index of the item if the array is sorted (as per `sort`).  I realize that such a function would be unusual, but one can say "it is what it is".

The only satisfactory alternative that I can think of (#517) has already been effectively rejected.

```
# binary search
# requires last/1
# WARNING: If the input is not sorted as per sort, bsearch will terminate but with
# irrelevant results.
#
def bsearch(target):
  if length == 0 then null
  elif length == 1 then (if .[0] == target then 0 else null end)
  else . as $in
    # state variable: [start, end, answer]
    # where start and end are the upper and lower offsets to use.
    | last( [0, length-1, null]
        | while ( .[0] <= .[1] ;
                  (if .[2] != null then (.[1] = -1) # i.e. break
                   else
                     ( ( (.[1] + .[0]) / 2 ) | floor ) as $mid
                     | $in[$mid] as $monkey
                     | if $monkey == target  then (.[2] = $mid)     # success
                       elif .[0] == .[1]     then (.[1] = -1)       # failure
                       elif $monkey < target then (.[0] = ($mid + 1))
                       else (.[1] = ($mid - 1))
                       end
                   end ))) | .[2]
    end;
```

eloycoto:
This commit fixes issue #527 and move the bsearch function to a native C-code.

The performance is a bit better:

Testing script:
```bash
clear
if [[ `uname` == Darwin ]]; then
    MAX_MEMORY_UNITS=KB
else
    MAX_MEMORY_UNITS=MB
fi

export TIMEFMT='%J   %U  user %S system %P cpu %*E total'$'\n'\
'avg shared (code):         %X KB'$'\n'\
'avg unshared (data/stack): %D KB'$'\n'\
'total (sum):               %K KB'$'\n'\
'max memory:                %M '$MAX_MEMORY_UNITS''$'\n'\
'page faults from disk:     %F'$'\n'\
'other page faults:         %R'

echo "JQ code bsearch"
time /usr/bin/jq -n '[range(30000000)] | bsearch(3000)'

echo "C code bsearch"
time ./jq -n '[range(30000000)] | bsearch(3000)'
````

Results:

```
JQ code bsearch
3000
/usr/bin/jq -n '[range(30000000)] | bsearch(3000)'   8.63s  user 0.77s system 98% cpu 9.542 total
avg shared (code):         0 KB
avg unshared (data/stack): 0 KB
total (sum):               0 KB
max memory:                823 MB
page faults from disk:     1
other page faults:         432828
C code bsearch
3000
./jq -n '[range(30000000)] | bsearch(3000)'   8.44s  user 0.74s system 99% cpu 9.249 total
avg shared (code):         0 KB
avg unshared (data/stack): 0 KB
total (sum):               0 KB
max memory:                824 MB
page faults from disk:     0
other page faults:         432766
```

The results may be better if we can use jvp_array_read, and there is no need to copy/free the input array in each iteration. I guess that is like that for API pourposes when the libjq is in use with multiple threads in place.

---

Closes #527

### CODE REVIEW

File: src/builtin.c [hunk 1]
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+/* Assuming the input array is sorted, bsearch/1 returns */
+/* the index of the target if the target is in the input array; and otherwise */
+/*  (-1 - ix), where ix is the insertion point that would leave the array sorted. */
+/* If the input is not sorted, bsearch will terminate but with irrelevant results. */
+static jv f_bsearch(jq_state *jq, jv input, jv target) {
+    assert(jv_get_kind(input) == JV_KIND_ARRAY);
+    assert(jv_get_kind(target) == JV_KIND_NUMBER);
+    int len = jv_array_length(jv_copy(input));
+    if (len == 0) {
+        return jv_number(-1);
```
emanuele6: leaking input, and target

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+    } else if (len == 1) {
+        jv val = jv_copy(jv_array_get(jv_copy(input), 0));
+        int result = jv_cmp(target, val);
+        jv_free(val);
+        if (result == 0 ) {
+            return jv_number(0);
```
emanuele6: leaking input

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        } else if (result > 0) {
+            return jv_number(-2);
```
emanuele6: leaking input

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        } else {
+            return jv_number(-1);
```
emanuele6: leaking input

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+            answer = jv_number(-1);
+            break;
+        } else if (result < 0 ) {
+            end = mid -1;
+        } else {
+            start = mid +1;
+        }
+    }
+
+    if (jv_equal(answer, jv_null())) {
```
emanuele6, line 853: can never be true?

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        jv val = jv_array_get(jv_copy(input), start);
+        int result = jv_cmp(target, val);
+        if (result < 0) {
+            answer =  jv_number(-1 - start);
+        }else {
+            answer =  jv_number(-2 - start);
+        }
+    }
+    return answer;
```
emanuele6: leaking input.
if the previous if did not run, answer has been freed before being returned

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+    }
+
+    int start = 0;
+    int end = len - 1;
+    jv answer = jv_null();
+    while (start <end) {
+        int mid = (start + end) / 2;
+
+        jv val = jv_array_get(jv_copy(input), mid);
+        int result = jv_cmp(target, val);
```
emanuele6: if this loops more than once, reusing target after free + frees target again after free

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        } else if (result < 0 ) {
+            end = mid -1;
+        } else {
+            start = mid +1;
+        }
+    }
+
+    if (jv_equal(answer, jv_null())) {
+        jv val = jv_array_get(jv_copy(input), start);
+        int result = jv_cmp(target, val);
```
emanuele6, line 855: never runs, but would be using target after free

emanuele6: if this runs, it uses target after free + frees target after free

src/builtin.c
```diff
@@ -807,6 +806,62 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+static jv f_bsearch(jq_state *jq, jv input, jv target) {
+    assert(jv_get_kind(input) == JV_KIND_ARRAY);
+    assert(jv_get_kind(target) == JV_KIND_NUMBER);
+    int len = jv_array_length(jv_copy(input));
+    if (len == 0) {
+        return jv_number(-1);
+    } else if (len == 1) {
+        jv val = jv_copy(jv_array_get(jv_copy(input), 0));
+        int result = jv_cmp(target, val);
+        jv_free(val);
```
emanuele6: useless copy+free?

nicowilliams: Yes.  Write this instead:

```C
        int result = jv_cmp(target, jv_copy(jv_array_get(jv_copy(input), 0)));
```

nicowilliams: Do we need to optimize for the case where `len == 1`?

src/builtin.c
```diff
@@ -807,6 +806,63 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+    assert(jv_get_kind(input) == JV_KIND_ARRAY);
+    assert(jv_get_kind(target) == JV_KIND_NUMBER);
+    int len = jv_array_length(jv_copy(input));
+    if (len == 0) {
+        jv_free(input);
+        jv_free(target);
+        return jv_number(-1);
+    } else if (len == 1) {
+        int result = jv_cmp(target, jv_array_get(jv_copy(input), 0));
+        jv_free(input);
```
emanuele6: no need to copy

src/builtin.c
```diff
@@ -807,6 +806,63 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+            answer = jv_number(-1);
+            break;
+        } else if (result < 0 ) {
+            end = mid -1;
+        } else {
+            start = mid +1;
+        }
+    }
+
+    if (jv_equal(answer, jv_null())) {
```
emanuele6: this `jv_equal` will free `answer` which is not good, since you want to return it if its not `null`, you should have copied it.

Also the best way to check if something is `null` is with `if (jv_get_kind(answer) == JV_KIND_NULL)`, this is one of the few `jv_` functions that does not free its argument automatically so no need to copy if you use that.

src/builtin.c
```diff
@@ -807,6 +806,63 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        int result = jv_cmp(jv_copy(target), jv_array_get(jv_copy(input), start));
```
emanuele6: could avoid copy here, and only free target input on else branch

src/builtin.c
```diff
@@ -807,6 +806,63 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        if (result < 0) {
+            answer =  jv_number(-1 - start);
+        }else {
+            answer =  jv_number(-2 - start);
```
emanuele6: this is fine, but I just wanted to point out it is only fine to not call `jv_free(answer)` before replacing it here because `answer` is a `jv_null()`, and nulls are one of the kinds of jvs that do not necessarily need to be freed.

src/builtin.c
emanuele6: I guess it is working out anyway because `answer` is always either `jv_null()` or a `jv_number()` which are never allocated.

eloycoto:
@emanuele6 I think this is ready

itchyny:
The `bsearch` filter should support searching from an array of any kind.
```
 $ jq -n '[{x:range(3),y:range(3)}] | debug(. == sort) | bsearch({x:1,y:1})'
["DEBUG:",true]
4
```

suisanT1 (approved):

eloycoto:
@emanuele6 @itchyny I've just refactored this, and merging should be okay.

src/builtin.c
```diff
@@ -780,6 +779,61 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+            end = mid -1;
+        } else {
+            start = mid +1;
+        }
+    }
+    if (jv_get_kind(answer) == JV_KIND_NULL) {
+        int result = jv_cmp(target, jv_array_get(jv_copy(input), start));
+        if (result < 0) {
+            answer =  jv_number(-1 - start);
+        }else {
```
itchyny:
```suggestion
        } else {
```

src/builtin.c
```diff
@@ -780,6 +779,61 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+        if (result == 0) {
+            answer = jv_number(mid);
+            break;
+        } else if (start == end ) {
+            answer = jv_number(-1);
+            break;
+        } else if (result < 0 ) {
+            end = mid -1;
+        } else {
+            start = mid +1;
```
itchyny:
```suggestion
        } else if (start == end) {
            answer = jv_number(-1);
            break;
        } else if (result < 0) {
            end = mid - 1;
        } else {
            start = mid + 1;
```

src/builtin.c
```diff
@@ -780,6 +779,61 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+            return jv_number(-2);
+        } else {
+            return jv_number(-1);
+        }
+    }
+
+    int start = 0;
+    int end = len - 1;
+    jv answer = jv_null();
+    while (start <end) {
```
itchyny:
```suggestion
    while (start < end) {
```

src/builtin.c
```diff
@@ -780,6 +779,61 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
...
+static jv f_bsearch(jq_state *jq, jv input, jv target) {
+    assert(jv_get_kind(input) == JV_KIND_ARRAY);
+    int len = jv_array_length(jv_copy(input));
+    if (len == 0) {
+        jv_free(input);
+        jv_free(target);
+        return jv_number(-1);
+    } else if (len == 1) {
+        int result = jv_cmp(target, jv_array_get(input, 0));
+        if (result == 0 ) {
```
itchyny:
```suggestion
        if (result == 0) {
```

tests/jq.test
```diff
@@ -1553,6 +1553,10 @@ bsearch(0,2,4)
 1
```
itchyny, line 1649: Could you change the query of this case to `bsearch(0,1,2,3,4)`? The output should be `-1 0 1 2 -4`.

itchyny:
I found a regression. On updating tests see my review comment above.
```
 $ jq -n '[1,2] | bsearch(2)'          
1
 $ ./jq -n '[1,2] | bsearch(2)'
-3
```

itchyny:
I think we can reduce special cases, simplify the logic.
```c
static jv f_bsearch(jq_state *jq, jv input, jv target) {
  if (jv_get_kind(input) != JV_KIND_ARRAY) {
    return type_error(input, "cannot be searched from");
  }
  int start = 0;
  int end = jv_array_length(jv_copy(input));
  jv answer = jv_invalid();
  while (start < end) {
    int mid = (start + end) / 2;
    int result = jv_cmp(jv_copy(target), jv_array_get(jv_copy(input), mid));
    if (result == 0) {
      answer = jv_number(mid);
      break;
    } else if (result < 0) {
      end = mid;
    } else {
      start = mid + 1;
    }
  }
  if (!jv_is_valid(answer)) {
    answer = jv_number(-1 - start);
  }
  jv_free(input);
  jv_free(target);
  return answer;
}
```

itchyny (approved):

src/builtin.c
```diff
@@ -780,6 +779,37 @@ static jv f_sort_by_impl(jq_state *jq, jv input, jv keys) {
   }
 }
 
+/* Assuming the input array is sorted, bsearch/1 returns */
+/* the index of the target if the target is in the input array; and otherwise */
+/*  (-1 - ix), where ix is the insertion point that would leave the array sorted. */
+/* If the input is not sorted, bsearch will terminate but with irrelevant results. */
+static jv f_bsearch(jq_state *jq, jv input, jv target) {
+  if (jv_get_kind(input) != JV_KIND_ARRAY) {
+    return type_error(input, "cannot be searched from");
```
emanuele6, line 847: Here you are leaking target; you need to free it before returning.
Otherwise looks good to me.

emanuele6 (dismissed):

src/builtin.c
emanuele6: Maybe also add a test with a non-array input, and a target that uses allocated memory (e.g. a string), so this can be caught by the CI?

emanuele6:
I just remembered that `(start + end) / 2` for binary search is not recommended because it can overflow.
This can be avoided using `start + (end - start) / 2` instead.
See https://en.wikipedia.org/wiki/Binary_search_algorithm#Implementation_issues

tests/jq.test
```diff
@@ -1547,12 +1547,22 @@ ascii_upcase
...
+2
 -4
 
+bsearch({x:1})
+[{ "x": 0 },{ "x": 1 },{ "x": 2 }]
+1
+
+bsearch(0)
+"aa"
+jq: error (at <unknown>): string ("aa") cannot be searched from
```
emanuele6: To test errors, you need something like this:
```suggestion
try ["OK", bsearch(0)] catch ["KO",.]
"aa"
["KO","string (\"aa\") cannot be searched from"]
```
Also you used a number as target which is not necessarily allocated, you should use something like an array or object, with elements

src/builtin.c
```diff
@@ -46,7 +46,6 @@ void *alloca (size_t);
 #include "jv_private.h"
 #include "util.h"
 
-
```
emanuele6: can you put this line back in please?

itchyny (approved):

wader:
Looks good to me, I'm a bit surprised of the small performance difference, calling overhead etc in both cases are larger than the binary search or something?

wader (approved):

itchyny:
> I'm a bit surprised of the small performance difference, calling overhead etc in both cases are larger than the binary search or something?

Because `[range(30000000)]` took most of the time in the entire query.
Reading the target array from file is better in comparing the performance,
but it still costs, so it cannot compare only of `bsearch`.

```
 $ jq -nc '[range(3000000)]' > /tmp/x.json
 $ hyperfine --warmup 10 'jq "bsearch(100)" /tmp/x.json' './jq "bsearch(100)" /tmp/x.json'
Benchmark 1: jq "bsearch(100)" /tmp/x.json
  Time (mean ± σ):     575.6 ms ±   5.4 ms    [User: 513.4 ms, System: 56.1 ms]
  Range (min … max):   570.1 ms … 588.7 ms    10 runs

Benchmark 2: ./jq "bsearch(100)" /tmp/x.json
  Time (mean ± σ):     478.9 ms ±   3.9 ms    [User: 419.8 ms, System: 52.6 ms]
  Range (min … max):   473.4 ms … 483.5 ms    10 runs

Summary
  ./jq "bsearch(100)" /tmp/x.json ran
    1.20 ± 0.01 times faster than jq "bsearch(100)" /tmp/x.json

 $ jq -nc '[range(10000000)]' > /tmp/x.json
 $ hyperfine --warmup 10 'jq "bsearch(100)" /tmp/x.json' './jq "bsearch(100)" /tmp/x.json'
Benchmark 1: jq "bsearch(100)" /tmp/x.json
  Time (mean ± σ):      2.002 s ±  0.056 s    [User: 1.759 s, System: 0.208 s]
  Range (min … max):    1.956 s …  2.121 s    10 runs

Benchmark 2: ./jq "bsearch(100)" /tmp/x.json
  Time (mean ± σ):      1.683 s ±  0.077 s    [User: 1.435 s, System: 0.203 s]
  Range (min … max):    1.612 s …  1.821 s    10 runs

Summary
  ./jq "bsearch(100)" /tmp/x.json ran
    1.19 ± 0.06 times faster than jq "bsearch(100)" /tmp/x.json
```

itchyny:
Merging, but with dropping the misleading profiling result comments.

PR #2946, opened 2023-11-15
simplify paths/0 and paths/1

asheiduk:
`recurse/0` already handles traversing objects and arrays, so it is more consistent to use that.
For `paths/1` it is easier to use the actual value returned by `recurse` instead of querying that value with `getpath/1` later.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -48,8 +48,8 @@ def indices($i): if type == "array" and ($i|type) == "array" then .[$i]
   else .[$i] end;
 def index($i):   indices($i) | .[0];       # TODO: optimize
 def rindex($i):  indices($i) | .[-1:][0];  # TODO: optimize
-def paths: path(recurse(if (type|. == "array" or . == "object") then .[] else empty end))|select(length > 0);
-def paths(node_filter): . as $dot|paths|select(. as $p|$dot|getpath($p)|node_filter);
+def paths: path(recurse)|select(length > 0);
+def paths(node_filter): path(recurse|select(node_filter))|select(length > 0);
 def isfinite: type == "number" and (isinfinite | not);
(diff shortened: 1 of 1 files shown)


emanuele6:
Could even use `..` instead of `recurse`.
`..` is syntax sugar that compiles to a call to `recurse/0`.

nicowilliams:
LGTM.

nicowilliams:
@emanuele6 did you want to require the use of `..`?

wader (approved):
LGTM but i noticed there is no tests for `paths/1`, time to add?

asheiduk:
@wader There is something in ` man.test`: 
```
[paths(type == "number")]
[1,[[],{"a":2}]]
[[0],[1,1,"a"]]
```

emanuele6:
> @emanuele6 did you want to require the use of `..`?

Well, I thought it would look neater with `..`...

Also, it would be easier to read since `path(..)` is a common pattern, and `..` is commonly used, while `recurse` with no arguments is basically never used, but fine either way I guess.

asheiduk:
> > @emanuele6 did you want to require the use of `..`?
> 
> Well, I thought it would look neater with `..`...
> 
> Also, it would be easier to read since `path(..)` is a common pattern, and `..` is commonly used, while `recurse` with no arguments is basically never used, but fine either way I guess.

@emanuele6 I think this is true for "user" code but here the definition of `recurse` is in the same file -- a file internal to jq itself. Having this textual direct link (e.g. via double-clicking) seemed more important to me. But that's only my take.

After all: Shall I change it so the PR can be merged or was this only an optional suggestion?

emanuele6:
Well, it is fine with `recurse`

emanuele6:
Thank you!

PR #2952, opened 2023-11-27
Extend fuzzing set up

DavidKorczynski:
Adds a parse function in `jv_parse.c` that enables parsing using custom flags for the parser. This is then used by two fuzzers added as well.

This is to make sure fuzzing hits various code parts currently not fuzzed, e.g. `stream_token`:
https://storage.googleapis.com/oss-fuzz-coverage/jq/reports/20231125/linux/src/jq/src/jv_parse.c.html#L241

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -234,6 +234,7 @@ enum {
 
 jv jv_parse(const char* string);
 jv jv_parse_sized(const char* string, int length);
+jv jv_parse_custom_flags(const char* string, int flags);
 
 typedef void (*jv_nomem_handler_f)(void *);
 void jv_nomem_handler(jv_nomem_handler_f, void *);
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -859,49 +859,6 @@ jv jv_parser_next(struct jv_parser* p) {
   }
 }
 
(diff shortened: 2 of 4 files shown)


DavidKorczynski:
Am not sure if adding `jv_parse_custom_flags` is too intrusive: if so, then I could wrap it in a macro that is only present for fuzzing builds?

emanuele6:
I think it's ok to add a new function, but maybe make `jv_parse_sized` a wrapper that calls `jv_parse_sized_custom_flags` with `0` as flags to avoid duplication.

Thank you for adding more fuzzing!

PS: Could you also add a fuzzer for `jv_dump` functions (`jv` => `JSON` serialisation)?

DavidKorczynski:
> maybe make jv_parse_sized a wrapper that calls jv_parse_sized_custom_flags with 0 as flags to avoid duplication.

Done!

> PS: Could you also add a fuzzer for jv_dump functions (jv => JSON serialisation)?

Added!

Once this is landed I'll add the new fuzzers to https://github.com/google/oss-fuzz/blob/master/projects/jq/build.sh

emanuele6:
Thank you!

PR #2953, opened 2023-11-27
libjq: extern C for C++

rockwotj:
If using libjq from C++ it would be nice to not need to do this at the
import site, so just extern "C" to the public headers for libjq

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -5,6 +5,10 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #if (defined(__GNUC__) && __GNUC__ >= 7) || \
     (defined(__clang__) && __clang_major__ >= 10)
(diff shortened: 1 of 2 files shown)


emanuele6 (approved):

src/jv.h
```diff
@@ -259,6 +263,9 @@ int jv_cmp(jv, jv);
 jv jv_group(jv, jv);
 jv jv_sort(jv, jv);
 
+#ifdef __cplusplus
+}
+#endif
 
 #endif
```
wader, line 271: Unrelated to this PR: i wonder if the comment below should be cleaned up?

wader (approved):

src/jv.h
emanuele6: Yeah, maybe!

emanuele6:
Thank you!

src/jv.h
rockwotj: In addition to that, would yall be open to some light comments on what these methods do for these two files?

PR #2955, opened 2023-11-29
website: use https URLs instead of http URLs in download page

emanuele6:
Also add markdown formatting for the decNumber URL so it gets rendered as a link in the html page.


itchyny (approved):

emanuele6:
|x|screenshot|
|---|---|
|before|file2|
|after | file|

wader (approved):

PR #2959, opened 2023-11-30
Extend fuzzing

DavidKorczynski:
Adds `jq_dump_disassembly` as a target for `jq_fuzz_compile.c`

Adds a new fuzzer where the primary purpose is to hit `jq_next`, which is currently not analyzed by the existing fuzzers.

### DIFF
--- a/tests/jq_fuzz_execute.cpp
+++ b/tests/jq_fuzz_execute.cpp
@@ -1,0 +1,40 @@
+#include <fuzzer/FuzzedDataProvider.h>
+#include <string>
+
+extern "C" {
+#include "jq.h"
+#include "jv.h"
+}
+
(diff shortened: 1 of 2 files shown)


tests/jq_fuzz_execute.cpp
```diff
@@ -0,0 +1,40 @@
+#include <fuzzer/FuzzedDataProvider.h>
+#include <string>
+
+extern "C" {
+#include "jq.h"
+#include "jv.h"
+}
```
emanuele6: I think that `extern "C"` is not needed thanks to #2953

emanuele6 (approved):

emanuele6:
Thank you!

PR #2960, opened 2023-12-02
jq_fuzz_execute: fix memory leak

DavidKorczynski:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64626

### DIFF
--- a/tests/jq_fuzz_execute.cpp
+++ b/tests/jq_fuzz_execute.cpp
@@ -27,8 +27,7 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
           jv_free(actual);
         }
         jv_free(next);
-      }
-      else {
+      } else {
         // Only free if input as invalid as otherwise jq_teardown
         // frees it.
         jv_free(input);
--- a/tests/jq_fuzz_execute.cpp
+++ b/tests/jq_fuzz_execute.cpp
@@ -28,7 +28,7 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
         }
         jv_free(next);
       } else {
-        // Only free if input as invalid as otherwise jq_teardown
+        // Only free if input is invalid as otherwise jq_teardown
         // frees it.
         jv_free(input);
       }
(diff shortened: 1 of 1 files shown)


tests/jq_fuzz_execute.cpp
```diff
@@ -28,8 +28,11 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
         }
         jv_free(next);
       }
-
-      // Do not free "input" as this is handled by jq_teardown.
+      else {
+        // Only free if input as invalid as otherwise jq_teardown
+        // frees it.
```
wader, line 32: So if understand it correctly: `jq_parse` will always return something allocated. If valid the input argument will end up on the stack by `jq_start` and freed during teardown. If invalid it needs to be freed explicitly?

tests/jq_fuzz_execute.cpp
```diff
@@ -28,8 +28,11 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
         }
         jv_free(next);
       }
-
-      // Do not free "input" as this is handled by jq_teardown.
+      else {
```
wader: nit pick: `} else {` to follow style in other files?

tests/jq_fuzz_execute.cpp
DavidKorczynski: Yes, that's how I understand it.

tests/jq_fuzz_execute.cpp
DavidKorczynski: Ah yes, thanks! Done!

wader (approved):
LGTM but maybe wait for one more approve

tests/jq_fuzz_execute.cpp
emanuele6: "is invalid"?

DavidKorczynski: > "is invalid"?

Ah yes, good catch. Fixed.

emanuele6:
Thank you!

PR #2963, opened 2023-12-03
jq_fuzz_execute.cpp: fix false assert

DavidKorczynski:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64632 by following the suggestion here: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64632#c2 -- the issue is fixed when this attribute is set.

I did not add any of the other suggestions in the comment: we haven't run into these issues in the fuzzer so far.

### DIFF
--- a/tests/jq_fuzz_execute.cpp
+++ b/tests/jq_fuzz_execute.cpp
@@ -16,9 +16,7 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
   jq_state *jq = NULL;
   jq = jq_init();
   if (jq != NULL) {
-    char *origin = strdup("/tmp/");
-    jq_set_attr(jq, jv_string("JQ_ORIGIN"), jv_string(origin));
-    free(origin);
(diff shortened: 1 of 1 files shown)


tests/jq_fuzz_execute.cpp
```diff
@@ -16,6 +16,10 @@ extern "C" int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {
   jq_state *jq = NULL;
   jq = jq_init();
   if (jq != NULL) {
+    char *origin = strdup("/tmp/");
+    jq_set_attr(jq, jv_string("JQ_ORIGIN"), jv_string(origin));
+    free(origin);
```
wader: Could this be `jq_set_attr(jq, jv_string("JQ_ORIGIN"), jv_string("/tmp/));`? skip strdup/free

wader (approved):
LGTM with maybe possible simplification that i mentioned

tests/jq_fuzz_execute.cpp
DavidKorczynski: Yes, thanks! Done

wader (approved):

PR #2964, opened 2023-12-05
Return error for string literal beginning with single quote

### LINKED ISSUE #501: "Invalid Numeric Literal" on colon following a single-quoted string (incorrect parser error message)
**EDIT (20150416): I KNOW IT'S INVALID JSON.  THE BUG IS ABOUT THE NONSENSICAL ERROR MESSAGE**

I get this from absent-mindedly and incorrectly using single-quotes:

```
$ echo "{'foo':'bar'}" | jq .foo
parse error: Invalid numeric literal at line 1, column 7
```

This is under 1.3.  Trying to reproduce under 1.4, but after 5 minutes of trying to build from source I'm just going to submit this and move on.

mattmeyers:
## Summary

In order for a string literal to be valid in JSON, the value must be surrounded by double quotes. This PR updates the literal validation function to return an error message when the literal begins with a single quote.

I'm not sure if this is the best solution since the line/col that gets reported is at the end of the literal and might be confusing. Another possible solution would be to mark the single quote character as invalid in the `classify` function. This would give the exact line/col where the invalid string begins, but I'm not sure if there would be other consequences to doing this. It appears that reporting the position after an invalid literal is seen elsewhere e.g. `echo '{"foo":flase}' | jq`.

## Testing

Run the command `echo "{'foo': 1}" | jq` and ensure the output mentions an invalid string literal. Compare this to the dev version of jq which mentions an invalid numeric literal.

Closes #501


src/jv_parse.c
```diff
@@ -512,6 +512,8 @@ static pfunc check_literal(struct jv_parser* p) {
   switch (p->tokenbuf[0]) {
   case 't': pattern = "true"; plen = 4; v = jv_true(); break;
   case 'f': pattern = "false"; plen = 5; v = jv_false(); break;
+  case '\'':
+    return "Invalid string literal";
```
wader: Maybe should even say something like "Invalid string literal. Try use double quote instead of single quote."

wader:
Probably good to add a test case for this to `tests/jq.test`:
```
try fromjson catch .
"{'a': 123}"
"Invalid numeric literal at line 1, column 5 (while parsing '{'a': 123}')"
```
(but with updated error message)

About other error cases: I wonder if the could somehow check the first character to know if this if something resembling a number so that we could distinguish between invalid number and other invalid things?

mattmeyers:
> Probably good to add a test case for this to `tests/jq.test`:

Of course! I got excited and committed before looking into how tests were handled :facepalm: 

> About other error cases: I wonder if the could somehow check the first character to know if this if something resembling a number so that we could distinguish between invalid number and other invalid things?

Not sure I follow. Isn't this what the parser already does around the code I added?

wader:
> Of course! I got excited and committed before looking into how tests were handled 🤦

No worries :)

> > About other error cases: I wonder if the could somehow check the first character to know if this if something resembling a number so that we could distinguish between invalid number and other invalid things?
> 
> Not sure I follow. Isn't this what the parser already does around the code I added?

Haven't look close att the current code but it seems like the clumps together a lots of different invalid jsons texts into "Invalid numeric literal" some i think we could probably split up into more user friendly errors and suggestions:

Did a script to try different strings:
```sh
$ for i in true false nan ident truea falsea nana aa1 1aa + - / "&" "*" '(' "'aa'"; do echo "$i:\t$(jq 2>&1 <<< "$i")"; done
true:	true
false:	false
nan:	null
ident:	jq: parse error: Invalid numeric literal at line 2, column 0
truea:	jq: parse error: Invalid literal at line 2, column 0
falsea:	jq: parse error: Invalid literal at line 2, column 0
nana:	jq: parse error: Invalid literal at line 2, column 0
aa1:	jq: parse error: Invalid numeric literal at line 2, column 0
1aa:	jq: parse error: Invalid numeric literal at line 2, column 0
+:	jq: parse error: Invalid numeric literal at line 2, column 0
-:	jq: parse error: Invalid numeric literal at line 2, column 0
/:	jq: parse error: Invalid numeric literal at line 2, column 0
&:	jq: parse error: Invalid numeric literal at line 2, column 0
*:	jq: parse error: Invalid numeric literal at line 2, column 0
(:	jq: parse error: Invalid numeric literal at line 2, column 0
'aa':	jq: parse error: Invalid numeric literal at line 2, column 0
```

Sidenote:
No idea what is going on with `-`:
```
$ echo - | jq
$

# weirdly gojq behaves the same
$ echo - | gojq
$
```

itchyny:
Recognized as echo command option.

wader:
> Recognized as echo command option.

Thanks 🤦

emanuele6:
LGTM if you add the test

src/jv_parse.c
emanuele6: @wader If you want that, add it to the lexer, or it will also be printed for `fromjson/0` or `tonumber/0` or JSON parsing errors

emanuele6: Maybe it could work if you rephrase it as something like
```
Invalid string literal: Got ', but was expecting "
```

mattmeyers:
Hey all, sorry for the delay. It's the busy part of the year. I went ahead and updated the error message, checked for a few more invalid literals, and added some tests. Let me know if there are any more changes you would like to see.

src/jv_parse.c
```diff
@@ -512,6 +512,12 @@ static pfunc check_literal(struct jv_parser* p) {
   switch (p->tokenbuf[0]) {
   case 't': pattern = "true"; plen = 4; v = jv_true(); break;
   case 'f': pattern = "false"; plen = 5; v = jv_false(); break;
+  case '\'':
+    return "Invalid string literal; expected \", but got '";
+  case '/':
+  case '*':
+  case '&':
+    return "Invalid literal";
```
emanuele6: Why?

foo.jq
```diff
@@ -0,0 +1 @@
+flase
```
emanuele6: ?

src/jv_parse.c
mattmeyers: In @wader's comment above, he found other character's that were triggering the `invalid numeric literal`. This check catches those. Would this cause an issue?

foo.jq
mattmeyers: That's embarrassing. That was just a test file that should not have been committed. Removed.

src/jv_parse.c
emanuele6: Why should it say "Invalid literal" instead of "Invalid numeric literal" for those three characters? How is that better?

mattmeyers: Does it make more sense for those to trigger `Invalid numeric literal`? Personally those don't seem like numeric characters (as compared to, say, `-` which if alone, could be an invalid negative number). I was just implementing the suggestions above. I'm happy to revert if this doesn't make sense.

emanuele6: "Invalid numeric literal" means that `jq` tried to parse a JSON number and it was not a number. so the error makes sense, yes.
"Invalid literal"? what does that even mean? :)

mattmeyers: Oh I see. I guess if we got to this point, then we've tried parsing it as a string, boolean, and null. So all that's left is parsing it as a number, hence that message. Maybe the better error message would be something like `Invalid character '&' encountered while parsing number`? I'm happy to revert this change for now though if everything else looks good.

wader: I guess that ppl search and find #501 (it's the most visited issue) because we mention numeric in many error cases and it's confusing

Had a go at compring jq to `JSON.parse`:
```sh
$ for i in ident truea falsea nana aa1 1aa + - / "&" "*" '(' "'aa'"; do echo "$i:\t$(jq 2>&1 <<< "$i")"; echo "        $(node -p "JSON.parse(\"$i\")" 2>&1 | grep Error | head -n1)"; done
ident:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token 'i', "ident" is not valid JSON
truea:	jq: parse error: Invalid literal at line 2, column 0
        SyntaxError: Unexpected non-whitespace character after JSON at position 4 (line 1 column 5)
falsea:	jq: parse error: Invalid literal at line 2, column 0
        SyntaxError: Unexpected non-whitespace character after JSON at position 5 (line 1 column 6)
nana:	jq: parse error: Invalid literal at line 2, column 0
        SyntaxError: Unexpected token 'a', "nana" is not valid JSON
aa1:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token 'a', "aa1" is not valid JSON
1aa:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected non-whitespace character after JSON at position 1 (line 1 column 2)
+:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token '+', "+" is not valid JSON
-:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: No number after minus sign in JSON at position 1 (line 1 column 2)
/:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token '/', "/" is not valid JSON
&:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token '&', "&" is not valid JSON
*:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token '*', "*" is not valid JSON
(:	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token '(', "(" is not valid JSON
'aa':	jq: parse error: Invalid numeric literal at line 2, column 0
        SyntaxError: Unexpected token ''', "'aa'" is not valid JSON
```

emanuele6: For me it's ok changing the default error to `Invalid character X` instead of `Invalid numeric literal`. It is not ok however to make `&`, `/` and `*` randomly error with "Invalid literal"; while most other character would error e.g. `boo`, `=`, `!` return `Invalid literal`.
Anyhow, I only just noticed that `Invalid literal` is an error that `jq` already uses instead of `Invalid numeric literal` for `f`, `n`, `t` because they are short circuited, so that should be changed to be more consistent I guess.
I am fine with the `'` patch as a temporary fix for now, but speecial casing `&`, `/`, `*` feels too random.

Should we just change the default error to `Invalid character` instead, and not even special case `'` since `Invalid character "'" at` is probably good enough?

mattmeyers: I'd be happy to implement a general `Invalid character "X"` kind of error message, but I may need someone to point me in the right direction. My understanding is that the `check_literal` function returns a `const char*` meaning we wouldn't be able to get that dynamic error message here without changing a bunch of code. If this change is beyond the scope of the issue, I can just revert back to the simple `'` special case.

itchyny:
This change looks good to merge. Any objections?

wader (approved):

tests/jq.test
```diff
@@ -2103,6 +2103,9 @@ try ["ok", setpath([1]; 1)] catch ["ko", .]
 {"hi":"hello"}
 ["ko","Cannot index object with number"]
 
+try fromjson catch .
+"{'a': 123}"
+"Invalid string literal; expected \", but got ' at line 1, column 5 (while parsing '{'a': 123}')"
```
itchyny, line 2108: Why `column 5`, not `column 2`?

itchyny (approved):
Thank you!

PR #2965, opened 2023-12-06
tests: add new fuzzer

DavidKorczynski:
This adds a new fuzzer that is similar in nature to `jq_fuzz_execute`. The difference is that the `jq_fuzz_execute` will seed both the `jq` program and input with fuzz data wheras this new `jq_fuzz_fixed` fuzzer has a predefined list of `jq` program and only the input will be seeded from the fuzzer. The benefit is that this fuzzer is closer to identifying issues that may happen in "generic jq programs" and much faster at exploring certain parts of the codebase. It does explore a subset of the `jq_fuzz_execute`, however, there is plenty of code to explore and this will be much more efficient at targeting some (relevant) parts of it.

The `jq_progs` was taken from the `/tests/*.test` files.

From a local run, this drastically improves code coverage.


emanuele6:
Thank you!

tests/jq_fuzz_fixed.cpp
```diff
@@ -0,0 +1,297 @@
...
+    "[.user, .projects[]]",
+    "{(.user): .titles}",
+    "{user, title: .titles[]}",
+    "utf8bytelength",
+    "walk(if type == \"array\" then sort else . end)",
+    "walk( if type == \"object\" then with_entries( .key |= sub( \"^_+\"; "
+    "\"\") ) else . end )",
+    "[while(.<100; .*2)]",
+    "with_entries(.key |= \"KEY_\" + .)",
+    ". - [\"xml\", \"yaml\"]",
```
wader, line 257: Are these generated from documentation or test? maybe could be?

DavidKorczynski: from the `tests/*.test`. This one specifically  is:https://github.com/jqlang/jq/blob/5029328d35f3e60037970d27f350a742af41aa02/tests/man.test#L170

Specifically, it was generated using a simple scraping "capture each line that is after an empty line in some of the `.test` files" with a bit of sorting etc.

Please feel free to add any new ones as well! The more the merrier, and also it's not a big deal (for the fuzzer at least) if some of them are not valid.

PR #2969, opened 2023-12-09
builtins: make ltrimstr and rtrimstr error for non-string inputs

emanuele6:
I also refactored the code in rtrimstr to make it more similar to the
code used by ltrimstr.

Previously, ltrimstr/rtrimstr would just let the input pass through for
non-string inputs or arguments.

That was happening because, they were leaking the errors returned by
startswith/endswith treating them as if they were jv_false().

This patch also fixes that memory leak discovered by oss-fuzz

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64946

### DIFF
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -2110,3 +2110,19 @@ try ltrimstr("x") catch "x", try rtrimstr("x") catch "x" | "ok"
 {"hey":[]}
 "ok"
 "ok"
+
+# ltrimstr/1 and rtrimstr/1 return an error for non-strings. #2969
+
+.[] as [$x, $y] | try ["ok", ($x | ltrimstr($y))] catch ["ko", .]
+[["hi",1],[1,"hi"],["hi","hi"],[1,1]]
+["ko","startswith() requires string inputs"]
+["ko","startswith() requires string inputs"]
+["ok",""]
+["ko","startswith() requires string inputs"]
+
+.[] as [$x, $y] | try ["ok", ($x | rtrimstr($y))] catch ["ko", .]
+[["hi",1],[1,"hi"],["hi","hi"],[1,1]]
+["ko","endswith() requires string inputs"]
+["ko","endswith() requires string inputs"]
+["ok",""]
+["ko","endswith() requires string inputs"]
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -294,37 +294,6 @@ static jv f_endswith(jq_state *jq, jv a, jv b) {
   return ret;
 }
 
-static jv f_ltrimstr(jq_state *jq, jv input, jv left) {
-  jv startswith = f_startswith(jq, jv_copy(input), jv_copy(left));
-  if (jv_get_kind(startswith) != JV_KIND_TRUE) {
-    jv_free(startswith);
-    jv_free(left);
-    return input;
-  }
-  /*
-   * FIXME It'd be better to share the suffix with the original input --
-   * that we could do, we just can't share prefixes.
-   */
-  int prefixlen = jv_string_length_bytes(left);
-  jv res = jv_string_sized(jv_string_value(input) + prefixlen,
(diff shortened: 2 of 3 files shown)


emanuele6:
`gojq` copies `ltrimstr`/`rtrimstr` behaving like `.` if the input or the argument is not a string; do we want to preserve that behaviour, and just free the error message instead of leaking it?

itchyny:
I thought the previous behavior is useful for recursive trimming (`.. |= ltrimstr`). But I also understand the inconsistency in the builtins, and using `strings` filter allows to trim recursively, so I'm okay with changing the behavior.

emanuele6:
Since `f_startswith` / `f_endswith` are basically
```jq
def ltrimstr($left): if startswith($left) then .[$left | length:] end;
def rtrimstr($right): if endswith($right) then .[:$right | -length] end;
```
defined in `C`, should we just rewrite them in `jq` in `src/builtin.jq`?

The memory "optimisation" that is mentioned in the FIXME in `f_ltrimstr` could be handled by string slices, if we even want it...

tests/jq.test
```diff
@@ -2091,3 +2091,18 @@ try ("foobar" | .[1.5]) catch .
 null
 "Cannot index string with number"
 
+# ltrimstr/1 and rtrimstr/1 don't leak, and return an error for
```
wader: move to where rest of ltrimstr/rstrimstr tests are?

emanuele6: Hmm, there isn't a `ltrimstr`/`rtrimstr` section, only two tests in the middle of "string operations"

```jq
[.[]|ltrimstr("foo")]
["fo", "foo", "barfoo", "foobar", "afoo"]
["fo","","barfoo","bar","afoo"]

[.[]|rtrimstr("foo")]
["fo", "foo", "barfoo", "foobar", "foob"]
["fo","","bar","foobar","foob"]
```

Wouldn't it be confusing if I moved this block with a comment there?

wader:
Agree on making them consistent with other string functions and it does not look like we have documented the passthru behaviour anywhere.

tests/jq.test
wader: That works too i guess, i'm thinking it's good to group tests per function at least? but don't have that strong opinion about it really

emanuele6: Hmm, maybe create a separate PR for that, all the string functions are in the same section currently

emanuele6:
What do you think about rewriting those functions in `jq` instead of keeping the C implemenations?

tests/jq.test
wader: Sounds good 👍

wader:
> What do you think about rewriting those functions in `jq` instead of keeping the C implemenations?

If no big performance difference i think it makes sense

wader (approved):
Wait to see if @itchyny is happy too?

itchyny (approved):
LGTM

itchyny:
If you could it's better to fix leaks in the patch release and change the behavior in 1.8.

emanuele6:
Should I just create a new PR and leave this open for 1.8 then?

emanuele6:
I think the next release should be `1.8` since we already merged non-bug fix changes like the one that allows using `# \` everywhere

itchyny:
> we already merged non-bug fix changes like the one that allows using # \ everywhere

That's the only feature addition in 1.7? I think we should still discuss on the patch release for security issues, reverting the comment changes.

emanuele6:
If you want to release a patch release before `1.8`, reverting the comment change, only keeping the bug fix that allows `CR` in comments is fine I guess

nicowilliams:
Any reason not to merge this?

wader:
LGTM unless the change of pass through is a backwards compatibility issue?

nicowilliams:
Yeah, I'm not concerned about the passthrough change.

PR #2971, opened 2023-12-10
jq_fuzz_fixed: fix test cases that use string interpolation

emanuele6:
`"\("` is implementation-defined in C++, and both clang and gcc treat it as equivalent to `"("`, not `"\\("`.
This patch replaces `"\("` with `"\\("` so that tests with string interpolation actually use string interpolation.


emanuele6:
I'll just merge this to make the fuzzer fuzz the right tests sooner.
cc @DavidKorczynski (in case my edit broke something `=)`)

PR #2977, opened 2023-12-11
ltrimstr/1+rtrimstr/1: don't leak on invalid input or arguments

emanuele6:
`ltrimstr`/`rtrimstr` was ignoring and leaking the error returned by `f_startswith()`/`f_endswith()`.

This also means that they just let the input pass through for non-string inputs or arguments.

Only fix the leak for now; in the next release, #2969 will make them rethrow the error returned by startswith/endswith.

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64946

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -295,7 +295,9 @@ static jv f_endswith(jq_state *jq, jv a, jv b) {
 }
 
 static jv f_ltrimstr(jq_state *jq, jv input, jv left) {
-  if (jv_get_kind(f_startswith(jq, jv_copy(input), jv_copy(left))) != JV_KIND_TRUE) {
+  jv startswith = f_startswith(jq, jv_copy(input), jv_copy(left));
+  if (jv_get_kind(startswith) != JV_KIND_TRUE) {
+    jv_free(startswith);
     jv_free(left);
     return input;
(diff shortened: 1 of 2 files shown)


emanuele6:
Next release will be 1.8

src/builtin.c
```diff
@@ -295,7 +295,9 @@ static jv f_endswith(jq_state *jq, jv a, jv b) {
 }
 
 static jv f_ltrimstr(jq_state *jq, jv input, jv left) {
-  if (jv_get_kind(f_startswith(jq, jv_copy(input), jv_copy(left))) != JV_KIND_TRUE) {
+  jv startswith = f_startswith(jq, jv_copy(input), jv_copy(left));
+  if (jv_get_kind(startswith) != JV_KIND_TRUE) {
+    jv_free(startswith);
```
itchyny, line 300: Should we free `startswith` outside this case too?

emanuele6: null, true, false, invalids without messages, and numbers without literals don't need to be freed

itchyny: Ah, I see.

itchyny (approved):

Issue #911, opened 2015-08-16
Win32 build is broken, needs to find PathIsRelativeA()


nicowilliams:
`linker.c` needs to include `<windows.h>` and `<shlwapi.h>`, and jq needs to link with `-lshlwapi` when building with mingw.  I'm not sure if the latter is something we should do unconditionally in `configure.ac`, so I'll not fix that (instead we can point out the need to add it to `LDFLAGS` in the ./configure incantation).

dtolnay:
Why would we not want to link with `-lshlwapi` unconditionally (on mingw, anyway)?

nicowilliams:
Ah yes, I'd be matching on the host triple, so yes, it should be unconditional in the mingw case.

svnpenn:
For any dummies out there, that means you need to do this to fix it:

    make clean
    make LDFLAGS=-lshlwapi

JanSchulz:
Seems that the `make LDFLAGS=...` is still needed: #1074

With that, I get:
```
  CC       src/linker.lo
src/linker.c: In function 'path_is_relative':
src/linker.c:32:13: warning: implicit declaration of function 'PathIsRelativeA' [-Wimplicit-function-declaration]
   int res = PathIsRelativeA(s);
             ^
```

Only a warning, but...

JanSchulz:
And on 32bit it it still sfails even with that: https://github.com/stedolan/jq/pull/1075#issuecomment-172264537

```
CC       src/jv_file.lo
  CC       src/jv_parse.lo
  CC       src/jv_print.lo
  CC       src/jv_unicode.lo
  CC       src/linker.lo
src/linker.c: In function 'path_is_relative':
src/linker.c:32:13: warning: implicit declaration of function 'PathIsRelativeA' [-Wimplicit-function-declaration]
   int res = PathIsRelativeA(s);
             ^
  CC       src/locfile.lo
  CC       src/util.lo
  CC       src/lexer.lo
  CC       src/parser.lo
  CCLD     libjq.la
libtool: warning: undefined symbols not allowed in x86_64-pc-mingw32 shared libraries; building static only
  CC       src/main.o
  CCLD     jq.exe
./.libs/libjq.a(linker.o): In function `path_is_relative':
C:\projects\jq/src/linker.c:32: undefined reference to `PathIsRelativeA'
collect2.exe: error: ld returned 1 exit status
Makefile:893: recipe for target 'jq.exe' failed
make[1]: *** [jq.exe] Error 1
[...]

PR #2979, opened 2023-12-11
builtin.inc: fix build with -Woverlength-strings

### LINKED ISSUE #1481: builtin.inc exceeds maximum length for C99
In order to publish my bindings I need to solve this compiler warning:

```
Found the following significant warnings:
  jq/builtin.inc:1:1: warning: string literal of length 14491 exceeds maximum length 4095 that ISO C99 compilers are required to support [-Woverlength-strings]
```

Is there a way to split this literal into smaller chunks or so?

emanuele6:
C99 only allows string literals with length up to 4095 characters.

This patch introduces a `gen_jq_builtins` generator program that generates the `jq_builtins` array in builtin.c as an array literal instead of a string literal to make compilation not fail when building with `-Woverlength-strings`.

Also add `-Woverlength-strings` to `CFLAGS` to verify that the fix works.

Fixes #1481

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,7 @@ LIBJQ_INCS = src/builtin.h src/bytecode.h src/compile.h                 \
         src/linker.h src/locfile.h src/opcode_list.h src/parser.y       \
         src/util.h src/decNumber/decContext.h src/decNumber/decNumber.h \
         src/decNumber/decNumberLocal.h src/jv_dtoa_tsd.h src/jv_thread.h \
-        src/jv_private.h
+        src/jv_private.h src/math-macos.h
 
 LIBJQ_SRC = src/builtin.c src/bytecode.c src/compile.c src/execute.c    \
         src/jq_test.c src/jv.c src/jv_alloc.c src/jv_aux.c              \
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,10 +19,13 @@ LIBJQ_SRC = src/builtin.c src/bytecode.c src/compile.c src/execute.c    \
 
 ### C build options
 
-AM_CFLAGS = -Wextra -Wall -Wno-unused-parameter -Wno-unused-function
+AM_CFLAGS = -Wextra -Wall -Woverlength-strings
+
+AM_CFLAGS += -Wno-unused-parameter -Wno-unused-function
 
 if WIN32
 AM_CFLAGS += -municode
+HOSTCFLAGS += -municode
 endif
 
 ACLOCAL_AMFLAGS = -I config/m4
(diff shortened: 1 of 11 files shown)


emanuele6:
I need to add `src/gen_jq_builtins.c` to dist

emanuele6:
I think I need to pass more flags to make it build on non-x86 systems.
And I also need to define `main()` in a special way to make it build on windows.

emanuele6:
I fixed the `macos (amd64)` build =)

Makefile.am
```diff
@@ -137,6 +144,13 @@ if ENABLE_ALL_STATIC
 jq_LDFLAGS += -all-static
 endif
 
+bin_PROGRAMS += src/gen_jq_builtins
+gen_jq_builtins_SOURCES = src/gen_jq_builtins.c
+
+if WIN32
+gen_jq_builtins_LIBADD = -lshlwapi
```
wader: Curious, what requires this? dugg around a bit and in https://github.com/jqlang/jq/issues/911 it seems to be needed for some `PathIsRelativeA`

emanuele6: *shruggie* I still have not fixed the win32 build =)

wader:
Huh strange failure `cannot execute binary file: Exec format error`, can't really see how that could happen... cross compile flags somehow?

emanuele6:
I fixed the windows builds, now I will try to remove `-lshlwapi`

wader:
Remembered now that we only run and tests on amd64 https://github.com/jqlang/jq/blob/master/.github/workflows/ci.yml#L168-L172 but it does not correlate with the failured ci jobs hmm. Thinking if gen_builtins is built as cross target binary somehow?

emanuele6:
It works without `-lshlwapi`, we should check if `jq` actually needs it later.

emanuele6:
`CI / linux (mipsr6el)`:
```
  CCLD     src/gen_jq_builtins
  GEN      src/version.h
file -- src/gen_jq_builtins
src/gen_jq_builtins: ELF 32-bit LSB executable, MIPS, MIPS32 rel6 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-mipsn8.so.1, BuildID[sha1]=6ae16e6a26ff4d90fa6e4c3e1b4dcdb949d11a63, for GNU/Linux 4.5.0, not stripped
  GEN      src/builtin.inc
/bin/bash: line 1: src/gen_jq_builtins: cannot execute binary file: Exec format error
```
🤔

emanuele6:
I think CIs are all running on amd64, and just cross compiling for a given architecture

wader:
`configure: WARNING: using cross tools not prefixed with host triplet` 🤔

emanuele6:
We should use an host compiler to build `gen_jq_builtins` somehow

wader:
> We should use an host compiler to build `gen_jq_builtins` somehow

Hmm define a `CC_HOST` in the ci env?

src/main-win32.h
```diff
@@ -0,0 +1,28 @@
+#ifndef MAIN_WIN32_H
+#define MAIN_WIN32_H
+
+#ifdef WIN32
+#include <stringapiset.h>
+#include <wchar.h>
+
+#define DEFINE_MAIN(ARGC, ARGV) \
```
wader, line 8: Haven't notice this code before. Is this used when building native for windows, not via emulation layers like msys2 or cygwin?

emanuele6: I think, when you compile with `-municode` on windows, `wmain()` is called instead of `main()` as entry point with `wchar_t*` strings instead of `char*` strings; then that code in `wmain()` converts the `wchar_t*` strings to `char*` strings encoding with utf-8.

emanuele6: I assume it is probably an hack to make the program get the correct utf-8 encoding on unicode characters you pass through `cmd.exe` because windows encodings are weird.
For example, recently, I added a test to `curl` that verifies that utf-8 bytes get correctly encoded to JSON; and I had to modify my test to get the utf-8 string from a file instead of from command line arguemnts because the windows CI was not passing the right bytes to `curl`. :/
https://github.com/curl/curl/pull/12434

emanuele6:
I didn't anticipate that adding a generator program would be so complex.
I think this PR is too complex for what it is trying to solve, and creates annoyances when cross-compiling.
It is probably cleaner to just rewrite the `def math: error;` functions in C instead of jq, and just serialise `src/builtin.jq` to a literal C array with `sed` as I mentioned in #1481.

PR #2982, opened 2023-12-11
ci.yml: use `sysctl -n hw.logicalcpu` instead of `nproc` on macos

emanuele6:
The "macos (arm64)" runner on github does not have nproc.

    /Users/runner/work/_temp/f44f0d9f-19eb-4a23-860e-26533d7efdfa.sh: line 10: nproc: command not found

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,7 +160,7 @@ jobs:
             --enable-static \
             --enable-all-static \
             CFLAGS="-O2 -pthread -fstack-protector-all"
-          make -j"$(nproc)"
+          make -j"$(sysctl -n hw.logicalcpu)"
           strip ./jq
           file ./jq
           cp ./jq jq-${{ env.SUFFIX }}


wader (approved):
But it somehow worked anyway?

emanuele6:
`make -j`'s argument is optional in some `make`; e.g. in GNU make if you omit it `-j` it runs jobs in parallel with no limit.

Maybe `macos` uses GNU make?

wader:
Aha it ended up with no stdout output on error

emanuele6:
Yes, it would have not have worked if `macos` was using BSD make for example:
```console
$ bmake -j"$(false)"
bmake: option requires an argument -- j
usage: bmake [-BeikNnqrSstWwX]
            [-C directory] [-D variable] [-d flags] [-f makefile]
            [-I directory] [-J private] [-j max_jobs] [-m directory] [-T file]
            [-V variable] [-v variable] [variable=value] [target ...]
```

PR #2985, opened 2023-12-12
jv_parse: let decNumberFromString/strtod parse complex nans as a NaN

emanuele6:
Before this patch (when using decNumber), "Nan123" was parsed as a NaN,
only if the first n was uppercase.

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -514,7 +514,7 @@ static pfunc check_literal(struct jv_parser* p) {
   case 'f': pattern = "false"; plen = 5; v = jv_false(); break;
   case 'n':
     // if it starts with 'n', it could be a literal "nan"
-    if (p->tokenpos != 3) {
+    if (p->tokenbuf[1] == 'u') {
       pattern = "null"; plen = 4; v = jv_null();
     }
   }
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1938,6 +1938,11 @@ tojson | fromjson
 {"a":nan}
 {"a":null}
 
+# also "nan with payload" #2985
+fromjson | isnan
+"nan1234"
+true
+
 
 # calling input/0, or debug/0 in a test doesn't crash jq
 


emanuele6:
Initially, I also added this test:
```jq
# also "nan with payload" #2985
fromjson | isnan
"nan1234"
true
```

But then I removed it because I realised "nan with payload" is parsed differently on the non-decnumber build (`nan(0x4D2)`), so the test would be decnumber specific.

Now I realise that many other tests in `jq.test` will only work in decnumber builds; we should probably move them to another file like we do for `onig.test`.

wader:
> Now I realise that many other tests in `jq.test` will only work in decnumber builds; we should probably move them to another file like we do for `onig.test`.

Yes probably a good idea to split them out. We do have https://github.com/jqlang/jq/blob/master/.github/workflows/decnum.yml but it seems to not run the standard tests... maybe it should?

src/jv_parse.c
```diff
@@ -514,7 +514,7 @@ static pfunc check_literal(struct jv_parser* p) {
   case 'f': pattern = "false"; plen = 5; v = jv_false(); break;
   case 'n':
     // if it starts with 'n', it could be a literal "nan"
-    if (p->tokenpos != 3) {
+    if (p->tokenbuf[1] == 'u') {
```
wader, line 517: Is this function performance critical or could it be made less tricky with strcmp:s etc? :)

emanuele6: It calls `strcmp` later on the `pattern`

wader:
Should we add some kind of tests also?

emanuele6:
It used to have tests, but I removed it because it would be decNumber specific; I guess I could revert it for now since other tests in `shtest` are decNumber specific

src/jv_parse.c
wader: Aha ok, maybe it was mostly the `'n'` case that was not that obvious what cases it covers

emanuele6: If strcmp fails it shortcircuits to saying it is an invalid literal, I guess it could try to the parse it as a number anyway, but that would change the error message to "Invalid number literal" from "Invalid literal" for `foof` for example.
So maybe let's refactor it later when we change the error for invalid characters

wader:
Yeap i think that would be good. After release we can maybe split out things and also make the decnum action (that builds without decnum support) run tests

src/jv_parse.c
wader: 👍 sorry should have been clearer that i didn't mean it was needed for this PR

PR #2997, opened 2023-12-18
Always build Docker image on CI, group Dependabot PRs for official actions

itchyny:
We should build the docker image to keep the CI workflow working, and just stop before pushing the image to the registry. Otherwise, we don't notice upgrading actions/download-artifact breaks the workflow. Also I improved the Dependabot configuration to group official actions.


wader (approved):

itchyny:
Thank you!

PR #3003, opened 2023-12-23
Add jq 1.7.1 to webpage

wader:
Make it the default download


wader:
Not tested locally yet. Also do we want to add 1.7.1 to https://github.com/jqlang/jq/blob/master/docs/content/manual/manual.yml#L6? and i guess we don't create a new version directory under docs/content for patch releases?

emanuele6 (approved):

PR #3008, opened 2024-01-10
Provide strptime implementation on all systems

george-hopkins:
Some operating systems, notably [Redox OS](https://redox-os.org), do not have an implementation of `strptime` and would benefit from the built-in fallback as well.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -451,7 +451,7 @@ jv jq_util_input_next_input(jq_util_input_state *state) {
   return value;
 }
 
-#if defined(WIN32) && !defined(HAVE_STRPTIME)
+#ifndef HAVE_STRPTIME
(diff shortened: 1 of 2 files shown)


src/util.c
```diff
@@ -451,7 +451,7 @@ jv jq_util_input_next_input(jq_util_input_state *state) {
   return value;
 }
 
-#if defined(WIN32) && !defined(HAVE_STRPTIME)
+#ifndef HAVE_STRPTIME
```
wader, line 454: Was added in https://github.com/jqlang/jq/commit/fdab39bc7b8d41c1ae410f03a42afc10a9322c99

@nicowilliams do you remember if there was some reason for provide replacement strptime only for WIN32?

nicowilliams: There's no reason.

nicowilliams:
Thanks!

PR #3013, opened 2024-01-20
Release dev artifacts too?

nicowilliams:
In the past I've not wanted to commit to the jv/jq APIs, but we've tried to stay backwards compatible for them for long enough, so why not?


nicowilliams:
I bet there's more edits needed here to get the release making to work.

PR #3026, opened 2024-01-30
doc: no need to include . in example

KnorpelSenf:
If I recall correctly, `jq .` was necessary a long time ago in order to process stdin. It is more intuitive to just pipe to `jq`, so I believe this should be the minimal example in the help text.

Would you also like to see a sentence stating that `.` is implied if no arguments are passed?

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -63,7 +63,7 @@ static void usage(int code, int keep_it_short) {
     "The simplest filter is ., which copies jq's input to its output\n"
     "unmodified except for formatting. For more advanced filters see\n"
     "the jq(1) manpage (\"man jq\") and/or https://jqlang.github.io/jq/.\n\n"
-    "Example:\n\n\t$ echo '{\"foo\": 0}' | jq .\n"
+    "Example:\n\n\t$ echo '{\"foo\": 0}' | jq\n"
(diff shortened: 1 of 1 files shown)


nicowilliams:
Allowing the program to be optional in some cases turned out to be a bad idea.  I don't think we should encourage that.  So I think we must reject this PR.

KnorpelSenf:
That's alright, understood. Why is it a bad idea? :)

pkoppstein:
@KnorpelSenf asked:
> Why is it a bad idea?

By most accounts, it's a fine idea but there's another fine idea that (in practice) causes complications of the "feature interaction" variety.  The other idea is that if invoked by itself (without anything on STDIN), jq should emit its help text.  The complications involve implementation details and portability, and achieving agreement on what "user expectations" are or should be assumed to be.

nicowilliams:
@KnorpelSenf the issue is that the rules for when the program can be elided and defaulted to `.` and when it can't aren't easy for users to internalize, so it has caused lots of issues to be opened here.  Thus I would like to rip out this feature.

nicowilliams:
That said, thank you for your contribution!

KnorpelSenf:
Fascinating, I have not run into that yet. Thanks for elaborating!

PR #3036, opened 2024-02-08
minor buglets

UlrichEckhardt:
Just two minor glitches I found accidentally:
- Use `#error` to generate an error, instead of a syntax error.
- Add missing semicolon.

In case you wonder how I found those, I'm trying to convert jq to use [bazel](https://bazel.build) as build system. This is mostly for training, I'm not directly planning to submit that as PR.

Cheers!

Uli

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -210,7 +210,7 @@ static jv f_ ## name(jq_state *jq, jv input, jv a, jv b, jv c) { \
 }
 #define LIBM_DDDD_NO(name) \
 static jv f_ ## name(jq_state *jq, jv input, jv a, jv b, jv c) { \
-  jv_free(c) \
+  jv_free(c); \
   jv_free(b); \
   jv error = jv_string("Error: " #name "/3 not found at build time"); \
   return ret_error2(input, a, error); \
--- a/src/jv_dtoa.c
+++ b/src/jv_dtoa.c
@@ -299,7 +299,7 @@ extern "C" {
 #endif
 
 #if defined(IEEE_8087) + defined(IEEE_MC68k) + defined(VAX) + defined(IBM) != 1
-Exactly one of IEEE_8087, IEEE_MC68k, VAX, or IBM should be defined.
+#  error Exactly one of IEEE_8087, IEEE_MC68k, VAX, or IBM should be defined.
 #endif
 
 typedef union { double d; ULong L[2]; } U;


src/builtin.c
```diff
@@ -210,7 +210,7 @@ static jv f_ ## name(jq_state *jq, jv input, jv a, jv b, jv c) { \
 }
 #define LIBM_DDDD_NO(name) \
 static jv f_ ## name(jq_state *jq, jv input, jv a, jv b, jv c) { \
-  jv_free(c) \
+  jv_free(c); \
```
emanuele6, line 213: This is a "minor buglet"? lol
I think it is rather major; thank you!

src/jv_dtoa.c
```diff
@@ -299,7 +299,7 @@ extern "C" {
 #endif
 
 #if defined(IEEE_8087) + defined(IEEE_MC68k) + defined(VAX) + defined(IBM) != 1
-Exactly one of IEEE_8087, IEEE_MC68k, VAX, or IBM should be defined.
+#  error Exactly one of IEEE_8087, IEEE_MC68k, VAX, or IBM should be defined.
```
emanuele6, line 302: This is a vendored file, but I guess we can change it.
This would be the only directive in the file that uses indentation though

emanuele6:
Thank you!

src/builtin.c
wader: How did this compile? 🤔

emanuele6: @wader This code is only used on systems that don't have one of the math functions with three arguments.
The only math function with 3 arguments that jq supports is [`fma`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fma.html).
I guess all our CI targets have `fma`.

src/jv_dtoa.c
UlrichEckhardt: I'm not sure what you mean with vendored, can you elaborate? In any case, the indentation can be removed, it's something I do without thinking about it.

src/builtin.c
wader: Aha i see, wonder if there is a sane way to do a CI build that disables all conditionally built builtins just to see if compiles at least 🤔

src/jv_dtoa.c
wader: I think he menas that `jv_dota.c` is a copy of the half-famous "dtoa.c" code

UlrichEckhardt: Nevermind, I saw you merged this already. :)

emanuele6: It is a library file that is included in the repository and we redistribute, but is not written by us. (this practice is commonly called "vendoring")

Issue #2868, opened 2023-08-29
`delpaths` sorts the given paths, but the sort is not path-aware and doesn't sort array _slices_ correctly


nicowilliams:
Reported by @emanuele6.

The C function `jv_sort()` is not appropriate for use in `jv_delpaths()` because it doesn't know how to sort array slices correctly:

```C
496 jv jv_delpaths(jv object, jv paths) {
497   if (jv_get_kind(paths) != JV_KIND_ARRAY) {
498     jv_free(object);
499     jv_free(paths);
500     return jv_invalid_with_msg(jv_string("Paths must be specified as an array"));
501   }
502   paths = jv_sort(paths, jv_copy(paths));
503   jv_array_foreach(paths, i, elem) {
  ..
526 }
```

capak07:
Hi, I would like to start working on this issue if it's not assigned to someone, thanks

wader:
@capak07 Hey! go for it, pretty sure no one is working on it

PR #3040, opened 2024-02-12
Update delpaths function to use sort with slicing

capak07:
Pull Request for #2868 
- Added `jv_array_slice` function to support slicing of the array

### DIFF
--- a/src/jv_aux.c
+++ b/src/jv_aux.c
@@ -500,8 +500,10 @@ jv jv_delpaths(jv object, jv paths) {
     jv_free(paths);
     return jv_invalid_with_msg(jv_string("Paths must be specified as an array"));
   }
-  paths = jv_sort(jv_array_slice(paths, 0, jv_array_length(jv_copy(paths))), jv_copy(paths));
-  jv_array_foreach(paths, i, elem) {
+  int paths_length = jv_array_length(jv_copy(paths));
+  paths = jv_sort(jv_array_slice(paths, 0, paths_length), jv_copy(paths));
+  for(int i = 0; i < paths_length; i++) {
+    jv elem = jv_array_get(jv_copy(paths), i);
     if (jv_get_kind(elem) != JV_KIND_ARRAY) {
       jv_free(object);
       jv_free(paths);
@@ -512,7 +514,7 @@ jv jv_delpaths(jv object, jv paths) {
     }
     jv_free(elem);
   }
-  if (jv_array_length(jv_copy(paths)) == 0) {
+  if (paths_length == 0) {
     // nothing is being deleted
     jv_free(paths);
     return object;
@@ -522,7 +524,7 @@ jv jv_delpaths(jv object, jv paths) {
     jv_free(paths);
     jv_free(object);
     return jv_null();
-  }
+  }  
   return delpaths_sorted(object, paths, 0);
 }
 
--- a/src/jv_aux.c
+++ b/src/jv_aux.c
@@ -500,7 +500,7 @@ jv jv_delpaths(jv object, jv paths) {
     jv_free(paths);
     return jv_invalid_with_msg(jv_string("Paths must be specified as an array"));
   }
-  paths = jv_sort(paths, jv_copy(paths));
+  paths = jv_sort(jv_array_slice(paths, 0, jv_array_length(jv_copy(paths))), jv_copy(paths));
   jv_array_foreach(paths, i, elem) {
     if (jv_get_kind(elem) != JV_KIND_ARRAY) {
       jv_free(object);
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1023,6 +1023,10 @@ try delpaths(0) catch .
 {}
 "Paths must be specified as an array"
 
+map(delpaths([[6], [7], [8], [9], [10]]))
+[[0, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5]]
+[[0, 1, 2, 3, 4, 5]]
+
 del(.), del(empty), del((.foo,.bar,.baz) | .[2,3,0]), del(.foo[0], .bar[0], .foo, .baz.bar[0].x)
 {"foo": [0,1,2,3,4], "bar": [0,1]}
 null


src/jv_aux.c
```diff
@@ -500,7 +500,7 @@ jv jv_delpaths(jv object, jv paths) {
     jv_free(paths);
     return jv_invalid_with_msg(jv_string("Paths must be specified as an array"));
   }
-  paths = jv_sort(paths, jv_copy(paths));
+  paths = jv_sort(jv_array_slice(paths, 0, jv_array_length(jv_copy(paths))), jv_copy(paths));
```
emanuele6: The order of evaluation of arguments in a funciton call is unspecified.
You need to use a separate variable to enforce an order of evaluation, or if the compiler decides to call `jv_array_slice(path, ...)` before `jv_copy(path)`, `jv_copy()` will copy something that was already freed.

emanuele6: Also I am not calling `jv_array_slice` here is fine; afaict it modifies the array inplace in some cases instead of returning a new array.

wader:
Can someone give an example when this happens? also some test for it would be good

src/jv_aux.c
capak07: @emanuele6 makes sense. I initially had this block, I think this would fix the potential issue, any thoughts?
```
jv paths_copy = jv_copy(paths);
jv paths_sliced = jv_array_slice(paths_copy, 0, jv_array_length(paths_copy));
jv sorted_paths = jv_sort(paths_sliced, paths_copy);
jv_free(paths);
paths = sorted_paths;
```

emanuele6: Hmm, that is wrong, because you are using the same copy of `paths` (`paths_copy`) three times.
You just need something like this:
```c
  jv slice = jv_array_slice(jv_copy(paths), 0, jv_array_length(jv_copy(paths)))
  paths = jv_sort(slice, paths);
```

emanuele6: I am not sure if you actually understood the problem, so I will try to explain it more clearly:

In C, if you have:
```c
  c = function(A(), B());
```
It is unspecified whether that gets executed as:
```c
  {
    jv a = A();
    jv b = B();
    c = function(a, b);
  }
```
or as:
```c
  {
    jv b = B();
    jv a = A();
    c = function(a, b);
  }
```

In this case, you have:
```c
  paths = jv_sort(jv_array_slice(paths, 0, jv_array_length(jv_copy(paths))), jv_copy(paths));
```

If the compiler generates code that runs this as:
```c
  {
    jv copy = jv_copy(paths);
    jv slice = jv_array_slice(paths, 0, jv_array_length(jv_copy(paths)));
    paths = jv_sort(slice, copy);
  }
```
Then it is fine because you are not using a value that has already been freed anywhere.

But if the compiler generates code that runs it is as:
```c
  {
    jv slice = jv_array_slice(paths, 0, jv_array_length(jv_copy(paths)));
    jv copy = jv_copy(paths);
    paths = jv_sort(slice, copy);
  }
```
Then it is not fine because `jv_array_slice(paths,...)` will have freed `paths`, and `jv_copy(paths)` will try to copy `paths` after it has been freed.

So you need to use an assignment to enforce the correct order e.g. using the code I posted in the previous comment, or with the code that uses two variables that I posted in this correct, or even with something like this:
```c
  jv copy = jv_copy(paths);
  paths = jv_sort(jv_array_slice(paths, 0, jv_array_length(jv_copy(paths))), copy);
```
You just need to somehow make sure the functions get called in the right sequence, so that `paths` does not get accidentally freed before being copied

emanuele6:
@wader Well, I checked my 2023-08-29 chat logs with @nicowilliams and that issue was about the following `jq` script (that is supposed to delete duplicate elements in the input array without sorting):
```jq
[6,5,1,2,5,3,4,6,7,6,5] |
del(
  . as $out |
  range(length) as $i
  .[$i+1:][] |
  select(. == $dot[$i])
)
```
Not working as intended because of its use of slices; output:
```console
$ jq -cn '[6,5,1,2,5,3,4,6,7,6,5] | del(. as $dot | range(length) as $i | .[$i+1:][] | select(. == $dot[$i]))'
[6,5,1,2,3,4,7,5]
$ jq -cn '[6,5,1,2,5,3,4,6,7,6,5] | delpaths([path(. as $dot | range(length) as $i | .[$i+1:][] | select(. == $dot[$i])) | debug])'
["DEBUG:",[{"start":1,"end":null},6]]
["DEBUG:",[{"start":1,"end":null},8]]
["DEBUG:",[{"start":2,"end":null},2]]
["DEBUG:",[{"start":2,"end":null},8]]
["DEBUG:",[{"start":5,"end":null},5]]
["DEBUG:",[{"start":8,"end":null},1]]
[6,5,1,2,3,4,7,5]
```
(The last duplicate `5` did not get deleted)

This is the usual problem that afflicts path expressions with negative indices and slices; the slice paths do not get canonicalised (or they are not sorted in an order that would make the deletions work correctly), so they end up pointing the wrong elements.

A version of that same script that uses a range to loop the array instead of a slice works correctly:
```jq
[6,5,1,2,5,3,4,6,7,6,5] |
del(
  . as $dot |
  range(length) as $i |
  .[range($i + 1; length)] |
  select(. == $dot[$i])
)
```
output:
```console
$ jq -cn '[6,5,1,2,5,3,4,6,7,6,5] | del(. as $dot | range(length) as $i | .[range($i + 1; length)] | select(. == $dot[$i]))'
[6,5,1,2,3,4,7]
$ jq -cn '[6,5,1,2,5,3,4,6,7,6,5] | delpaths([ path(. as $dot | range(length) as $i | .[range($i + 1; length)] | select(. == $dot[$i])) | debug ])'
["DEBUG:",[7]]
["DEBUG:",[9]]
["DEBUG:",[4]]
["DEBUG:",[10]]
["DEBUG:",[10]]
["DEBUG:",[9]]
[6,5,1,2,3,4,7]
```

---

I don't really understand this patch, so I am not sure what @capak07 had in mind with it. This does not solve the issue #2868 was created for, but maybe they are trying to solve a different bug?

capak07:
@emanuele6 I apologize for the oversight, the description for the items said "it doesn't know how to sort array slices correctly"
based on your suggestion
I have these changes
```
 paths = jv_sort(paths, jv_copy(paths));
  for (int i = 0; i < jv_array_length(jv_copy(paths)); i++) {
    jv elem = jv_array_get(jv_copy(paths), i);
```
 I replaced the jv_array_foreach loop with a traditional for loop that iterates over the length of the array.
 Any thoughts on this?

emanuele6:
I don't understand what you mean

src/jv_aux.c
```diff
@@ -500,8 +500,10 @@ jv jv_delpaths(jv object, jv paths) {
     jv_free(paths);
     return jv_invalid_with_msg(jv_string("Paths must be specified as an array"));
   }
-  paths = jv_sort(paths, jv_copy(paths));
-  jv_array_foreach(paths, i, elem) {
+  int paths_length = jv_array_length(jv_copy(paths));
+  paths = jv_sort(jv_array_slice(paths, 0, paths_length), jv_copy(paths));
```
nicowilliams, line 504: This can't be the fix, right?  I mean, the problem is that we need to canonicalize array indices in `paths` first, no?

nicowilliams: `jv_array_slice(a, 0, length_of_a)` should be a no-op, too.

emanuele6, line 504: What are you doing? This still has the order of operation problem; if it gets evaluated like this you have a use after free:
```c
  {
    jv a = jv_array_slice(paths, 0, paths_length);
    jv b = jv_copy(paths);
    paths = jv_sort(a, b);
  }
```

emanuele6: I don't understand what this patch is trying to do either

capak07: @nicowilliams 
is this what you're suggesting?
```
jv canonicalize_indices(jv paths) {
  int paths_length = jv_array_length(jv_copy(paths));
  for (int i = 0; i < paths_length; i++) {
    jv path = jv_array_get(jv_copy(paths), i);
    int index = jv_number_value(path);
    if (index < 0) {
      index = paths_length + index;
    }
    if (index < 0 || index >= paths_length) {
      jv_free(paths);
      return jv_invalid_with_msg(jv_string("Invalid array index"));
    }
    paths = jv_array_set(paths, i, jv_number(index));
  }
  return paths;
}
```

nicowilliams: To canonicalize the paths we need to refer to the tree they refer to.  The idea is to take any negative indices in the paths and normalize them to positive indices, but to do this you need to know the length of the corresponding array in the tree for that negative index in the path.

emanuele6: `if (index < 0) index = paths_length + index;`? Why would you want to do that? What bug did you solve with those patches?

capak07: Okay so to convert a negative index into a positive index, we need the length of the subtree, this could work then
```
jv canonicalize_indices(jv paths, jv tree) {
  int paths_length = jv_array_length(jv_copy(paths));
  for (int i = 0; i < paths_length; i++) {
    jv path = jv_array_get(jv_copy(paths), i);
    jv subtree = tree;
    int path_length = jv_array_length(jv_copy(path));
    for (int j = 0; j < path_length; j++) {
      jv step = jv_array_get(jv_copy(path), j);
      if (jv_is_number(step)) {
        int index = jv_number_value(step);
        if (index < 0) {
          int subtree_length = jv_array_length(jv_copy(subtree));
          index = subtree_length + index;
          if (index < 0) {
            // Invalid index, handle error
            jv_free(paths);
            return jv_invalid_with_msg(jv_string("Invalid array index"));
          }
          path = jv_array_set(path, j, jv_number(index));
        }
        subtree = jv_array_get(jv_copy(subtree), index);
      } else if (jv_is_string(step)) {
        subtree = jv_object_get(jv_copy(subtree), step);
      } else {
        // Invalid path step, handle error
        jv_free(paths);
        return jv_invalid_with_msg(jv_string("Invalid path step"));
      }
    }
    paths = jv_array_set(paths, i, path);
  }
  return paths;
}
```

emanuele6: With canonicalising a path, we mean converting:
```json
["foo",{"start":10,"end":null},{"start":-200,"end":-2},20]
["foo",-20,"hello"]
```
To (if foo is an array with 120 elements)
```json
["foo",30]
["foo",100,"hello"]
```
You need to know the input to do that; if foo were an array with 300 elements you would want this instead:
```json
["foo",120]
["foo",280,"hello"]
```

nicowilliams: Yeah, it's a bit messy.  Because path elements could overlap because of the slice specifiers, we might have to explode `{"start":...,"end":...}` into the individual indices that match those slice specifiers.

@capak07 the context here is that internally a "slice" specifier like `$something[start:end]` is represented as an object with a `"start"` and `"end"` keys, and that the start and end indices can be negative to mean "starting from the end of this string or array".  Here `jv_sort()` just can't possibly do the right thing.  We should either have a `jv_sort_paths()` that does, or we should normalize the `paths` to be given to `jv_sort()`, and the latter seems like the better path forward for now because the former will likely require that we use `qsort_r()` (which isn't portable, so we might have to import one from a BSD).

emanuele6: You should not reject object indices (they are slices), and `del(.[100])` on a short array should not be an error, so you should not start throwing an invalid path error

tests/jq.test
```diff
@@ -1023,6 +1023,10 @@ try delpaths(0) catch .
 {}
 "Paths must be specified as an array"
 
+map(delpaths([[6], [7], [8], [9], [10]]))
```
emanuele6, line 1026: This case already works

emanuele6:
Hmm, I am closing this PR since it seems the patch does not solve any issue, and otherwise it does not accomplish anything useful.

nicowilliams:
This is a tricky bug to fix.

PR #3044, opened 2024-02-18
implement borrowing

MaxBrandtner:
This commit adds borrowing to the jv c-api

As discussed this adds a borrowed flag in the jv struct.

Adds the following functions
`jv jv_borrow(jv)`
`jv jv_is_borrowed(jv)`

Changes jv_copy() and jv_free() to return early if the jv is borrowed

still need to add documentation for the two functions

### DIFF
--- a/src/jv.c
+++ b/src/jv.c
@@ -150,11 +150,7 @@ jv jv_invalid_with_msg(jv err) {
   jvp_invalid* i = jv_mem_alloc(sizeof(jvp_invalid));
   i->refcnt = JV_REFCNT_INIT;
 
-  if(jv_is_borrowed(err)){
-    i->errmsg = jv_unborrow(err);
-  }else{
-    i->errmsg = err;
-  }
+  i->errmsg = jv_return(err);
 
   jv x = {JVP_FLAGS_INVALID_MSG, 0, 0, 0, 0, {&i->refcnt}};
   return x;
@@ -1002,13 +998,8 @@ jv jv_array_set(jv j, int idx, jv val) {
   // copy/free of val,j coalesced
   jv* slot = jvp_array_write(&j, idx);
   jv_free(*slot);
-  if(jv_is_borrowed(val)){
-    *slot = jv_unborrow(val);
-  }else{
-    *slot = val;
-  }
-  if(jv_is_borrowed(j)) return jv_unborrow(j);
-  return j;
+  *slot = jv_return(val);
+  return jv_return(j);
 }
 
 jv jv_array_append(jv j, jv val) {
@@ -1433,19 +1424,15 @@ jv jv_string_slice(jv j, int start, int end) {
 }
 
 jv jv_string_concat(jv a, jv b) {
-  a = jv_return(a);
-  b = jv_return(b);
-  a = jvp_string_append(a, jv_string_value(b),
+  a = jvp_string_append(jv_return(a), jv_string_value(b),
                         jvp_string_length(jvp_string_ptr(b)));
   jv_free(b);
   return a;
 }
 
-jv jv_string_append_buf(jv input, const char* buf, int len) {
-  jv a = jv_return(input);
-
+jv jv_string_append_buf(jv a, const char* buf, int len) {
   if (jvp_utf8_is_valid(buf, buf+len)) {
-    a = jvp_string_append(a, buf, len);
+    a = jvp_string_append(jv_return(a), buf, len);
   } else {
     jv b = jvp_string_copy_replace_bad(buf, len);
     a = jv_string_concat(a, b);
@@ -1453,12 +1440,10 @@ jv jv_string_append_buf(jv input, const char* buf, int len) {
   return a;
 }
 
-jv jv_string_append_codepoint(jv input, uint32_t c) {
-  jv a = jv_return(input);
-
+jv jv_string_append_codepoint(jv a, uint32_t c) {
   char buf[5];
   int len = jvp_utf8_encode(c, buf);
-  a = jvp_string_append(a, buf, len);
+  a = jvp_string_append(jv_return(a), buf, len);
   return a;
 }
 
@@ -1798,11 +1783,10 @@ jv jv_object_set(jv input, jv key, jv value) {
 jv jv_object_delete(jv input, jv key) {
   assert(JVP_HAS_KIND(input, JV_KIND_OBJECT));
   assert(JVP_HAS_KIND(key, JV_KIND_STRING));
-  jv object = jv_unborrow(input);
-  jv_free(input);
-  jvp_object_delete(&object, key);
+  input = jv_return(input);
+  jvp_object_delete(&input, key);
   jv_free(key);
-  return object;
+  return input;
 }
 
 int jv_object_length(jv object) {
--- a/src/jv.c
+++ b/src/jv.c
@@ -164,7 +164,7 @@ jv jv_invalid_get_msg(jv inv) {
 
   jv x;
   if (JVP_HAS_FLAGS(inv, JVP_FLAGS_INVALID_MSG)) {
-    x = jv_copy(((jvp_invalid*)inv.u.ptr)->errmsg);
+    x = jv_unborrow(((jvp_invalid*)inv.u.ptr)->errmsg);
   }
   else {
     x = jv_null();
@@ -862,7 +862,7 @@ static jv* jvp_array_write(jv* a, int i) {
     int j;
     for (j = 0; j < jvp_array_length(*a); j++) {
       new_array->elements[j] =
-        jv_copy(array->elements[j + jvp_array_offset(*a)]);
+        jv_unborrow(array->elements[j + jvp_array_offset(*a)]);
     }
     for (; j < new_length; j++) {
       new_array->elements[j] = JV_NULL;
@@ -974,7 +974,7 @@ jv jv_array_get(jv j, int idx) {
   jv* slot = jvp_array_read(j, idx);
   jv val;
   if (slot) {
-    val = jv_copy(*slot);
+    val = jv_unborrow(*slot);
   } else {
     val = jv_invalid();
   }
@@ -1636,8 +1636,8 @@ static jv jvp_object_unshare(jv object) {
     struct object_slot* new_slot = jvp_object_get_slot(new_object, i);
     *new_slot = *old_slot;
     if (jv_get_kind(old_slot->string) != JV_KIND_NULL) {
-      new_slot->string = jv_copy(old_slot->string);
-      new_slot->value = jv_copy(old_slot->value);
+      new_slot->string = jv_unborrow(old_slot->string);
+      new_slot->value = jv_unborrow(old_slot->value);
     }
   }
 
--- a/src/jv.c
+++ b/src/jv.c
@@ -1433,6 +1433,8 @@ jv jv_string_slice(jv j, int start, int end) {
 }
 
 jv jv_string_concat(jv a, jv b) {
+  a = jv_return(a);
+  b = jv_return(b);
   a = jvp_string_append(a, jv_string_value(b),
                         jvp_string_length(jvp_string_ptr(b)));
   jv_free(b);
@@ -1785,17 +1787,12 @@ jv jv_object_set(jv input, jv key, jv value) {
   assert(JVP_HAS_KIND(input, JV_KIND_OBJECT));
   assert(JVP_HAS_KIND(key, JV_KIND_STRING));
 
-  jv object = jv_unborrow(input);
-  jv_free(input);
+  input = jv_return(input);
   // copy/free of object, key, value coalesced
-  jv* slot = jvp_object_write(&object, key);
+  jv* slot = jvp_object_write(&input, jv_return(key));
   jv_free(*slot);
-  if(jv_is_borrowed(value)){
-    *slot = jv_unborrow(value);
-  }else{
-     *slot = value;
-  }
-  return object;
+  *slot = jv_return(value);
+  return input;
 }
 
 jv jv_object_delete(jv input, jv key) {
--- a/src/jv.c
+++ b/src/jv.c
@@ -995,7 +995,11 @@ jv jv_array_set(jv j, int idx, jv val) {
   // copy/free of val,j coalesced
   jv* slot = jvp_array_write(&j, idx);
   jv_free(*slot);
-  *slot = val;
+  if(jv_is_borrowed(val)){
+    *slot = jv_unborrow(val);
+  }else{
+    *slot = val;
+  }
   if(jv_is_borrowed(j)) return jv_unborrow(j);
   return j;
 }
@@ -1774,7 +1778,11 @@ jv jv_object_set(jv object, jv key, jv value) {
   // copy/free of object, key, value coalesced
   jv* slot = jvp_object_write(&object, key);
   jv_free(*slot);
-  *slot = value;
+  if(jv_is_borrowed(value)){
+    *slot = jv_unborrow(value);
+  }else{
+     *slot = value;
+  }
   if(jv_is_borrowed(object)) return jv_unborrow(object);
   return object;
 }
--- a/src/jv.c
+++ b/src/jv.c
@@ -114,10 +114,10 @@ const char* jv_kind_name(jv_kind k) {
   return "<unknown>";
 }
 
-const jv JV_NULL = {JVP_FLAGS_NULL, 0, 0, 0, {0}};
-const jv JV_INVALID = {JVP_FLAGS_INVALID, 0, 0, 0, {0}};
-const jv JV_FALSE = {JVP_FLAGS_FALSE, 0, 0, 0, {0}};
-const jv JV_TRUE = {JVP_FLAGS_TRUE, 0, 0, 0, {0}};
+const jv JV_NULL = {JVP_FLAGS_NULL, 0, 0, 0, 0, {0}};
+const jv JV_INVALID = {JVP_FLAGS_INVALID, 0, 0, 0, 0, {0}};
+const jv JV_FALSE = {JVP_FLAGS_FALSE, 0, 0, 0, 0, {0}};
+const jv JV_TRUE = {JVP_FLAGS_TRUE, 0, 0, 0, 0, {0}};
 
 jv jv_true() {
   return JV_TRUE;
@@ -151,7 +151,7 @@ jv jv_invalid_with_msg(jv err) {
   i->refcnt = JV_REFCNT_INIT;
   i->errmsg = err;
 
-  jv x = {JVP_FLAGS_INVALID_MSG, 0, 0, 0, {&i->refcnt}};
+  jv x = {JVP_FLAGS_INVALID_MSG, 0, 0, 0, 0, {&i->refcnt}};
   return x;
 }
 
@@ -590,7 +590,7 @@ static jv jvp_literal_number_new(const char * literal) {
     return JV_INVALID;
   }
 
-  jv r = {JVP_FLAGS_NUMBER_LITERAL, 0, 0, JV_NUMBER_SIZE_INIT, {&n->refcnt}};
+  jv r = {JVP_FLAGS_NUMBER_LITERAL, 0, 0, 0, JV_NUMBER_SIZE_INIT, {&n->refcnt}};
   return r;
 }
 
@@ -674,7 +674,7 @@ jv jv_number(double x) {
 #else
     JV_KIND_NUMBER,
 #endif
-    0, 0, 0, {.number = x}
+    0, 0, 0, 0, {.number = x}
   };
   return j;
 }
@@ -806,7 +806,7 @@ static jvp_array* jvp_array_alloc(unsigned size) {
 }
 
 static jv jvp_array_new(unsigned size) {
-  jv r = {JVP_FLAGS_ARRAY, 0, 0, 0, {&jvp_array_alloc(size)->refcnt}};
+  jv r = {JVP_FLAGS_ARRAY, 0, 0, 0, 0, {&jvp_array_alloc(size)->refcnt}};
   return r;
 }
 
@@ -869,7 +869,7 @@ static jv* jvp_array_write(jv* a, int i) {
     }
     new_array->length = new_length;
     jvp_array_free(*a);
-    jv new_jv = {JVP_FLAGS_ARRAY, 0, 0, new_length, {&new_array->refcnt}};
+    jv new_jv = {JVP_FLAGS_ARRAY, 0, 0, 0, new_length, {&new_array->refcnt}};
     *a = new_jv;
     return &new_array->elements[i];
   }
@@ -1091,7 +1091,7 @@ static jv jvp_string_copy_replace_bad(const char* data, uint32_t length) {
   length = out - s->data;
   s->data[length] = 0;
   s->length_hashed = length << 1;
-  jv r = {JVP_FLAGS_STRING, 0, 0, 0, {&s->refcnt}};
+  jv r = {JVP_FLAGS_STRING, 0, 0, 0, 0, {&s->refcnt}};
   return r;
 }
 
@@ -1102,7 +1102,7 @@ static jv jvp_string_new(const char* data, uint32_t length) {
   if (data != NULL)
     memcpy(s->data, data, length);
   s->data[length] = 0;
-  jv r = {JVP_FLAGS_STRING, 0, 0, 0, {&s->refcnt}};
+  jv r = {JVP_FLAGS_STRING, 0, 0, 0, 0, {&s->refcnt}};
   return r;
 }
 
@@ -1110,7 +1110,7 @@ static jv jvp_string_empty_new(uint32_t length) {
   jvp_string* s = jvp_string_alloc(length);
   s->length_hashed = 0;
   memset(s->data, 0, length);
-  jv r = {JVP_FLAGS_STRING, 0, 0, 0, {&s->refcnt}};
+  jv r = {JVP_FLAGS_STRING, 0, 0, 0, 0, {&s->refcnt}};
   return r;
 }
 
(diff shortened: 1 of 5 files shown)


MaxBrandtner:
The documentation is now already upstreamed(seeing as you can apparently not create a pr on the docs, but edit them directly)

Would still be nice if someone reviewed it to check that there are no issues with it

src/jv.h
```diff
@@ -33,7 +33,8 @@ struct jv_refcnt;
    Really. Do not play with them. */
 typedef struct {
```
emanuele6, line 34: You cannot use bitfields in a public struct; their order is not defined; the compiler can order however it wants.

src/jv.h
```diff
@@ -51,8 +52,13 @@ jv_kind jv_get_kind(jv);
 const char* jv_kind_name(jv_kind);
 static int jv_is_valid(jv x) { return jv_get_kind(x) != JV_KIND_INVALID; }
 
-jv jv_copy(jv);
-void jv_free(jv);
+void jv__free(jv);
+jv jv__copy(jv);
+
+static inline jv jv_copy(jv a){if(a.borrowed) return a; return jv__copy(a);}
+static inline void jv_free(jv a){if(!a.borrowed) jv__free(a);}
```
emanuele6, line 59: Making those functions static wrechs the api, now the `jv_copy` and `jv_free` symbols don't exist anymore.
Also likely to be broken since a `static` function imported into different compilation units is accessing a bitfield without calling `libjq` to let it do it.

src/jv.h
emanuele6: If `libjq` and the program that uses `libjq` are compiled with different compilers there are no guarantees of what bits will be used, so this is likely to not work.
Even if they are compiled with the same compiler, it may still not work because the order of bit-fields needs to only be consistent in individual compilation units; the compiler may feel like ordering them differently for optimisation reasons.

At most you can have a public struct with bit-fields, and required that the user does not touch those fields manually, so they are only used by `libjq` functions, but that is an error-prone, and confusing way to define the layout of a public struct in my opinion. (Also you would have to make sure that all the functions that use those bit-fields are in the same compilation unit.)
And in this case it is not possible since there are `static` functions defined in `jv.h` that access bit-fields directly without calling to `libjq`.

Bit-fields don't just subdivide the bits of a struct in the specified order, like regular fields do for bytes: they are different, and their order is not enforced or stable. Only really suitable for internal structs. They are not a very useful C feature because of this.
The only way to do something similar to this with a public struct is with manual masking, maybe facilitated by macros.

emanuele6: I noticed that all compilers I tried were suspiciously all using the same layout for bit-fields.
I just learned that apparently the SysV ABI (3.1.2; for amd64) specifies how bit-fields must be layed out in memory for public structs; they need to be layed out in order like for regular fields dividing the byte from most significant bit to least significant bit. Interesting!
Still you probably should avoid using bit fields in a public struct since standard C does not support them, and `libjq` does not only target `amd64` on UNIX-like systems (I didn't bother to check whether other standard ABIs support bit-fields in structs).

nicowilliams: > You cannot use bitfields in a public struct; their order is not defined; the compiler can order however it wants.

C99 says, § 6.7.2.1, constraint 10:

> 10. An implementation may allocate any addressable storage unit large enough to hold a bit-
field. If enough space remains, a bit-field that immediately follows another bit-field in a
structure shall be packed into adjacent bits of the same unit. If insufficient space remains,
whether a bit-field that does not fit is put into the next unit or overlaps adjacent units is
implementation-defined. The order of allocation of bit-fields within a unit (high-order to
low-order or low-order to high-order) is implementation-defined. The alignment of the
addressable storage unit is unspecified.

So, yes, we can't have bit-fields in public structures when either those bits are meant to be accessed by the application or by `inline` functions or macros exported by the header.  But, as you note, _platforms_ often add their own specifications around some of these issues in their ABIs.  SysV, for example, does, and so does [Windows](https://learn.microsoft.com/en-us/cpp/c-language/c-bit-fields?view=msvc-170):

> Bit fields defined as `int` are treated as `signed`. A Microsoft extension to the ANSI C standard allows `char` and `long` types (both `signed` and `unsigned`) for bit fields. Unnamed bit fields with base type `long`, `short`, or `char` (`signed` or `unsigned`) force alignment to a boundary appropriate to the base type.

> Bit fields are allocated within an integer from least-significant to most-significant bit.

> The Windows ABI convention packs bit fields into single storage integers, and doesn't straddle storage units.

Which takes care of all concerns here for _Windows_.  I believe we'll find the same is true for Solaris/Illumos, Linux, and OS X.  I believe all of those follow SVID to some degree in their ABIs, except maybe OS X.

That said, I've no problem with using a bit of `pad_` w/o using bit fields just in case.

src/jv.h
nicowilliams: Making these `static inline` was my suggestion.  We can still export the original `jv_copy()`/`jv_free()` from libjq using a bit of pre-processor conditionals.

src/jv.h
```diff
@@ -70,6 +76,8 @@ jv jv_true(void);
 jv jv_false(void);
 jv jv_bool(int);
 
+static inline jv jv_is_borrowed(jv a){if(a.borrowed) return jv_true(); return jv_false();}
```
nicowilliams: Does anything need this function?  I think probably not.

nicowilliams:
The commit needs a longer message showing what is the intended use of this.

nicowilliams:
Thanks by the way.  This will make `jv` functions more ergonomic in out-of-tree applications.

src/jv.h
emanuele6: Well, but what about non-amd64 architectures?

Anyway, those bit-fields should be replaced with masking with get/set macros, yes.
Also, I suggest using the most signficant bit for the boolean flag and the least significant bits for the integer instead of the other way around.

nicowilliams:
Hmm, this feature should be awesome, but we need to be very careful.  Consider this:

```C
  o = jv_object_set(jv_copy(o), jv_copy(k), jv_copy(v));
```

What happens if all of `o`, `k`, and `v` are borrowed and have reference count == 1?

Well, we do get a _new_ object, so if after the borrow is undone the original `o` is `jv_free()`ed elsewhere, that's not so bad, but if you look at `jvp_object_unshare()` you'll see that we do additional `jv_copy()`s there, and so...

I strongly suspect that this just isn't going to work unless we use a variant of `jv_copy()`/`jv_free()` that ignores the `borrowed` flag and use those judiciously.  Basically, whenever we take a reference but don't actually store the referenced copy anywhere that will outlive the local scope, then we can use the modified `jv_copy()`/`jv_free()`, but whenever we'll store the "copy" in any slot that will outlive the current local scope then we must actually increment the reference count.  And that brings us to:

 - this PR needs a test
 - and changes to `jv_object_set()`/`jv_object_delete()` (or the corresponding `jvp_object_*()` functions), and `jv_array_set()` to make this safe

src/jv.h
MaxBrandtner: while I can't think of much of a usecase I implemented it just in case

MaxBrandtner:
> I strongly suspect that this just isn't going to work unless we use a variant of `jv_copy()`/`jv_free()` that ignores the `borrowed` flag and use those judiciously. Basically, whenever we take a reference but don't actually store the referenced copy anywhere that will outlive the local scope, then we can use the modified `jv_copy()`/`jv_free()`, but whenever we'll store the "copy" in any slot that will outlive the current local scope then we must actually increment the reference count. And that brings us to:

I guess jv__copy() could be used in such cases, but maybe it would be better to rename the original jv__copy() function to something more descriptive

Should I add a jv_unborrow() function that increments the reference count and returns a jv with the borrow flag unset? This would behave like jv_copy() only that it also resolves the issue regarding borrowed elements being added to an unborrowed jv

nicowilliams:
> Should I add a jv_unborrow() function that increments the reference count and returns a jv with the borrow flag unset? This would behave like jv_copy() only that it also resolves the issue regarding borrowed elements being added to an unborrowed jv

Yes, I think that's a good name.  The rule would be that any time you want a dynamic scope reference you use `jv_copy()`, and any time you want an indefinite lifetime reference you use `jv_unborrow()`.

MaxBrandtner:
@nicowilliams would the borrow issue just apply to jv_object_set() or also to other functions

MaxBrandtner:
I did not find jv_copy() in jv_object_set()

Here are all the functions that return jv or some form of jvp that contain jv_copy() where it could be relevant

```c
static jv jvp_object_unshare(jv); 
jv jv_array_get(jv, int);
static jv* jvp_array_write(jv*, int);
jv jv_invalid_get_msg(jv);
```

MaxBrandtner:
are there any functions outside jv.c that need jv_unborrow?

nicowilliams:
> @nicowilliams would the borrow issue just apply to jv_object_set() or also to other functions

It would apply to all container mutation functions.  There's only arrays and objects as containers.  You should arrange that the values stored by `jv_object_set()` and `jv_array_set()` are always non-borrowed, which means they'll automatically always be "copied" when you `jv_object_get()`, `jv_array_get()`, or iterate them.

nicowilliams:
> are there any functions outside jv.c that need jv_unborrow?

Until we start making use of `jv_borrow()` in `src/execute.c`, then no, I don't think so.  If we start making use of `jv_borrow()` in `src/execute.c` then everything to do with setting `$binding`s in the VM will have to unborrow.

nicowilliams:
> I did not find jv_copy() in jv_object_set()
> 
> Here are all the functions that return jv or some form of jvp that contain jv_copy() where it could be relevant
> 
> ```c
> static jv jvp_object_unshare(jv); 
> jv jv_array_get(jv, int);
> static jv* jvp_array_write(jv*, int);
> jv jv_invalid_get_msg(jv);
> ```

See earlier comment.  As long as you store unborrowed values in containers, then you don't have to alter the container read/iteration paths.

(Ah, yes, of course, `jv_invalid_with_msg()` is a container too, albeit of just one value.)

MaxBrandtner:
what edge-cases do you think are relevant to test for in the test. Of course we should test the main use-case(skipping jv_copy() and jv_free()), as well as perhaps jv_object_set() and jv_array_set(). Is there anything I'm missing?

src/jv.h
MaxBrandtner: Also it should exist for the sake of consistency as jv_get_refcount() also exists

nicowilliams:
> what edge-cases do you think are relevant to test for in the test. Of course we should test the main use-case(skipping jv_copy() and jv_free()), as well as perhaps jv_object_set() and jv_array_set(). Is there anything I'm missing?

I think you want to write a test where you do build objects/arrays/invalids-with-messages with keys/values that are borrowed, then later free everything, and when run with ASAN/valgrind this should yield no errors.  You should also pass borrowed values to other `jv_*()` functions, like `jv_sort()`, `jv_cmp()`, `jv_array/object_get()`, etc, and again, there should be no memory errors.

MaxBrandtner:
I think the following functions should be checked

```c
jv_sort()
jv_cmp() //unsure
JV_ARRAY()
jv_invalid_get_msg() //unsure
jv_invalid_with_msg() //unsure
jv_equal()
jv_getpath()
jv_keys()
jv_setpath()
jv_array_append()
jv_array_concat()
jv_array_get()
jv_array_length() //unsure
jv_object_get()
```

Are there any other functions (particularly jv_array_ / object functions that should be tested)?

MaxBrandtner:
`jv_array_concat()` and `jv_array_append()` produce borrowed outputs. `jv_setpath()` produces memory errors

MaxBrandtner:
well it seem as if to integrate `jv_borrow()` instead of just returning the input as the output(and basically assuming the input to be consumed), at any point were that occurs jv_unborrow() needs to be called if the jv is borrowed and return the value directly otherwise. 

Or jv_unborrow() could just consume its input. Meaning wherever in the source code there is a function like this
```c
jv func(jv i){
  //modify i
  return i;
}
```

`return jv_unborrow(i);` should be called instead

MaxBrandtner:
So basically it would look like this: `static inline jv jv_unborrow(jv a){if(!a.borrowed) return a; a.borrowed = 0; return jv_copy(a);}`

(Note: copying a borrowed jv is still required as otherwise a borrowed jv could be freed)

nicowilliams:
> So basically it would look like this: `static inline jv jv_unborrow(jv a){if(!a.borrowed) return a; a.borrowed = 0; return jv_copy(a);}`

A function that will be "modifying" a given `jv` value, if the `jv` has `refcnt > 1 || borrowed`, then it must make a copy of that `jv` value.

MaxBrandtner:
modifying `jv_unborrow()` to not be a form of `jv_copy()` will cause issues in other places

nicowilliams:
> modifying `jv_unborrow()` to not be a form of `jv_copy()` will cause issues in other places

Basically you need two operations: one for the `jv` that will be "modified" by `jv_array_set()` or `jv_object_set()`, and one for the `jv` that will be stored by `jv_array_set()`, `jv_object_set()`, and `jv_invalid_with_msg()`.

MaxBrandtner:
This should work now.
The only remaining issue *should* be that a `jv_return()` function ought to be added I think instead of specifying that all external functions that return a modified input need to add the following logic

**currently**
```c
jv function(jv i){
 //modify i
 if(jv_is_borrowed(i)) return jv_unborrow(i);
 return i;
}
```

**with jv_return**
```c
jv function(jv i){
   //modify i
   return jv_return(i);
}
```

Furthermore, the `jv_return()` pattern is also used in a lot of places inside of jv.

MaxBrandtner:
And the docs need to be improved regarding borrowing

src/jv.c
```diff
@@ -164,7 +164,7 @@ jv jv_invalid_get_msg(jv inv) {
 
   jv x;
   if (JVP_HAS_FLAGS(inv, JVP_FLAGS_INVALID_MSG)) {
-    x = jv_copy(((jvp_invalid*)inv.u.ptr)->errmsg);
+    x = jv_unborrow(((jvp_invalid*)inv.u.ptr)->errmsg);
```
nicowilliams: This should not be needed.  `jv`s contained in others should never be borrowed, I'd think.  They need to be owned.

src/jv.c
```diff
@@ -862,14 +862,14 @@ static jv* jvp_array_write(jv* a, int i) {
     int j;
     for (j = 0; j < jvp_array_length(*a); j++) {
       new_array->elements[j] =
-        jv_copy(array->elements[j + jvp_array_offset(*a)]);
+        jv_unborrow(array->elements[j + jvp_array_offset(*a)]);
```
nicowilliams: Ditto.

src/jv.c
```diff
@@ -974,7 +974,7 @@ jv jv_array_get(jv j, int idx) {
   jv* slot = jvp_array_read(j, idx);
   jv val;
   if (slot) {
-    val = jv_copy(*slot);
+    val = jv_unborrow(*slot);
```
nicowilliams: Ditto.

src/jv.c
```diff
@@ -995,7 +995,12 @@ jv jv_array_set(jv j, int idx, jv val) {
   // copy/free of val,j coalesced
   jv* slot = jvp_array_write(&j, idx);
   jv_free(*slot);
-  *slot = val;
+  if(jv_is_borrowed(val)){
+    *slot = jv_unborrow(val);
```
nicowilliams: Aha, see, the elements of arrays aren't borrowed, so no need to unborrow when reading them.

src/jv.c
nicowilliams: Need to check that `jvp_refcnt_unshared()` returns false when the `jv` is borrowed...

src/jv.c
```diff
@@ -1013,6 +1018,7 @@ jv jv_array_concat(jv a, jv b) {
     a = jv_array_append(a, elem);
   }
   jv_free(b);
+  if(jv_is_borrowed(a)) return jv_unborrow(a);
```
nicowilliams: Superfluous.  The result of `jv_array_set()` on a borrowed array should always be unborrowed, right?

src/jv.c
```diff
@@ -1636,8 +1642,8 @@ static jv jvp_object_unshare(jv object) {
     struct object_slot* new_slot = jvp_object_get_slot(new_object, i);
     *new_slot = *old_slot;
     if (jv_get_kind(old_slot->string) != JV_KIND_NULL) {
-      new_slot->string = jv_copy(old_slot->string);
-      new_slot->value = jv_copy(old_slot->value);
+      new_slot->string = jv_unborrow(old_slot->string);
```
nicowilliams: The keys and values stored in objects should always be owned, not borrowed, so no need to unborrow here.

src/jv.c
```diff
@@ -1711,8 +1717,7 @@ static int jvp_object_equal(jv o1, jv o2) {
     if (jv_get_kind(slot->string) == JV_KIND_NULL) continue;
     jv* slot2 = jvp_object_read(o2, slot->string);
     if (!slot2) return 0;
-    // FIXME: do less refcounting here
-    if (!jv_equal(jv_copy(slot->value), jv_copy(*slot2))) return 0;
+    if (!jv_equal(jv_borrow(slot->value), jv_borrow(*slot2))) return 0;
```
nicowilliams, line 1717: Nice!

src/jv.c
```diff
@@ -1772,7 +1777,12 @@ jv jv_object_set(jv object, jv key, jv value) {
   // copy/free of object, key, value coalesced
   jv* slot = jvp_object_write(&object, key);
   jv_free(*slot);
-  *slot = value;
+  if(jv_is_borrowed(value)){
```
nicowilliams: Can this not be written as `*slot = jv_unborrow(value)` because that's the same as `*slot = value` when `value` is not borrowed?

src/jv.c
```diff
@@ -1781,6 +1791,7 @@ jv jv_object_delete(jv object, jv key) {
   assert(JVP_HAS_KIND(key, JV_KIND_STRING));
   jvp_object_delete(&object, key);
   jv_free(key);
+  if(jv_is_borrowed(object)) return jv_unborrow(object);
```
nicowilliams: I think we need to do the unborrow before the `jvp_object_delete()`, no?  Because we're modifying the `object`, therefore we must own the reference.

src/jv.h
nicowilliams: Let's get the rest of this PR in good enough shape to rebase-and-merge and then we can resolve this comment.

src/jv.h
nicowilliams: We still need to address this.  When this PR is good shape I'll propose a patch.

src/jv_aux.c
```diff
@@ -151,13 +151,16 @@ jv jv_get(jv t, jv k) {
     jv_free(t);
     jv_free(k);
   }
+
+  if(jv_is_borrowed(v)) return jv_unborrow(v);
```
nicowilliams: This is superfluous.  The value we're returning is necessarily already not borrowed.

src/jv_aux.c
```diff
@@ -151,13 +151,16 @@ jv jv_get(jv t, jv k) {
...
   return v;
 }
 
 jv jv_set(jv t, jv k, jv v) {
   if (!jv_is_valid(v)) {
     jv_free(t);
     jv_free(k);
+    if(jv_is_borrowed(v)) return jv_unborrow(v);
```
nicowilliams: Hmmm, before this PR `v` here would be an owned or copied invalid value, and we would return... the same thing, without incrementing or decrementing the reference count on `v`.  We should do the same here, but indeed, if `v` is borrowed we need to return it not-borrowed, but... _with or without incrementing the reference count_?

src/jv_aux.c
```diff
@@ -233,6 +236,7 @@ jv jv_set(jv t, jv k, jv v) {
     jv_free(v);
     t = err;
```
nicowilliams: This looks like a bug in master.  We need to `jv_free(t)` in this case.

emanuele6: There is a `jv_free(t)` 3 lines above this one

nicowilliams: Ah, duh, I'm blind.

src/jv.c
MaxBrandtner: does that always need/ought to be checked? If so it should be added to jv_borrow()

src/jv.c
MaxBrandtner: No because if you pass an owned value to jv_unborrow() it will behave like jv_copy()

MaxBrandtner: but it could be replaced with jv_return() (which does not copy owned values)

src/jv.c
MaxBrandtner: Perhaps? I'll add a test for it and see from there

src/jv_aux.c
MaxBrandtner: We absolutely need to increment the reference count else the following could and has occurred(this caused me quite some pain)

```c
jv a = jv_string("to borrow");
jv b = JV_ARRAY(jv_borrow(a));
jv_free(a);

jv c = jv_getpath(b, JV_ARRAY(jv_number(0))); // trying to read deleted memory
```

src/jv.c
MaxBrandtner: Actually, no jv_invalid_with_msg() currently does not unborrow its errormessage. This should be changed and is a bug(for an explanation see why jv_set has to increment the refcount)

MaxBrandtner: But yeah with jv_invalid_with_msg() fixed all jv functions should be consistent in unborrowing when writing, thus there is no need to unborrow when reading

src/jv.c
MaxBrandtner: indeed it fails

src/jv.c
MaxBrandtner: But we are writing val to *slot. We need to increment the refcount and unborrow there

src/jv.c
MaxBrandtner: could jv_array_slice() cause the same issue?

MaxBrandtner: nvm added a test for it and it passes just fine

MaxBrandtner: jv_object_set() should be problematic as well

And maybe jv_string_append_buf() / jv_string_append_codepoint() /jv_string_append_str() / jv_string_slice() / jv_string_split()

src/jv.h
MaxBrandtner: This function is actually needed a lot(at least within jv) Outside everything should be able to be done with jv_return() and jv_unborrow() , but I think this ought to stay there

src/jv.c
MaxBrandtner: I've now gone through the entire jv api (at least every function I deemed relevant on https://github.com/jqlang/jq/wiki/C-API:-jv)

Sorry for having forgotten some functions initially

MaxBrandtner:
The test should definitely be refactored

src/jv.h
MaxBrandtner: If bitfields are an issue we could use mod operations instead and perhaps rename the pad_ var to bitfield

I would like to avoid that solution though as a.borrowed is more intuitive than a.bitfield % 2

MaxBrandtner:
Perhaps 
- jq_set_attrs()
- jq_set_attr()
- jq_get_attr()
- jq_halt()
- jq_start()

need to also be adjusted

MaxBrandtner:
Actually, `jv_return()` only needs to be used in a very limited number of circumstances
- when using jvp
- when returning a jv* to which the borrowed jv was added 
- when returning an struct to which the borrowed jv was added
- when returning the borrowed jv

src/jv.c
nicowilliams: No.

src/jv.c
nicowilliams: Yes, but we don't need to unborrow when _reading_ the slot elsewhere _because_ it's always not-borrowed once written into the slot.

nicowilliams:
> Perhaps
> 
>     * jq_set_attrs()
> 
>     * jq_set_attr()
> 
>     * jq_get_attr()
> 
>     * jq_halt()
> 
>     * jq_start()
> 
> 
> need to also be adjusted

I don't think so.  Those ultimately use arrays/objects/invalids to store their values.

src/jv.c
nicowilliams: I'm not saying _this_ code here should change.  I'm saying the readers don't need to unborrow what they find in arrays/objects/invalids -- they're already not borrowed.

src/jv.c
MaxBrandtner: Actually, this solution was improper to begin with. `jvp_*` should not do borrow checking and unborrowing is unnecessary since the elements of the array will already not be borrowed. I've reverted the change to `jvp_array_write()` and have added the proper check for `jv_array_set()`. I don't really understand what the `jvp_refcnt_unshared()` was all about, but since borrowing is now all handled outside of `jvp_*`  it should not matter.

MaxBrandtner:
Also what checks tests should be in the final pr?
I think the following would be good to check
- setting jv_array_set / jv_object_set
- borrow safety(aka assigning a jv to an array or object, then freeing it and seeing if the value is still accessible in the array/object) 
- I guess we could check for every function that does some borrow safety stuff whether it is still borrow safe, but that would make the check quite long

MaxBrandtner:
I rewrote the check now to check the following functions comprehensively

- jv_copy
- jv_borrow
- jv_free

- jv_object_set
- jv_array_set
- jv_invalid_with_msg
- jv_string_concat
- jv_string_append_buf

src/jv.h
MaxBrandtner: Just saw that [mongoose](https://mongoose.ws/documentation/#struct-mg_connection) uses bitfields as well. Seeing as mongoose is quite widely used and they showcase it being used on different architectures, I do not think bitfields will be an issue.

MaxBrandtner:
I think this PR probably should be officially NACKed. While I still think there might be performance benefits to not writing / reading a pointer all the time, this proposal adds way too much complexity to be merged into the project.

PR #3055, opened 2024-03-01
improve tonumber/0 performance by parsing input as number literal

itchyny:
Previously, the tonumber/0 filter parses the input as JSON values, but this is less-performant on large non-number strings. Parsing the input string as number literal fixes the performance issue. Also, this fix changes the filter to reject numbers with white spaces.


tests/jq.test
```diff
@@ -2000,8 +2000,8 @@ null
 2
 
 .[] |= try tonumber
-["1", "2a", "3", 4]
-[1, 3, 4]
+["1", "2a", "3", " 4 ", "5.67", ".89", "-876", "+5.43", 21]
+[1, 3, 5.67, 0.89, -876, 5.43, 21]
```
emanuele6, line 2004: Note that spaces before (but not after) the number are still accepted in `--disable-decnum` builds; e.g. `" 4"`

emanuele6 (approved):

wader (approved):

nicowilliams:
Thanks!

PR #3056, opened 2024-03-01
Add trim/0, ltrim/0 and rtrim/0 that trims leading and trailing whitespace

wader:


src/jv_unicode.c
```diff
@@ -118,3 +118,16 @@ int jvp_utf8_encode(int codepoint, char* out) {
   assert(out - start == jvp_utf8_encode_length(codepoint));
   return out - start;
 }
+
+// space codepoints for unicode basic latin and latin-1 supplement
```
wader: This is same as what golang `strings.TrimSpace`/`unicode.IsSpace` considers whitespace https://pkg.go.dev/unicode#IsSpace

docs/content/manual/manual.yml
```diff
@@ -1772,6 +1772,27 @@ sections:
...
+
+          Whitespace is
+          `" "`,
+          `"\n"`
+          `"\t"`,
+          `"\r"`,
+          `"\f"`,
+          `"\u000b"` (vertical tab),
+          `"\u0085"` (next line) and
+          `"\u00a0"` (no-break space). These are the whitespace characters in
```
nicowilliams: Is this exhaustive for Unicode, or just Latin scripts in Unicode?  If the latter, why not be more exhaustive?

src/builtin.c
```diff
@@ -1184,6 +1184,47 @@ static jv f_string_indexes(jq_state *jq, jv a, jv b) {
   return jv_string_indexes(a, b);
 }
 
+static jv f_string_trim(jq_state *jq, jv a) {
```
nicowilliams: We could jq-code `trim`...

docs/content/manual/manual.yml
wader: Is only whitespace from basic block and latin-1, so not exhaustive. Reason is mostly to match what other implementations do. There are quite a lot of other whitespace characters in other blocks, wikipedia has a good list https://en.wikipedia.org/wiki/Whitespace_character.

Reading about PCRE's `\s`-class it seem to match something similar.

src/builtin.c
wader: Main reason was i couldn't come up with something that i thought would be reasonably performant. Any ideas?   reduce/slice somehow? fq's trim is `def trim: capture("^\\s*(?<str>.*?)\\s*$"; "m").str;` atm but i fear that compiling a regex each time might not be great (maybe a regex cache would be good?)

Maybe time to measure?

docs/content/manual/manual.yml
wader: Rust `trim`/`is_whitespace` uses characters with `White_Space` property https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt maybe more resonable?

src/jv_unicode.c
itchyny: Note that Go's `strings.TrimSpace` considers `White_Space` property as well.

src/builtin.c
nicowilliams: Good point.

docs/content/manual/manual.yml
nicowilliams: Maybe we should document that the list is not stable, that we may add to it.

nicowilliams: > Rust `trim`/`is_whitespace` uses characters with `White_Space` property https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt maybe more resonable?

I do like that better, yeah, especially in light of @itchyny's comment.

itchyny: Wait for utf8proc to be included for upcase/downcase?

wader: Update to `White_Space` but i guess we can wait. Also update docs and tests.

itchyny:
I found [a list of languages on this function](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(string_functions)#trim). I have no objection to name this trim not strip, because we already have ltrimstr. Does anyone want ltrim and rtrim as well?

wader:
> I found [a list of languages on this function](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(string_functions)#trim). I have no objection to name this trim not strip, because we already have ltrimstr. Does anyone want ltrim and rtrim as well?

Personally mostly have needed `trim`, and i agree that also having `ltrim`/`rtrim` would fits well with `ltrimstr`/`rtrimstr`.

wader:
Added `ltrim/0` and `rtrim/0`

src/builtin.c
```diff
@@ -1197,6 +1197,58 @@ static jv f_string_indexes(jq_state *jq, jv a, jv b) {
...
+
+static jv string_trim(jv a, int op) {
+  if (jv_get_kind(a) != JV_KIND_STRING) {
+    return ret_error(a, jv_string("trim input must be a string"));
+  }
+
+  int len = jv_string_length_bytes(jv_copy(a));
+  const char* start = jv_string_value(a);
+  const char* trim_start = start;
+  const char* end = trim_start + len;
```
emanuele6: `char *`

src/builtin.c
```diff
@@ -1197,6 +1197,58 @@ static jv f_string_indexes(jq_state *jq, jv a, jv b) {
...
+  const char *trim_end = end;
+  int c;
+
+  if (op & TRIM_LEFT) {
+    for (;;) {
+      const char* ns = jvp_utf8_next(trim_start, end, &c);
```
emanuele6: `char *`

emanuele6 (approved):
Looks good to me

wader:
@emanuele6 thanks for review! maybe wait for one more apporval? thinking adding new functions might be good idea with more consensus

About waiting for utf8proc from #2547: i think we can merge this separately and fix that later

itchyny (approved):
LGTM
After utf8proc is included later, let's see whether we can clean up (or get rid of) jv_unicode.c.

wader:
> LGTM After utf8proc is included later, let's see whether we can clean up (or get rid of) jv_unicode.c.

Yeap, had a quick look at it a few week ago and it seemed like it would be easy, something like `utf8proc_get_property(c)->category == UTF8PROC_CATEGORY_ZS `

Issue #1614, opened 2018-03-04
[feature request] system(command; args) filter


dkrieger:
It would be nice to allow external commands that take JSON in and return JSON out to be embedded in a pipeline of jq filters. an example use case would be mapping over an array of objects, converting an object member to a Longest Common Substring -matched element from a reference array using an external command, and picking up jq execution with that output.

"command" would be a string representing the name/path of the command
args would be an object representing arguments to the command.

parsing semantics demonstration:
```
echo '[10,20,30,40]' | jq 'map(system(mylookup; {"v": false, "f": true, "dict": "mydict.tsv", "o": "json"}))'
```
would feed each item of our input array to stdin of `mylookup -f --dict="mydict.tsv" -o "json"`

charles-dyfis-net:
I would really hope that we *wouldn't* support `system()` as a conventional single-string-argument call that just invokes `['sh', '-c', 'somestring']`; such a usage is almost impossible to use securely. At bare minimum, any such interface should allow multiple arguments, so someone can call `'map(execpipe(["bash", "-c", "mylookup \"$1\"", "_", .]))'`, thus passing an untrusted value (the `.`) in a position where it can't be parsed by the shell as source code. (Consider data values containing substrings like `$(rm -rf ~)` as exemplars of what we want to avoid).

Even then, since `jq` code has been written historically by folks who assume that the language isn't capable of security-sensitive operations, I'd hope for an `--unsafe` argument to be required before this or any equivalent is available.

nicowilliams:
Although I agree that `system(3)` is an awful interface, it's also trivial to add FFIs for, so as it happens I have that already in PR #1843.  Help me kick the tires on that!

(I'll see about not using `system(3)`, instead using `_spawn()` on Windows, or `posix_spawn()` on Unix.  But that's quite a bit more work.)

nichtich:
In particular this feature would also allow to collect and merge data via web APIs with curl. The `--unsafe` option is another issue, I'd prefer a flag to enable/disable *all* external communication via side-channels (including `$ENV`, `env`, `now`, and `strptime`).

nicowilliams:
@nichtich yes, the thought has occurred :)

dosmanak:
I get this feature is controverse, but awk (`awk 'BEGIN{system("date")}'`) and also sed (`echo "date" | sed e`) allows to exec a command 👹 . 

So if `jq` shall be part of this family it should be implemented.

nicowilliams:
[...]

PR #3057, opened 2024-03-02
Add functions to work with filesystem paths: `expand_path` , `get_home`, `get_abspath`

Florents-Tselai:
This PR brings to the public interface three functions already available in `util.h`.
* `get_home`
* `get_abspath`
* `expand_path`

Along with the already available `env` function, they can make jq play more nicely with the filesystem.

Here's what they look like.

image

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1149,6 +1149,28 @@ static jv f_env(jq_state *jq, jv input) {
   return env;
 }
 
+static jv f_expand_path(jq_state *jq, jv input) {
+    if (jv_get_kind(input) != JV_KIND_STRING) {
+        jv_free(input);
+        return type_error(input, "expand_path/1: string required");
+    }
+    return expand_path(input);
+}
+
+static jv f_get_home(jq_state *jq, jv input) {
+    jv_free(input);
+    return get_home();
+}
+
+static jv f_get_abspath(jq_state *jq, jv input) {
+    if (jv_get_kind(input) != JV_KIND_STRING) {
+        jv_free(input);
+        return type_error(input, "get_abspath/1: string required");
+    }
+    return jq_realpath(input);
+}
+
+
 static jv f_halt(jq_state *jq, jv input) {
   jv_free(input);
   jq_halt(jq, jv_invalid(), jv_invalid());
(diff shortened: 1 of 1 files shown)


Florents-Tselai:
I've chosen slightly different names than those in `util.h`.
Also I'm not entirely sure that `expand_path` is worth having.
If this looks reasonable, I'll get to add a couple of tests too.

wader:
The `get_`-prefix feel a bit redundant? maybe `get_home` -> `hoemdir`, `get_abspath` -> `abspath`?

Currently in function names "path" is used to refer to a path expression, could this make things confusing?

@nicowilliams how would this fit naming-wise with the other IO stuff you worked on?

Florents-Tselai:
> The `get_`-prefix feel a bit redundant? maybe `get_home` -> `hoemdir`, `get_abspath` -> `abspath`?
> 
> Currently in function names "path" is used to refer to a path expression, could this make things confusing?
> 

I'm for `homedir` and `absdir`

> @nicowilliams how would this fit naming-wise with the other IO stuff you worked on?

src/builtin.c
```diff
@@ -1149,6 +1149,28 @@ static jv f_env(jq_state *jq, jv input) {
   return env;
 }
 
+static jv f_expand_path(jq_state *jq, jv input) {
+    if (jv_get_kind(input) != JV_KIND_STRING) {
+        jv_free(input);
+        return type_error(input, "expand_path/1: string required");
```
emanuele6: You should remove `jv_free(input)` here; otherwise it is a double-free.
EDIT: also in the other function that calls `type_error`: I forgot to add the second comment to the review.

emanuele6:
Also those functions have `0` arguments (just the input), not  `1`.
The error messages should say `/0`, not `/1`.

nicowilliams:
How about some docs?

wader:
Some tests would be nice too

itchyny:
`homedir` is good, it simply represents what it returns. `abspath` looks it converts any relative paths to absolute paths regardless of the existence, but current implementation does not seem to work with non-existing path. The command `realpath` prints the absolute path regardless of existence, so worth discussing which it should behave. `expand_path` looks an ambiguous term; expanding the home dir of the current user is what it does now (`os.path.expanduser` in Python), but there are possibility of misinterpreted as expanding wildcards, or expanding embed variables, or [something else called shell expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Expansions.html).

Florents-Tselai:
Yeah, you're right @itchyny; from these, only `home_dir` has clear semantics. The rest can be a bit flaky.
How about ditching the rest, and keep `realpath` and set behavior similar to https://www.man7.org/linux/man-pages/man3/realpath.3.html ?

itchyny:
I'm thinking this can be archived if system() (#1614) is included. Long term feature request though.

jemc:
If my PR for the `--sandbox` flag (#3092) is merged first, then this PR should also be updated so that these new filesystem features are blocked by the `--sandbox` flag.

Issue #2566, opened 2023-04-01
Built-in function power10 doesnt seem to exist


hansonchar:
**Describe the bug**
Under [List-of-Built-in-Functions](https://github.com/stedolan/jq/wiki/jq-Language-Description#List-of-Built-in-Functions):
>pow10/0

But 
```bash
$ jq -nc 'pow10'
jq: error (at <unknown>): Error: pow10/0 not found at build time
```

Furthermore, under [Math](https://stedolan.github.io/jq/manual/#Math), `pow10` is listed as a one-input C-math function, but

```bash
$ jq -nc 'pow10(1)'
jq: error: pow10/1 is not defined at <top-level>, line 1:
pow10(1)
jq: 1 compile error
```

**Environment (please complete the following information):**
```
$ uname -a
Darwin ... 21.6.0 Darwin Kernel Version 21.6.0: Mon Dec 19 20:44:01 PST 2022; root:xnu-8020.240.18~2/RELEASE_X86_64 x86_64

$ jq --version
jq-1.6
```

emanuele6:
> Furthermore, under Math, `pow10` is listed as a one-input C-math function, but

Exactly, it only has one input, so you use it as `2|pow10`, not as `pow10(2)` just like the sine function is used as `2|sin`.


Note that a `pow10/0` does in fact exist in your jq, otherwise the script would not compile with a syntax error, and that `jq` only errors when `pow10` is run:
```console
~ $ jq -n '2|try pow10'
~ $ jq -n '2|try pow10a'
jq: error: pow10a/0 is not defined at <top-level>, line 1:
2|try pow10a
jq: 1 compile error
~ $ jq -n '2|select(false)|pow10'
~ $ jq -n '2|select(true)|pow10'
jq: error (at <unknown>): Error: pow10/0 not found at build time
```

Your `jq` does have `pow10/0`, but it is implemented effectively as:
```jq
def pow10:
    error("Error: pow10/0 not found");
```

Support of libc math functions depends on the underlying libc since those `jq` functions are just wrappers around them.
If libc doesn't provide one of those functions, `jq` still provides the `jq` wrapper as a function that error when they are called.

Your system's libc does not support `pow10`, like most other modern systems don't; `pow10` is not defined by the C standard and is an old extension.

This is a duplicate of #2456. You can also read that for more details.

hansonchar:
Interesting!

Issue #2456, opened 2022-07-18
Cannot use the pow() nor pow10() functions


sebma:
**Describe the bug**
I want to divide an integer representing a throughput in `bits/sec` by `10^9` to get `Gbits/sec` but I get `pow/1 is not defined` errors.

**To Reproduce**
```shell
$ echo 7223210000 | jq -r './pow(10,9) | tostring + " bits/sec"'
jq: error: pow/1 is not defined at <top-level>, line 1:
./pow(10,9) | tostring + " bits/sec"
jq: 1 compile error
$ echo 7223210000 | jq -r './pow10(9) | tostring + " bits/sec"'
jq: error: pow10/1 is not defined at <top-level>, line 1:
./pow10(9) | tostring + " bits/sec"
jq: 1 compile error
```

**Expected behavior**
```shell
$ echo 7223210000 | jq -r './pow(10,9) | tostring + " bits/sec"'
7.22321 bits/sec
$ echo 7223210000 | jq -r './pow10(9) | tostring + " bits/sec"'
7.22321 bits/sec
```

**Environment (please complete the following information):**
 - OS and Version: Linux CentOS 7.9
 - jq version 1.5

**Additional context**
`jq` has been built with maths functions ( `-lm`) : 
```
$ ldd $(which jq) | grep libm
        libm.so.6 => /usr/lib64/libm.so.6 (0x00007fa9155a9000)
$ rpm -qf /usr/lib64/libm.so.6
glibc-2.17-325.el7_9.x86_64
```

wader:
Hey, jq uses ; as argument separator so try pow(10; 9)

emanuele6:
`pow(10,9)` is the same as `pow(10),pow(9)` it calls `pow/1` twice (and `pow/1` doesn't exist).
You want to call `pow/2` with two argument, so you should use `pow(10;9)`.

```
bash-5.1$ jq -n 'pow(10; 9)'
1000000000
```
---

`pow10/1` does not exist. `pow10` exists as `pow10/0`: it takes no arguments and takes a number as input, and outputs the result; you would use it as `9 | pow10`.
But keep in mind that `pow10/0` doesn't work because it is not built correctly:
```
bash-5.1$ jq -n '9 | pow10'
jq: error (at <unknown>): Error: pow10/0 not found at build time
```

emanuele6:
Also i think it would make more sense to just use a number literal.
If you don't want to have to write `1000000000`, you can just use `1e9`.

```bash
[...]

PR #3059, opened 2024-03-07
Remove non-standard pow10 filter in favor of exp10

itchyny:
The `pow10(3)` function is a non-standard glibc extension removed in 2.27. Listing in the manual often confuses users like  #2456, #2566. The `exp10(3)` function was a glibc extension too, but is standardized in TS 18661-4:2015 and expected to be available widely.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -201,7 +201,6 @@ AC_CHECK_MATH_FUNC(lgamma_r)
 AC_CHECK_MATH_FUNC(nearbyint)
 AC_CHECK_MATH_FUNC(nextafter)
 AC_CHECK_MATH_FUNC(nexttoward)
-AC_CHECK_MATH_FUNC(pow10) # Not available with glibc version >= 2.27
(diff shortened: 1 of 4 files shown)


emanuele6 (approved):

emanuele6:
Maybe we could add a note in the manual that explains that `exp10` is equivalent to `pow10`, but more likely to be implemented?

wader:
Some reason `pow10` can't be an alias for `exp10`?

itchyny:
I prefer making implemenetation clean, there are `exp/0`, `exp2/0` so `exp10/0` is easy to remember (even if `pow/2` exists), another reason is `pow10/0` has not been available in many environments, so I don't think it's important to keep compatibility.

wader:
@itchyny Ok makes sense, i'm fine with removing it

wader (approved):

itchyny:
Thank you.

PR #3060, opened 2024-03-08
remove repetitive words

wellweek:


wader (approved):
Looks good. But not sure what we had said about changes to "vendored" code like decNumber

emanuele6:
Really not much point in fixing those problems in those vendored (`src/decNumber/*`) files in my opinion, but it's fine.

emanuele6:
Thank you

PR #3065, opened 2024-03-12
Use codepoint index for indices/1, index/1 and rindex/1

### LINKED ISSUE #1430: index() function returns wrong offset for non-ascii chars

### LINKED ISSUE #1624: substring of index/rindex doesn't work for utf8 inputs

### LINKED ISSUE #3064: `indices` reports byte offsets instead of character offsets
**Describe the bug**
jq uses _characters_ to index strings.
To see that, we can run `"🇬🇧oo" | .[0 : 1,2,3,4]`, which yields "🇬" "🇬🇧" "🇬🇧o" "🇬🇧oo".
Note that 🇬🇧 is actually two characters and 8 bytes, as we can see from `"🇬🇧" | length, utf8bytelength`.
However, the `indices` filter returns _byte offsets_  to the pattern in the string.
The documentation does not specify the behaviour of `indices` for UTF-8 strings, but given that `length` and `.[x:y]` use character counts to index strings, it is likely that this is a bug and not just undocumented behaviour.

**To Reproduce**
$ ./jq-linux-amd64-1.7.1 -nc '"🇬🇧oo" | indices("o")'
[8,9]
$ ./jq-linux-amd64-1.7.1 -nc '"ƒoo" | indices("o")'
[2,3]

**Expected behavior**
$ ./jq-linux-amd64-1.7.1-fixed -nc '"🇬🇧oo" | indices("o")'
[2,3]
$ ./jq-linux-amd64-1.7.1-fixed -nc '"ƒoo" | indices("o")'
[1,2]

The problem is probably caused in [jv_string_indexes](https://github.com/jqlang/jq/blob/c95b34ff827d05a2d262f00280a4891a295ed0ed/src/jv.c#L1272).

wader:
Previsouly byte index was used.

Fixes #1430, fixes #1624, fixes #3064.

### DIFF
--- a/src/jv.c
+++ b/src/jv.c
@@ -1274,15 +1274,21 @@ jv jv_string_indexes(jv j, jv k) {
   assert(JVP_HAS_KIND(k, JV_KIND_STRING));
   const char *jstr = jv_string_value(j);
   const char *idxstr = jv_string_value(k);
-  const char *p;
+  const char *p, *lp;
   int jlen = jv_string_length_bytes(jv_copy(j));
   int idxlen = jv_string_length_bytes(jv_copy(k));
   jv a = jv_array();
 
   if (idxlen != 0) {
-    p = jstr;
+    int n = 0;
+    p = lp = jstr;
     while ((p = _jq_memmem(p, (jstr + jlen) - p, idxstr, idxlen)) != NULL) {
-      a = jv_array_append(a, jv_number(p - jstr));
+      while (lp < p) {
+        lp += jvp_utf8_decode_length(*lp);
+        n++;
+      }
+
+      a = jv_array_append(a, jv_number(n));
       p++;
     }
   }
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1345,6 +1345,22 @@ indices(", ")
 "a,b, cd,e, fgh, ijkl"
 [3,9,14]
 
+index("!")
+"здравствуй мир!"
+14
+
+.[:rindex("x")]
+"正xyz"
+"正"
+
+indices("o")
+"🇬🇧oo"
+[2,3]
+
+indices("o")
+"ƒoo"
+[1,2]
+
 [.[]|split(",")]
 ["a, bc, def, ghij, jklmn, a,b, c,d, e,f", "a,b,c,d, e,f,g,h"]
 [["a"," bc"," def"," ghij"," jklmn"," a","b"," c","d"," e","f"],["a","b","c","d"," e","f","g","h"]]


src/jv.c
```diff
@@ -1274,15 +1274,21 @@ jv jv_string_indexes(jv j, jv k) {
...
   int idxlen = jv_string_length_bytes(jv_copy(k));
   jv a = jv_array();
 
   if (idxlen != 0) {
-    p = jstr;
+    int n = 0;
+    p = lp = jstr;
     while ((p = _jq_memmem(p, (jstr + jlen) - p, idxstr, idxlen)) != NULL) {
-      a = jv_array_append(a, jv_number(p - jstr));
+      while (lp < p) {
```
wader, line 1286: To make this even more efficient i guess we would need to count codepoints inside memmem somehow

wader:
Haven't entirely convinced myself yet that it should be fine to look for matches using the byte representation. Assuming both the needle and haystack is valid utf-8 i'm thinking it should be fine because of utf-8's self-synchronization property.

Update: now looking at `jv_string_slice` https://github.com/jqlang/jq/blob/c95b34ff827d05a2d262f00280a4891a295ed0ed/src/jv.c#L1374 i'm not sure anymore if one can assume strings are valid utf-8 or is the invalid utf-8 checks not really needed?

emanuele6 (approved):

itchyny:
I'd like to include this. Any objection on changing the behavior in 1.8?

wader:
Ok to merge for me but would be great if someone could have a look or know if my assumption about strings always being valid utf-8 is true.

pkoppstein:
@itchyny asked:

> Any objection on changing the behavior in 1.8

This is a major breaking change and it has been my understanding for some years that such changes would have to wait until jq 2.0.  Certainly if we were following a strict SemVer policy that would be the case.  Since we don't seem to be doing so, the situation is not black-and-white, but if the change is incorporated into 1.8, we should be sure to highlight it.   

@wader wrote:
> if my assumption about strings always being valid utf-8 is true.

Based on past experience, such an assumption would not be warranted, so the question is: could the proposed changes make anything worse?  I suppose the major issue would be whether (in the presence of invalid utf-8) the old index would give an accurate byte count but the new version might give an inaccurate codepoint count. 

Perhaps a starting point would be `"a\uDD1Ec"`:
```
echo '"a\uDD1Ec"' | jaq -r .
Error: failed to parse: invalid character with index 56606

echo '"a\uDD1Ec"' | jq -c '[index("c"), length]'
[4,3]
```

wader:
> @itchyny asked:
> 
> > Any objection on changing the behavior in 1.8
> 
> This is a major breaking change and it has been my understanding for some years that such changes would have to wait until jq 2.0. Certainly if we were following a strict SemVer policy that would be the case. Since we don't seem to be doing so, the situation is not black-and-white, but if the change is incorporated into 1.8, we should be sure to highlight it.

I can't see how the current behaviour for non-ASCII strings makes any sense or could even be useful in any resonable way? so for me it feels more like a bug.

> @wader wrote:
> 
> > if my assumption about strings always being valid utf-8 is true.
> 
> Based on past experience, such an assumption would not be warranted, so the question is: could the proposed changes make anything worse? I suppose the major issue would be whether (in the presence of invalid utf-8) the old index would give an accurate byte count but the new version might give an inaccurate codepoint count.
> 
> Perhaps a starting point would be `"a\uDD1Ec"`:
> 
> ```
> echo '"a\uDD1Ec"' | jaq -r .
> Error: failed to parse: invalid character with index 56606
> 
> echo '"a\uDD1Ec"' | jq -c '[index("c"), length]'
> [4,3]
> ```

This is an incomplete surrogates pair? yeap stuff like this i'm concerned about also.

pkoppstein:
@wader wrote:

> I can't see how the current behaviour for non-ASCII strings makes any sense or could even be useful in any reasonable way?

Currently:
```
echo '"a\udc37c"' | jq -c '[., length, index("\udc37")]'
["a�c",3,1]
```
With the proposed change, would this result be different?

In any case, as I've said, as far as I'm concerned, the main thing is that if the change is made in 1.8, it should be highlighted, preferably as a possibly "breaking" change.

wader:
With this change:
```sh
$ echo '"a\uDD1Ec"' | ./jq -c '[index("c"), length]'
[2,3]
```
Seems correct assuming broken surrogates codepoints should be allowed. But I think i'm mostly concern if there is any way to produce jq strings that has a byte buffer that is not valid utf-8. If so use of `jvp_utf8_decode_length` might end up out-of-sync codepoint-wise or pointing outside the byte buffer.

itchyny (approved):

wader:
As @nicowilliams also expressed this is a bug https://github.com/jqlang/jq/issues/1430#issuecomment-347628258 ill merge this now

thaliaarchi:
Should the docs be updated to make it clear that it's a codepoint index, instead of byte index?

wader:
Good point, i did quick skimming of the docs and it seems like we don't say it's byte offset anywhere but we don't make it that clear that it's codepoints either. So maybe mention for the index-functions and possible also under "Array/String Slice" and/or "Types and Values"? for the regexp functions we do mention things are in codepoints.

PR #3068, opened 2024-03-15
strftime/1: fix validation of non-string argument with number input

emanuele6:
There was a incorrect `else`, that caused jq to not ensure that the argument to `strftime/1` is a string when the input is a number; this ends up calling `jv_string_value` on a non-string value, which does not work, and causes an assert failure.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67403

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1599,9 +1599,9 @@ static jv f_strftime(jq_state *jq, jv a, jv b) {
     }
   } else if (jv_get_kind(a) != JV_KIND_ARRAY) {
     return ret_error2(a, b, jv_string("strftime/1 requires parsed datetime inputs"));
-  } else if (jv_get_kind(b) != JV_KIND_STRING) {
-    return ret_error2(a, b, jv_string("strftime/1 requires a string format"));
   }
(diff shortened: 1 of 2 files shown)


src/builtin.c
```diff
@@ -1599,9 +1599,9 @@ static jv f_strftime(jq_state *jq, jv a, jv b) {
     }
   } else if (jv_get_kind(a) != JV_KIND_ARRAY) {
     return ret_error2(a, b, jv_string("strftime/1 requires parsed datetime inputs"));
-  } else if (jv_get_kind(b) != JV_KIND_STRING) {
-    return ret_error2(a, b, jv_string("strftime/1 requires a string format"));
   }
+  if (jv_get_kind(b) != JV_KIND_STRING)
+    return ret_error2(a, b, jv_string("strftime/1 requires a string format"));
```
wader, line 1604: Same issue in `f_strflocaltime` it seems?

emanuele6: Yes!

wader (approved):

PR #3070, opened 2024-03-15
builtin.c: fix signed integer overflow in jv2tm

emanuele6:
`jv2tm` now properly clamps large number values to a signed 32-bit integer and rejects nan.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65885


emanuele6:
This does not solve the problem fully; it still underflows if you use `-1e31` instead of `1e31` because of the `tm->tm_year -= 1900;` in `jv2tm`

src/builtin.c
```diff
@@ -1461,30 +1461,29 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
+    offsetof(struct tm, tm_yday),
+  };
+
+  jv_array_foreach(a, i, n) {
+    if (jv_get_kind(n) != JV_KIND_NUMBER || jvp_number_is_nan(n))
+      return 0;
+    double d = jv_number_value(n);
+    jv_free(n);
+    if (i == 0) /* year */
+      d -= 1900;
```
emanuele6, line 1488: to fix the UB underflow I performed the `-= 1900` on the double value before the `i32` conversion. this does not trigger UB errors, and is how `binop_minus` implements subtraction

src/builtin.c
```diff
@@ -1461,30 +1461,31 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
+    offsetof(struct tm, tm_mon),
+    offsetof(struct tm, tm_mday),
+    offsetof(struct tm, tm_hour),
+    offsetof(struct tm, tm_min),
+    offsetof(struct tm, tm_sec),
+    offsetof(struct tm, tm_wday),
+    offsetof(struct tm, tm_yday),
+  };
+
+  jv_array_foreach(a, i, n) {
```
emanuele6: I replaced the `TO_TM_FIELD` macro repeated 8 times with a loop over `offsets`

src/builtin.c
```diff
@@ -1461,30 +1461,31 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
+    if ((size_t)i >= (sizeof offsets / sizeof *offsets))
```
emanuele6: I can loop more efficiently

tests/jq.test
```diff
@@ -1568,6 +1568,10 @@ strftime("%A, %B %d, %Y")
 1435677542.822351
 "Tuesday, June 30, 2015"
 
+try ["OK", strftime("%Y-%m-%dT%H:%M:%SZ")] catch ["KO", .]
```
emanuele6: I am using `try ["OK",` because the test throws an error in the windows CI, and `--run-tests` does not output exceptions to the log, so this is the only way to check what is going on

emanuele6: `["KO","strftime/1: unknown system failure"]`; windows's lib c (lib posix?) is probably more strict than gnu's so it does not allow `mday=0`; i will make sure to at least set `mday` so the test is portable
```
The <time.h> header shall declare the tm structure, which shall include at least the following members:

int    tm_sec   Seconds [0,60]. 
int    tm_min   Minutes [0,59]. 
int    tm_hour  Hour [0,23]. 
int    tm_mday  Day of month [1,31]. 
int    tm_mon   Month of year [0,11]. 
int    tm_year  Years since 1900. 
int    tm_wday  Day of week [0,6] (Sunday =0). 
int    tm_yday  Day of year [0,365]. 
int    tm_isdst Daylight Savings flag. 
```

src/builtin.c
```diff
@@ -1461,30 +1461,35 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
...
+    offsetof(struct tm, tm_min),
+    offsetof(struct tm, tm_sec),
+    offsetof(struct tm, tm_wday),
+    offsetof(struct tm, tm_yday),
+  };
+
+  for (size_t i = 0; i < (sizeof offsets / sizeof *offsets); ++i) {
+    jv n = jv_array_get(jv_copy(a), i);
+    if (!jv_is_valid(n))
+      break;
```
wader, line 1480: this is essentially the `i < length(a)` check?

wader (approved):

src/builtin.c
emanuele6: yes

itchyny:
Why did you leave `mktime` not accepting the input `strftime` can recognize?
```sh
 $ ./jq -n '[2024,7,20] | todate, mktime'           
"2024-08-20T00:00:00Z"
jq: error (at <unknown>): mktime requires parsed datetime inputs
```

emanuele6:
I didn't touch `todate`/`strftime` directly; it looks like `mktime` doesn't accept those values because there was an unnecessary extra check that requires the length of the input array to be at least 6:
```c
  if (jv_array_length(jv_copy(a)) < 6)
    return ret_error(a, jv_string("mktime requires parsed datetime inputs"));
```

That can be removed now

PR #3071, opened 2024-03-15
util.c: fix _WIN32 port of strptime

emanuele6:
In _WIN32 builds, TIME_MAX needs to be INT32_MAX.

Also fix UB overflow, and useless check in NetBSD's `strptime("%s")`:
```c
(time_t_value * 10) <= TIME_MAX
```
is always true since `time_t` is signed, and `TIME_MAX` is its maximum value.

And, if it is not optimised out, it causes a signed integer overflow (undefined in standard C) for inputs that start with a sequence of characters between "922337203685477581" and "999999999999999999" (inclusive, 64-bit `time_t`).

Also, since the check does not do what it is supposed to do, even if we assume that signed integer overflow is defined like for unsigned integers, on builds where `time_t` is `int32_t`, and `TIME_MAX` is `INT32_MAX`, this will cause `strptime("%s")` to accept 99999999999 as a valid timestamp, equivalent to 1410065398 (`Sun  7 Sep 04:49:58 UTC 2014`) instead of rejecting it.
This works because `floor(log10(UINT32_MAX)) == floor(log10(INT32_MAX))` in 32-bit.

Noticed thanks to a compiler warning in the windows build CI.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -818,40 +818,47 @@ recurse:
             LEGAL_ALT(ALT_O);
             continue;
 
-#ifndef TIME_MAX
-#define TIME_MAX	INT64_MAX
-#endif
-        case 's':	/* seconds since the epoch */
-            {
-                time_t sse = 0;
-                uint64_t rulim = TIME_MAX;
-
-                if (*bp < '0' || *bp > '9') {
-                    bp = NULL;
-                    continue;
-                }
-
-                do {
-                    sse *= 10;
-                    sse += *bp++ - '0';
-                    rulim /= 10;
-                } while ((sse * 10 <= TIME_MAX) &&
-                     rulim && *bp >= '0' && *bp <= '9');
-
-                if (sse < 0 || (uint64_t)sse > TIME_MAX) {
-                    bp = NULL;
-                    continue;
-                }
+        case 's': {     /* seconds since the epoch */
 #ifdef _WIN32
-                if (localtime_s(tm, &sse) == 0)
+            const time_t TIME_MAX = INT32_MAX;
 #else
-                if (localtime_r(&sse, tm))
+            const time_t TIME_MAX = INT64_MAX;
 #endif
-                    state |= S_YDAY | S_WDAY | S_MON | S_MDAY | S_YEAR;
-                else
-                    bp = NULL;
+            time_t sse;
+            time_t d;
+
+            if (*bp < '0' || *bp > '9') {
+                bp = NULL;
+                continue;
             }
+
+            sse = *bp++ - '0';
+            while (*bp >= '0' && *bp <= '9') {
+                d = *bp++ - '0';
+                if (sse > TIME_MAX/10) {
+                    bp = NULL;
+                    break;
+                }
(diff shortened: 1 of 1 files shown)


emanuele6:
Now, jq/libjq builds without warnings on the windows CI job

file

wader:
This was sent upstream as https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=58041 or? wait for that before merge?

emanuele6:
This was fixed in NetBSD rewriting the loop in a different way; I could port the NetBSD fix to jq instead of using mine.

wader:
> This was fixed in NetBSD rewriting the loop in a different way; I could port the NetBSD fix to jq instead of using mine.

Ok, if equivalent fix i think using the netbsd version will be less confusing, thinking if there will be backports in the future etc

wader (approved):

emanuele6:
The NetBSD code triggers another warning:
```
ake[2]: Entering directory '/d/a/jq/jq'
  CC       src/jv_dtoa_tsd.lo
  CC       src/lexer.lo
  CC       src/parser.lo
  GEN      jq.1
  CC       src/main.o
  CC       src/builtin.lo
  CC       src/bytecode.lo
  CC       src/compile.lo
  CC       src/execute.lo
  CC       src/jq_test.lo
  CC       src/jv.lo
  CC       src/jv_alloc.lo
  CC       src/jv_aux.lo
  CC       src/jv_dtoa.lo
  CC       src/jv_file.lo
  CC       src/jv_parse.lo
  CC       src/jv_print.lo
  CC       src/jv_unicode.lo
  CC       src/linker.lo
  CC       src/locfile.lo
  CC       src/util.lo
  CC       src/decNumber/decContext.lo
  CC       src/decNumber/decNumber.lo
src/util.c: In function 'strptime':
src/util.c:843:25: warning: comparison of integer expressions of different signedness: 'time_t' {aka 'long int'} and 'long unsigned int' [-Wsign-compare]
  843 |                 if (sse > TIME_MAX - d) {
      |                         ^
  CCLD     libjq.la
  CCLD     jq.exe
make[2]: Leaving directory '/d/a/jq/jq'
make[1]: Leaving directory '/d/a/jq/jq'
```

wader:
> The NetBSD code triggers another warning:

the joy! 🥳

wader (approved):

src/util.c
```diff
@@ -818,40 +818,47 @@ again:		switch (c = *fmt++) {
...
+            const time_t TIME_MAX = INT32_MAX;
+#else
+            const time_t TIME_MAX = INT64_MAX;
 #endif
-        case 's':	/* seconds since the epoch */
-            {
-                time_t sse = 0;
-                uint64_t rulim = TIME_MAX;
+            time_t sse;
+            time_t d;
```
emanuele6, line 828: Instead of adding a cast, I declared `d` as `time_t` instead of `unsigned` so the signedness/size of `TIME_MAX - d` is always correct.

wader (approved):

PR #3080, opened 2024-03-27
@base64d: fix unhandled overflow

emanuele6:
I replaced `(3*l)/4` with `3*(l/4)` to prevent overflows:

```console
$ ./jq-before -n '238609295*"|||"|@base64d|"."'
src/builtin.c:718:29: runtime error: signed integer overflow: 715827885 * 3 cannot be represented in type 'int'
jq: error: cannot allocate memory
Aborted (core dumped)
```

```console
$ ./jq-after -n '238609295*"|||"|@base64d|"."'
jq: error (at <unknown>): string ("||||||||||...) is not valid base64 data
```

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67640

---

I didn't add a test for this, because this takes ages to run on my laptop even if I use something like
```sh
</dev/zero tr \\0 \| | head -c715827883 | ./jq -R '@base64d|"."
```
to generate the input string, and even without `valgrind`.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -715,7 +715,7 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
-    size_t decoded_len = (3 * len) / 4; // 3 usable bytes for every 4 bytes of input
+    size_t decoded_len = (3 * (size_t)len) / 4; // 3 usable bytes for every 4 bytes of input
     char *result = jv_mem_calloc(decoded_len, sizeof(char));
     memset(result, 0, decoded_len * sizeof(char));
     uint32_t ri = 0;


emanuele6:
> I replaced `(3*l)/4` with `3*(l/4)` to prevent overflows:

Hmm, right, I can't do that or the result will be wrong for e.g. `l=3`; I will just add a cast instead.

src/builtin.c
```diff
@@ -715,7 +715,7 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
-    size_t decoded_len = (3 * len) / 4; // 3 usable bytes for every 4 bytes of input
+    size_t decoded_len = (3 * (size_t)len) / 4; // 3 usable bytes for every 4 bytes of input
```
wader, line 718: I guess too late to change now, but is there a reason why `jv_*_length` etc don't return `size_t`?

wader (approved):

src/builtin.c
emanuele6: @wader I think it was done that was to make it fit in a `jv`; `jv`s are 128-bit values (so they can be returned by value without using the stack with `rax` and `rdx` in amd64), and they use 8-bit for flags, 8-bit for pad, 16-bit for offset, 32-bit for size (weirdly as sized `int` instead of unsigned), and 64-bit for either a pointer to an allocate value  or an embedded `double`. If both `size` and `u.ptr` were 64-bits (size `size_t`), you would not have space for flags, etc., and you could not efficiently return `jv`s by value, having to use something like `jv_get(&output, input, path)` instead of `output = jv_get(input, path)`.

Though, I am not sure why `size` has to be in `jv`, and not in the allocated value; I have never read that code.

wader: Huh interesting! i wonder much performance is gained by doing it that way? 🤔 gojq uses a `any` as value type which is two pointers https://github.com/golang/go/blob/bf86aec25972f3a100c3aa58a6abcbcc35bdea49/src/runtime/runtime2.go#L143-L146 are there any comprehensive benchmarks about this? jaq has some https://github.com/01mf02/jaq?tab=readme-ov-file#performance which seems to suggest jq, gojq and jaq are not that different, i would have expected more difference :)

emanuele6: Two pointers are 2\*64-bit = 128-bit, so the same size as a `jv`.

If it was not clear, what I meant to say is that with the C amd64 calling convention, that is the largest "integer" value you can return efficiently by value just using registers; declaring a function that returns a larger value results in code similar to something like this:
```c
struct foo { size_t x[3]; }; /* 192-bit */
struct foo bar(int);
/* ... */
    struct foo v = bar(123);
/* automatically compiles to */
    struct foo v;
    (void)bar(&v, 123); /* returns &v */

/* declaring an extern function that is
  struct foo bar(int);
as
  struct foo *bar(struct foo *out, int);
*should* work because they are called equivalently */
```

Since it uses memory instead of just registers, it is going to be less efficient, so for something that is returned frequently by value it is probably better to not go above 128-bit.

My guess was that `size` was intentionally made a 32-bit integer and not a `size_t` to fit along with flags, value, etc in a `jv` since `jv` is pretty obviously intentionally 128-bit so it can be returned by value.

(Page 27 (pdf) of https://cs61.seas.harvard.edu/site/pdf/x86-64-abi-20210928.pdf for formal specification of return by value convention)

wader: 👍  (continued on discord)

PR #3081, opened 2024-03-28
jv_array_indexes: remove a bunch of unnecessary copies/gets

emanuele6:


emanuele6:
I found another `jv_array_foreach` loop that only conditionally uses the value, and could be replaced by a regular index loop

wader (approved):

PR #3090, opened 2024-04-07
Modified _modify method to not hold | onto reference

### LINKED ISSUE #2706: `|=` still holds on to a reference on the value to be modified
```jq
def _modify(paths; update):
    reduce path(paths) as $p ([., []];
        . as $dot
      | null
      | label $out
      | ($dot[0] | getpath($p)) as $v
      | (
          (   $$$$v   # <----- here we let go of $v, but $dot[0] at $p still has the same value
            | update
            | (., break $out) as $v
            | $$$$dot
            | setpath([0] + $p; $v)
          ),
          (
              $$$$dot
            | setpath([1, (.[1] | length)]; $p)
          )
        )
    ) | . as $dot | $dot[0] | delpaths($dot[1]);
```

We should `setpath([0] + $p; null)` before the `$$$$v | update`.

EDIT: This is by code inspection.  I haven't tested this.

capak07:
- Fixes #2706 
- Updated _modify method so use set path before updating

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -18,6 +18,7 @@ def _modify(paths; update):
       | null
       | label $out
       | ($dot[0] | getpath($p)) as $v
+      | setpath([0] + $p; null)
       | (
           (   $$$$v
             | update


wader:
Trying to understand the issue. Is the problem that a reference is leaked causing a memory leak or that a reference is hold longer than needed? Is there any way to verify this works?

emanuele6:
I don't get it

itchyny:
No further explanation given, so closing.

PR #3091, opened 2024-04-07
A couple of teeny tweaks

petdance:
Two tweaks:
* Don't get HOMEDRIVE environment variable unless needed
* Use put_char instead of put_str for single-character strings


wader (approved):

itchyny (approved):
Looks good, thanks.

Issue #1361, opened 2017-03-05
Security considerations


lmeyerov:
We're considering allowing user-defined transforms specified in jq, but it has been difficult for us to ascertain how safe it is from code injection & privilege escalation, by design & in practice. As far as I can tell, while jq supports some user-defined functions, they are largely sandboxed. The exception is the module imports, which would need some thing for restricting.

Can anyone clarify the security model? May help to link the answer as part of the README.

nicowilliams:
Excellent question.  So far we've held back from adding the sorts of builtins that might make it easier to break out of the jq sandbox.

Real or potential issues to be concerned with:

 - CPU consumption -- jq is Turing complete (but single-threaded)
 - Row-hammer attacks -- there's no reason to think that they are not possible to code in jq
 - Potential security bugs in jq -- we've not yet applied AFL, but others have and we know that there exist JSON parser bugs
 - Potential security bugs in bison where the jq program could source itself can exploit them
 - It is possible to read arbitrary files down from the current directory using `include` and `import` directives -- this is a big deal if you don't trust the jq code you're running
 - We do want to add more I/O builtins in the future -- you can't count on jq being sandboxed forever

Now, it is my intention to require the use of a command-line switch to enable future I/O builtins that could allow reading or writing files, or executing external programs.  But still, there's no way to guarantee that's so.

I think it's probably desirable to add a command-line option to disallow the main program from including files or modules, or more generally a sandbox option.

nicowilliams:
See also #1215, #660, and others.

lmeyerov:
Ah, excellent. Thanks! 

Out of those, the include and import statements seem concerning under most threat models (including ours.) +1 for sandboxed mode flag for this and future features :) 

For the interim, would it be sufficient for us to do a blacklist for "include" and "import"?

nicowilliams:
@lmeyerov Yes.  You might even edit `src/parser.y` and remove the `include`/`import` functionality.

Thinking about how to disable `include`/`import`... we could just shove that into `struct locfile`, or add another parse param for this, and the in the rules for `include`/`import` simply error out if they are disabled.

nicowilliams:
There's a lot of potential sandboxing options:

 - allow importing modules, but not data
 - allow importing modules, but not data, and don't allow modules to import data either
 - don't allow importing anything

Eventually we'll have fairly capable I/O builtins.  We might have a need for many sandboxing options.

[...]

PR #3092, opened 2024-04-12
Add `--sandbox` flag to prevent dynamic loading of other files/data.

jemc:
Some use cases for `jq` may involve accepting untrusted input. See discussion in #1361 for some security considerations that may be relevant for those use cases.

This commit adds a `--sandbox` flag which is meant to mitigate one category of security issue with untrusted input: features of jq which are meant to let the jq filter access files/data other than the direct input data given to the CLI.

Specifically, the new `--sandbox` flag blocks the implicit loading of `$HOME/.jq`, and also blocks the use of `import` and `include` for loading other `jq` files.

If other features are added to `jq` in the future which allow for reading files/data as part of the filter syntax, it is intended that the `--sandbox` flag would also gate access to those.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -396,6 +396,28 @@ if $VALGRIND $Q $JQ -L ./tests/modules --sandbox -n 'import "a" as a; empty'; th
     exit 1
 fi
 
+## Test environment variable access
+
+if [ "$(FOO=foo $VALGRIND $Q $JQ -nr '$ENV.FOO')" != foo ]; then
+    echo "couldn't read an environment variable via \$ENV" 1>&2
+    exit 1
+fi
+
+if [ "$(FOO=foo $VALGRIND $Q $JQ --sandbox -nr '$ENV.FOO')" != null ]; then
+    echo "\$ENV should have been empty due to the sandbox flag" 1>&2
+    exit 1
+fi
+
+if [ "$(FOO=foo $VALGRIND $Q $JQ -nr 'env.FOO')" != foo ]; then
+    echo "couldn't read an environment variable via env" 1>&2
+    exit 1
+fi
+
+if [ "$(FOO=foo $VALGRIND $Q $JQ --sandbox -nr 'env.FOO')" != null ]; then
+    echo "env should have been empty due to the sandbox flag" 1>&2
+    exit 1
+fi
+
 ## Halt
 
 if ! $VALGRIND $Q $JQ -n halt; then
--- a/tests/shtest
+++ b/tests/shtest
@@ -381,6 +381,21 @@ if ! $VALGRIND $Q $JQ -L tests/modules -ne 'import "test_bind_order" as check; c
     exit 1
 fi
 
+if HOME="$mods/home1" $VALGRIND $Q $JQ --sandbox -nr fg; then
+    echo "home module was loaded when it should have been prevented by sandbox flag" 1>&2
+    exit 1
+fi
+
+if HOME="$mods/home2" $VALGRIND $Q $JQ --sandbox -n 'include "g"; empty'; then
+    echo "module was included when it should have been prevented by sandbox flag" 1>&2
+    exit 1
+fi
+
+if $VALGRIND $Q $JQ -L ./tests/modules --sandbox -n 'import "a" as a; empty'; then
+    echo "module was imported when it should have been prevented by sandbox flag" 1>&2
+    exit 1
+fi
+
 ## Halt
 
 if ! $VALGRIND $Q $JQ -n halt; then
--- a/docs/content/manual/manual.yml
+++ b/docs/content/manual/manual.yml
@@ -297,6 +297,11 @@ sections:
         operations that would allow the filter code to access data other
         than the input data that is explicitly specified in the invocation.
 
+        This flag also hides all environment variables from the enviroment
+        where jq was run by setting `$ENV` and `env` to be an empty object.
+        If you need to pass named arguments to a sandboxed jq filter, use the
+        `--arg` and/or `--argjson` options to pass them explicitly.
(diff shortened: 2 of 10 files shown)


itchyny:
Should this flag block use of [env](https://jqlang.github.io/jq/manual/#$env-env)?

wader:
> Should this flag block use of [env](https://jqlang.github.io/jq/manual/#$env-env)?

Good point, block i think. Would `env` and `$ENV` just be empty objects, null or not defined at all?

Any special consideration around the `input*` functions?

jemc:
I will have a crack at updating the PR to make `env` and `$ENV` be empty objects when the sandbox flag is present.

Regarding the `input*` functions, my thinking was that because these related to letting the jq filter interact with the inputs that were explicitly given to the filter by the invoker, I saw this as being "within the sandbox".

jemc:
I've updated the PR to clear the `$ENV` value and `env` builtin, including tests and documentation changes to match. Thanks for the feedback!

Let me know if there is anything else.

tests/shtest
```diff
@@ -396,6 +396,28 @@ if $VALGRIND $Q $JQ -L ./tests/modules --sandbox -n 'import "a" as a; empty'; th
...
+    exit 1
+fi
+
+if [ "$(FOO=foo $VALGRIND $Q $JQ -nr 'env.FOO')" != foo ]; then
+    echo "couldn't read an environment variable via \env" 1>&2
+    exit 1
+fi
+
+if [ "$(FOO=foo $VALGRIND $Q $JQ --sandbox -nr 'env.FOO')" != null ]; then
+    echo "\env should have been empty due to the sandbox flag" 1>&2
```
emanuele6:
```console
$ mksh
$ echo "\env should have been empty due to the sandbox flag"
v should have been empty due to the sandbox flag
$ echo '\env' | od -vAn -tc                                                    
 033   n   v  \n
```

itchyny:
The `input`, `inputs` are safe, but I think wader is talking about `input_filename` and `input_line_number`.

tests/shtest
jemc: Thanks for noticing the copy-paste mistake here.

I've pushed an amended version of the commit to fix that issue.

jemc:
I've pushed an amended commit that fixes the copy/paste mistake in the test echo strings noticed by @emanuele6.

Regarding `input_filename` and `input_line_number`:

There's possibly an argument for saying that `input_filename` should hide the real filename(s) of the input, but I personally don't find it that compelling. However, if someone here wants to specify a special behavior for that case, I will happily implement it as specified. For reference, note that `input_filename` is the string `"<stdin>"` when STDIN is used as the input.

For `input_line_number` I don't think it's leaking any unexpected security-sensitive information. It lets you examine an aspect of the input which is outside of the scope of JSON data structures, but it's still part of the input, so I think it's definitely fair game for a sandboxed filter. Please note that it also works properly when used with STDIN input, so it's not a feature that needs to peek at a real filesystem file to work properly - it's only looking at the string of the input.

jemc:
@itchyny @wader @emanuele6 - can any of you give this PR a deeper review and approve it (or give further comments on things to be improved)?

emanuele6:
I am not personally particularly interested in this feature as it is implemented, so I can't comment much on it.

itchyny:
Sorry for waiting, but I think we need more discussion on this topic. Using `seccomp(2)` could be better, or simple `chroot(2)` might be enough. Yes, platform-independent block like this PR could be the best. For cli usage we already provide Docker image so that might be enough. Anyway I think we need to collect more opinions, we are unpaid volunteers with the fear of being blamed for security issues.

wader:
Good points @itchyny. It would be great to hear what @nicowilliams thinks and how it would affect other future sandbox efforts.

One thing might be to very clearly document what `--sandbox` actually means and also mention other methods to use if stronger isolation is needed like wasm, chroot, seccomp etc.

jemc:
If there's a fear that `--sandbox` "implies too much" about the security properties of the feature, or otherwise create confusion, I could rename it to more specifically pinpoint what it actually does.

Naming is hard, but perhaps something like `--disable-external-file-access` would be less objectionable?

The main point here is that `jq` has some features which allow for external file access, and at minimum I just want a way to turn those features off.

pkoppstein:
In the interests of making some progress on this feature, which in some way or another has been discussed since at least [March 2017](https://github.com/jqlang/jq/issues/1361#issuecomment-284206217), I'd like to propose that --sandbox be added as an experimental feature, or at least one with explicit indications that certain details are subject to change.

A related thought is that, since there are so many different considerations and criteria regarding "security" and "sandbox", the "sandbox" flag could refer to a JSON object that specifies which "security flags" are to be enabled.  E.g. {"now": false, "env": false} would mean that access to `now` and `env` are disabled.

jemc:
> A related thought is that, since there are so many different considerations and criteria regarding "security" and "sandbox", the "sandbox" flag could refer to a JSON object that specifies which "security flags" are to be enabled. E.g. {"now": false, "env": false} would mean that access to now and env are disabled.

Are there any other existing examples in `jq` of a CLI option that takes a JSON object as its value? I didn't find one.

Seems more idiomatic to have separate flags that control separate features.

Again, I'm only interested in disabling the features that allow for arbitrary file access from within the filter, so if I need to rename this flag to only mention the file restriction, I'm happy to rename it. Other flags could be added for controlling other features.

I just need guidance from an authoritative maintainer on what the flag name and desired behavior should be, and I can adapt the PR to match.

jemc:
It's been about 6 months, so I'm checking in again.

Any authoritative maintainer available to give direction on what changes are needed to get this across the finish line?

PR #3095, opened 2024-04-18
main.c: remove a FIXME

emanuele6:
Using an "invalid" variable name is fine, it just means you will only be
able to access that argument from $ARGS.named

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -479,7 +479,6 @@ int main(int argc, char* argv[]) {
         options |= EXIT_STATUS;
         if (!short_opts) continue;
       }
-      // FIXME: For --arg* we should check that the varname is acceptable
       if (isoption(argv[i], 0, "args", &short_opts)) {
         further_args_are_strings = 1;
         further_args_are_json = 0;


wader:
Huh didn't know. Help says `--arg name value      set $name to the string value;` so maybe we should check?

emanuele6:
I don't see the point in restring that; `jq -n ... '$ARGS.named'` is a fairly common pattern to construct objects from some shell variables; e.g.
```sh
jq -n \
    --arg foo "$(something "$baz")" \
    --arg bar "$thisthing" \
    '$ARGS.named'
```
We would be preventing users from constructing objects with `this-cool-key` using this method; it is likely that there are scripts that already rely on this.

Maybe the `--help` description could be extended? (Wouldn't that take too much space?)

The man page entry for `--arg` (but only the one for `--arg` even though it also works with `--argjson`, `--slurpfile`, etc) mentions that the arguments are also available in the `$ARGS.named` object, but does not say anything about `$name` only being assigned if `name` is a valid identifer
> --arg name value:
>
> &nbsp;&nbsp;&nbsp;&nbsp;This option passes a value to the jq program as a predefined
> &nbsp;&nbsp;&nbsp;&nbsp;variable. If you run jq with --arg foo bar, then $foo is available
> &nbsp;&nbsp;&nbsp;&nbsp;in the program and has the value "bar". Note that value will be
> &nbsp;&nbsp;&nbsp;&nbsp;treated as a string, so --arg foo 123 will bind $foo to "123".
>
> &nbsp;&nbsp;&nbsp;&nbsp;Named arguments are also available to the jq program as
> &nbsp;&nbsp;&nbsp;&nbsp;$ARGS.named.

wader:
Good points, yeap maybe enough to add some note about it in the more detailed arg docs?

thaliaarchi:
We should reopen and merge this. If I were to do it in my PR, I’d make a new commit, rebased to the front of the series, which would be effectively the same as merging this.

thaliaarchi (commented):
- Should update the manual to note that invalid names are accessible in `$ARGS.named`. I don’t think this detail needs to go in the shorter `--help`.
- All variable names are added to the bytecode as `$`-definitions regardless of validity. Does this matter?

thaliaarchi:
Turns out gojq errors for this:

```
gojq: compile error: invalid variable name: #name#
```

jaq does not error and has all named variables in `$ARGS.named`.

PR #3103, opened 2024-04-30
doc(readme): Add docker usage doc

sheepster1:
Hey, added a short doc on running the tool inline with docker

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,17 @@ Download the latest releases from the [GitHub release page](https://github.com/j
 
 Pull the [jq image](https://github.com/jqlang/jq/pkgs/container/jq) to start quickly with Docker.
 
+
+### Run with Docker
+##### Example: Extracting the version from a `package.json` file
+```bash
+docker run --rm -i ghcr.io/jqlang/jq:latest < package.json '.version'
+```
+##### Example: Extracting the version from a `package.json` file with a mounted volume
+```bash
+docker run --rm -i -v "${PWD}":/pwd -w /pwd ghcr.io/jqlang/jq:latest '.version' package.json
+```
+
 ### Building from source
 
 #### Dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,17 @@ Download the latest releases from the [GitHub release page](https://github.com/j
 
 Pull the [jq image](https://github.com/jqlang/jq/pkgs/container/jq) to start quickly with Docker.
 
+
+### Run with Docker
```
itchyny:
```suggestion
#### Run with Docker
```

This part should be in the section of `Docker Image`, right?

itchyny:
This is nice improvement.

emanuele6:
Thank you

README.md
```diff
@@ -17,6 +17,17 @@ Download the latest releases from the [GitHub release page](https://github.com/j
...
 
+
+#### Run with Docker
+##### Example: Extracting the version from a `package.json` file
+```bash
+docker run --rm -i ghcr.io/jqlang/jq:latest < package.json '.version'
+```
+##### Example: Extracting the version from a `package.json` file with a mounted volume
+```bash
+docker run --rm -i -v "${PWD}":/pwd -w /pwd ghcr.io/jqlang/jq:latest '.version' package.json
```
wader, line 28: One possible improvement could be to do `-v "$PWD:$PWD" -w "$PWD"` to make paths work more as expected

sheepster1: Not sure I understand, Can you maybe share an example command that will act in an unexpected way?

wader: If one would happen to use absolute host paths in arguments etc they won't work, ex:
```sh
docker run --rm -i -v "${PWD}":/pwd -w /pwd ghcr.io/jqlang/jq:latest -f $PWD/file.jq ...
```
but even with the -v -w trick above paths outside host $PWD will not work... so not perfect either.

itchyny: I think this is a slight difference no need to fix. `-f` accepts relative path, `env.PWD` is different, but should be ok.

sheepster1: Opened a PR https://github.com/jqlang/jq/pull/3105

wader: @itchyny yeap not big thing but it has hit me a couple of times before :)

Issue #3112, opened 2024-05-07
tonumber doesn't work on `true` or `false`


davidfetter:
It should because arithmetic on booleans is often a clear way to get complicated things done.

itchyny:
Any example how useful it is? I'm not sure how common this conversion is, and I'm afraid using this tends to make query less readable.

pkoppstein:
> Any example ...

As it happens, I was just working on [`def kruskal`](https://rosettacode.org/wiki/Selection_bias_in_clinical_sciences#jq)
which requires converting a (strictly) boolean array to an array of 0s and 1s.  I would have most likely written
`map(tonumber)` if that had been possible. 

However, I am not sure the pros in this case outweigh the cons, partly because "tonumber" was mainly intended for
parsing numeric strings.  If it were to be extended to boolean, why not to other types, notably `"null"`?   Also, even in cases such as `def krusal`, there is a potential downside to using `tonumber` in that a stray value such as `2` or `"2"` would not be caught as an error.  

I'm not trying to make a general case against polymorphism, but there is a general case to be made against *increasing* polymorphism - the point being that it raises issues about "backward compatibility" if broadly understood.

itchyny:
How about making `length` work on booleans? I've never heard of length of booleans, but neither of length of null, which already works.

davidfetter:
> > Any example ...
> 
> As it happens, I was just working on [`def kruskal`](https://rosettacode.org/wiki/Selection_bias_in_clinical_sciences#jq) which requires converting a (strictly) boolean array to an array of 0s and 1s. I would have most likely written `map(tonumber)` if that had been possible.

I've been managing to punch out of my weight in speeding things up by sneaking around the monster known to speeders-up of code as `The Branch Predictor`. Looking at the code folks have written along this line, it's started to look like a nice way to do things even when speeding-up is not the highest goal.

I suppose I could invent some BS about how mapping bools to bits is somehow important in LLMs, but I just won't do that for fashion. I figure it's cool and shouldn't make things worse for those who don't use it.

> However, I am not sure the pros in this case outweigh the cons, partly because "tonumber" was mainly intended for parsing numeric strings.

Stuff gets repurposed all the time, so "primarily intended" goes to a reverence for the past that would be a lot easier to justify if `jq` were all about the strict mathematical definitions, program proving, and so on. Is that the direction the project is going?

[...]

PR #3113, opened 2024-05-07
Make tonumber work on `true` and `false` #3112

davidfetter:
As title and bug report.


tests/man.test
```diff
@@ -814,8 +814,8 @@ true
 [null, 1]
 
 [.[] | tonumber?]
-["1", "invalid", "3", 4]
-[1, 3, 4]
+["1", "invalid", "3", 4, true, false]
+[1, 3, 4, 1, 0]
```
kirkoman, line 818: I'm just a random passerby in this repo, but having spent an obscene amount of time trying (failing) to get doc-building working yesterday (on Windows, long story), I think what you need to do here is make the change to [docs/content/manual/manual.yml:2539](https://github.com/davidfetter/jq/blob/4698694e37920d13409e8394237a4b0d0d25b254/docs/content/manual/manual.yml#L2539-L2540) and then rebuild the docs per [docs/README.md](/davidfetter/jq/blob/4698694e37920d13409e8394237a4b0d0d25b254/docs/README.md).

Maybe some words and another example in the [`tonumber` section of the manpage](/davidfetter/jq/blob/4698694e37920d13409e8394237a4b0d0d25b254/docs/content/manual/manual.yml#L1454-L1464) might be nice, too?

davidfetter: Thanks! Getting that patch modified as you've suggested.

PR #3114, opened 2024-05-07
source ~/.jq file on windows (fixes #3104)

### LINKED ISSUE #3104: ~/.jq is not sourced on windows
On Windows, `~/.jq` when it is present as a file should be sourced automatically, but it is not.

To be clear, `$HOME/.jq` *is* sourced, however `HOME` is not a standard environment variable on Windows, you must set it manually.  The corresponding environment variable on Windows is `USERPROFILE`.

**To Reproduce**
```jq
# ~/.jq:
def a:
  "a";
```
```
C:\Users\me>jq -n "a"
jq: error: a/0 is not defined at <top-level>, line 1:
a
jq: 1 compile error

C:\Users\me>set HOME=%USERPROFILE%

C:\Users\me>jq -n "a"
"a"
```

**Environment**

- Windows 11
- jq 1.7.1

kirkoman:
Sourcing of `~/.jq` (file) was hardwired to use the `HOME` environment variable, which is not defined when running in a native Windows shell (PowerShell, CMD).  This change lets it check the `USERPROFILE` environment variable and other standard fallbacks.  `HOME` still has priority if it is set (e.g., in MSYS2 shell).

### DIFF
--- a/src/linker.c
+++ b/src/linker.c
@@ -420,14 +420,16 @@ int load_program(jq_state *jq, struct locfile* src, block *out_block) {
     return 1;
   }
 
-  char* home = getenv("HOME");
-  if (home) {    // silently ignore no $HOME
-    /* Import ~/.jq as a library named "" found in $HOME */
+  jv home = get_home();
+  if (jv_is_valid(home)) {
+    /* Import ~/.jq as a library named "" found in $HOME or %USERPROFILE% */
     block import = gen_import_meta(gen_import("", NULL, 0),
         gen_const(JV_OBJECT(
             jv_string("optional"), jv_true(),
-            jv_string("search"), jv_string(home))));
+            jv_string("search"), home)));
     program = BLOCK(import, program);
+  } else {    // silently ignore if home dir not determined
+    jv_free(home);
   }
 
   nerrors = process_dependencies(jq, jq_get_jq_origin(jq), jq_get_prog_origin(jq), &program, &lib_state);
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -3691,7 +3691,8 @@ sections:
       For example, with `-L$HOME/.jq` a module `foo` can be found in
       `$HOME/.jq/foo.jq` and `$HOME/.jq/foo/foo.jq`.
 
-      If `$HOME/.jq` is a file, it is sourced into the main program.
+      If `.jq` exists in the user's home directory, and is a file (not a
+      directory), it is automatically sourced into the main program.
 
     entries:
       - title: "`import RelativePathString as NAME [<metadata>];`"
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -4109,7 +4109,7 @@ Consecutive components with the same name are not allowed to avoid ambiguities (
 For example, with \fB\-L$HOME/\.jq\fR a module \fBfoo\fR can be found in \fB$HOME/\.jq/foo\.jq\fR and \fB$HOME/\.jq/foo/foo\.jq\fR\.
 .
 .P
-If \fB$HOME/\.jq\fR is a file, it is sourced into the main program\.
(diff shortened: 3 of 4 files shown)


docs/content/manual/manual.yml
```diff
@@ -3662,7 +3662,7 @@ sections:
       For example, with `-L$HOME/.jq` a module `foo` can be found in
       `$HOME/.jq/foo.jq` and `$HOME/.jq/foo/foo.jq`.
 
-      If `$HOME/.jq` is a file, it is sourced into the main program.
+      If `~/.jq` is a file, it is sourced into the main program.
```
wader: Maybe we want some separate comment about windows here? do windows users usually known where `%USERPROFILE%/.jq` would be in the file system, give some hint? (i'm a windows noob so have no idea :) )

wader:
Possible to add a test for this? we have some $HOME related tests here https://github.com/jqlang/jq/blob/master/tests/shtest#L355 but not sure how messy to emulate a more normal windows env?

Otherwise code change look fine with me and thanks for looking into this!

kirkoman:
Ok yeah I did see those tests and briefly puzzled over them running in an msys environment. I will see if I can get at least something simple in there exercising the `USERPROFILE` case.

docs/content/manual/manual.yml
kirkoman: I could use plain language applicable to everyone, like this?:
```suggestion
      If `.jq` is a file in the user's home directory, it is sourced
      into the main program.
```
But my opinion is that `~` is going to be understandable to most Windows command-line users, especially those who would be using `jq`.

While historically a Unix-ism, and not expanded by the older CMD shell, it has been adopted by PowerShell.  Also, `~` expansion is touched on a few lines up from here in the model search path.

Let me know if you want me to change it.

src/linker.c
```diff
@@ -420,13 +420,13 @@ int load_program(jq_state *jq, struct locfile* src, block *out_block) {
...
+  jv home = get_home();
+  if (jv_is_valid(home)) {    // silently ignore if home dir not determined
+    /* Import ~/.jq as a library named "" found in $HOME or %USERPROFILE% */
     block import = gen_import_meta(gen_import("", NULL, 0),
         gen_const(JV_OBJECT(
             jv_string("optional"), jv_true(),
-            jv_string("search"), jv_string(home))));
+            jv_string("search"), home)));
     program = BLOCK(import, program);
   }
```
wader, line 433: Is the valgrind complaint about a missing `else jv_free(home);` here? and i guess it showed up now because shtest on linux etc now runs without HOME set? maybe the window and USERPROFILE test should be separate and not affect the exiting ones?

kirkoman: Yes, I had a memory leak.  Valgrind found it because my test didn't correctly restore `HOME` back to its former value (I only `HOME=$HOME_BAK` instead of `export HOME=$HOME_BAK`).  This caused the following test to pass through the `HOME`-not-found codepath.

I've fixed the memory leak and the test.

docs/content/manual/manual.yml
kirkoman: I went ahead with the language change.

kirkoman:
Test added (after some minor flopping and twitching).

docs/content/manual/manual.yml
wader: 👍 either way works for me for me

wader (approved):
Looks good to me! let's see if some other maintainer has opinions

itchyny (approved):
Thank you!

itchyny:
@kirkoman Could you rebase the changes? The dev manual should be updated.

wader:
Forced pushed conflict fix (felt guilty as i was the one who moved the file)

itchyny:
Thanks!

kirkoman:
Sorry wasn't tuned in. Thanks for the rebase and the merge!

wader:
@kirkoman 🥳

PR #3120, opened 2024-05-14
Make tests pass when --disable-decnum

nicowilliams:
`make check` fails when `--disable-decnum` is used.  This PR makes `make check` pass in that case, though at the price of uglifying the manual a bit because the decnum-dependent mantests in the manual need to use a new builtin in a conditional expression to select an expected result.


src/builtin.c
```diff
@@ -1744,6 +1753,8 @@ BINOPS
   {f_now, "now", 1},
   {f_current_filename, "input_filename", 1},
   {f_current_line, "input_line_number", 1},
+  {f_have_decnum, "have_decnum", 1},
+  {f_have_decnum, "have_literal_numbers", 1},
```
wader, line 1831: If we want to pollute the namespace less we could have one function or a binding that is an object with various build configuration stuff? could also move `--build-configuration` there. Ex something like `jq -n buildconf` -> `{"decnum": true, ..., "args:" "--disable-debug..."}`

wader (approved):

src/builtin.c
nicowilliams: We already have `$JQ_BUILD_CONFIGURATION`!

```
: ; jq -cn '$JQ_BUILD_CONFIGURATION'
"--prefix=/tmp/jq --with-oniguruma=builtin --enable-maintainer-mode --srcdir=.. 'CFLAGS=-g -ggdb3 -gdwarf-4 -O0 -pthread' --disable-decnum --enable-docs"
: ; 
```

nicowilliams: However, `$JQ_BUILD_CONFIGURATION` is not an object.  It's a string, and while one could parse it, that would be a bit gross, though at least `./config.status --config` emits something that is easy enough to parse.  E.g., `./configure --prefix=/tmp/jq --with-oniguruma=builtin --enable-maintainer-mode --srcdir=.. CFLAGS="-g -ggdb3 -gdwarf-4 -O0 -pthread" --disable-decnum --enable-docs` becomes `--prefix=/tmp/jq --with-oniguruma=builtin --enable-maintainer-mode --srcdir=.. 'CFLAGS=-g -ggdb3 -gdwarf-4 -O0 -pthread' --disable-decnum --enable-docs` in the output of `./config.status --config`.

Anyways, to answer your question, we still need some short-cut for the manual page examples, and what I have now there is rather ugly.  The only way I can think of to do better would be to add some sort of decoration in the manual to indicate that some test is only expected to pass if `USE_DECNUM` or only expected to pass if `!USE_DECNUM`.  WHat would that decoration look like?  Ideas?

I'd like to push this and then maybe look for ways to make the mantests in the manual neater.

nicowilliams:
Why aren't the GitHub Actions running?  Just because of the "conflict" in `jq.1.prebuilt`?

nicowilliams:
Rebased.

src/builtin.c
itchyny: We already split the manual test cases to man.test and manonig.test. We do so by matching the programs, but we can also by adding a field to the manual file indicating the feature support. For the purpose of making test pass without decnum support, that's enough. If there is anyone really want to switch their filters based on the decnum support availability, builtin function will be useful.

PR #3123, opened 2024-05-17
Allow tests to use slurpfile and rawfile arguments

jadutter:
Fixed it so that `jq` can use `--slurpfile` and `--rawfile` when running tests.
I think it's useful to be able to reference a file containing some data instead of needing to repeatedly copy-paste the same input to different test cases. Particularly since if the input is a larger object, then it's difficult to read since it needs to be all on one line in the test file. 

## Simple example

Given the following files

`raw.csv`
```raw.csv
id,name
1,Alice
2,Bob
```

`slurp.json`
```slurp.json
{"foo":"bar"}
```

and this test file

`myjq.test`
```myjq.test
$raw|split("\n")|.[1]|split(",")|.[1]
null
"Alice"

$slurp|.[0]|.foo
null
"bar"

map(($slurp|.[0]|.foo)+.)
["ista","yonx"]
["barista","baryonx"]
```

then this command 
```sh
jq --rawfile raw ./raw.csv --slurpfile slurp ./slurp.json  --run-tests ./myjq.test
```
should output
```
Test #1: '$raw|split("\n")|.[1]|split(",")|.[1]' at line number 1
Test #2: '$slurp|.[0]|.foo' at line number 5
Test #3: 'map(($slurp|.[0]|.foo)+.)' at line number 9
3 of 3 tests passed (0 malformed, 0 skipped)
```

### DIFF
--- a/src/jq_test.c
+++ b/src/jq_test.c
@@ -9,12 +9,12 @@
 #include "jq.h"
 
 static void jv_test();
-static void run_jq_tests(jv, int, FILE *, int, int);
+static void run_jq_tests(jv, jv, int, FILE *, int, int);
 #ifdef HAVE_PTHREAD
 static void run_jq_pthread_tests();
 #endif
 
-int jq_testsuite(jv libdirs, int verbose, int argc, char* argv[]) {
+int jq_testsuite(jv progargs, jv libdirs, int verbose, int argc, char* argv[]) {
   FILE *testdata = stdin;
   int skip = -1;
   int take = -1;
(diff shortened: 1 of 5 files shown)


wader:
I've thought about some kind of binding support in the .test format, either inline or read from a file. Could that be an alternative?

Can't come up with a neat syntax now but something like this:

```
# read json value inline
%%BIND $input
{
   "a": 123
}

.a
$input
123

type
$input
"object"

# read json value from file.json
%%BIND $input file.json
...

```

wader:
Doh i got it wrong, probably more like:
```
# read json value inline
%%BIND $input
{
   "a": 123
}

$input | .a
null
123

$input | type
null
"object"
```

PR #3125, opened 2024-05-21
Add a build step to deal with spurious compile errors when building from source

jbrains:
When I tried to build a new version in the same working directory I'd used to build a previous version, I saw build errors. After a few attempts, I remembered that "make clean" might get past the problem. It seems like a nice thing to suggest to everyone who might fall into the same trap.


wader:
Fine with me. I wonder if we should align all the comments in the code block somehow?

jbrains:
> Fine with me. I wonder if we should align all the comments in the code block somehow?

Done.

wader (approved):

Issue #3127, opened 2024-05-25
The code example for trim ltrim and rtrim from the manual do not work


riknor:
This example from the manual do not work 
https://jqplay.org/jq?q=trim%2C%20ltrim%2C%20rtrim&j=%22%20abc%20%22

![image](https://github.com/jqlang/jq/assets/859655/74ea73e2-3c7c-41bf-a09f-e22bb0c5a350)

wader:
Hello, dup of https://github.com/jqlang/jq/issues/3078 and probably another reason why the online documentation should point to the latest release instead of development version 🤔

If you need a trim function in latest stable release you can define one something like this:
```sh
$ jq -n 'def trim: capture("^\\s*(?<s>.*?)\\s*$"; "m").s; " abc " | trim'
"abc"
```

wader:
Now fixed by #3130 but /dev manual will still have possibly broken jqplay examples

Issue #3078, opened 2024-03-25
Trim is not working well


realtica:
Please test it in: [jqplay](https://jqplay.org/jq?q=trim%2C%20ltrim%2C%20rtrim&j=%22%20abc%20%22)

_The error is general:_
jq: error: trim/0 is not defined at <top-level>, line 1:
trim, ltrim, rtrim
jq: error: ltrim/0 is not defined at <top-level>, line 1:
trim, ltrim, rtrim      
jq: error: rtrim/0 is not defined at <top-level>, line 1:
trim, ltrim, rtrim             
jq: 3 compile errors
exit status 3

wader:
Hey! jqplay uses jq 1.7.1 and the trim functions are current only in the development version of jq and will most likely be in next release.

wader:
It would be nice if jqplay could support different versions of jq, maybe even a "nightly" somehow. But that discussion should probably be at https://github.com/owenthereal/jqplay

wader:
Maybe the main confusion comes from that https://jqlang.github.io/jq/manual/ is docs for the development version and not latest stable release?

rrrnld:
Oh, I agree. This bit me today as well! I used the manual as documentation and tried it in the command line, which didn't work. I assumed that the manual is for the latest stable version and that I had some older version installed. I tried it in the playground -- same result, which confused me even more. My next thought was that maybe I'm looking docs for a fork, or that the docs were generated by an LLM and unreliable (oh no)… only to find this issue. Maybe it makes sense to keep the docs for the stable version, and offer a nightly version of the docs when explicitly asked?

wader:
@heyarne sorry for the confusion, i just made #3130 to hopefully clear things up a bit

wader:
Now fixed by #3130 but /dev manual will still have possibly broken jqplay examples

PR #3130, opened 2024-05-27
website: Make latest release (1.7) be default manual

wader:
Fixes issue with confusion about features that has not been
released yet.

Move deveopment manual.yml to dev/manual.yml
Symlink manual.yml to v1.7/manual.yml
Refactor to share history header with links to other manual versions.
Change man.test and man page generation to use dev/manual.yml

Related to https://github.com/jqlang/jq/issues/3078 https://github.com/jqlang/jq/issues/3127


wader:
Looks like this:

Screenshot 2024-05-27 at 20 41 58

Tested locally doing `jq/docs$ pipenv run python3.12 build_website.py` and serve `output` directory using this Caddyfile:
```
{
	debug
}

http://localhost:3000 {
	root /jq/* output
	uri strip_prefix /jq
	file_server
}
```

docs/templates/manual.html.j2
```diff
@@ -33,7 +33,17 @@
...
-        {{ history | markdownify }}
+        <p>
+          <em>
+            For other versions, see
+            <a href="/jq/manual/v1.7/">1.7</a>,
+            <a href="/jq/manual/v1.6/">1.6</a>,
+            <a href="/jq/manual/v1.5/">1.5</a>,
+            <a href="/jq/manual/v1.4/">1.4</a>,
+            <a href="/jq/manual/v1.3/">1.3</a> or
+            <a href="/jq/manual/dev/">development version</a>.
```
wader: Made these same and absolute for all for less maintenance work and so that links are correct at both /jq/manual/ and /jq/manual/<version>/

itchyny: The `/jq` looks like the `--root` and the links don't work locally?

wader: Wops i messed up the markdown. It was suppose to say "...are correct at both `/jq/manual/` and `/jq/manual/<version>/`". If we want to use a symlink to point /jq/manual/ to current release then that manual will be accessable at two locations so `../` will not work.

With the caddy config that maps `/jq/<path>` to  `<docs output root>/<path>` (same as https://jqlang.github.io/jq/) links work fine for me locally. Or did you mean something else?

I think we want to keep `/jq/manual/` working and not break links?

itchyny: As mentioned in [docs/README.md](https://github.com/jqlang/jq/tree/master/docs) and  [manual](https://jqlang.github.io/jq/download/), we can specify the root path with `--root` to make links working with the `output/` dir, serving with Python easily. So I think the links don't work locally. But the `--root` is a mess, I forgot why I couldn't make links work without this option, so you can cleanup the option.

wader: Ah i see, didn't know. I fixed the link to be relative to root now so i think we can keep it for now.

wader:
I wonder what to do about jqplay links in the dev manual that won't work 🤔 add some warning somewhere or make jqplay version-aware? (might confuse ppl also)

itchyny (commented):
This is nice. Thank you!

My slight concern is the abbreviated name `dev` in the path, but I think that's ok. BTW should we consider follow the trend of inclusiveness and change the default branch name? If we choose `develop`, I'd like to align the path of the manual, but if `main`, I don't care.

wader:
> This is nice. Thank you!

Yeap hope it will address some confusion

> My slight concern is the abbreviated name `dev` in the path, but I think that's ok.

Same as branch name might be good? but ok to merge this now as is and rename later if we decide to rename default branch?

> BTW should we consider follow the trend of inclusiveness and change the default branch name? If we choose `develop`, I'd like to align the path of the manual, but if `main`, I don't care.

Fine with me too. Only concern i have would be that there are build scripts etc out there explicitly specifying "master"?

itchyny:
We should avoid changing the path as much as possible, we'll have to implement redirect for the links in other websites.

wader:
> We should avoid changing the path as much as possible, we'll have to implement redirect for the links in other websites.

Good point, i was mostly concerned to not break the "default" manual `/jq/manual` URL in this change. So should we change /jq/manual/dev to /jq/manual/master in this PR?

itchyny:
I don't think `/master/` in the path is a good name. I was just worried about the case `/dev/` for `develop` branch. But anyway, I think it's better to fix the manual confusion soon rather than waisting time for misalignment against the default branch name in the future. So I'm ok with merging this now. Any thoughts from other maintainers?

itchyny (approved):

wader:
Merge?

wader:
@itchyny ok to merge?

emanuele6 (approved):

PR #3131, opened 2024-05-28
fix option parsing of `--binary` on non-Windows platforms

calestyo:
`-b`/`--binary` should be accepted on all platforms so that they can be used in code that is meant to be portable.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -440,8 +440,8 @@ int main(int argc, char* argv[]) {
         _setmode(fileno(stdin),  _O_BINARY);
         _setmode(fileno(stdout), _O_BINARY);
         _setmode(fileno(stderr), _O_BINARY);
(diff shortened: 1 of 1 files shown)


wader:
Hey, thanks! look good to me. I wonder if we should add a regression test for this in `shtest`?

calestyo:
For this particular case I, personally, would rather it's not worth a test.

The test would only tell when the option parsing itself fails again, which I guess is unlikely, but even if, it would be something that's quickly noticed.

I'd rather like to see tests that check whether the different combinations of `--binary` and the `--raw*` options produce the same results per platform and with single/multiple values returned and with values which contain encoded LF and/or CRLF-newlines (see the point I've mentioned in #3132).

itchyny (approved):
LGTM

PR #3137, opened 2024-06-08
Omit unnecessary matrix variables in CI workflows

itchyny:
We don't have to define `matrix` variables if we list all configurations under the `include` key.
See <https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations>.


itchyny:
Oh, it looks the branch protection rules need to be updated :(

PR #3144, opened 2024-06-24
Implement `add/1`

### LINKED ISSUE #2595: builtins: Add lazy generator versions of `add/0`: `add/1` and `add/2`
(Github doesn't support moving PRs from one repo to another, so I've copy-pasted the text of @emanuele6 's PR from the deleted prior `jqlang/jq` repository below. The relevant branch is [emanuele6:niceadd](https://github.com/emanuele6/jq/tree/niceadd) )

@emanuele6 says:

---

And reimplement `add/0` in terms of those.

Also add documentation.

The `add/2` version may appear redundant since it is mostly equivalent to `add($initial, generator)`, but it is not really because initial is a `$`-argument, so `add/2` will output a result for each of the outputs of the first expression; consider the following examples:
```console
$ echo 1 2 | jq --jsonarg -n 'add(inputs; $ARGS.positional[])' -- 3 4
8
9
$ echo 1 2 | jq --jsonarg -n 'add(inputs, $ARGS.positional[])' -- 3 4
10
$
```
```console
$ jq -n '[] | add(.[], 1)'
1
$ jq -n '[] | add(.[]; 1)'
$ # nothing
```

---
If this gets merged, the implementation of `walk/1` can be fixed using the fix proposed in stedolan/jq#2584.

I think these functions should definitely be builtins since they are very useful to have, it makes sense to have them since they are just a generalisation of the existing `add/0`.
Also I had to write manual `reduce` loops almost every day when solving last year's advent of code puzzles in jq because `add/0` was too unnecessarily slow/heavy for the large inputs, which felt a bit lame. :/

`{min,max}/{1,2}` (and maybe also `{min,max}_by/{2,3}`) should also be added since those would be useful too, but since the implementation of `min/0` and `max/0` implementation seems to be written in C, I will leave removing and rewriting the C implementation in jq in terms of `max/2` and `min/2` to another PR.

myaaaaaaaaa:
Improves the ergonomics of `add` by avoiding the need for wrapper arrays. Assignments should be easier now too:

```diff
-... | .field = ([ ... ] | add) | ...
+... | .field = add(...) | ...
```

Related to #2595


wader:
I like it. What do other ppl say? from the original issue there was also talk about generator variants of `min`/`max`, add those too?

myaaaaaaaaa:
> from the original issue there was also talk about generator variants of `min`/`max`, add those too?

Done

src/builtin.jq
```diff
@@ -6,9 +6,22 @@ def sort_by(f): _sort_by_impl(map([f]));
...
 def unique: group_by(.) | map(.[0]);
 def unique_by(f): group_by(f) | map(.[0]);
-def max_by(f): _max_by_impl(map([f]));
-def min_by(f): _min_by_impl(map([f]));
-def add: reduce .[] as $x (null; . + $x);
+
+def _minmax_by(gen; p; $ismin):
+    reduce (gen | [., p]) as $elem (null;
+      if (. == null) or (($elem[1] < .[1]) == $ismin) then $elem end
+    ) | .[0];
```
wader: I haven't checked but it would probably be a good idea to see what kind of performance impacts it is to do this in jq instead of C, i fear it's quite a lot slower. Maybe it's worth using separate implementation for array and generator?

pkoppstein:
Adding a stream-oriented `add` is long-overdue, so my hope is that its introduction will not be delayed by potentially contentious issues.

Adding stream-oriented versions of min, max, min_by, and max_by would be fine, but it seems to me that existing implementations in C should be retained, both for the sake of efficiency and perhaps to facilitate future C-oriented development.

Maybe the best would be to have one narrowly focused PR for `add`; and one or more separate ones for the rest.

myaaaaaaaaa:
> Adding a stream-oriented `add` is long-overdue, so my hope is that its introduction will not be delayed by potentially contentious issues.
> 

I'm fine with separating the `min`/`max` changes to another PR if there's desire to get `add/1` merged quickly

src/builtin.jq
myaaaaaaaaa: I personally prefer the elegance of a unified implementation (and for what it's worth, the jq version should have better memory usage since it doesn't create a temporary array containing all the `path_exp` results).

~How about leaving the C implementation in (but commented out) for now? That way, this can be easily reverted if the performance impact turns out to affect anyone.~

myaaaaaaaaa:
It looks like `min/1` and `max/1` have been attempted before, and met the same concerns regarding jq vs C performance:
* #2679

Perhaps it would be prudent to discuss the best approach for `min/1` and `max/1` in #2595 first?

For now, I'll revert this PR to focus solely on `add/1`

myaaaaaaaaa:
Friendly ping

emanuele6 (approved):

wader (approved):

itchyny (approved):
Thank you!

PR #3152, opened 2024-07-11
Avoid ctype.h abuse: cast char to unsigned char first.

### LINKED ISSUE #3151: Undefined behaviour in use of <ctype.h> functions
**Describe the bug**
When the argument to the standard C isspace or isalpha functions from &lt;ctype.h> is neither (a) a value representable by the type unsigned char, nor (b) the value of the macro EOF, the behaviour is undefined:

> In all cases the argument is an `int`, the value of which shall be representable as an `unsigned char` or shall equal the value of the macro `EOF`.  If the argument has any other value, the behavior is undefined.
>
> C11, Sec. 7.4 Character handling `<ctype.h>`, p. 200, clause 1

The effect of this undefined behaviour may be wrong answers from dereferencing negative array offsets into other random heap data, crashes from dereferencing negative array offsets into unmapped oblivion, or demons flying out of your nose.

The function f_strptime may pass an arbitrary value of type char, which is controlled by user data input, into the &lt;ctype.h> isspace function, leading to data-controlled undefined behaviour:

https://github.com/jqlang/jq/blob/e2ffb5360a43e9683a90b88070cefa82a9ed2a5b/src/builtin.c#L1477

The function main may pass an arbitrary value of type char, which is controlled by user command-line input, into the &lt;ctype.h> isalpha function, leading to command-line-controlled undefined behaviour:

https://github.com/jqlang/jq/blob/e2ffb5360a43e9683a90b88070cefa82a9ed2a5b/src/main.c#L129

**To Reproduce**

Data-controlled undefined behaviour:
```
echo '"20240711é"' | jq -r 'strptime("%Y%m%d")'
```

Command-line-controlled undefined behaviour:
```
jq -é
```

**Expected behavior**
jq should explicitly cast values of type `char` to `unsigned char`.  This adjusts the values so that they meet the intended domain of the &lt;ctype.h> functions (which are designed for use with functions like fgetc that return EOF-or-unsigned-char values via type int).

The other callers of &lt;ctype.h> functions I found, in src/util.c, already cast values of type `char` to `unsigned char`, or work on the contents of `unsigned char *` arrays.

any

**Additional context**
The &lt;ctype.h> API is confusing and almost everyone gets this wrong when trying to use it to process `char *` strings.

riastradh:
fix #3151

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1474,7 +1474,7 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
   }
 #endif
   const char *end = strptime(input, fmt, &tm);
-  if (end == NULL || (*end != '\0' && !isspace(*end))) {
+  if (end == NULL || (*end != '\0' && !isspace((unsigned char)*end))) {
     return ret_error2(a, b, jv_string_fmt("date \"%s\" does not match format \"%s\"", input, fmt));
   }
   jv_free(b);
--- a/src/main.c
+++ b/src/main.c
@@ -126,7 +126,7 @@ static void die() {
 }
 
 static int isoptish(const char* text) {
-  return text[0] == '-' && (text[1] == '-' || isalpha(text[1]));
+  return text[0] == '-' && (text[1] == '-' || isalpha((unsigned char)text[1]));
 }
 
 static int isoption(const char* text, char shortopt, const char* longopt, size_t *short_opts) {


emanuele6:
Basically the problem is that, if `char` is `signed char` on the target system, and the byte that is being checked is non-ascii e.g. `0x80-0xff`, it will be sign extended to `0xffffffxx` instead of being passed as `0x000000xx`, and that is undefined behaviour.

It is wrong to call `isspace(x)` with a value of `x>0x7f` anyway because it would be checking a single byte instead of a character, but at least it is not undefined behaviour; let's merge this.

Thank you for the patch.

PR #3162, opened 2024-08-20
mktime/0: remove unnecessary  length>=6  input requirement

emanuele6:
This check was redundant and useless since `jv2tm()` used to require 8 elements, and errored anyway if not enough elements were provided.
After commit bc96146, this check also prevents `mktime/0` from accepting now supported inputs with less than 8 elements.

Ref: https://github.com/jqlang/jq/pull/3070#issuecomment-2299902072

Reported-by: itchyny <itchyny@cybozu.co.jp>


itchyny:
Is it intended behavior that `[] | todate` is also accepted?

emanuele6:
@itchyny Yes. I made it so, if a value is missing, `0` is assumed; that is the same as `[0,0,0,0,0,0,0,0]`

itchyny (approved):
Thank you.

PR #3164, opened 2024-08-24
Fix docs for `strptime(fmt)`

cjlarose:
* `strptime(fmt)` returns the "broken down time" representation
* `mktime` consumes the "broken down time" representation and produces a Unix timestamp
* `gmtime` consumes a Unix timestamp and produces the "broken down time" representation

However, the docs for `strptime(fmt)` mixed up the functionality of `mktime` and `gmtime`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,7 +16,7 @@ you have `pipenv` installed, you can install the dependencies by running
 Also, you may need to run `virtualenv -p /usr/bin/python3 venv/` and
 then `source venv/bin/activate`, and only then `pipenv sync`.
 
-Once this is done, rerun `./configure` in the jq root directory and then
+Once this is done, rerun `./configure --enable-docs` in the jq root directory and then
 the `Makefile` will be able to generate the jq manpage.  You can just run
 `make jq.1` to build the manpage manually, and `make tests/man.test` to
 update the manual tests.
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -2258,7 +2258,7 @@ sections:
 
           The `strptime(fmt)` builtin parses input strings matching the
           `fmt` argument.  The output is in the "broken down time"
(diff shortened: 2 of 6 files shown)


docs/README.md
```diff
@@ -16,7 +16,7 @@ you have `pipenv` installed, you can install the dependencies by running
 Also, you may need to run `virtualenv -p /usr/bin/python3 venv/` and
 then `source venv/bin/activate`, and only then `pipenv sync`.
 
-Once this is done, rerun `./configure` in the jq root directory and then
+Once this is done, rerun `./configure --enable-docs` in the jq root directory and then
```
wader, line 19: This needed?

wader (approved):
Looks good, just not sure about adding `--enable-docs`

docs/README.md
cjlarose: At least so far as I understand it, if you don't run `./configure` with `--enable-docs`, then  you won't be able to generate the docs at all.

The relevant part of the [Makefile.am](https://github.com/jqlang/jq/blob/7b15708778e89b1de4d8088eedb217291ea8c293/Makefile.am#L180-L187) shows that an error message printed unless `ENABLE_DOCS` is set.

`ENABLE_DOCS` is provided by [this](https://github.com/jqlang/jq/blob/7b15708778e89b1de4d8088eedb217291ea8c293/configure.ac#L121) use of `AM_CONDITIONAL` in the `configure.am` described in the [GNU automake docs](https://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html).

Lastly, the GitHub Actions configuration specifies the [exact same flag](https://github.com/jqlang/jq/blob/7b15708778e89b1de4d8088eedb217291ea8c293/.github/workflows/manpage.yml#L40) when it's checking if the manpages are up-to-date.

I think it just wasn't really documented anywhere.

itchyny (approved):
Thank you.

PR #3167, opened 2024-08-30
Implement `--experimental-fileargs` cli option

myaaaaaaaaa:
A combination of `--jsonargs` and `--slurpfile`. Slurps all json files passed as a command line argument, and makes them available under `$ARGS.named[filename]`.

```sh
$ jq -n '{"a":"a"}' >a.json
$ jq -n '{"b":"b"}' >b.json
$ jq -n --experimental-fileargs '$ARGS.named' *.json

{
	"a.json": [
		{
			"a": "a"
		}
	],
	"b.json": [
		{
			"b": "b"
		}
	]
}
```

Using this over an `input` builtin allows easy access to filesystem functionality such as `find` and globbing, without having to implement them as jq builtins. Files can also be written-in-place when using snapshots ( #3165 ), since all inputs are read before any writes happen.

Depends on #3165 for tests, marking as draft until that one is merged.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1346,7 +1346,38 @@ static jv f_input(jq_state *jq, jv input) {
     return v;
   return jv_invalid_with_msg(jv_string("break"));
 }
+static jv f_snapshot(jq_state *jq, jv input, jv filename, jv data) {
+  if (jv_get_kind(filename) != JV_KIND_STRING)
+    return ret_error(filename, jv_string("filename must be a string"));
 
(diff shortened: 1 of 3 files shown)


wader:
Maybe not superelegant but one can do this:
```sh
$ jq -n '[inputs | {(input_filename): .}] | add' <(echo 1) <(echo 2)
{
  "/dev/fd/11": 1,
  "/dev/fd/12": 2
}
```

PR #3177, opened 2024-09-14
[Docs] Clarify the behvior of first, last, and nth on empty generators

mousetail:
I was confused by this behavior so thought it might be good to document it.

### DIFF
--- a/docs/content/manual/v1.7/manual.yml
+++ b/docs/content/manual/v1.7/manual.yml
@@ -3005,8 +3005,8 @@ sections:
           Note that `nth(n; expr)` doesn't support negative values of `n`.
 
           `first` and `nth` produce no values if the generator has insufficient
-          elements. `last` currently returns `null` instead but this is
-          a bug and may be changed in a future release.
+          elements. `last` currently returns `null` instead. The behavior
+          of `last` is a bug and may be changed in a future release.
 
         examples:
           - program: "[first(range(.)), last(range(.)), nth(./2; range(.))]"
--- a/docs/content/manual/v1.7/manual.yml
+++ b/docs/content/manual/v1.7/manual.yml
@@ -3005,8 +3005,8 @@ sections:
           Note that `nth(n; expr)` doesn't support negative values of `n`.
 
           `first` and `nth` produce no values if the generator has insufficient
-          elements. `last` currently returns `null` instead. The behavior
-          of `last` is a bug and may be changed in a future release.
+          elements. `last` currently returns `null` instead. `last` might be
+          modified to produce no values in the future.
 
         examples:
           - program: "[first(range(.)), last(range(.)), nth(./2; range(.))]"
--- a/docs/content/manual/v1.7/manual.yml
+++ b/docs/content/manual/v1.7/manual.yml
@@ -3004,10 +3004,16 @@ sections:
           The `nth(n; expr)` function extracts the nth value output by `expr`.
           Note that `nth(n; expr)` doesn't support negative values of `n`.
 
+          `first` and `nth` output `empty` if the generator has insufficient
+          elements. `last` returns `null` instead.
+
         examples:
           - program: "[first(range(.)), last(range(.)), nth(./2; range(.))]"
             input: "10"
             output: ["[0,9,5]"]
+          - program: "[[first(range(.))], [last(range(.))], [nth(5; range(.))]]"
+            input: "0"
+            output: ["[[],[null],[]]"]
 
       - title: "`first`, `last`, `nth(n)`"
         body: |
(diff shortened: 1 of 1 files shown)


docs/content/manual/v1.7/manual.yml
```diff
@@ -3017,10 +3017,16 @@ sections:
 
           The `nth(n)` function extracts the nth value of any array at `.`.
 
+          `first` and `nth` output `empty` if the generator has insufficient
```
wader: I wonder if "produces no output" or "outputs no value" is better then "output empty"? seems to be what is used in other places

itchyny:
Maybe you want to update the docs for `last/1` but editing the wrong place. `jq "[range(.)] | [[first], [last], [nth(5)]]" <<< "0"` yields `[[null],[null],[null]]`.

docs/content/manual/v1.7/manual.yml
mousetail: I found the documentation for `try` and `break` also uses the phrase "output/produce `empty`". In this case I think it's a bit more explicit than "outputs no value".

mousetail:
@itchyny I moved the doc update to the correct section

docs/content/manual/v1.7/manual.yml
wader: Ok, i mostly want consistency but i think i personally prefer "no output" as mentioning empty feel a bit confusing, ex would we document the function `empty` as producing `empty`? if we want to use the word empty maybe it should not be written as a `function` but instad be something like "the empty generator" or "empty output" but then i think i prefer "no output" 🤔

mousetail: I have updated the docs to say "produce no values" instead

wader: 👍 let's see what the other maintainers think

But i do wonder what the reasoning was for `last/1` to return `null`?

mousetail: I assume it was just the easiest to implement. `last` is implemented as:
```jq
def last(g): reduce g as $item (null; $item);
```
Reduce can't really return `empty` fundamentally so you'd need to do something more complex to match the behavior of `first` and `nth`.

At this point it would probably break too many to change it so it's better to just document the current behavior.

wader: True and yeap can't be changed now, i would probably have prefer no output for all of them, also now when look closer maybe `first/1` is the weird one:
```sh
$ jq -cn '[] | [[first], [first(.[])], [last], [last(.[])]]'
[[null],[],[null],[null]]
```

mousetail: One potential "fixed" last that outputs no values for an empty generator would be this:

```jq
def last(g): reduce g as $item ([0,null]; [first+1,$item])|if first>0 then last else empty end;
```

A lot less elagent than the current implementation though.

itchyny: See #1869. This is a bug that should be fixed in the future.

wader: @itchyny 👍  oh long discussion, seems fascinatingly messy :)

mousetail: That branch is 5 years old, I doubt it will still be merged

itchyny: It's better to use byte code to avoid boxing like the PR even if it makes more messy.

wader (approved):

pkoppstein:
My understanding is that PR currently adds the text: 
```
`last` returns `null` instead.
```

Of course, that is true of the current implementation, but if this behavior is to be documented in the manual, there should be a disclaimer to the effect that the behavior may change in the future.

(As @itchyny recently pointed out:
> See https://github.com/jqlang/jq/pull/1869. This is a bug that should be fixed in the future.
)

wader (approved):
I'm debating with myself if it's consistent that `first/0` and `last/0` output `null` for an empty array. It makes sense because they operate on an array instead of a generator?

mousetail:
`first/0` having the same behavior as `.[0]` seems logical to me.

The existence of `first`/`last`/`nth` seems a bit pointless anyways since you can just use `.[0]`, `.[-1]`, `.[n]` which are shorter. But that's an entirely seperate point.

pkoppstein (commented):
Thanks for adding a disclaimer.  Perhaps, however, it would be better if the documentation were agnostic as to whether it's a bug or not, though I would agree that simply describing the behavior as "unspecified" would be a bit silly.  I believe it would be sufficient to say that currently `last(empty)` evaluates to null, but that in future it is likely to evaluate to the empty stream.

wader (approved):

itchyny:
I submitted a patch for the issue: #3179.

PR #3179, opened 2024-09-18
fix: make last(empty) yield no output values

itchyny:
This patch fixes `last/1` to produce no output value when the argument yields nothing. For example, `[last(empty)]` and `[last(range(0))]` now produce `[]` instead of `[null]`. I fixed the issue using byte-coded definition, to avoid unnecessary value boxing. This closes #1869 and closes #3177.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1879,6 +1879,33 @@ BINOPS
 #undef LIBM_DD
 #undef LIBM_DA
 
+// This is a hack to make last(g) yield no output values,
+// if g yields no output values, without using boxing.
+static block gen_last_1() {
+  block last_var = gen_op_var_fresh(STOREV, "last");
+  block is_empty_var = gen_op_var_fresh(STOREV, "is_empty");
+  block init = BLOCK(gen_op_simple(DUP),
+                     gen_const(jv_null()),
+                     last_var,
+                     gen_op_simple(DUP),
+                     gen_const(jv_true()),
+                     is_empty_var);
+  block call_arg = BLOCK(gen_call("arg", gen_noop()),
+                         gen_op_simple(DUP),
+                         gen_op_bound(STOREV, last_var),
+                         gen_const(jv_false()),
+                         gen_op_bound(STOREV, is_empty_var),
+                         gen_op_simple(BACKTRACK));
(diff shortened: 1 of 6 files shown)


mousetail (approved):

src/builtin.c
```diff
@@ -1879,6 +1879,32 @@ BINOPS
 #undef LIBM_DD
 #undef LIBM_DA
 
+// This is a hack to make last(empty) yield no output values without using boxing.
```
wader: Maybe should say "last(g) where g does not output any value" so that it does not sound like it's a special case for exactly argument "empty"?

src/builtin.c
```diff
@@ -1879,6 +1879,32 @@ BINOPS
...
+                         gen_op_bound(STOREV, is_empty_var),
+                         gen_op_simple(BACKTRACK));
+  block if_empty = gen_op_simple(BACKTRACK);
+  return BLOCK(init,
+               gen_op_target(FORK, call_arg),
+               call_arg,
+               BLOCK(gen_op_bound(LOADVN, is_empty_var),
+                     gen_op_target(JUMP_F, if_empty),
+                     if_empty,
+                     gen_op_bound(LOADVN, last_var)));
```
wader, line 1906: Trying to understand what kind of "pesudo"-jq this translates, will
```jq
last(g)
```
be something like (vm is not limited by scoping rules?)
```jq
( true as $is_empty
| ( g as $last
  | false as $is_empty
  | empty
  )
// if $is_empty then empty else $last end
)
```

src/builtin.c
itchyny: I fixed the comment.

src/builtin.c
itchyny: Closer to write with `reduce`;
```jq
def last(g): reduce g as $item (
  { last: null, is_empty: true };
  .last = $item | .is_empty = false
) | if .is_empty then empty else .last end;
```

wader (approved):

itchyny:
Thank you.

01mf02:
This is a very welcome change and brings jq and jaq closer together. Thank you!

PR #3180, opened 2024-09-21
Add `filter` command as an alias for map(select())

bseidenberg:
I (and I've heard others) use this often enough that it makes sense to have more user-friendly syntax.

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -1207,6 +1207,21 @@ sections:
             output: ['{"id": "second", "val": 2}']
 
 
+      - title: "`filter(boolean_expression)`"
+        body: |
+
+          The function `filter(f)` takes an array and returns an array
+          with only the elements for which `f` returns true for that element.
+
+          The behavior is identical to map(select(`f`)) but it's more user
+          friendly.
+
+        examples:
+          - program: 'filter(. >= 2)'
+            input: '[1,5,3,0,7]'
+            output: ['[5,3,7]']
+
+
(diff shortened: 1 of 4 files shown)


bseidenberg:
Not sure what is needed to get this PR reviewed - please let me know.

wader:
Hi, the code itself looks ok but don't think i've heard anyone ask for such addition before. Maybe other maintainers have? we're usually quite conservative with adding new functions.

pkoppstein:
@wader wrote:
> we're usually quite conservative with adding new functions.

Not so long ago, there was a major performance issue that justified such conservatism. It's my understanding that the performance overhead problem has been largely resolved.  Does anyone know what the performance overhead these days is?

Even setting aside general considerations concerning bloat, the argument against adding `filter` is  quite strong, so let me articulate two points that may not be obvious.  One is that 'select' is a `filter` that actually filters in a way that is already quite confusing to anyone steeped in SQL's SELECT, so adding 'filter' to jq can be seen as making a potentially confusing situation even more complex.  

Another, though admittedly lesser, point concerns the wisdom of having two ways to say (in words) the same thing.  ("In words" because I do not mean to include symbolic expressions like `[.[]]` as a word or phrase.) The cadence of maps and selects in a program has a certain appeal that would be interrupted by an interloper such as `filter`.

wader:
> Not so long ago, there was a major performance issue that justified such conservatism. It's my understanding that the performance overhead problem has been largely resolved. Does anyone know what the performance overhead these days is?

Don't know but it's my understand also that performance of builtins.jq is not a big issue now. I was mostly thinking about the same issues you also pointed out, adding on to already confusing names and partially overlapping functionality.

PR #3181, opened 2024-09-22
add skip/2 as the counterpart to limit/2

itchyny:
This PR implements `skip/2` as the counterpart of `limit/2`. I also changed the behavior of `limit/2` with negative count to emit an error, like `nth/2` does. Also, I redefined `nth/2` using `skip/2` to minimize the impact of growth of builtin filters.


wader (approved):
Like it 👍 

BTW i'm looking at #3179 but i'm new to the jq vm so taking some times to understand it

itchyny:
Any comments from another maintainer?

pkoppstein:
Please merge! Thank you!

itchyny:
Thank you!

Issue #3183, opened 2024-09-26
jq manual remake


01mf02:
I would like to extend the jq manual with information about diverging behaviour between different jq implementations (jq, gojq, jaq, ...). For this, I will have to go through the whole manual once, which I think would be a nice opportunity to adapt its format. 

Currently, the jq manual is written as a YAML file, which mostly contains Markdown blocks and examples.
I thought about transforming it to a pure Markdown file, and then creating HTML from it via [Pandoc](https://pandoc.org/).
The advantages of this would be:

- syntax highlighting of JSON / jq code
- smaller input file
- better support for editing (e.g. spellchecking)

To show the feasibility of it, I recreated [a part of the "Conditionals and Comparisons" section of the jq manual in Markdown](https://github.com/user-attachments/files/17150178/conds.md). You can use

    pandoc conds.md -s -o conds.html --section-divs --lua-filter filter.lua

to render this to the [following HTML](https://github.com/user-attachments/files/17150259/jq.manual.excerpt.pdf) (sorry, GitHub doesn't let me upload HTML directly, so I exported the HTML to PDF and uploaded that).

This requires the file `filter.lua`, which I wrote today in about 2 hours (I don't know Lua ^^) and which currently looks like this:

~~~ lua
function Header(el)
  if el.content[1].text == "Examples" then
    --print(dump(el))
    el.attr.classes:insert("examples")
  end
  return el
end

function Code(code)
  code.classes[1] = "jq"
  --print(dump(code))
  return code
end

function CodeBlock(block)
  --print(dump(block))
  --if block.classes[1] == "jq-test" then
  rows = {}
  categories = {"Filter", "Input", "Output"}
  local i = 1
  for line in block.text:gmatch("[^\n]+") do
    local code = pandoc.Code(line)
    local lang = "json"
    if i == 1 then lang = "jq" end
    code.classes[1] = lang
    table.insert(rows, {pandoc.Plain(categories[i] or ""), code})
    i = i + 1
  end
  simple_table = pandoc.SimpleTable(
    "", -- caption
    {pandoc.AlignDefault, pandoc.AlignDefault},
    {0, 0}, -- let pandoc determine col widths,
    {}, -- headers
    rows
  )
  return pandoc.utils.from_simple_table(simple_table)
end
~~~

I am fairly confident that I would be able to convert the whole manual to this Markdown format.
But I'm only going to do this if it's likely that this is going to be merged into jq.
So the question is: Would you consider merging such a change?

[...]

PR #3184, opened 2024-09-27
Escape some code parts in manual.

01mf02:
This is needed to make the Markdown -> HTML conversion work with Pandoc, see https://github.com/jqlang/jq/issues/3183.

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -1011,15 +1011,15 @@ sections:
 
           The key difference between `map(f)` and `map_values(f)` is
           that the former simply forms an array from all the values of
-          `($x|f)` for each value, $x, in the input array or object,
+          `($x|f)` for each value, `$x`, in the input array or object,
           but `map_values(f)` only uses `first($x|f)`.
 
           Specifically, for object inputs, `map_values(f)` constructs
           the output object by examining in turn the value of
-          `first(.[$k]|f)` for each key, $k, of the input.  If this
+          `first(.[$k]|f)` for each key, `$k`, of the input.  If this
           expression produces no values, then the corresponding key
           will be dropped; otherwise, the output object will have that
-          value at the key, $k.
+          value at the key, `$k`.
 
           Here are some examples to clarify the behavior of `map` and
           `map_values` when applied to arrays. These examples assume the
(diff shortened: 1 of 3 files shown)


wader (approved):

wader (approved):
I also always forget the man page 😬

itchyny:
Why do you leave older versions (v1.6, v1.5...)? If you will change all the manuals to Markdown, older versions should be converted as well.

01mf02:
> Why do you leave older versions (v1.6, v1.5...)? If you will change all the manuals to Markdown, older versions should be converted as well.

I originally converted the `v1.7` manual during my tests, but I then noticed that the `dev` manual had some new content compared to `v1.7`, so I converted that as well.
However, as I've tried to motivate in https://github.com/jqlang/jq/issues/3183#issuecomment-2378925555, I think the best solution would be to replace the current version-specific manuals by a single manual that documents all versions. That way, all versions of jq can benefit from an up-to-date manual, and maintenance becomes much easier. I'm motivated to do this work myself, starting from the `dev` manual.

pkoppstein:
@01mf02 wrote:
> all versions of jq can benefit from an up-to-date manual

Do you mean going all the way back to v1.3?  That would be an enormous amount of work, and might well make the manual very difficult to understand.  One of the nice things about the manual, even now, is that it is quite succinct.  

Perhaps with hyperlinks it would be possible to make the result readable, but the risk of introducing inaccuracies would be substantial.   There's also the risk that all the work that you put into this project might not reach a critical mass of acceptance.  

Since I don't understand all the issues involved, let me just ask whether there isn't a less onerous and risky alternative.

01mf02:
> Do you mean going all the way back to v1.3? That would be an enormous amount of work, and might well make the manual very difficult to understand. One of the nice things about the manual, even now, is that it is quite succinct.
> 
> Perhaps with hyperlinks it would be possible to make the result readable, but the risk of introducing inaccuracies would be substantial. There's also the risk that all the work that you put into this project might not reach a critical mass of acceptance.
> 
> Since I don't understand all the issues involved, let me just ask whether there isn't a less onerous and risky alternative.

Indeed, I think I found such an alternative. See #3186.

pkoppstein:
@01mf02 wrote:
> Indeed, I think I found such an alternative. See https://github.com/jqlang/jq/pull/3186.

Excellent!  Thank you.

itchyny (approved):
Aside from the ongoing PR changing various ways about the docs (build system, overall contents...), this small patch itself looks good for me to improve the current manual a bit, so I'll merge this. Thank you.

PR #3187, opened 2024-10-04
Move closer to the standard autotools idiom.

dag-erling:
* Create `src/config.h` instead of passing everything on the compiler command line.

* To reduce the magnitude of this change, add `--include src/config.h` to CFLAGS instead of adding `#include "src/config.h"` at the top of each source file.  Not all compilers support this, but I don't think we care about anything other than gcc and clang at this point.

* Rather than generate `src/config_opts.inc`, emit `JQ_CONFIG` directly into `src/config.h`.

* Rather than generate `src/version.h`, define `JQ_VERSION` as an alias for the standard `PACKAGE_VERSION`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,8 @@ jq.1
 
 # Generated source
 src/builtin.inc
-src/config_opts.inc
+src/config.h.in
+src/config.h
 *.pc
 
 # Autotools junk
@@ -38,8 +39,6 @@ jq-*.zip
 configure
 aclocal.m4
 Makefile.in
-version.h
-.remake-version-h
 config.cache
 *.rpm
 m4/libtool.m4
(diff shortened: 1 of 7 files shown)


itchyny (approved):
This is a nice improvement, thank you.

itchyny:
@dag-erling 
Sorry, I didn't check the behavior changes of updating `version.h` before merging. The version string should be followed by the git commit hash, or `-dirty` suffix if any of the source code is changed. So if I checkout different branch and then `make`, (or even if I edit one line of `main.c` and then `make`), the version string should change. But after merging your PR, it seems that the version string is cached and even `./configure` doesn't change the version. This is critical for maintainers because we can't trust the version string of the issue reports. Can you fix this problem?

itchyny:
Can we use `CONFIGURE_DEPENDENCIES`?

itchyny:
Hmm, I configure as `CONFIG_STATUS_DEPENDENCIES = $(SOURCES)`, but `PACKAGE_VERSION` is cached in `autom4te.cache/` and the version is not re-evaluated. The package version in `AC_INIT` is fairly static, and we can't easily allow `make` to update the version when the git revision (or dirty state) has changed. I understand why we had special rules to make `version.h`, and regret not confirming well before merging this PR, but let's wait for the author's comment.

PR #3188, opened 2024-10-04
Improve reliability of NO_COLOR tests.

dag-erling:
Use `env` to ensure that `NO_COLOR` is set to the desired value (or unset) regardless of any initialization that may occur when `script` forks a shell to run jq in.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -14,6 +14,9 @@ JQ="$JQ -b"
 
 PATH=$JQBASEDIR:$PATH $JQBASEDIR/tests/jq-f-test.sh > /dev/null
 
+SHELL=/bin/sh
+export SHELL
+
 if [ -f "$JQBASEDIR/.libs/libinject_errors.so" ]; then
   # Do some simple error injection tests to check that we're handling
   # I/O errors correctly.
--- a/tests/shtest
+++ b/tests/shtest
@@ -586,17 +586,17 @@ if $test_no_color && command -v script >/dev/null 2>&1; then
   od -tc $d/expect
   od -tc $d/color
   cmp $d/color $d/expect
-  NO_COLOR= faketty $JQ_NO_B -n . > $d/color
+  faketty env NO_COLOR= $JQ_NO_B -n . > $d/color
   printf '\033[0;90mnull\033[0m\r\n' > $d/expect
   od -tc $d/expect
   od -tc $d/color
   cmp $d/color $d/expect
-  NO_COLOR=1 faketty $JQ_NO_B -n . > $d/color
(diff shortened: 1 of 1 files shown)


emanuele6 (dismissed):

dag-erling:
Do you target any platforms where it's not supported?

I suppose I can remove the `-u` and revert the removal of the `unset` line and we'll be no worse off than before. I don't dare try `env -i` (which is portable) because there might be circumstances where we need `PATH` etc. to be set up just right to run the tests correctly. Another option is to wrap the command in `sh -c "..."` so we can set and unset variables without relying on `env` (e.g. `faketty sh -c "unset NO_COLOR; $JQ_NO_B -n ."` and `faketty sh -c "NO_COLOR=1 $JQ_NO_B -n ."`). What do you think?

(The problem I'm trying to fix is that the `NO_COLOR=1` case is failing on my machine because `NO_COLOR` is not being passed through `faketty` and I can't for the life of me figure out why or reproduce it outside of the context of `shtest`.)

emanuele6:
> Do you target any platforms where it's not supported?

I don't know about others, but one of the sytems I usually use to test things is my sdf.org shell account that runs a ~15yo NetBSD, and `env` there doesn't have it.

> I suppose I can remove the `-u` and revert the removal of the `unset` line and we'll be no worse off than before. I don't dare try `env -i` (which is portable) because there might be circumstances where we need `PATH` etc. to be set up just right to run the tests correctly. Another option is to wrap the command in `sh -c "..."` so we can set and unset variables without relying on `env` (e.g. `faketty sh -c "unset NO_COLOR; $JQ_NO_B -n ."` and `faketty sh -c "NO_COLOR=1 $JQ_NO_B -n ."`). What do you think?

You can't exactly do `faketty sh -c ''`: look at the `faketty() { script -qec "$*" /dev/null; }` implementation, but you can probably do something like that tweaking the implementations a bit.

> (The problem I'm trying to fix is that the `NO_COLOR=1` case is failing on my machine because `NO_COLOR` is not being passed through `faketty` and I can't for the life of me figure out why or reproduce it outside of the context of `shtest`.)

This test does not look very robust, I am also noticing that it does not set `SHELL` to `/bin/sh` before invoking `script`.
At least the `util-linux` implementation of `script` invokes `$SHELL`, not `/bin/sh`, to run the specified command, so it will be invoking the login shell of the current user that may not be an `sh` shell (e.g. `tcsh`, `zsh`, `fish`) or may be `/bin/false` or something else. That should also be fixed.

itchyny (approved):

PR #3189, opened 2024-10-09
Overhaul of the Manual

01mf02:
In the last days, I've completely overhauled the jq manual. Some of the most important changes:

* I've regrouped the chapters. I tried to make it such that the chapters build on each other; for example, to explain basic filters such as `|`, it is useful to know how to construct values first, and to understand paths such as `.posts[0]`, it is useful to understand `|` first.
* To help with finding the information, the TOC can now be expanded to show subsections of sections.
* I've annotated many filters with compatibility notes that document divergent behaviour for other implementations. Currently, this covers mostly my own jq interpreter jaq, but I expect this to be useful too for gojq.
* I've reworked the visual appearance of the man page, such that longer blocks (such as examples or notes) are visible as such.
* I've corrected bugs in the manual as I went along; for example, the behaviour of arithmetic-update assignments.
* I've also added a large number of new examples.
* I significantly reworked the sections on paths, definitions, and assignments. These are nearly rewrites.

All in all, this PR adds a lot of new information to the jq manual and tries to improve its structure.

To see it in all its glory, you need Pandoc >3.0. Go to `jq/docs`, run `pipenv sync`, then

    pipenv run python3 build_website.py --root /output && python3 -m http.server

and open http://0.0.0.0:8000/output/manual in your browser.

Please let me know if I can do something to get this merged.
This builds on and supersedes my previous PR (#3186).

Screenshot 2024-10-09 at 16-40-21 jq Manual
Screenshot_2024-10-09_16-37-28


pkoppstein:
Very impressive!

I followed your directions for viewing the documentation and would just point out that they assume 'markdown' is already installed.

Here is a batch of comments and suggestions based on a quick perusal.

(1) Re:

Note
The booleans can be defined by:
def  true: 0 == 0;
def false: 0 != 0;

This information box should be deleted, mainly because it isn't really
true as can be seen by running:

jq -n 'def  true: 0 == 0 | debug; true'

(Perhaps you meant 'might have been defined'
but even that would be too much of a distraction here.)

(2) Re :

 but in expressions such 1E1234567890, 

Should read:  ... such as 1E1234567890, 

(3) Re: jq disposes of many builtin functions f

Should read: jq defines many ...

(4) Re: Parenthesis work as a grouping operator 

Should read: Parentheses work ...

(Parenthesis is a singular noun.)

(5) Re: complex path

I think that the phrase "compound path" would be more suitable and
less confusing here.  (It's potentially confusing in part because the
jq manual uses the phrase "complex assignment" appropriately, and the
two concepts should be kept distinct.)

(6) Re:

These filters only produce the values true and false, and so are only useful for genuine Boolean operations, 

Suggestion: omit the clause:

and so are only useful for genuine Boolean operations, 

as it suggests there is something wrong about using `and` and `or`
with non-boolean arguments.  

(7) Re: Note  The expressions f and g and f or g ...

Suggestion: omit the box entirely.

(First, it's a big distraction;  second, the point has been made
elsewhere; third, the last line ("This yields twice true ...")
is confusing. ("This yields true twice, thus certifying the equivalence."
would be clearer.)

01mf02:
Thanks for your remarks, @pkoppstein --- I've integrated all of them in [471b16e](https://github.com/jqlang/jq/pull/3189/commits/471b16ee655d9b18d17e58346a9df4b4baa06d53).

I suppose that `pipenv sync` suffices to install the `markdown` dependency?

pkoppstein:
@01mf02  wrote:
>  I've integrated all of them 

I hope that means you found them useful :-)

> markdown

I just installed it.

Here's Batch #2:

8) Re: 

> It is also possible to define functions in jq, which is used pervasively to define jq’s standard library. (In fact, many jq functions such as map and select are written in jq.)

The above is a bit ungrammatical.  Also, is "standard library" the same as builtin.jq
or does it mean all the defined functions, including builtin.jq?  Anyway it might be 
best to go for brevity here, perhaps something along the lines of

It is also possible to define functions in jq itself. In fact, many of jq's built-in functions,
including `map` and `select`, are written in jq.

9) Re:

> So, there’s generally a cleaner way to solve most problems in jq than defining variables.

Comment: I realize this sentence sentence comes from older versions of
the manual, but I think it's a good time to improve it (especially in
light of the need for variables described in the "Defining Functions"
subsection) or even retire it completely.  After all, the first
paragraph of this subsection already states that "variables aren’t
usually necessary in order to use a value twice".  How about combining
it with the sentence that follows it, perhaps along the lines:

So, variables are often unnecessary and sometimes even best avoided,
but jq does let you define variables using the syntax: f as $x. 


10) Re: def while1( .... 

Comment: Introducing while1 here is confusing.  As I understand it, you
want to juxtapose the two defs for ease of comparison. One way to do this
would be with "#" comments, but if you want to avoid additional commentary,
I would suggest simply renaming "while1" to "while_defined_naively",
or some such.

Also, these wide code blocks (the ones that typically cause a slider to appear)
make reading very difficult.  In this particular case, the problem would
be solved if the two snippets were written using one of the "indentation" forms
of `if then else end` e.g.

   if ...   then ...
   else
   end


11) Re: The filter a = b is equivalent to b as $x | a |= b.

My guess is that you meant `b as $x | a |= $x` here
otherwise it would be pointless at true and incorrect at worst (e.g. if `b` is `input`).
Also, should some caveat or caution be introduced about the assumptions behind the choice of the name of the variable?

12) Re: 
> jaq uses a different approach than jq and gojq to run assignments, which does not construct compound paths during assignments.

There are a couple of English-language problems with this sentence. 
First, one would not normally say "run assignments" here; and secondly,
the "which" is in apposition to "assignments" but is intended to refer
back to jaq.  

How about reworking the entire "Compatibility: note along the following lines:  

jaq's approach to handling assignments is quite different from that of
jq and gojq.  Specifically, jaq executes assignments without
constructing compound path expressions.  This means that jaq does not
allow certain filters on the left-hand side of assignments, notably
`f?` and `label $x | f`.  jaq’s approach is generally more
performant, but in certain scenarios, jaq and jq will produce different 
results, in particular when using f |= empty.  However, for the
examples in this section, jq and jaq yield the same outputs.

01mf02:
I have now also integrated all remarks from your second batch.

> I hope that means you found them useful :-)

Yes, tremendously!

> Also, these wide code blocks (the ones that typically cause a slider to appear) make reading very difficult.

Are you reading the jq manual on a phone? Because I typically do not see a slider appearing ...
Anyway, I followed your suggestion and broke the code example into shorter lines.

> My guess is that you meant `b as $x | a |= $x` here otherwise it would be pointless at true and incorrect at worst (e.g. if `b` is `input`). Also, should some caveat or caution be introduced about the assumptions behind the choice of the name of the variable?

There must be some evil spell around `=` and `+=` that makes every attempt to define these operators erroneous. :)
Thank you for spotting this, I have corrected my mistake. I also clarified that `$x` must be fresh at this point.

> There are a couple of English-language problems with this sentence.

Thank you again so much for taking the time to reformulate my not-so-native sentences. I highly appreciate it.

pkoppstein:
Batch #3 (actually just one comment!)

13) Re:
> When using input, it is generally necessary to invoke jq with the -n command-line option, otherwise the first entity will be lost.

At first blush, this gives the impression that the warning does not apply to `inputs`.
Also, since the first example immediately below shows the opposite case, it might be worthwhile coming
up with a more helpful phrasing, perhaps:

When using `input` and/or `inputs`, it is often necessary to invoke jq with the -n command-line option to avoid losing the first value in the input stream.

01mf02:
> > When using input, it is generally necessary to invoke jq with the -n command-line option, otherwise the first entity will be lost.
> 
> At first blush, this gives the impression that the warning does not apply to `inputs`. Also, since the first example immediately below shows the opposite case, it might be worthwhile coming up with a more helpful phrasing, perhaps:
> 
> When using `input` and/or `inputs`, it is often necessary to invoke jq with the -n command-line option to avoid losing the first value in the input stream.

Good catch. I have integrated your suggestion.

01mf02:
It is now also possible to generate a PDF version of the manual. For this, you need [Typst](https://github.com/typst/typst) (I use 0.11.1).

    pandoc content/manual/manual.md --lua-filter filters/filter.lua -o manual.typ
    typst c manual.typ

Here is the [current output](https://github.com/user-attachments/files/17380695/manual.pdf). The appearance is still a bit rough at some places (especially example tables) --- keep in mind that I have not applied any Typst-specific styling. With a bit of work, this could be made to look quite nice. Right now, this is more of a demo to show that the new manual enables a wide range of output formats with relatively little effort.

01mf02:
Sorry to spam you again, but: @wader, @nicowilliams, @itchyny, @pkoppstein, @emanuele6, may I ask you what the status is on this PR? I've invested quite a significant amount of time into this, and it would be nice to have at least some basic feedback (which @pkoppstein has already given). If you have concerns about certain parts of this PR, then I'd be happy to address these.

docs/filters/filter.lua
```diff
@@ -0,0 +1,106 @@
...
+  return el
+end
+
+function example(test)
+  local _, _, filter, input, output = test:find("([^\n]+)\n([^\n]+)\n(.*)")
+
+  if FORMAT == "man" then
+    filter = "'" .. filter .. "'"
+    input = "'" .. input .. "'"
+    return pandoc.CodeBlock("$ jq " .. filter .. " \\\n  <<< " .. input .. "\n" .. output)
```
wader, line 75: For the web version multiple outputs seems to end up on same line

Screenshot 2024-10-19 at 16 59 03

wader: Maybe example sections should have background color? same as note/compatibility?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+Therefore, jq provides shorthand syntax for these combinations.
+For example:
+
+* `.key[]` for `.key | .[]`,
+* `.[].key?` for `.[] | .key?`,
+* `.[]?[]` for `.[]? | .[]`,
+* `.a.b` for `.a | .b`,
+* `.a.b.c` for `.a | .b | .c`, and so on.
+
+We call such a combination a _compound path_.
```
wader, line 1090: gojq (and jqjq) and [xq](https://github.com/MiSawa/xq) calls them term suffix. Would be great to settle on one. I don't think i have any preference which one really.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+[collect the output of a filter into an array](#arrays).
+Generally, things that would be done with loops and iteration
+in other languages are just done by gluing filters together in jq.
+
+We can run a filter `FILTER` using `jq FILTER`, e.g. `jq .foo`.
+For large filters, it may be more convenient to
+write it into some `FILE` and to
+run it with `jq -f FILE`, e.g. `jq -f filter.jq`.
+
+::: Note
```
wader, line 37: `:::` is a pandoc markdown extension? maybe an alternative could be [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)? pandoc seems to support them. Would make it a bit nicer to read/edit the markdown directly, but maybe as we should not see this markdown as a "consumption" format?

```
> [!NOTE]
>
> When using a [variable](#variables) `$x` as key without surrounding parentheses,
> you can omit a value after `$x`, i.e. you can write `{$x}`. 
> This is equivalent to `{x: $x}`, which constructs
> an object with a field `"x"` that takes the value of `$x`.
```

Looks like:

> [!NOTE]
>
> When using a [variable](#variables) `$x` as key without surrounding parentheses,
> you can omit a value after `$x`, i.e. you can write `{$x}`. 
> This is equivalent to `{x: $x}`, which constructs
> an object with a field `"x"` that takes the value of `$x`.

(weirdly in github comments i get lines breaks per ">" line but in vscode markdown preview it looks better hmm)

But not sure if it's feasible to use for examples also? also not sure how they gets translate for man pages?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+For instance, running the command `jq 'map(.price) | add'`
+takes an array of JSON objects as input and
+returns the sum of their "price" fields.
+
+A jq program, such as `map(.price) | add`, is called a _filter_.
+Each filter takes an input value and produces a _stream_ of output values.
+For instance, when the input value is an array,
+the filter [`.[]`](#iteration-operator) yields all the elements of the array.
+Even literals like `"hello"` or `42` are filters ---
+they take an input and produce the same literal as output.
```
wader: Maybe be even more explicit like "they take an input, that is ignored, and produce the same literal output"?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+However, the jq programming language is quite rich and
+allows for much more than just validation and pretty-printing.
+
+There are many filters for various standard tasks, such as
+[extracting a particular field of an object](#indexing-operator), or
+[converting a number to a string](#tostring).
+
+Filters can be combined in various ways.
+For example, you can
+[feed the output of one filter to another filter](#composition), or
```
wader, line 27: Maybe "into" another filter?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+[collect the output of a filter into an array](#arrays).
+Generally, things that would be done with loops and iteration
+in other languages are just done by gluing filters together in jq.
+
+We can run a filter `FILTER` using `jq FILTER`, e.g. `jq .foo`.
+For large filters, it may be more convenient to
+write it into some `FILE` and to
+run it with `jq -f FILE`, e.g. `jq -f filter.jq`.
```
wader: Maybe "run it using"? (we seem to use "using" more often for run examples)

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+  Terminates argument processing.  Remaining arguments are not
+  interpreted as options.
+
+## Input options
+
+* `--null-input` / `-n`:
+
+  Don't read any input at all. Instead, the filter is run once
+  using `null` as the input. This is useful when using jq as a
+  simple calculator or to construct JSON data from scratch.
```
wader: Wonder if we should mention `input`/`inputs` here?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+  ::: Compatibility
+  jaq does not support this option.
+  :::
+
+## Output options
+
+* `--compact-output` / `-c`:
+
+  By default, jq pretty-prints JSON output. Using this option
+  will result in more compact output by instead putting each
```
wader: Maybe s/putting/outputting/?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+  ::: Compatibility
+  jaq does not support this option.
+  :::
+
+* `--sort-keys` / `-S`:
+
+  Output the fields of each object with the keys in sorted order.
+
+  ::: Compatibility
```
wader, line 235: Mention that gojq always sorts?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+  JSON texts in jq's input and output.  This means that an ASCII
+  RS (record separator) character is printed before each value on
+  output and an ASCII LF (line feed) is printed after every
+  output.  Input JSON texts that fail to parse are ignored (but
+  warned about), discarding all subsequent input until the next
+  RS.  This mode also parses the output of jq without the `--seq`
+  option.
+
+  ::: Compatibility
+  jaq does not support this option.
```
wader: Also not supported by gojq

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+  Use the given number of spaces (no more than 7) for indentation.
+
+* `--unbuffered`:
+
+  Flush the output after each JSON object is printed.
+  This is useful if you pipe a slow data source into jq and
+  pipe jq's output elsewhere.
+
+  ::: Compatibility
+  jaq does not support this option.
```
wader: Also not supported by gojq

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+* `--ascii-output` / `-a`:
+
+  jq usually outputs non-ASCII Unicode codepoints as UTF-8, even
+  if the input specified them as escape sequences (like
+  "\u03bc"). Using this option, you can force jq to produce pure
+  ASCII output with every non-ASCII character replaced with the
+  equivalent escape sequence.
+
+  ::: Compatibility
+  jaq does not support this option.
```
wader: Also not supported by gojq

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+  :::
+
+* `--binary` / `-b`:
+
+  Windows users using WSL, MSYS2, or Cygwin, should use this option
+  when using a native jq.exe, otherwise jq will turn newlines (LFs)
+  into carriage-return-then-newline (CRLF).
+
+  ::: Compatibility
+  jaq does not support this option.
```
wader: Also not supported by gojq

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+## Development options
+
+* `--build-configuration`:
+
+  Output the build configuration of jq and exit with zero.
+  This output has no supported format or structure and may change
+  without notice in future releases.
+
+  ::: Compatibility
+  jaq does not support this option.
```
wader: Also not supported by gojq

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+- 37 (white)
+
+::: Compatibility
+jaq does not consider `JQ_COLORS`.
+:::
+
+
+# Types and Values
+
+jq supports the same set of datatypes as JSON ---
```
wader: Maybe this should be "data types"? seems to be much more common

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+Moreover, build configuration options can alter how jq processes numbers.
+
+The following remarks are therefore offered with the understanding that they
+are intended to be descriptive of the current version of jq and
+should not be interpreted as being prescriptive:
+
+(1) Any arithmetic operation on a number that has not
+already been converted to an IEEE754 double precision
+representation will trigger a conversion to the IEEE754
+representation.
```
wader, line 462: I guess jq will end up using whatever the C compiler uses for double and that is most likely in practice IEEE754? maybe a bit too picky to mention

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+Inside a string, you can put a filter inside parentheses after a backslash,
+such as:
+
+    "Hello \(.name) of planet \(.planet)!"
+
+The output of the filter will be interpolated into the string.
+The example above is equivalent to:
+
+    "Hello " + (.name) + " of planet " + (.planet) + "!"
```
wader: Maybe too much details to say that it's `.name| tostring`?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+::: Examples
+
+~~~
+[.user, .projects[]]
+{"user":"stedolan", "projects": ["jq", "wikiflow"]}
+["stedolan", "jq", "wikiflow"]
+~~~
+
+~~~
+[ .[] | . * 2]
```
wader: Extra whitespace after `[`

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+[2, 4, 6]
+~~~
+
+:::
+
+
+
+## Objects: `{}`
+
+Like in JSON, `{}` is for constructing objects (aka dictionaries or hashes), as in:
```
wader: Above in array section "as in" is used without ":", same here?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+We say that a key is _identifier-like_ when it
+does not start with a digit and
+consists only of alphanumeric characters and underscores;
+for example, `foo` is identifier-like.
+If a key is an identifier-like string, such as `"foo"`,
+then the quotes can be omitted.
+For example, you may write `{"a":42, "b":17}` more succinctly as `{a:42, b:17}`.
+
+Key and value expressions are evaluated with the input of `{}`.
```
wader: Confused me at first, maybe "with the input of the object literal " or "with the input of the `{}`-literal"?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+::: Example
+
+The filter `{("a", "b"): 1}` outputs `{"a": 1}` and `{"b": 1}`.
+
+:::
+
+If an input object contains an entry `"k": v` where `k` is an identifier-like key,
+then we can create a new object `{"k": v}` by writing `{k}`.
+You can use this to select particular fields of an object.
```
wader, line 653: I noticed now that we're using fields and keys a bit interchangeably, makes sense to settle on one?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+~~~
+
+:::
+
+## Function call
+
+jq provides [many builtin functions](#builtin-functions) for a variety of tasks,
+and you can also [define your own functions](#definitions).
+
+Each function has an _arity_ that specifies how many _arguments_ that function takes.
```
wader: Wonder if we should be even more explicit, like "how many _arguments_, in addition to input, that..."?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+~~~
+
+:::
+
+## Recursive descent: `..` {#recursive-descent}
+
+Recursively descends `.`, producing every value.  This is the
+same as [the zero-argument `recurse` function](#recurse).
+This is intended to resemble the XPath `//` operator.
+Note that `..a` does not work; use `.. | .a` instead.
```
wader, line 877: If didn't know jq I would probably guess it means `.a | ..` 🤔

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+In the example below we use `.. | .a?` to find
+all the values of object keys "a" in any object found "below" `.`.
+
+This is particularly useful in conjunction with
+[`path(EXP)`](#path) and the
+[`?` operator](#error-suppression).
```
wader, line 883: My most common use case is `.. | select(<expr>)?`, even have a `grep_by/1` function in fq for it :)

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+In this section, we will show three very frequently used operators, namely for
+[iteration](#iteration-operator),
+[indexing](#indexing-operator), and
+[slicing](#slicing-operator).
+These operators serve to obtain parts of values.
+Furthermore, we will see how to [combine these operators](#compound-paths).
+
+## Iteration operator: `.[]` {#iteration-operator}
+
+The operator `.[]` returns the values contained inside the input value.
```
wader, line 907: Is there any good rule when to use "outputs" instead of returns? i'm thinking for clarity return only for functions that is guaranteed to output one value (or throw error)

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+.foo[]
+{"foo":[1,2,3]}
+1
+2
+3
+~~~
+
+~~~
+.[]
+{"a": 1, "b": 1}
```
wader: Use `"b": 2` for clarity?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+.[2]
+[{"name":"JSON", "good":true}, {"name":"XML", "good":false}]
+null
+~~~
+
+~~~
+.[-2]
+[1,2,3]
+2
+~~~
```
wader, line 1009: Maybe some examples showing that `f` can be an expression outputting multiple keys?
```
{a:1, b:2}
.["a","b"]
1
2
```

```
.[keys[] | select(test("^b"))]
{a:1, b:2, c:3}
2
```

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+.[:3]
+["a","b","c","d","e"]
+["a", "b", "c"]
+~~~
+
+~~~
+.[-2:]
+["a","b","c","d","e"]
+["d", "e"]
+~~~
```
wader, line 1062: Same as above, some advanced example?
```
 .[:range(1;length+1)]
[1,2,3]
[1]
[1,2]
[1,2,3]
```

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+     | "[", t, ":", t, "]"
+     | "[", t, ":",    "]"
+     | "[",    ":", t, "]"
+     ;
+~~~
+
+Here,
+`ident` refers to an [identifier-like key](#objects),
+`t` refers to a filter, and
+`atomic` refers to an _atomic_ filter, such as
```
wader, line 1120: Is literals part of atomic? ex: `{a:1}.a`

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+Then `f * g` outputs the values `3, 6, 4, 8`.
+:::
+
+::: Compatibility
+For any operator in this section such as `+`,
+in jq , `f + g` is equivalent to `g as $y | f as $x | $x + $y`, whereas
+in jaq, `f + g` is equivalent to `f as $x | g as $y | $x + $y`.
+That means that in the above example, jaq outputs the values
+`3, 4, 6, 8` instead of `3, 6, 4, 8`.
+Note that this difference shows only when both `f` and `g` produce multiple values.
```
wader, line 1327: Wonder if jq's behaviour is accidental or on purpose? think personally i would expect jaq's behaviour. For paths something similar have bitten me a few times `[[1,2],[3,4]][0,1][0,1]` -> `1 3 2 4` for jq, which also reminds me of this weirdness `[1,2] | debug[0,1]` in jq, will debug print twice.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+:::
+
+## Ordering: `>`, `>=`, `<=`, `<` {#ordering}
+
+The ordering operators `>`, `>=`, `<=`, `<` return whether
+their left argument is greater than, greater than or equal
+to, less than or equal to or less than their right argument
+(respectively).
+
+Values are ordered as follows, by increasing order:
```
wader: Maybe something like "If two values of different type are ordered the types are ordered instead in the following increasing order"?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+sorted order), and if their keys are equal then the values
+are compared key by key.
+
+::: Examples
+
+~~~
+. < 5
+2
+true
+~~~
```
wader, line 1524: One more advacned and one mixed type example?
```
[1,3] > [1,2]
null
true
```
```
"jq" > false
null
true
```

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+# Boolean filters
+
+Every value can be converted to a boolean --- in particular,
+the values `false` and `null` have the _boolean value_ `false`,
+all other values have the boolean value `true`.
+This section describes several filters that analyze the boolean value of values.
+
+You can negate the boolean value of a value with the builtin function [`not`](#not).
+It is called as a filter to which things can be piped
+rather than with special syntax, as in `.foo and .bar | not`.
```
wader, line 1550: A word of caution about precedence? `false and false | not` -> `true` vs `false and (false | not)` -> `false`.

This also makes me wish jq had `!` as it already has unary `-`

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+"truthiness" than is found in JavaScript or Python, but it
+means that you'll sometimes have to be more explicit about
+the condition you want.  You can't test whether, e.g. a
+string is empty using `if .name then A else B end`; you'll
+need something like `if .name == "" then A else B end` instead.
+:::
+
+More cases can be added to an if using `elif A then B` syntax.
+
+`if A then B end` is shorthand for `if A then B else .  end`.
```
wader: Maybe "is a shorthand for"?

wader: Extra whitespace after "else ."

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+~~~
+
+~~~
+.[] | if . then ., .+1 else . end
+[false, 1, null]
+false
+1
+2
+null
+~~~
```
wader, line 1606: Maybe a multiple output `i` example?
```
if range(0;4) % 2 == 0 then "even" else "odd" end
null
"even"
"odd"
"even"
"odd"
```

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+## `and`, `or`
+
+The filter `f and g` returns true if both `f` and `g` return an output with boolean value `true`.
+The filter `f or g` returns true if either `f` or `g` return an output with boolean value `true`.
+Otherwise, both filters return `false`.
+
+Given two filters `f` and `g`,
+their conjunction `f and g` and
+their disjunction `f  or g`
```
wader: Extra whitespace before "or"

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+null
+42
+42
+1
+~~~
+
+:::
+
+
+# Error handling
```
wader, line 1753: Maybe short summary when errors can happen?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+[.[] | tonumber?]
+["1", "invalid", "3", 4]
+[1, 3, 4]
+~~~
+
+:::
+
+## label-break
+
+A convenient use of try/catch is to break out of control
```
wader: Should this example be moved to above try-catch section? in the current documentation this is under "Breaking out of control structures" which i guess include try-catch?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+:::
+
+The filter `f as $x | g` runs `f` on its input, and
+for each output `y` produced by `f`,
+it runs `g` with the original input and with `$x` set to `y`.
+Thus `as` functions as something of a foreach loop.
+
+Just as `{foo}` is a handy way of writing `{foo: .foo}`, so
+`{$foo}` is a handy way of writing `{foo: $foo}`.
```
wader: Unrelated?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+The filter
+
+    . as {realnames: $names, posts: [$first, $second]} | g
+
+is equivalent to the filter
+
+    .realnames as $names  |
+    .posts     as $p1     |
+    ($p1.[0])  as $first  |
```
wader: What's the reason for use of a temp binding here instead of `.posts[0] as $first`?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+    . as {("a", "b"): $x, ("c", "d"): $y} | [$x, $y]
+
+produces four outputs, namely `[1, 3] [1, 4] [2, 3] [2, 4]`.
+
+:::
+
+When using a filter `(f)` as object key in a pattern, then
+`f` is run with the input that was matched
+by its parent object pattern, not
+by the whole pattern.
```
wader, line 2057: Huh didn't expect that, assumed it was similar to compund/suffix path where all parts get the same input

wader: Ah now i maybe see why you used temp variable above. But maybe in that pattern example case, with only simple indexing, it is easier to follow when not using a temp variable and instead just just "full" paths?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+jq has reduction operators, which can be used to run a filter on
+every element of a stream while keeping some intermediate state.
+These operators are used to define some bits of jq's standard library,
+such as [`add`](#add).
+
+## `reduce`
+
+The `reduce` syntax allows you to combine all of the results of
+an expression by accumulating them into a single answer.
+The form is `reduce EXP as $var (INIT; UPDATE)`.
```
wader, line 2211: In other parts of the documentation we now use single letter to describe part of a syntax. Maybe settle one on way on consistency?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+then the expression is equivalent to:
+
+    foreach (EXP as PATTERN | {$x1, ..., $xn}) as $x (
+      INIT;
+      $x as {$x1, ..., $xn} | UPDATE;
+      $x as {$x1, ..., $xn} | EXTRACT
+    )
+
+Here, `$x` must be chosen to be a fresh variable;
+in particular, it must not occur in `UPDATE` or `EXTRACT`.
```
wader: Not sure i follow what fresh and must not occur means here

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+["foo", "bar", "baz"]
+{"index":1,"item":"foo"}
+{"index":2,"item":"bar"}
+{"index":3,"item":"baz"}
+~~~
+
+:::
+
+
+# Definitions
```
wader: Maybe "Function definitions"? would make it clearer and easer to find i think

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+## Generators and iterators
+
+Some jq operators and functions are actually generators in
+that they can produce zero, one, or more values for each
+input, just as one might expect in other programming
+languages that have generators.  For example, `.[]`
+generates all the values in its input (which must be an
+array or an object), `range(0; 10)` generates the integers
+between 0 and 10, and so on.
```
wader: between 0 and 9?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+# Assignment
+
+jq provides a number of binary assignment operators, such as `|=` and `=`.
+These replace parts of the input at positions
+given by the  left-hand side with outputs
+given by the right-hand side, then return the updated input.
+
+::: Example
+The filter `{a: 1, b: 2} | (.a = 3)` outputs `{a: 3, b: 2}`.
```
wader: Some reason for the parentheses?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+We can use any kind of [compound path](#compound-paths) that starts with `.`
+on the left-hand side of an assignment, such as
+`.[].a` or `.[0]`.
+We'll discuss usage of other filters on the left-hand side
+in [complex assignments](#complex-assignments).
+
+## Update assignment: `|=`
+
+For every position returned by `p`, the update operator `p |= f`
```
wader: Define what "position" means here? or use "value"? something like:

For every value returned by `p`, the update operator `p |= f` replaces that value with the output of `f` applied to the value.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+## Plain assignment: `=`
+
+The plain assignment operator `=` differs from `|=` in two main points:
+First, the input to the right-hand side is the same as the input to
+the left-hand side, not the current value returned by the left-hand side.
+Second, when the right-hand side returns multiple values, then
+the operation is performed for each of these values.
+
+::: Example
```
wader, line 2657: Some very basic examples also? like `{a:1} | .b = 2`?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+which filters can be used on the left-hand side of assignments.
+
+The following filters are path expressions:
+
+* `.` (identity)
+* `..` (recursive descent)
+* compound path: if it starts with some `f`, then `f` must be a path expression
+  * (`.[]` is a path expression because it starts with `.`, which is a path expression)
+  * (`{}[]` is _not_ a path expression, because it starts with `{}`, which is no path expression)
+* `if i then t else e end`: if `i` and `e` are path expressions
```
wader: if `t` and `e` are path expressions?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+::: Compatibility
+
+jaq's approach to handling assignments is quite different from that of jq and gojq.
+Specifically, jaq executes assignments without constructing compound paths.
+This means that jaq does not allow certain filters on the left-hand side of assignments,
+notably `f?` and `label $x | f`.
+jaq's approach is generally more performant, but in certain scenarios,
+jaq and jq will produce different results, in particular when using `f |= empty`.
+However, for the examples in this section, jq and jaq yield the same outputs.
```
wader, line 2904: How does jaq keep track of where it is in the input? for jqjq i internally use a [value,path] pair and when output a value to the "user" it pick the value and for `path/1` it instead picks that path

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+
+:::
+
+
+# Managing large programs
```
wader, line 2909: Maybe in the future this could also include a section about debug techniques etc?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+`$HOME/.jq/foo.jq` and `$HOME/.jq/foo/foo.jq`.
+
+If `.jq` exists in the user's home directory, and is a file (not a directory),
+it is automatically sourced into the main program.
+
+
+
+# Builtin functions
+
+This section documents all named filter functions that are
```
wader: Can skip "named"?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5424 @@
...
+available by default in any jq program.
+
+## Basic functions
+
+### `empty`
+
+`empty` returns no results. None at all. Not even `null`.
+
+It's useful on occasion. You'll know if you need it :)
```
wader, line 3072: Maybe we could mention backtracking and use in recursive functions here? or refer to the section about functions and recursion?

wader (commented):
Nice work! note that as i'm reading thru nearly all of the document I might have added comments about texts that are already present.

Some questions/issues/reflections:
- Read/Edit directly in markdown is much nicer than yaml i would say
- What to do with old versions of the manual? Keep as is? Try move things into one big manual with version compatibility section etc?
- How to get it merged? approval from 3 maintainers is enough? also i think personally i would rather merge early, as it includes so many improvements, and then fix up stuff in further PRs rather than keep iterating on the same PR for a long time. That would also maker it easier to work in parallell.

pkoppstein:
@wader wrote:

>  also i think personally i would rather merge early, as it includes so many improvements

I wholeheartedly agree.  In particular, I hope that any potential revisions regarding gojq won't hold things up.  Indeed, there's a lot to be said for waiting until the jq+jaq version of the manual has been merged before embarking on a gojq-related project. Perhaps that's something @itchyny might be willing to undertake?

01mf02:
> Nice work! note that as i'm reading thru nearly all of the document I might have added comments about texts that are already present.

Sorry to answer a bit late --- I have been quite busy with other stuff, and did not have the time so far to integrate your comments.

> Read/Edit directly in markdown is much nicer than yaml i would say

Yes, that's also why I did this change in the first place. :)

> What to do with old versions of the manual? Keep as is? Try move things into one big manual with version compatibility section etc?

I would propose to keep the old versions of the manual up to (including) 1.7, and to start annotating the current version with compatibility notes for added/changed functions. That way, finding out which functions are available in older versions of jq should become much easier, because you have the information in the manual. (Nowadays, you would have to browse through all manuals from previous versions until you do not find a reference to a function anymore.)

> How to get it merged? approval from 3 maintainers is enough? also i think personally i would rather merge early, as it includes so many improvements, and then fix up stuff in further PRs rather than keep iterating on the same PR for a long time. That would also maker it easier to work in parallell.

Of course, I do not have a say in this, but I wholeheartedly agree with this. It would be really great if some other maintainers would speak up here ...

wader:
> Sorry to answer a bit late --- I have been quite busy with other stuff, and did not have the time so far to integrate your comments.

No worries! hope they are useful

> I would propose to keep the old versions of the manual up to (including) 1.7, and to start annotating the current version with compatibility notes for added/changed functions. That way, finding out which functions are available in older versions of jq should become much easier, because you have the information in the manual. (Nowadays, you would have to browse through all manuals from previous versions until you do not find a reference to a function anymore.)

Yes that is a good idea i think. Would that be per implementation?

> Of course, I do not have a say in this, but I wholeheartedly agree with this. It would be really great if some other maintainers would speak up here ...

Yeap would be great

docs/filters/filter.lua
01mf02: > For the web version multiple outputs seems to end up on same line

This is now corrected in 52ce7951cfc0d122a800177af77d9029c027cd29.

01mf02: > Maybe example sections should have background color? same as note/compatibility?

Example sections are different, because they can be quite long and they are not shown by default. For that reason, I would not color them.

docs/content/manual/manual.md
01mf02: When you have a term like `add[].a`, it seems to me that "term suffix" refers to `[].a`, whereas the concept that I defined as "compound path" refers to `add[].a`. Or am I mistaken?

docs/content/manual/manual.md
01mf02: I just had a shot with GitHub's alerts syntax, and as far as I can see, these only support a small set of note types. So this is not easy to adapt to notes with a custom title, such as "Compatibility" or "Example".

docs/content/manual/manual.md
01mf02: Added in 80ebbfe.

docs/content/manual/manual.md
01mf02: I'm not sure about this: https://en.wiktionary.org/wiki/feed_into
I think in this context, it's more like "feeding food to someone".

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: I am not knowledgeable about this point, but this is taken from the mainstream docs, so I would change this at a later point.

docs/content/manual/manual.md
01mf02: No, you're right, this should be mentioned. Done.

docs/content/manual/manual.md
01mf02: Good catch! Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: I added `{...}` to the header of the section and write "the `{...}` literal" to make this clearer. Same for arrays.

docs/content/manual/manual.md
01mf02: Given that both terms are used quite a bit among the docs (already in the current upstream version), I just clarified that both terms are the same.

docs/content/manual/manual.md
01mf02: I now wrote: "Each function takes an input, as well as a fixed number of _arguments_.
The number of arguments that a function takes is called _arity_."

docs/content/manual/manual.md
01mf02: Do you suggest adding this to the docs?

docs/content/manual/manual.md
01mf02: I think it's best to consider "returns" and "outputs" as synonyms; this allows writing:

> Given two filters `f` and `g`, their concatenation
`f, g` first returns the outputs of `f`, then the outputs of `g`.

If we say that "returns" means "outputs at most one value", then this would need to be written as:

> Given two filters `f` and `g`, their concatenation
`f, g` first outputs the outputs of `f`, then the outputs of `g`.

And that feels quite awkward to me.

docs/content/manual/manual.md
01mf02: Good idea. Done.

docs/content/manual/manual.md
01mf02: The more, the merrier. I added your examples.

docs/content/manual/manual.md
01mf02: Added your example.

docs/content/manual/manual.md
01mf02: Oh god. I did not know about the behaviour of `[[1,2],[3,4]][0,1][0,1]`. That would make the description of compound paths a nightmare. Just checked, and gojq also does it like that. Is there a reason for this?

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Integrated.

docs/content/manual/manual.md
wader: Aha ok, let's use `:::` 👍

docs/content/manual/manual.md
wader: Ok not an expert either, think we can leave it as is

docs/content/manual/manual.md
wader: Maybe something like
```
[`?` operator](#error-suppression), ex `.. | select(EXP)?` will recursively select all values fulfilling 
`EXP` ignoring any errors like indexing a number etc.
```

docs/content/manual/manual.md
01mf02: I documented this behaviour in 4bbc505.

docs/content/manual/manual.md
01mf02: > Maybe "is a shorthand for"?

I found [this](https://www.reddit.com/r/grammar/comments/139idw6/is_it_as_shorthand_for_or_as_a_shorthand_for/), which seems to suggest that "is shorthand for" is quite a bit more common.

01mf02: > Extra whitespace after "else ."

Good catch, corrected.

docs/content/manual/manual.md
01mf02: Added.

docs/content/manual/manual.md
01mf02: This is the beauty of it --- jaq does not have to explicitly keep track of where it is in the input. The information is there implicitly, in the stack. :)

docs/content/manual/manual.md
01mf02: You have eyes like an eagle, my friend. 1539242

docs/content/manual/manual.md
wader: Aha that is a good point, then it makes sense

docs/content/manual/manual.md
wader: Yes agree

docs/content/manual/manual.md
wader: Don't know if accidental or not. Think i talked to @nicowilliams about this on discord some months ago, also about the multi-eval behaviour. I will try find the transcript.

wader: Found this from discord:

Screenshot 2024-11-06 at 13 32 17

... some confusion and debugging cut away ...

Screenshot 2024-11-06 at 13 31 50

docs/content/manual/manual.md
wader: I see! that makes sense, and now i realize this is probably described in your jq papers? i hope i can get more time to read and understand the jaq internals more

docs/content/manual/manual.md
01mf02: Yes, it is. I added a note about this.

docs/content/manual/manual.md
01mf02: Thanks for posting the transcript. I think that this counts as a bug, because it's not actually necessary to evaluate `inputs` as often as it is done here.

docs/content/manual/manual.md
01mf02: Added.

docs/content/manual/manual.md
01mf02: I actually did this on purpose to make the symmetry more visible in the source, but I'll remove it anyway because it is not pretty in the rendered document.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
wader: 😬

docs/content/manual/manual.md
01mf02: Good suggestion, I moved it.

docs/content/manual/manual.md
01mf02: Yes, I removed it.

docs/content/manual/manual.md
01mf02: To make this more general --- when replacing `posts` with `(f)`, for example, this desugaring is still correct, whereas `.[f][0]` could yield wrong results (what if `f` is `input`, for example?).

docs/content/manual/manual.md
01mf02: > Huh didn't expect that, assumed it was similar to compund/suffix path where all parts get the same input

I also got tripped by this. :)

> Ah now i maybe see why you used temp variable above. But maybe in that pattern example case, with only simple indexing, it is easier to follow when not using a temp variable and instead just just "full" paths?

You're probably right; I changed it.

docs/content/manual/manual.md
01mf02: Yes, it's done.

docs/content/manual/manual.md
wader: Yes I think so too and "fixing" it could only affect someone using input functions and multi index? i don't think there are other functions with side-effects? or could it affect error handling etc?

docs/content/manual/manual.md
01mf02: I'm not sure about adding this to the docs right now. I think I will try not to add too many new things to the docs now, because this PR is already getting a bit too large for my taste ...

docs/content/manual/manual.md
01mf02: This would be good, but quite a bit of work. I'd leave this for later.

docs/content/manual/manual.md
01mf02: I now write:

> Here, the name of `$x` must be chosen such that `$x` does not shadow any existing variable at this point.

Is that clearer?

docs/content/manual/manual.md
01mf02: I made this clearer now.

docs/content/manual/manual.md
01mf02: No, I just thought that it might be easier to understand for a beginner. I removed them now.

docs/content/manual/manual.md
01mf02: Thanks for your suggestion, I used it.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: That would be a nice idea.

docs/content/manual/manual.md
01mf02: Yes. Done.

docs/content/manual/manual.md
01mf02: I personally do not see a special link between `empty` and backtracking/recursive functions. `empty` can be useful in so many ways. I'm actually not sure whether `empty` or `.` is my favourite filter. ;)
I would leave it as it is for now, given that this is what the original docs say about `empty`.

01mf02:
OK, I have now integrated all relevant suggestions by @wader, and updated the man page / tests. From my point of view, this is now ready.

01mf02:
(I also integrated all changes to the manual in `master` since the opening of this PR.)

docs/content/manual/manual.md
01mf02: Yes, I described this in detail in my [jq language specification](https://github.com/01mf02/jq-lang-spec/), section 6.

docs/content/manual/manual.md
01mf02: I'm not familiar enough to talk about how jq handles this stuff, but I can think of a few other functions with side effects, like `halt` and `debug`. Error handling should not be affected, I think, but again, given my very limited knowledge of jq internals, I'm not very sure about this.

wader:
What do ppl think about this? think i'm happy about it. Merging this i think will unblock motivations to improve the docs even more and maybe also cleanup/simplify the docs generating processing (maybe generate html directly from md?)

@01mf02 could you do a rebase/regenerate to fix the conflicts?

pkoppstein:
@wader - Yes, long overdue.

01mf02:
@wader, it's done!

wader (approved):

itchyny (changes requested):
A few things I noticed in the manual.

- Example styles are lost; the Examples link should be in `var(--bs-secondary-color)` and the outputs should be aligned each lines. The Run buttons lack the border and `target="_blank"` because they are links to the external website.
- Anchor link icons are lost, this is important for users to notice we have anchor links each sections and filters. Also, I want to keep the existing anchor links; `#recurse` to the recurse filter section.

pkoppstein:
@01mf02 - A tiny enhancement request whenever you get around to it:

The very first line in the subsection on Modules currently reads:

> jq has a library/module system. Modules are files whose names end in .jq.

I think it would be very helpful to add something to the effect that a module cannot have a "main" program. 
 That should make it clear that not every .jq file is a valid jq module, while also making it clear why the word "module" is used.

01mf02:
@itchyny, thanks a lot for your improvement suggestions!

> Example styles are lost; the Examples link should be in `var(--bs-secondary-color)` and the outputs should be aligned each lines. The Run buttons lack the border and `target="_blank"` because they are links to the external website.

I adapted the run buttons and the examples links (although adapting the latter to precisely match the current manual would probably take quite a bit more time, just let me know if the current solution is an issue). Concerning the output alignment, I'm not sure whether I understand the issue, because outputs seem aligned to me:
Screenshot 2024-12-04 at 10-14-05 jq Manual

> Anchor link icons are lost, this is important for users to notice we have anchor links each sections and filters. Also, I want to keep the existing anchor links; `#recurse` to the recurse filter section.

I added the anchor link icons. I also tried to preserve more of the existing anchor links; however, this is sometimes not possible where the structure between the old and the new version has changed too much. Still, much more links should match their old IDs now. (Including your example, `#recurse`.)

01mf02:
> @01mf02 - A tiny enhancement request whenever you get around to it:
> 
> The very first line in the subsection on Modules currently reads:
> 
> > jq has a library/module system. Modules are files whose names end in .jq.
> 
> I think it would be very helpful to add something to the effect that a module cannot have a "main" program. That should make it clear that not every .jq file is a valid jq module, while also making it clear why the word "module" is used.

Good suggestion. Let us do this once this PR is merged. I'm reluctant to make content changes at this point.

docs/content/manual/manual.yml
```diff
@@ -1 +0,0 @@
-v1.7/manual.yml
```
itchyny: The default manual should be kept for the released version 1.7, see #3130. I think you're going to change only the development version to Markdown.

docs/build_website.py
```diff
@@ -91,5 +91,9 @@ def copy_public_files(root=''):
 os.makedirs('output', exist_ok=True)
 copy_public_files()
 
+os.system("./build_manual.sh > content/manual/manual.yml")
+
 for fn in glob.glob('content/**/*.yml', recursive=True):
     generate_file(env, fn)
+
+os.remove("content/manual/manual.yml")
```
itchyny: This path should be in .gitignore.

docs/filters/filter.lua
```diff
@@ -0,0 +1,119 @@
...
+end
+
+function Div(el)
+  local class = el.classes[1]
+  if class == "Examples" then
+    el = el:walk{CodeBlock = function(block) return example(block.text) end}
+
+    if FORMAT == "html" then
+      local summary = pandoc.Plain{
+        pandoc.RawInline("html",  '<summary class="text-body-secondary link-body-emphasis">'), pandoc.Str 'Examples',
```
itchyny, line 51: Screenshot 2024-12-05 at 20 58 03
Using summary tag duplicates the triangle icon on Safari. Also, the font size is big.

docs/templates/manual.html.j2
```diff
@@ -12,17 +12,43 @@
...
+          {%- if not loop.first %}
+              </ul>
+            </details>
+          </li>
+          {%- endif %}
+          <li>
+            <details>
+              <summary>
+                <a href="#{{ section.id }}" class="text-decoration-none py-2">{{ section.title }}</a>
+              </summary>
```
itchyny, line 29: Screenshot 2024-12-05 at 21 01 37
Triagnle icon is duplicated.

itchyny (changes requested):

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+  Like `--stream`, but invalid JSON inputs yield array values
+  where the first element is the error and the second is a path.
+  For example, `["a",n]` produces `["Invalid literal at line 1,
+  column 7",[1]]`.
+
+  Implies `--stream`.  Invalid JSON inputs produce no error values
+  when `--stream` without `--stream-errors`.
+
+  ::: Compatibility
+  jaq does not support this option.
```
itchyny: gojq does not support this option too.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+- 31 (red)
+- 32 (green)
+- 33 (yellow)
+- 34 (blue)
+- 35 (magenta)
+- 36 (cyan)
+- 37 (white)
+
+::: Compatibility
+jaq does not consider `JQ_COLORS`.
```
itchyny, line 409: gojq supports `GOJQ_COLORS`.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+::: Compatibility
+jaq does not provide this function.
+:::
+
+### `input_line_number`
+
+Returns the line number of the input currently being filtered.
+
+::: Compatibility
+jaq does not provide this function.
```
itchyny: gojq does not support this filter.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+
+Note that certain flags may also be specified within REGEX, e.g.
+
+    jq -n '("test", "TEst", "teST", "TEST") | test("(?i)te(?-i)st")'
+
+evaluates to `true`, `true`, `false`, `false`.
+
+::: Compatibility
+jaq uses the [`regex` library](https://docs.rs/regex/latest/regex/) instead of Oniguruma,
+which can result in subtle differences in regex execution.
```
itchyny: gojq uses the [`regexp` package](https://pkg.go.dev/regexp) so it is different from Oniguruma.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+`$x` does not shadow any existing variable at this point.)
+
+A similar transformation can be made for `reduce`.
+
+:::
+
+::: Compatibility
+
+Similarly as for `reduce`,
+jq  considers only the _last_ output of `UPDATE` for the next iteration, whereas
```
itchyny: gojq is compatible with jq in this behavior.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+[EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form):
+
+~~~ ebnf
+path = atomic, part
+     | ".", ident
+     | path, part
+     | path, part, "?"
+     ;
+
+part = ".", ident
```
itchyny: `".", string` is also a `part`.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+
+The rules for what constitutes a compound path are surprisingly complex.
+Therefore, we define it via a formal grammar in
+[EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form):
+
+~~~ ebnf
+path = atomic, part
+     | ".", ident
+     | path, part
+     | path, part, "?"
```
itchyny: How can you parse `.foo??` in this definition?

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+
+~~~
+ascii_upcase
+"useful but not for é"
+"USEFUL BUT NOT FOR é"
+~~~
+
+:::
+
+## String formatting functions
```
itchyny, line 4408: I think these are called string formats. They can be used like filters, but when placed before strings they can format interpolations, so different concept from functions.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+
+~~~
+map(has(2))
+[[0,1], ["a","b","c"]]
+[false, true]
+~~~
+
+:::
+
+### `in`
```
itchyny: `in` is arity 1, so maybe `in(x)` should be correct.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+
+Mind the precedences.
+In particular, the filter
+`false and false | not` is equivalent to
+`(false and false) | not` and yields `true`, whereas
+`false and (false | not)` yields `false`.
+
+:::
+
+## if-then-else-end {#if-then-else}
```
itchyny: Anchor is `if-then-else-end`. I think `end` keyword is necessary.

docs/content/manual/manual.md
```diff
@@ -0,0 +1,5538 @@
...
+
+is equivalent to the filter
+
+    .[0]  as $x  |
+    .[1]  as $p1 |
+    $p1.a as $y  |
+    $x, $y
+
+Given the input `{a: 1, b: []}`, it returns twice `null`,
+because neither `.b` nor `.c[0]` are present in the input.
```
itchyny: You mean `.[0]` nor `.[1].a`?

docs/content/manual/manual.yml
01mf02: Done.

docs/build_website.py
01mf02: Done.

docs/filters/filter.lua
01mf02: I tried to rectify this now.

docs/templates/manual.html.j2
01mf02: Done.

docs/content/manual/manual.md
01mf02: `.foo?` is a path, and `.foo??` is equivalent to `(.foo?)?`; that is, catching all errors from the path `.foo?`. We can see this by the following:

~~~
$ jq -n '.[error]?'
jq: error (at <unknown>) (not a string): null
$ jq -n '.[error]??'
$ jq -n '(.[error]?)?
~~~

(Here, the last two commands do not output anything, because the second `?` silences all errors of the path, including the error from `error`.)

docs/content/manual/manual.md
01mf02: I changed the example, it was wrong. Thanks a lot for spotting this!

docs/content/manual/manual.md
01mf02: Thanks for spotting this!

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: The jq 1.7 manual says: "The @foo syntax is used to format and escape strings [...]", but it never really gives an explicit name for `@...`. You are right that they are not completely like functions, because they can be used before strings (unlike regular functions), but they can be *also* used like functions. I added a note about this to the manual now, but left the name of the section unchanged. Please let me know if you are happy with that, otherwise I can still change it.

docs/content/manual/manual.md
01mf02: Good catch! Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done, and good to know this!

docs/content/manual/manual.md
01mf02: Done.

docs/content/manual/manual.md
01mf02: Done.

01mf02:
Thanks a lot for all your comments, @itchyny!

yochem:
From @wader:
> maybe generate html directly from md?

I think this is a very good idea, and a much more elegant solution than the current manual conversion of md -> yml -> html with the Python code. You could either (1) use a static site generator, like Jekyll (GitHub pages native) or Hugo (faster, more modern, bit easier to use), or (2) since pandoc is already used, just use that to convert everything from md to html and add that to the template. I think (1) is probably the easier approach. It has builtin support for TOC creation, templates, includes and more. I could create a POC for this if you want.

Another benefit would be that you can have a website repo with just the website code, that pulls it's content (the markdown files) from this repo on changes. This separates all markup and website code from this repo, while still keeping the actual documentation close to the code.

wader:
@yochem yeap would be great to look into such things after we get this PR merged, it's already quite large and dragged on for a long time 😅

yochem:
Yes definitely haha! I was not suggesting to include that in this PR, just a suggestion for after that. Maybe would've been better to create a separate issue for it. Sorry for the noice!

01mf02:
> I think this is a very good idea, and a much more elegant solution than the current manual conversion of md -> yml -> html with the Python code. You could either (1) use a static site generator, like Jekyll (GitHub pages native) or Hugo (faster, more modern, bit easier to use), or (2) since pandoc is already used, just use that to convert everything from md to html and add that to the template. I think (1) is probably the easier approach. It has builtin support for TOC creation, templates, includes and more. I could create a POC for this if you want.
> 
> Another benefit would be that you can have a website repo with just the website code, that pulls it's content (the markdown files) from this repo on changes. This separates all markup and website code from this repo, while still keeping the actual documentation close to the code.

I also agree with you, but the problem is that the manual for the older jq versions also need to be generated somehow. And they are in YAML. So while the current solution in this PR is a bit clumsy, it ensures backwards compatibility. If the older manuals were not around, then generating HTML from Markdown directly would become feasible.

docs/filters/filter.lua
```diff
@@ -0,0 +1,119 @@
...
+  if class == "Examples" then
+    el = el:walk{CodeBlock = function(block) return example(block.text) end}
+
+    if FORMAT == "html" then
+      local summary = pandoc.Plain{
+        pandoc.RawInline("html",  '<summary class="text-body-secondary link-body-emphasis small">'), pandoc.Str 'Examples',
+        pandoc.RawInline("html", "</summary>")
+      }
+      return {
+        pandoc.RawBlock("html",  '<details class="pb-3">'), summary, el,
```
itchyny: Can you apply the `small` class to this details tag instead of the summary tag?

docs/content/manual/manual.md
itchyny: A path should be followed by `part` before `?` in your definition, but `.foo??` cannot be parsed because `part` is missing. If any path can be followed by `?`, it should be `path = path "?"`?

docs/filters/filter.lua
01mf02: Yes - done in 7382953.

docs/content/manual/manual.md
01mf02: I agree with you, the EBNF I gave was botched. I have tried to extend it a bit in 549dcdca3e5463901b13d27283b7f7de36616465, please let me know what you think.

01mf02: After some experimentation with paths, I have found that the details of the syntax are really quite tricky and even for same small filters, different implementations return different results. This concerns in particular `??`: For example, `.foo??[]` is rejected by jq 1.7.1 as syntax error, whereas gojq 0.12.16 accepts it. It is also not completely clear to me which kinds of (atomic) filters can be at the head of a path. Therefore, I have removed the formal path syntax from the documentation for now. Perhaps, after this PR is merged, we can clarify what a path actually is, but for now, this should remove one more roadblock to merging.

01mf02:
May I ask what is the status of this PR? @itchyny, are you happy with my answers to your change requests?

wader:
Would be great to get this merged. Are there any blockers or concerns left?

PR #3194, opened 2024-10-29
Parse short options in order given

thaliaarchi:
Fixes:
- Short options cannot be repeated in the same group: e.g., `-nn` fails, but `--null-input --null-input` works
- Short options are processed in an arbitrary fixed order, not the order given by the user: e.g., `-hV`, `-Vh`, `-h -V`, and `--help --version` are processed as `-h`, but `-V -h` and `--version --help` are processed as `-V`
- `-L` cannot be a part of a short option group (as the last one): e.g., `-nLlib` and `-nL lib` fail

Questions:
- [x] `-L` is the only option without a long form and, with this refactor, it would be easy to add. What should it be called?
    - `--module`: this matches the terminology in the manual, but doesn't match the `L` mnemonic
    - `--library`: this matches `-L`
    - `--import-lib`: `-L` doesn't actually import, but rather adds to the search path
    - `--library-path`: this is the conventional flag used by linkers, the inspiration for this option
    - `--library-dir`: avoids confusion with thinking it's the filename of a .jq file or a `:`-separated path
- [x] Needs regression tests for the above fixes. Where should they go?

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -129,20 +129,18 @@ static int isoptish(const char* text) {
   return text[0] == '-' && (text[1] == '-' || isalpha((unsigned char)text[1]));
 }
 
-static int isoption(const char* text, char shortopt, const char* longopt, size_t *short_opts) {
-  if (text[0] != '-' || text[1] == '-')
-    *short_opts = 0;
-  if (text[0] != '-') return 0;
-
-  // check long option
-  if (text[1] == '-' && !strcmp(text+2, longopt)) return 1;
-  else if (text[1] == '-') return 0;
-
-  // must be short option; check it and...
-  if (!shortopt) return 0;
-  if (strchr(text, shortopt) != NULL) {
-    (*short_opts)++; // ...count it (for option stacking)
-    return 1;
+static int isoption(const char** text, char shortopt, const char* longopt, int is_short) {
+  if (is_short) {
+    if (shortopt && **text == shortopt) {
+      (*text)++;
+      if (!**text) *text = NULL;
+      return 1;
+    }
+  } else {
+    if (longopt != NULL && !strcmp(*text, longopt)) {
+      *text = NULL;
+      return 1;
+    }
   }
   return 0;
 }
@@ -344,9 +342,8 @@ int main(int argc, char* argv[]) {
   int further_args_are_json = 0;
   int args_done = 0;
   int jq_flags = 0;
-  size_t short_opts = 0;
   jv lib_search_paths = jv_null();
-  for (int i=1; i<argc; i++, short_opts = 0) {
+  for (int i=1; i<argc; i++) {
     if (args_done || !isoptish(argv[i])) {
       if (!program) {
         program = argv[i];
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -366,223 +363,172 @@ int main(int argc, char* argv[]) {
...
+          options |= UNBUFFERED_OUTPUT;
+        } else if (isoption(&text, 'S', "sort-keys", is_short)) {
+          options |= SORTED_OUTPUT;
+        } else if (isoption(&text, 'R', "raw-input", is_short)) {
+          options |= RAW_INPUT;
+        } else if (isoption(&text, 'n', "null-input", is_short)) {
+          options |= PROVIDE_NULL;
+        } else if (isoption(&text, 'f', "from-file", is_short)) {
+          options |= FROM_FILE;
+        } else if (isoption(&text, 'L', "library-path", is_short)) {
```
wader, line 405: Add to help text?

src/main.c
```diff
@@ -366,223 +363,172 @@ int main(int argc, char* argv[]) {
...
+      int is_short;
+      // First '-' already checked by isoptish
+      if (text[1] == '-') {
+        text += 2;
+        is_short = 0;
+      } else {
+        text++;
+        is_short = 1;
+      }
+      int raw; // Temporary for --rawfile
```
wader, line 375: maybe `is_rawfile`? or keep it more local as in current code?

src/main.c
```diff
@@ -366,223 +363,172 @@ int main(int argc, char* argv[]) {
...
-        ret = jq_testsuite(lib_search_paths,
-                           (options & DUMP_DISASM) || (jq_flags & JQ_DEBUG_TRACE),
-                           argc - i, argv + i);
-        goto out;
-      }
-
-      // check for unknown options... if this argument was a short option
-      if (strlen(argv[i]) != short_opts + 1) {
-        fprintf(stderr, "%s: Unknown option %s\n", progname, argv[i]);
-        die();
```
wader, line 585: Hmm not sure i follow how repeated short option end up here in current code :)

wader (commented):
Nice cleanup :+1: 

> Needs regression tests for the above fixes. Where should they go?

As `tests/shtest` is already quite crowded (should maybe be split up?) i think a new "shtest" file might be a good idea

src/main.c
thaliaarchi: Unfortunately, it can't be made more local and keep the if-else chain. The chain removes the need for all the conditional continues, so I prefer an extra variable over that.

Honestly, I don't think `is_rawfile` gives any extra clarity locally 🤷‍♀️.

src/main.c
thaliaarchi: For short options, the way the current code processes flags is by checking in order: Does it have `L`? Does it have `s`? Does it have `r`? etc. For each of those checks, it scans with `strchr` until the first occurrence of the short option and increments the count `int short_opts`. That means multiple occurrences of the same flag under-count and leave `short_opts <= strlen(argv[i])`.

Either a loop around the option checks is needed, or a loop around the `strchr` scanning. Since some options have effects which matter when processed in a different order (e.g., `-h` and `-V`), it is more correct to process short options in order from the front just like long options.

src/main.c
thaliaarchi: Good catch. I updated the manual too.

thaliaarchi:
> As tests/shtest is already quite crowded (should maybe be split up?) i think a new "shtest" file might be a good idea

I ended up adding tests to tests/shtest, since it seems to fit with other regression tests there. Splitting shtest ought to be a separate PR, if we do it.

thaliaarchi:
On further thought, I also like `--library-dir`, since it avoids confusion with thinking it's the filename of a .jq file or a `:`-separated path.

This option evidently comes from the tradition of ld. [GNU ld](https://www.man7.org/linux/man-pages/man1/ld.1.html), [mold](https://manpages.ubuntu.com/manpages/kinetic/man1/ld.mold.1.html), and [gold](https://manpages.ubuntu.com/manpages/focal/en/man1/gold.1.html) have a corresponding `--library-path` long option and a similar, but unrelated `-l`/`--library` option. lld and rustc have only `-L`. For consistency, I'll stick with `--library-path`.

src/main.c
wader: Ok! work for me, have no strong opinion about it 👍

wader:
> > As tests/shtest is already quite crowded (should maybe be split up?) i think a new "shtest" file might be a good idea
> 
> I ended up adding tests to tests/shtest, since it seems to fit with other regression tests there. Splitting shtest ought to be a separate PR, if we do it.

Good with me

wader (approved):

thaliaarchi:
Just pushed a small fix for insufficient `atoi` error handling for `--indent`.

src/main.c
```diff
@@ -366,223 +363,176 @@ int main(int argc, char* argv[]) {
...
+        } else if (isoption(&text, 'e', "exit-status", is_short)) {
+          options |= EXIT_STATUS;
+        } else if (isoption(&text, 0, "args", is_short)) {
+          further_args_are_strings = 1;
+          further_args_are_json = 0;
+        } else if (isoption(&text, 0, "jsonargs", is_short)) {
+          further_args_are_strings = 0;
+          further_args_are_json = 1;
+        } else if (isoption(&text, 0, "arg", is_short)) {
+          // FIXME: For --arg and --argjson we should check that the varname is acceptable
```
emanuele6: This FIXME is wrong: we most definitely should not do that. I made a PR to remove it, but then it wasn't merged.

Can we at least not readd this FIXME with an updated text? ^^

emanuele6: #3095

thaliaarchi: I was thinking of removing it for the same reason, since the comment predates `$ARGS`. However, I looked into it and saw that syntactically invalid names are still added to the bytecode. Should there be validation for that or does it not matter?

thaliaarchi: Turns out I was wrong in assuming that named args are in the bytecode of a program, so I see no need to add validation:
```
$ jq -n 42 --arg 'bad!' 12345 --debug-dump-disasm
0000 TOP
0001 LOADK 42
0003 RET

42
```
I've pushed a commit removing the comment and updating the manual, and I reverted the change to the text of the comment in my refactor.

We should be good to merge now!

wader (approved):

thaliaarchi:
I've dropped commit “Handle input errors for --indent” from the patch series, because I found a similar number parsing bug with `tonumber`. It will appear in a separate PR shortly.

How much reviewer consensus is needed for a merge? I'm ready on my end.

emanuele6 (approved):

PR #3199, opened 2024-11-08
Clarify -f argument parsing

thaliaarchi:
This clarifies the behavior of `-f` / `--from-file` in the manual and usage. It does not actually take an argument, as documented and commonly believed; it instead modifies the required filter positional argument.

Thus, the following are perfectly valid, though one might assume they are not with the previous documentation:
- `jq -fn f.jq`
- `jq -f -n f.jq`
- `jq f.jq -f`
- `jq --args -f f.jq`
- `jq --args -f -- f.jq`
- `jq --args f.jq -f`
- `jq --jsonargs -f f.jq`

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -221,10 +221,11 @@ sections:
         RS.  This mode also parses the output of jq without the `--seq`
         option.
 
-      * `-f filename` / `--from-file filename`:
+      * `-f` / `--from-file`:
 
-        Read filter from the file rather than from a command line, like
-        awk's -f option.
+        Read the filter from a file rather than from a command line,
+        like awk's -f option. This changes the filter argument to be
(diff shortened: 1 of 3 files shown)


emanuele6:
Thank you; can you also fix `exec jq --args -MRnf "$0" -- "$@"` in the example script?

`"$0"` should go after `--` so `exec jq --args -MRfn -- "$0" "$@"`

thaliaarchi:
Good point. Fixed.

emanuele6 (approved):

wader (approved):

01mf02:
Thank you for clarifying this; I was also under the misconception that `-f` takes an argument. Now I have corrected this in my alternative jq interpreter jaq https://github.com/01mf02/jaq/commit/2ab2e9472810bec39270fc01a14d5c9d64526224. The code has even become smaller due to this change. :)

PR #3203, opened 2024-11-14
fix: populate timezone data when formatting time

### LINKED ISSUE #1912: Wrong timezone names in strftime/strflocaltime for DST

### LINKED ISSUE #2429: strflocaltime always outputs +0000 for %Z
**Describe the bug**
jq outputs wrong timezone information when using strflocaltime with the `%Z` formatter:
```
└─$ timedatectl
               Local time: So 2022-05-08 13:48:24 CEST
           Universal time: So 2022-05-08 11:48:24 UTC
                 RTC time: So 2022-05-08 11:48:24
                Time zone: Europe/Berlin (CEST, +0200)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no

└─$ jq -rn 'now | strflocaltime("%Y-%m-%dT%H:%M:%S%z")' 
2022-05-08T13:48:33+0000
```

**To Reproduce**
`TZ=Asia/Tokyo jq -rn 'now | strflocaltime("%Y-%m-%dT%H:%M:%S%z")'`
```
└─$ TZ=Asia/Tokyo jq -rn 'now | strflocaltime("%Y-%m-%dT%H:%M:%S%z")' 
2022-05-08T20:51:59+0000
```

**Expected behavior**
Output should be like the `date` command:
```
└─$ TZ=Asia/Tokyo date +"%Y-%m-%dT%H:%M:%S%z"
2022-05-08T20:51:20+0900
```

 - OS and Version: Linux; Reproduced on Manjaro and Debian Bullseye
 - jq version: 1.6

**Additional context**
Might be related to https://github.com/stedolan/jq/issues/1912

### LINKED ISSUE #2475: strftime uses wrong time zone for %Z
**Describe the bug**
strftime (in jq) is documented to always format times in the UTC time zone.  However, if %Z is part of the format, will be replaced with the systems local time zone, which is inconsistent with the rest of the time formatting, resulting in unexpected outputs

**To Reproduce**
$ echo '1661932759' | TZ="Europe/Zurich" jq '. | strflocaltime("%Y-%m-%dT%H:%M:%S_%Z"), strftime("%Y-%m-%dT%H:%M:%S_%Z")'
"2022-08-31T09:59:19_CET"
"2022-08-31T07:59:19_CET"

Note: the output is very misleading... the time passed in is NOT 07:59:18 *CET*

**Expected behavior**
$ echo '1661932759' | TZ="Europe/Zurich" jq '. | strflocaltime("%Y-%m-%dT%H:%M:%S_%Z"), strftime("%Y-%m-%dT%H:%M:%S_%Z")'
"2022-08-31T09:59:19_CET"
"2022-08-31T07:59:19_UTC"

Note: the second line correct shows that the printed time is in UTC

 - Linux Debian testing
 - jq-1.6 (Debian package version 1.6-2.1)

marcin-serwin:
The `jv2tm` function was zeroing fields of `struct tm` that were not specified by the standard. However, depending on the libc this produced incorrect timezone data when used together with formatting functions. This change tries to fill the timezone data using either `mktime`, `timegm`, or manually.

fixes #2475; fixes #2429; fixes #1912.

Example:
```shell
TZ=Asia/Tokyo jq -rn '1731627341 | strflocaltime("%F %T %z %Z"), strftime("%F %T %z %Z")'
```

Before:
```
2024-11-15 08:35:41 +0000 JST
2024-11-14 23:35:41 +0000 JST
```

After:
```
2024-11-15 08:35:41 +0900 JST
2024-11-14 23:35:41 +0000 GMT
```

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1594,7 +1594,7 @@ static jv f_strptime(jq_state *jq, jv a, jv b) {
   return r;
 }
 
-static int jv2tm(jv a, struct tm *tm) {
+static int jv2tm(jv a, struct tm *tm, int localtime) {
   memset(tm, 0, sizeof(*tm));
   static const size_t offsets[] = {
     offsetof(struct tm, tm_year),
@@ -1624,13 +1624,25 @@ static int jv2tm(jv a, struct tm *tm) {
     jv_free(n);
   }
 
-  // We use UTC everywhere (gettimeofday, gmtime) and UTC does not do DST.
-  // Setting tm_isdst to 0 is done by the memset.
-  // tm->tm_isdst = 0;
+  if (localtime) {
+    tm->tm_isdst = -1;
+    mktime(tm);
+  } else {
+#ifdef HAVE_TIMEGM
+    timegm(tm);
+#elif HAVE_TM_TM_GMT_OFF
+    // tm->tm_gmtoff = 0;
+    tm->tm_zone = "GMT";
+#elif HAVE_TM___TM_GMT_OFF
+    // tm->__tm_gmtoff = 0;
+    tm->__tm_zone = "GMT";
+#endif
+    // tm->tm_isdst = 0;
 
-  // The standard permits the tm structure to contain additional members. We
-  // hope it is okay to initialize them to zero, because the standard does not
-  // provide an alternative.
+    // The standard permits the tm structure to contain additional members. We
+    // hope it is okay to initialize them to zero, because the standard does not
+    // provide an alternative.
+  }
 
   jv_free(a);
   return 1;
@@ -1642,7 +1654,7 @@ static jv f_mktime(jq_state *jq, jv a) {
   if (jv_get_kind(a) != JV_KIND_ARRAY)
     return ret_error(a, jv_string("mktime requires array inputs"));
   struct tm tm;
-  if (!jv2tm(a, &tm))
+  if (!jv2tm(a, &tm, 0))
     return jv_invalid_with_msg(jv_string("mktime requires parsed datetime inputs"));
   time_t t = my_mktime(&tm);
   if (t == (time_t)-1)
@@ -1740,7 +1752,7 @@ static jv f_strftime(jq_state *jq, jv a, jv b) {
   if (jv_get_kind(b) != JV_KIND_STRING)
     return ret_error2(a, b, jv_string("strftime/1 requires a string format"));
   struct tm tm;
-  if (!jv2tm(a, &tm))
+  if (!jv2tm(a, &tm, 0))
     return ret_error(b, jv_string("strftime/1 requires parsed datetime inputs"));
 
   const char *fmt = jv_string_value(b);
@@ -1771,7 +1783,7 @@ static jv f_strflocaltime(jq_state *jq, jv a, jv b) {
   if (jv_get_kind(b) != JV_KIND_STRING)
     return ret_error2(a, b, jv_string("strflocaltime/1 requires a string format"));
   struct tm tm;
-  if (!jv2tm(a, &tm))
+  if (!jv2tm(a, &tm, 1))
     return ret_error(b, jv_string("strflocaltime/1 requires parsed datetime inputs"));
   const char *fmt = jv_string_value(b);
   size_t alloced = strlen(fmt) + 100;
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1758,7 +1758,21 @@ static jv f_strftime(jq_state *jq, jv a, jv b) {
   const char *fmt = jv_string_value(b);
   size_t alloced = strlen(fmt) + 100;
   char *buf = alloca(alloced);
+#ifdef __APPLE__
+  /* Apple Libc (as of version 1669.40.2) contains a bug which causes it to
+   * ignore the `tm.tm_gmtoff` in favor of the global timezone. To print the
+   * proper timezone offset we temporarily switch the TZ to UTC. */
+  char *tz = getenv("TZ");
+  setenv("TZ", "UTC", 1);
+#endif
   size_t n = strftime(buf, alloced, fmt, &tm);
+#ifdef __APPLE__
+  if (tz) {
+    setenv("TZ", tz, 1);
+  } else {
+    unsetenv("TZ");
+  }
+#endif
   jv_free(b);
   /* POSIX doesn't provide errno values for strftime() failures; weird */
   if (n == 0 || n > alloced)
--- a/tests/shtest
+++ b/tests/shtest
@@ -694,4 +694,26 @@ printf '[\n      {\n            "a": 1\n      }\n]\n' > $d/expected
 $JQ --indent 6 -n "[{a:1}]" > $d/out
 cmp $d/out $d/expected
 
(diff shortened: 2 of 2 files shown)


itchyny:
I'm afraid the timezone offset is not fixed even with this patch.
```
# 1.7.1
 $ TZ=Asia/Tokyo jq -rn '1731627341 | strflocaltime("%F %T %z %Z"), strftime("%F %T %z %Z")'  
2024-11-15 08:35:41 +0900 JST
2024-11-14 23:35:41 +0900 JST

# marcin-serwin:strftime-fix
 $ TZ=Asia/Tokyo ./jq -rn '1731627341 | strflocaltime("%F %T %z %Z"), strftime("%F %T %z %Z")'
2024-11-15 08:35:41 +0900 JST
2024-11-14 23:35:41 +0900 UTC

# 1.7.1
 $ TZ=Europe/Paris jq -rn '1731627341 | strflocaltime("%F %T %z %Z"), strftime("%F %T %z %Z")'
2024-11-15 00:35:41 +0100 CET
2024-11-14 23:35:41 +0100 CET

# marcin-serwin:strftime-fix
 $ TZ=Europe/Paris ./jq -rn '1731627341 | strflocaltime("%F %T %z %Z"), strftime("%F %T %z %Z")'
2024-11-15 00:35:41 +0100 CET
2024-11-14 23:35:41 +0100 UTC

 $ uname -v
Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:41 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T8103
```

marcin-serwin:
This issue is caused by a bug in the Apple's Libc implementation. Specifically in [this line](https://github.com/apple-oss-distributions/Libc/blob/Libc-1669.40.2/stdtime/FreeBSD/strftime.c#L507) the `__DARWIN_UNIX03` macro is negated which means that the `strftime` is ignoring the time offset data present in the `tm struct` in favor of the older heuristic needed by legacy standards. Looking at other uses of this macro, it seems to enable non-legacy behavior so the negation here is likely a bug since this means that the presence of this macro opts into the legacy behavior. The FreeBSD libc, from which the Apple's one is forked, does not have this issue.

I've added a workaround in the second commit that temporarily sets the global timezone to UTC if it is compiled for apple platforms.

wader:
Nice digging 👍 jq can be used as a library does that make it a bad idea to use `setenv`? 🤔 

Possible to add some tests for this?

marcin-serwin:
> jq can be used as a library does that make it a bad idea to use `setenv`? 🤔

`setenv` is already used in the `my_mktime` function for a similar purpose. It can cause issues in multi-threaded programs but I don't know how to avoid that.

> Possible to add some tests for this?

I've added some tests in `tests/shtest`, is this the right place?

marcin-serwin:
I've disabled the tests on windows. The ucrt on Windows always uses the global settings for determining the timezone, which is documented in a note near the end of [the strftime docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strftime-wcsftime-strftime-l-wcsftime-l?view=msvc-170). I tried to make them work similarly to Apple workaround by using `setenv` and `_tzset` but I couldn't make it work and I gave up trying to figure out why. Given that it's a documented behavior for Windows and the `jq` docs warn that `strftime` output depends on the operating system I think it's ok to leave it as is.

wader:
> > jq can be used as a library does that make it a bad idea to use `setenv`? 🤔
> 
> `setenv` is already used in the `my_mktime` function for a similar purpose. It can cause issues in multi-threaded programs but I don't know how to avoid that.

I see 👍

> > Possible to add some tests for this?
> 
> I've added some tests in `tests/shtest`, is this the right place?

Is fine i think

About the apple libc bug: Will things break once/if apple fix it?

Otherwise i think this looks fine but would be great with some more eyes on it

marcin-serwin:
> About the apple libc bug: Will things break once/if apple fix it?

It will still work. With this bug fixed, the code will ignore the `TZ` variable and just look inside the struct for the timezone thus behaving like glibc.

itchyny (commented):
I think this is okay to merge, but sorry for shtest conflict.

itchyny (approved):
LGTM, thank you.

src/builtin.c
```diff
@@ -1740,13 +1752,27 @@ static jv f_strftime(jq_state *jq, jv a, jv b) {
...
     return ret_error(b, jv_string("strftime/1 requires parsed datetime inputs"));
 
   const char *fmt = jv_string_value(b);
   size_t alloced = strlen(fmt) + 100;
   char *buf = alloca(alloced);
+#ifdef __APPLE__
+  /* Apple Libc (as of version 1669.40.2) contains a bug which causes it to
+   * ignore the `tm.tm_gmtoff` in favor of the global timezone. To print the
+   * proper timezone offset we temporarily switch the TZ to UTC. */
+  char *tz = getenv("TZ");
```
itchyny, line 1765: @marcin-serwin 
Oh, we have to copy the string (just like how `my_mktime` does), otherwise we end up failing to restore `TZ`.
See the following code.
```cpp
#include <stdio.h>
#include <stdlib.h>

int main(void) {
  // Set to UTC and restore.
  char *tz = getenv("TZ");
  printf("%s\n", tz); // Etc/GMT+7
  setenv("TZ", "UTC", 1);
  printf("%s\n", tz); // Etc/GMT+7
  setenv("TZ", tz, 1);

  // Set to UTC and try to restore but fail.
  tz = getenv("TZ");
  printf("%s\n", tz); // Etc/GMT+7
  setenv("TZ", "UTC", 1);
  printf("%s\n", tz); // UTC (expected)
  setenv("TZ", tz, 1);

  tz = getenv("TZ");
  printf("%s\n", tz); // UTC
}
```
```sh
 $ TZ=Etc/GMT+7 ./a.out
Etc/GMT+7
Etc/GMT+7
Etc/GMT+7
UTC
UTC
```
This failure appears in the following interesting jq query (of course on macOS).
```sh
 $ TZ=Etc/GMT+7 ./jq -nc '1731627341 | .,. | [strftime("%FT%T"),strflocaltime("%FT%T%z")]'
["2024-11-14T23:35:41","2024-11-14T16:35:41-0700"]
["2024-11-14T23:35:41","2024-11-14T23:35:41+0000"]
```

marcin-serwin: You're right, I've opened https://github.com/jqlang/jq/pull/3264 with the fix.

PR #3204, opened 2024-11-15
Update compile.c to correctly inform user making substitutions with sub("a","b") instead of sub("a";"b")

jidhub:
echo '{"f":"a"}' | jq '.f | sub("a","b")'
above line generates the following unclear error message:

jq: error: sub/1 is not defined at &lt;top-level>, line 1:
.f | sub("a","b")     
jq: 1 compile error

This patch is changing the error message to:
jq: error: sub/1 is not defined, also happens when you separate arguments by comma (,) instead of semicolon (;), at &lt;top-level>, line 1:
.f | sub("a","b")     
jq: 1 compile error

because the correct command is:

echo '{"f":"a"}' | jq '.f | sub("a";"b")'

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -1159,7 +1159,7 @@ static int expand_call_arglist(block* b, jv args, jv *env) {
         else if (curr->op == LOADV)
           locfile_locate(curr->locfile, curr->source, "jq: error: $%s is not defined", curr->symbol);
         else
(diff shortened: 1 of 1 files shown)


itchyny:
Could this message show only when seached for 1 arity filter (no semicolon), and there is a filter with the same name but the arity if commas are replaced by semicolon?

itchyny:
I know that this is a syntax that is easily mistaken, but I am against unconditionally including it in the error message.

PR #3205, opened 2024-11-15
fix: reduce/foreach state variable should not be reset each iteration

itchyny:
When the UPDATE query in a reduce or foreach syntax does not emit
a value, it should simply skip updating the state variable. But
currently it resets the state variable to null.

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -832,7 +832,7 @@ block gen_reduce(block source, block matcher, block init, block body) {
   block loop = BLOCK(gen_op_simple(DUPN),
                      source,
                      bind_alternation_matchers(matcher,
-                                  BLOCK(gen_op_bound(LOADVN, res_var),
+                                  BLOCK(gen_op_bound(LOADV, res_var),
                                         body,
                                         gen_op_bound(STOREV, res_var))),
                      gen_op_simple(BACKTRACK));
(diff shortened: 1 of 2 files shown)


emanuele6 (approved):

wader (approved):

itchyny:
Now these syntax behaves like the following pseudo code.
Previously, `$x` was assigned `null` when it is used after `($x|UPDATE)`.
```ruby
# reduce EXP as $v (INIT; UPDATE)
for $x in INIT {
  for $v in EXP {
    for $u in ($x|UPDATE) {
      $x = $u
    }
  }
  yield $x
}

# foreach EXP as $v (INIT; UPDATE; EXTRACT)
for $x in INIT {
  for $v in EXP {
    for $u in ($x|UPDATE) {
      $x = $u
      for $e in ($x|EXTRACT) {
        yield $e
      }
    }
  }
}
```
These procedural codes explain any cases the filters emit multiple or no values.
The EXTRACT filter is evaluated against each UPDATE output, and the last value is used for
the next iteration. And if UPDATE emits no value, it skips updating the intermediate value.
I wondered whether we should change the behavior to how jaq does (multiple values
in UPDATE replicates successive iteration, resulting into exponentially many results,
and no values aborts the entire loop). But I now think the current behavior
(after this PR fix about the intermediate value reset) looks well-defined.
Although the jq manual does not explains well about these cases,
the manual shouldn't be too much detailed about such minor use cases.
I think the current example codes makes it easy to understand these syntaxes,
especially for beginners.
Also, I noticed a slight difference in gojq's reduce syntax about INIT, I'll fix this soon.

wader:
Update docs in this PR?

itchyny:
I'd like to make this PR focus on the null reset. The change is too detailed to write in the current user manual in my opinion.

PR #3207, opened 2024-11-20
Omit unnecessary matrix variables in the CI workflow

itchyny:
Ref: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,24 +12,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch:
-          - amd64
-          - arm64
-          - armel
-          - armhf
-          - i386
-          - mips
-          - mips64
-          - mips64el
-          - mips64r6
-          - mips64r6el
-          - mipsel
-          - mipsr6
-          - mipsr6el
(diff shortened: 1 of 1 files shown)


wader (approved):

itchyny:
Ah, this change requires the branch protection rule to be changed, and existing PRs need to be rebased after merging 😢

wader:
@itchyny not sure i followed how this was connected to branch protection rules? is it about that github actions is not smart enough to see that it won't change what jobs to run etc?

itchyny:
@wader 
The required status checks are configured by the check names, and they are generated by the matrix variables. Now it is configured as `linux (amd64)` is required, but the status checks reported by this PR was `linux (amd64, x86_64-linux-gnu)`. Probably if the check names are generated by the include configurations if the matrix variables do not exist. We can't configure OR condition for the required status checks (if possible, it would be great), so possible solution (not to force others rebase their PRs) is to remove the requirements for while and then re-configure the check requirements.

wader:
Aha i see, that is messy hmm, thanks for explaining

PR #3212, opened 2024-12-02
build: fix insecure RUNPATH

orbea:
In Gentoo -static-libtool-libs causes a QA Notice.

 * QA Notice: The following files contain insecure RUNPATHs
 *  Please file a bug about this at https://bugs.gentoo.org/
 *  with the maintainer of the package.
 *   /var/tmp/portage/app-misc/jq-1.7.1/image/usr/bin/jqn    RPATH: /var/tmp/portage/app-misc/jq-1.7.1/work/jq-jq-1.7.1/.libs

Gentoo-Issue: https://bugs.gentoo.org/945698

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -115,11 +115,10 @@ CLEANFILES = src/builtin.inc
 
 bin_PROGRAMS = jq
 jq_SOURCES = src/main.c
-jq_LDFLAGS = -static-libtool-libs
 jq_LDADD = libjq.la -lm
 
 if ENABLE_ALL_STATIC
-jq_LDFLAGS += -all-static
+jq_LDFLAGS = -all-static
 endif
 
 ### Tests (make check)


itchyny:
@orbea Sorry, I'm trying to catch up the issue but I can't understand well. Could you explain the issue and why we have to remove the flag?

orbea:
The `-static-libtool-libs` flag causes GNU libtool to add a rpath to the `.libs` build directory which is intended for internal use by the libtool implementation and is not installed in the resulting package causing Gentoo to print a QA notice. I can't say what the original intent in adding the flag in the first place was, it doesn't seem very useful to me.

wader:
Was added in https://github.com/jqlang/jq/commit/5d9ec838051c3451a59f9aa70eecde5c238a201c unclear why. Seem ok to me to remove. I wonder if we should have a CI test that verify that the resulting binary actually is static? inspect the binary or maybe run the binary in a scratch container etc?

orbea:
> Was added in [5d9ec83](https://github.com/jqlang/jq/commit/5d9ec838051c3451a59f9aa70eecde5c238a201c) unclear why. Seem ok to me to remove. I wonder if we should have a CI test that verify that the resulting binary actually is static? inspect the binary or maybe run the binary in a scratch container etc?

If the intent is that the binary is built entirely statically then I think the correct flag is `-static` where it will be entirely static excluding system libraries (Unless those are also built statically by the user). Where with `-all-static` it will result in failing to link any non-static system libraries like `libonig`.

itchyny (approved):
Thanks for explanation. LGTM!

itchyny:
Since I merged this PR, libtool is warning that `complete static linking is impossible` on macOS.
Compare https://github.com/jqlang/jq/actions/runs/13084445639/job/36513669856 and https://github.com/jqlang/jq/actions/runs/13085885383/job/36516955652.
```
libtool: warning: complete static linking is impossible in this configuration
```
How can I fix this problem?

itchyny:
I learned a few things about this topic, and found `--disable-shared` fixes the warning.

PR #3220, opened 2024-12-11
actions/attest-build-provenance

lectrical:
Adding https://github.com/actions/attest-build-provenance to the ci builds so that the release assets and docker image for the next release tag generate signed build provenance attestations for workflow artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,8 +374,8 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and release Docker image
-        id: attestation
         uses: docker/build-push-action@v6
+        id: build-push
         with:
           context: .
           push: ${{ startsWith(github.ref, 'refs/tags/jq-') }}
@@ -388,7 +388,7 @@ jobs:
         uses: actions/attest-build-provenance@v2
         with:
           subject-name: ghcr.io/${{ github.repository }}
-          subject-digest: ${{ steps.attestation.outputs.digest }}
+          subject-digest: ${{ steps.build-push.outputs.digest }}
           push-to-registry: true
 
   release:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,11 +7,6 @@ on:
       - "jq-*"
   pull_request:
 
-permissions:
-  id-token: write
-  contents: read
-  attestations: write
-
 jobs:
   linux:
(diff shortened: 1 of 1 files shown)


emanuele6:
It is erroring with:
```
Error: Failed to get ID token: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
```

lectrical:
@emanuele6 that seems a permissions issue on the pr actions. It works in the local repo (except docker push which was working 

https://github.com/lectrical/jq/actions/runs/12274537270

It's the normal error you expect if you forgot to add these https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries

Perhaps it should be skipped for pr requests anyway?

xximwon (approved):

xximwon (approved):

lectrical:
So according to this https://github.com/actions/attest-build-provenance/issues/99 the issue is expected.

I think I can maybe make it skip this on a pr originating from a fork?

xximwon (approved):

xximwon (approved):

xximwon (approved):

lectrical:
So that worked. The step is skipped unless a tag was pushed. I think that will only happen for a new release?

.github/workflows/ci.yml
```diff
@@ -343,6 +374,7 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and release Docker image
+        id: attestation
```
itchyny: I don't like this id because this isn't the step of attestation.

lectrical: The original example has `id: push` 

https://github.com/actions/attest-build-provenance?tab=readme-ov-file#container-image

A previous step has 

```
        uses: docker/metadata-action@v5
        id: metadata
```

So how about something like, which essentially uses a partial of the action name itself. Consistent with `metadata`? : 

```
        uses: docker/build-push-action@v6
        id: build-push
```

itchyny: Looks good.

lectrical: done

itchyny (changes requested):
I think adding attestation to the artifacts can be moved to the release job. This minimizes the permission. Also, this adds attestation to the aliased artifacts like jq-linux64.

PR #2256, opened 2021-01-26
Fix `--indent 0` implicitly enabling compact-output

amarshall:
Now it will retain pretty-printing, just removing any indentation. Fixes https://github.com/stedolan/jq/issues/1465 and fixes https://github.com/jqlang/jq/issues/2498.

PR #3232, opened 2025-01-23
Fix `--indent 0` implicitly enabling compact-output

### LINKED ISSUE #1465: --indent 0 removes newlines
I initially thought it would be easy to get output to look like this guy wants his to look:
https://github.com/stedolan/jq/issues/1462

However, using `--indent 0` from commandline `jq` seems to have simply disabled prettyprinting entirely, and in addition to removing indent characters, it removes newlines as well, which is surprising and unintuitive.

### LINKED ISSUE #2498: `--indent 0` omits more than indentations
**Describe the bug**

`jq --indent 0` behaves like `jq --compact-output` rather than only omit the indentations.

**To Reproduce**

`jq --indent 0 -n "[1,2]"` produces
```
[1,2]
```

**Expected behavior**

I would expect `jq --indent 0` to keep lines separated and only avoid adding the indentations.

For example, `jq --indent 0 -n "[1,2]"` should produce
```
[
1,
2
]
```

In other words, `jq --indent 0 ...` should be equivalent to `jq ... | sed 's/^\s*//'` on Linux.

**Environment:**
 - OS and Version: Windows, Linux - WSL
 - jq version 1.6

itchyny:
Cherry-pick of #2256 and #3202 with modifications.
Fixes #1465 and also fixes #2498.
Closes #2256 and also closes #3202.

### DIFF
--- a/src/jv.h
+++ b/src/jv.h
@@ -226,7 +226,7 @@ enum jv_print_flags {
   JV_PRINT_SPACE2   = 1024,
 };
 #define JV_PRINT_INDENT_FLAGS(n) \
-    ((n) < 0 || (n) > 7 ? JV_PRINT_TAB | JV_PRINT_PRETTY : (n) == 0 ? 0 : (n) << 8 | JV_PRINT_PRETTY)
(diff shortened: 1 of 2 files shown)


wader (approved):

itchyny:
Thank you. I'll merge next weekend if no one objects.

PR #3233, opened 2025-01-24
tests: refactor constant folding tests

itchyny:
As the title says, I refactored the constant folding tests.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -37,77 +37,17 @@ $VALGRIND $Q $JQ -Rse '. == "a\u0000b\nc\u0000d\ne"' $d/input
 $VALGRIND $Q $JQ -Rne '[inputs] == ["a\u0000b", "c\u0000d", "e"]' $d/input
 
 ## Test constant folding
-
-## XXX If we add a builtin to list the program's disassembly then we can
-## move all of these into tests/all.test
-
-# String constant folding (addition only)
 nref=$($VALGRIND $Q $JQ -n --debug-dump-disasm '"foo"' | wc -l)
-n=$($VALGRIND $Q $JQ -n --debug-dump-disasm '"foo" + "bar"' | wc -l)
-if [ $n -ne $nref ]; then
-  echo "Constant expression folding for strings didn't work"
-  exit 1
-fi
-
-# Numeric constant folding (binary operators)
-n=$($VALGRIND $Q $JQ -n --debug-dump-disasm '1+1' | wc -l)
-if [ $n -ne $nref ]; then
-  echo "Constant expression folding for numbers didn't work"
-  exit 1
-fi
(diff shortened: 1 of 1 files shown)


tests/shtest
```diff
@@ -37,77 +37,17 @@ $VALGRIND $Q $JQ -Rse '. == "a\u0000b\nc\u0000d\ne"' $d/input
...
-  echo "Constant expression folding for numbers didn't work"
-  exit 1
-fi
-n=$($VALGRIND $Q $JQ -n --debug-dump-disasm '9>=3' | wc -l)
-if [ $n -ne $nref ]; then
-  echo "Constant expression folding for numbers didn't work"
-  exit 1
-fi
+for exp in '1+1' '1-1' '2*3' '9/3' '9%3' '9==3' '9!=3' \
+  '9<3' '9>3' '9<=3' '9>=3' '1+2*3-4/5%6' '"foo" + "bar"'; do
```
wader, line 42: Do we constant fold `and`/`or` with constants? unary operators? something else?

wader (approved):

tests/shtest
itchyny: We don't, unfortunately.
```sh
 $ jq -n --debug-dump-disasm 'false and false'
0000 TOP
0001 DUP
0002 LOADK false
0004 JUMP_F 0019
0006 POP
0007 LOADK false
0009 JUMP_F 0015
0011 LOADK true
0013 JUMP 0017
0015 LOADK false
0017 JUMP 0022
0019 POP
0020 LOADK false
0022 RET

false

 $ jq -n --debug-dump-disasm 'false or false'
0000 TOP
0001 DUP
0002 LOADK false
0004 JUMP_F 0011
0006 POP
0007 LOADK true
0009 JUMP 0022
0011 POP
0012 LOADK false
0014 JUMP_F 0020
0016 LOADK true
0018 JUMP 0022
0020 LOADK false
0022 RET

false

 $ jq -n --debug-dump-disasm '(-1)'
0000 TOP
0001 LOADK 1
0003 CALL_BUILTIN _negate
0006 RET

-1
```

itchyny: https://github.com/jqlang/jq/blob/588ff18/src/parser.c#L468-L497

itchyny: https://github.com/jqlang/jq/blob/588ff1874c8c394253c231733047a550efe78260/src/parser.y#L202-L231

wader: Thanks for digging, at lest for `and`/`or` i guess it's very uncommon that filters in practice have such constant expressions

itchyny: Yeah, unary operator should be constant folded.

itchyny:
Thanks for review!

PR #3234, opened 2025-01-24
Move oniguruma and decNumber to vendor directory

itchyny:
I would like to move dependency libraries (oniguruma and decNumber) to `vendor/` directory. I was annoyed by pointless PRs that fixes comments of decNumber. I think it is better to move these to `vendor/` to clarify that these are third-party libraries. What do you think, maintainers?


wader (approved):
Makes sense to me

itchyny:
Thank you. I'll merge next weekend if no one objects.

PR #3239, opened 2025-01-26
Fix redundant subexpression in Generators and iterators example in doc

### LINKED ISSUE #3222: Redundant subexpression in Generators and iterators example in doc
**Describe the bug**
[jq 1.7 Manual / Generators and iterators](https://jqlang.github.io/jq/manual/#generators-and-iterators)

**To Reproduce**
In the first example:
`jq 'def range(init; upto; by): def _range: if (by > 0 and . < upto) or (by < 0 and . > upto) then ., ((.+by)|_range) else . end; if by == 0 then init else init|_range end | select((by > 0 and . < upto) or (by < 0 and . > upto)); range(0; 10; 3)'`
1. Redundant operation: `else .` produces an extra value beyond `upto`, then `| select((by > 0 and . < upto) or (by < 0 and . > upto))` filters it.
2. Should result in `empty` when `init == upto`.

**Expected behavior**
Can be simplified and corrected to:
`jq 'def range(init; upto; by): def _range: if (by > 0 and . < upto) or (by < 0 and . > upto) then ., ((.+by)|_range) else empty end; if init == upto then empty elif by == 0 then init else init|_range end; range(0; 10; 3)'`

/

**Additional context**
/

jhcarl0814:
Fixes #3222.
`jq 'def range(init; upto; by): def _range: if (by > 0 and . < upto) or (by < 0 and . > upto) then ., ((.+by)|_range) else . end; if by == 0 then init else init|_range end | select((by > 0 and . < upto) or (by < 0 and . > upto)); range(0; 10; 3)'`
Can be simplified and corrected to:
`jq 'def range(init; upto; by): def _range: if (by > 0 and . < upto) or (by < 0 and . > upto) then ., ((.+by)|_range) else empty end; if init == upto then empty elif by == 0 then init else init|_range end; range(0; 10; 3)'`

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -3223,9 +3223,8 @@ sections:
                     def _range:
                         if (by > 0 and . < upto) or (by < 0 and . > upto)
                         then ., ((.+by)|_range)
-                        else . end;
-                    if by == 0 then init else init|_range end |
-                    select((by > 0 and . < upto) or (by < 0 and . > upto));
+                        else empty end;
+                    if init == upto then empty elif by == 0 then init else init|_range end;
                 range(0; 10; 3)'
             input: 'null'
             output: ['0', '3', '6', '9']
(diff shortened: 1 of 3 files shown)


itchyny (approved):
Thank you.

PR #3240, opened 2025-01-27
docs: mention the behavior of scan using regex with capturing groups

### LINKED ISSUE #3086: `scan("(a)(b)")` returning an array is not documented
**Describe the bug**
If the regular expression of `scan/1`/`scan/2` contains a capture group, arrays contains the values of the capture groups are returned instead of strings that are the full match of the regular expression.

I use this feature all the time, and I have just noticed it is not documented!
> scan(regex), scan(regex; flags)
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Emit a stream of the non-overlapping substrings of the input that match
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;the regex in accordance with the flags, if any have been specified.  If
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;there  is no match, the stream is empty. To capture all the matches for
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;each input string, use the idiom [ expr ], e.g. [ scan(regex) ].
>
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;jq ´scan("c")´
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"abcdefabc"
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;=> "c", "c"

**To Reproduce**
```console
$ jq -cn '"foo bar xzbox" | scan(".b.")'
" ba"
"zbo"
$ jq -cn '"foo bar xzbox" | scan("(.)b(.)")'
[" ","a"]
["z","o"]
```
**Expected behavior**
The feature should be documented.

- OS and Version: Arch Linux
- jq version: 1.7.1

**Additional context**
This feature has been supported by `jq` for many years, and it also works in `gojq`.

itchyny:
This fixes #3086.

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -2699,12 +2699,17 @@ sections:
           that match the regex in accordance with the flags, if any
           have been specified.  If there is no match, the stream is empty.
           To capture all the matches for each input string, use the idiom
-          `[ expr ]`, e.g. `[ scan(regex) ]`.
+          `[ expr ]`, e.g. `[ scan(regex) ]`.  If the regex contains capturing
+          groups, the filter emits a stream of arrays, each of which contains
+          the captured strings.
 
         examples:
           - program: 'scan("c")'
             input: '"abcdefabc"'
             output: ['"c"', '"c"']
+          - program: 'scan("(a+)(b+)")'
+            input: '"abaabbaaabbb"'
+            output: ['["a","b"]', '["aa","bb"]', '["aaa","bbb"]']
 
       - title: "`split(regex; flags)`"
         body: |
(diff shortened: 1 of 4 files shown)


wader (approved):
I guess name capture groups are treated as unnamed capture groups on purpose? not sure i have any good ide how it would work otherwise

itchyny:
Thank you. They are not different in capturing matches so I don't see any issue here.

PR #3245, opened 2025-02-02
fix: `jv_number_value` should cache the double value of literal numbers

itchyny:
The code of `jv_number_value` is intended to cache the double value of
literal numbers, but it does not work because the function accepts the
`jv` struct by value. This patch fixes the behavior by checking if the
double value is `NaN`, which is the initial value. This patch apparently
does not cache the value if the literal is `NaN`, but improves the
performance of major use cases; e.g. `range(1000000)` runs 25% faster.

### DIFF
--- a/src/jv.c
+++ b/src/jv.c
@@ -206,9 +206,6 @@ enum {
   JVP_NUMBER_DECIMAL = 1
 };
 
-#define JV_NUMBER_SIZE_INIT      (0)
-#define JV_NUMBER_SIZE_CONVERTED (1)
-
 #define JVP_FLAGS_NUMBER_NATIVE       JVP_MAKE_FLAGS(JV_KIND_NUMBER, JVP_MAKE_PFLAGS(JVP_NUMBER_NATIVE, 0))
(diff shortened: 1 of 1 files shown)


wader (approved):
Maybe comment to future code reads about NaN being used as uncached value?

itchyny:
Is it okay for `jv_number_with_literal` to return non-literal number if the input is NaN?
```diff
diff --git a/src/jv.c b/src/jv.c
index b77e2d2..ebce513 100644
--- a/src/jv.c
+++ b/src/jv.c
@@ -589,6 +589,10 @@ static jv jvp_literal_number_new(const char * literal) {
     jv_mem_free(n);
     return JV_INVALID;
   }
+  if (decNumberIsNaN(&n->num_decimal)) {
+    jv_mem_free(n);
+    return jv_number(NAN);
+  }
 
   jv r = {JVP_FLAGS_NUMBER_LITERAL, 0, 0, JV_NUMBER_SIZE_INIT, {&n->refcnt}};
   return r;
```

itchyny:
I updated the branch to apply the patch in my above comment. We don't distinguish NaNs and we are not interested in the payload or signaling NaNs. So using non-literal NaN is reasonable to avoid repeating double conversion of NaN.

PR #3247, opened 2025-02-03
Add opengraph meta tags

wader:
Makes the site preview a bit nicer when sharing

### DIFF
--- a/docs/templates/shared/_head.html.j2
+++ b/docs/templates/shared/_head.html.j2
@@ -2,6 +2,11 @@
     <meta charset="utf-8">
     <title>{{ headline }}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta property="og:title" content="{{ headline }}">
+    <meta property="og:description" content="jq is a lightweight and flexible command-line JSON processor">
+    <meta property="og:type" content="website">
+    <meta property="og:image" content="{{ url }}{{ root }}/icon.png">
+    <meta property="og:url" content="{{ url }}/{{ permalink }}">
     <link rel="icon" href="{{ root }}/icon.svg" type="image/svg+xml">
     <link rel="apple-touch-icon" href="{{ root }}/icon.png" type="image/png">
     <link rel="canonical" href="{{ url }}/{{ permalink }}">


docs/templates/shared/_head.html.j2
```diff
@@ -2,6 +2,11 @@
     <meta charset="utf-8">
     <title>{{ headline }}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta property="og:title" content="{{ headline }}">
+    <meta property="og:description" content="jq is a lightweight and flexible command-line JSON processor">
+    <meta property="og:type" content="website">
+    <meta property="og:image" content="{{ root }}/icon.svg">
```
itchyny: We can't use SVG for og:image.
https://indieweb.org/The-Open-Graph-protocol#Does_not_support_SVG_images

wader: 👍  Fixed

emanuele6: In the paragraph before that one, it also says you can't use relative paths: https://indieweb.org/The-Open-Graph-protocol#Disallows_relative_URLs
I think you need to prefix `{{ url }}`

wader: Also fixed 👍  generated locally and it looks like this:
```html
    <meta property="og:title" content="jq">
    <meta property="og:description" content="jq is a lightweight and flexible command-line JSON processor">
    <meta property="og:type" content="website">
    <meta property="og:image" content="https://jqlang.org/icon.png">
    <meta property="og:url" content="https://jqlang.org/">
```

docs/templates/shared/_head.html.j2
```diff
@@ -2,6 +2,11 @@
     <meta charset="utf-8">
     <title>{{ headline }}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta property="og:title" content="{{ headline }}">
+    <meta property="og:description" content="jq is a lightweight and flexible command-line JSON processor">
+    <meta property="og:type" content="website">
+    <meta property="og:image" content="{{ url }}/icon.png">
```
emanuele6: `{{ url }}{{ root }}/icon.png` I think; root ~~is empty~~**is `/jq`** by default, and the icon is in `{{ root }}/icon.png` below

wader: Ah true, fixed and generated didn't change so seems it's empty

emanuele6: @wader you changed `og:url` instead of `og:image`; yes, the default is empty now, it used to be `'/jq'` but it was recently changed when `build_website.py` was modified for `jqlang.org`

wader: Oh sorry i'm confused. Looks correct now?

docs/templates/shared/_head.html.j2
```diff
@@ -2,6 +2,11 @@
     <meta charset="utf-8">
     <title>{{ headline }}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta property="og:title" content="{{ headline }}">
+    <meta property="og:description" content="jq is a lightweight and flexible command-line JSON processor">
+    <meta property="og:type" content="website">
+    <meta property="og:image" content="{{ url }}{{ root }}/icon.png">
+    <meta property="og:url" content="{{ url }}{{ root }}/{{ permalink }}">
```
wader: Ah this should probably not have root, fixing

emanuele6:
Should we just remove `{{ root }}` since it's not used? Now `<ref name=canonical>` is `{{ url }}/{{ permalink }}` while `<meta property=og:url>` is `{{ url }}{{ root }}/{{ permalink }}`; I don't know which one is right

emanuele6:
Never mind, now they are the same :)

wader:
> Never mind, now they are the same :)

👍 yeah i noticed my own confusion hehe

itchyny (approved):
Thank you!

PR #3249, opened 2025-02-08
fix: `jv_dump_string_trunc` should not break unicode characters

itchyny:
As the title says, `jv_dump_string_trunc`, which is used for error messages, should not break Unicode characters.

### DIFF
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1788,6 +1788,19 @@ try -. catch .
 "very-long-string"
 "string (\"very-long-...) cannot be negated"
 
+try (.-.) catch .
+"very-long-string"
+"string (\"very-long-...) and string (\"very-long-...) cannot be subtracted"
+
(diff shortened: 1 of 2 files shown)


tests/jq.test
```diff
@@ -1788,6 +1788,19 @@ try -. catch .
 "very-long-string"
 "string (\"very-long-...) cannot be negated"
 
+try (.-.) catch .
+"very-long-string"
+"string (\"very-long-...) and string (\"very-long-...) cannot be subtracted"
+
+"x" * range(0; 12; 2) + "☆" * 5 | try -. catch .
```
wader, line 1795: Would step size 1 test more `jvp_utf8_backtrack` cases?

wader (approved):

tests/jq.test
itchyny: I don't think so. The current test case covers code enough IMO.

wader: 👍

PR #3260, opened 2025-02-16
Refactor splits/2 using foreach syntax (remove private filter _nwise)

### LINKED ISSUE #3148: oss-fuzz/issues/detail?id=68056
**Describe the bug**

OSS-FUZZ is unhappy with a filter that uses _nwise(0). 

Note that `_nwise(0)` typically behaves like `recurse([])`, and one could argue that the def of _nwise/1
is just fine.  If the current behavior is judged to be acceptable, what should be done about [68056](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68056)?

**To Reproduce**
```
[1] | _nwise(0)
```

**Expected behavior**
If the current behavior is deemed unsatisfactory, how should the def be altered?
Plausible alternatives would include:

```
def _nwise($n):
  def n: if length <= $n then . else .[0:$n] , (.[$n:] | n) end;
  if $n <= 0 then empty
  else n
  end;
```

and similar alternatives, e.g. replacing `empty` above with a call to `error`. 
(Since the guard is outside `def n`, the performance impact would be negligible.)

**Additional context**
Neither gojq nor jaq define `_nwise/1`.

itchyny:
This commit refactors `splits/2` implementation using `foreach` syntax
and reduces intermediate arrays. This refactoring removes an unnecessary
private (and undocumented) filter `_nwise` and closes #3148.

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -98,26 +98,16 @@ def scan($re; $flags):
       else .string
       end;
 def scan($re): scan($re; null);
-#
-# If input is an array, then emit a stream of successive subarrays of length n (or less),
-# and similarly for strings.
-def _nwise($n):
-  def n: if length <= $n then . else .[0:$n] , (.[$n:] | n) end;
-  n;
-def _nwise(a; $n): a | _nwise($n);
-#
+
 # splits/1 produces a stream; split/1 is retained for backward compatibility.
-def splits($re; flags): . as $s
-#  # multiple occurrences of "g" are acceptable
-  | [ match($re; "g" + flags) | (.offset, .offset + .length) ]
-  | [0] + . +[$s|length]
-  | _nwise(2)
-  | $s[.[0]:.[1] ] ;
+def splits($re; $flags):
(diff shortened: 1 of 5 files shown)


src/builtin.jq
```diff
@@ -98,26 +98,16 @@ def scan($re; $flags):
...
 # splits/1 produces a stream; split/1 is retained for backward compatibility.
-def splits($re; flags): . as $s
-#  # multiple occurrences of "g" are acceptable
-  | [ match($re; "g" + flags) | (.offset, .offset + .length) ]
-  | [0] + . +[$s|length]
-  | _nwise(2)
-  | $s[.[0]:.[1] ] ;
+def splits($re; $flags):
+  .[foreach (match($re; $flags+"g"), null) as {$offset, $length}
+      (null; {start: .next, end: $offset, next: ($offset+$length)})];
```
wader, line 105: Nice

wader (approved):

itchyny:
Thank you for reviewing!

PR #3261, opened 2025-02-17
Disable shared libraries on macOS to avoid conflict with -all-static

itchyny:
After merging #3212, libtool warns that complete static linking is impossible
in this configuration on macOS. We are not interested in building shared
libraries on macOS, so I disabled on CI to suppress the warning (it is disabled
already on Windows). I also noticed that `--enable-static` on macOS is no-op
because macOS does not provide static system libraries. But I left it as it is
for consistency.


itchyny:
This is just for CI. Merging.

PR #3263, opened 2025-02-17
Fix compiler warning `type-limits` in `found_string` (ref #2911)

itchyny:
Whether `char` has a sign is implementation dependent, and `c >= 0` may
result in a compiler warning. We can use bitwise operator to check
whether the character is a control character, regardless of the sign.

### DIFF
--- a/tests/shtest
+++ b/tests/shtest
@@ -94,6 +94,20 @@ if printf '1\n' | $JQ -cen --seq '[inputs] == []' >/dev/null 2> $d/out; then
 fi
 cmp $d/out $d/expected
 
+# Test control characters for #2909
+cat > $d/expected <<EOF
+jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 1, column 3
+EOF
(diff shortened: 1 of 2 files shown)


tests/shtest
```diff
@@ -94,6 +94,20 @@ if printf '1\n' | $JQ -cen --seq '[inputs] == []' >/dev/null 2> $d/out; then
 fi
 cmp $d/out $d/expected
 
+# Test control characters for #2909
+cat > $d/expected <<EOF
+jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 1, column 3
+EOF
+for i in 1 2 30 31; do
```
wader, line 101: Include 0?

wader (approved):

tests/shtest
itchyny: We can't. Unfortunately.

wader: Ah gets null terminated doh

itchyny:
Thank you!

PR #3265, opened 2025-02-17
Replace jqplay.org with play.jqlang.org

owenthereal:
As discussed in [GitHub Discussion #3244](https://github.com/jqlang/jq/discussions/3244), we are migrating jqplay to the jqlang GitHub organization. This PR updates all references from `jqplay.org` to `play.jqlang.org` to reflect the new official domain.

For continuity, all `jqplay.org/*` requests are already configured to redirect to `play.jqlang.org` and will remain so for the foreseeable future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 ## Documentation
 
 - **Official Documentation**: [jqlang.org](https://jqlang.org)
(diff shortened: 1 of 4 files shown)


itchyny (approved):
Thank you!

Issue #3227, opened 2025-01-05
`reduce` and `foreach` with backtracking init have observable `DUPN`


kanwren:
The fix for https://github.com/jqlang/jq/issues/618 (7243989c52cea7933ace6535eae914c19eb1210a) introduced `DUPN` to the compilation of `reduce f as $x (init; update)` and `foreach f as $x (init; update; extract)` to address extra references being held.

However, the value that `DUPN` dupes (that is, `.` as it is passed to `f`) comes from before a forkpoint, which makes `DUPN` unsafe here; if we backtrack to the `FORK` (in particular, if `init` backtracks), then `f` will see `.` mutated to `null`. This is a problem, as to my understanding the mutation in `DUPN` (and the other `-N` instructions) should be a non-observable optimization.

# To reproduce

```
$ jq --null-input  '[] | reduce .[] as $x (0, 0; .)'
0
jq: error (at <unknown>): Cannot iterate over null (null)

$ jq --null-input '[] | foreach .[] as $x (0, 0; .)'
jq: error (at <unknown>): Cannot iterate over null (null)
```

The init contains a `,` and thus backtracks. `.[]` then gets `[]` on the first pass, but `null` when backtracking.

# Expected behavior

`reduce` and `foreach` should not break backtracks that run back through the body. That is to say, we should see:

```
$ jq --null-input '[] | reduce .[] as $x (0, 0; .)'
0
0

$ jq --null-input '[] | foreach .[] as $x (0, 0; .)'
<empty>
```
[^1]

Put another way, it should be the same as e.g. `reduce [][] as $x (0, 0; .)`.

[^1]: This is the output from local testing of a potential fix (removing `FORK` from `foreach`, using `STOREV`/`LOADVN` instead of `DUPN` in `reduce`)

# Environment

- OS: Linux (NixOS 25.05)
- `jq` version: tested against current `master` (588ff1874c8c394253c231733047a550efe78260) and 1.7.1

# Additional context

I noticed this issue while inspecting the bytecode, which gives some insights towards addressing it, which I've noted below.

From digging in thread history, it seems that @leonid-s-usov noticed the same issues with the bytecode while looking for a way to remove the `DUPN` instruction for another reason. They give a draft of potential changes in https://github.com/jqlang/jq/pull/2059, which from my testing would be sufficient to fix this bug.

## `foreach`

This shouldn't be an issue with `foreach` in the first place, and should be fixable without having to compromise. Inspecting the bytecode for the example above:

```
[...]

PR #3266, opened 2025-02-17
fix: `foreach` should not break init backtracking with `DUPN`

kanwren:
Simplifies the compilation of `foreach` to avoid the problem in https://github.com/jqlang/jq/issues/3227 by removing unnecessary `FORK ... BACKTRACK`. Adds a simple regression test to check that the `init` clause is allowed to backtrack.

Does not address `reduce`, which is less trivial, and may require changes to the instruction set or other tradeoffs.


itchyny (approved):
LGTM, Thank you!

PR #3269, opened 2025-02-20
Disable Valgrind by default during testing

itchyny:
Previously, the configure script automatically enables Valgrind during
testing when the valgrind command is found. However, running tests with
Valgrind is quite slow, so it is better to disable it by default, and
to enable it only when the user specifies the `--enable-valgrind` flag.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,11 +39,6 @@ if test "$USE_MAINTAINER_MODE" = yes; then
   AC_CHECK_PROGS(LEX, flex lex)
 fi
 
-dnl Check for valgrind
-AC_CHECK_PROGS(valgrind_cmd, valgrind)
-if test "x$valgrind_cmd" = "x" ; then
-    AC_MSG_WARN([valgrind is required to test jq.])
-fi
 AC_CHECK_FUNCS(memmem)
 
 AC_CHECK_HEADER("sys/cygwin.h", [have_cygwin=1;])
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -54,10 +49,9 @@ dnl Running tests with Valgrind is slow. It is faster to iterate on
 dnl code without Valgrind until tests pass, then enable Valgrind and
 dnl fix leaks.
 AC_ARG_ENABLE([valgrind],
-   AS_HELP_STRING([--disable-valgrind],[do not run tests under Valgrind]))
+   AS_HELP_STRING([--enable-valgrind],[enable Valgrind during testing]))
```
wader, line 52: Just to understand: With this configure will not check if valgrind is actually available when using `--enable-valgrind`?

wader (approved):

configure.ac
itchyny: Yes, but in the setup script the command availability is checked already.

wader: I see, the check in tests/setup? as i understand the will silently skip using valgrind if missing?

itchyny: Exactly. The behavior is not changed in this PR.

wader: Ok, i guess it could possible fail if valgrind was explicitly enabled but missing? but if so could be done in a separate PR i think

itchyny:
Thank you.

PR #3271, opened 2025-02-23
Fix segmentation fault on `strftime/1` and `strflocaltime/1`

itchyny:
When we use a large format string to `strftime/1` and `strflocaltime/1`,
we get segmentation fault. This happens because the result buffer is
allocated on the stack using `alloca` function, and `strftime` function
writes out of the space when the size is not enough to format time.
I fixed the segmentation fault issue by allocating the result buffer on
the heap using `malloc` function, because `alloca` function does not
provide a way to detect insufficient space.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -8,21 +8,6 @@
 #include <sys/time.h>
 #include <stdlib.h>
 #include <stddef.h>
-#ifdef HAVE_ALLOCA_H
-# include <alloca.h>
-#elif !defined alloca
-# ifdef __GNUC__
-#  define alloca __builtin_alloca
-# elif defined _MSC_VER
-#  include <malloc.h>
-#  define alloca _alloca
-# elif !defined HAVE_ALLOCA
-#  ifdef  __cplusplus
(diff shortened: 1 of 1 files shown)


src/builtin.c
```diff
@@ -1778,9 +1763,13 @@ static jv f_strftime(jq_state *jq, jv a, jv b) {
 #endif
   jv_free(b);
   /* POSIX doesn't provide errno values for strftime() failures; weird */
-  if (n == 0 || n > alloced)
+  if ((n == 0 && *fmt) || n > max_size) {
```
wader, line 1766: Ugg `strftime` error handling seems like a mess 😬 so n == 0 might mean error but we only conceder it an error if the format string is not empty? reading the man page it seems %p usage might also return 0?

wader (approved):

src/builtin.c
itchyny: Year, ugly. Hope we can handle error properly but it does not set errno. I think using %p only for the format is very rare use case.

PR #3274, opened 2025-02-24
Support CRLF line breaks in filters

### LINKED ISSUE #1219: Support \r as whitespace in filters

itchyny:
This PR closes #1219, and related to #1673.


src/lexer.l
```diff
@@ -130,7 +130,7 @@ struct lexer_param;
 \.[a-zA-Z_][a-zA-Z_0-9]*  { yylval->literal = jv_string(yytext+1); return FIELD;}
 \$([a-zA-Z_][a-zA-Z_0-9]*::)*[a-zA-Z_][a-zA-Z_0-9]*  { yylval->literal = jv_string(yytext+1); return BINDING;}
 
-[ \n\t]+  {}
+[ \r\n\t]+  {}
```
wader, line 133: Possible to use a white space class like `[:space:]`? that would probably include some more white space characters like `\v` i think

wader (approved):

src/lexer.l
itchyny: That's available, but I'm not sure how it's meaningful for jq script to support such spaces that JSON disallows.

wader: 👍 have no strong opinion about it, maybe good to keep what is considered whitespace between jq filter and JSON in sync?

itchyny:
Thank you.

PR #3242, opened 2025-01-30
fix: preserve numerical precision on unary negation

itchyny:
This commit fixes #2704.
```
 $ ./jq -n '[1234567890987654321,-1234567890987654321 | tojson]'
[
  "1234567890987654321",
  "-1234567890987654321"
]
```

PR #3275, opened 2025-02-24
Preserve numerical precision on `length/0`

itchyny:
While `abs` preserves the numerical precision (thanks to #3242), why not `length`?


wader (approved):
As the docs for length says "The length of a number is its absolute value" i think this makes prefect sense

itchyny:
Thank you.

PR #3277, opened 2025-02-26
Enable z/OS support

sachintu47:
This patch resolves compilation issues and includes platform-specific adjustments for IBM z/OS.

pthread_t is a structure, and its __ field represents the thread ID.

### DIFF
--- a/src/jq_test.c
+++ b/src/jq_test.c
@@ -382,7 +382,11 @@ static void run_jq_pthread_tests() {
 
     // wait for all threads
     for(a = 0; a < NUMBER_OF_THREADS; ++a) {
+#ifdef __MVS__
+        if (threads[a].__ != 0) {
+#else
         if (threads[a] != 0) {
+#endif
             pthread_join(threads[a], NULL);
         }
     }


src/jv_alloc.c
```diff
@@ -150,6 +150,15 @@ void* jv_mem_alloc_unguarded(size_t sz) {
...
 void* jv_mem_calloc(size_t nemb, size_t sz) {
+
+#ifdef __MVS__ 
+  if (sz == 0) 
+    sz = 1;
+
+  if(nemb == 0)
+    nemb = 1;
+#endif
+
```
itchyny: malloc/calloc with size 0 is undefined behavior and I think we should find better fix. What problem did you see before this fix?

sachintu47: All the test cases failed due to allocation error. One such example below

FAIL: tests/mantest
===================

jq: error: cannot allocate memory
 1: 0x27195E82 jv_mem_calloc+0x17a
 2: 0x27127200 jq_compile_args+0x434
 3: 0x27142D38 jq_compile+0x44
 4: 0x27115E06 jq_testsuite+0x584
 5: 0x272BCADE main+0x26a2
 6: 0x272BCEFA CELQINIT+0x1aca [CELQINIT:]
l�terminatingl�Test #1: '.' at line number 1
FAIL tests/mantest (exit status: 131)

itchyny: I fixed calloc users in #3280. Can you check the master branch?

sachintu47: Thanks @itchyny for the fix. It works perfectly. Have remove the patch from my commit

itchyny (approved):
Thank you!

PR #3279, opened 2025-03-01
Fix behavior of `--slurp --stream` when input has no trailing newline character

### LINKED ISSUE #3273: Strange behaviours of slurping streaming input without line terminators at the end
**Describe the bug**
When the input steam has no line terminator (and space) at the end, the combination of `--slurp` and `--stream` flags behaves oddly.

**To Reproduce**
```sh
 $ printf "[1]\n" | jq -c -s --stream .
[[[0],1],[[0]]] # ok

 $ printf "[1]" | jq -c -s --stream .  
[[[0],1]]
[[0]] # unexpected
```

**Expected behavior**
The second command outputs as same as the first command.

- OS and Version: not related
- jq version: 1.7.1 (and 1.7, 1.6)

**Additional context**
This is a close issue of #1534 but with the streaming parser flag.

itchyny:
This commit fixes a strange behavior of the combination of `--slurp` and
`--stream` flags when the input has no trailing newline character.
This fixes #3273.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -271,12 +271,11 @@ static int jq_util_input_read_more(jq_util_input_state *state) {
         fclose(state->current_input);
       }
       state->current_input = NULL;
-      jv_free(state->current_filename);
-      state->current_filename = jv_invalid();
-      state->current_line = 0 ;
     }
     const char *f = next_file(state);
     if (f != NULL) {
+      jv_free(state->current_filename);
+      state->current_line = 0;
       if (!strcmp(f, "-")) {
         state->current_input = stdin;
         state->current_filename = jv_string("<stdin>");
@@ -288,7 +287,6 @@ static int jq_util_input_read_more(jq_util_input_state *state) {
           state->failures++;
         }
       }
-      state->current_line = 0;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/util.c
```diff
@@ -429,10 +425,6 @@ jv jq_util_input_next_input(jq_util_input_state *state) {
       }
       value = jv_parser_next(state->parser);
       if (jv_is_valid(state->slurped)) {
-        // When slurping an input that doesn't have a trailing newline,
-        // we might have more than one value on the same line, so let's check
-        // to see if we have more data to parse.
-        has_more = jv_parser_remaining(state->parser);
```
wader, line 435: `jv_parser_remaining` somehow ended up returning > 0 when there was nothing remaning but 0 if there was 1 or more trailing whitespace characters?

itchyny: The problem of using remaining byte count for detecting more value existing to slurp is that it does not work for `--stream` where it sometimes has array/object closing value. E.g., `printf '[1]' | jq --stream --slurp`, `jv_parser_next` emits `[[0],1]` and `has_more` is 0, even if there is another remaining value `[0]`. I fixed the issue by changing `is_last` value so that it works both for non-stream and stream inputs.

wader: Ah now i see 👍

wader (approved):

itchyny:
Thank you!

PR #3280, opened 2025-03-02
Avoid zero-length `calloc`

itchyny:
Using `calloc` function with a count of zero elements is implementation
defined and it may cause allocation errors on some platforms (ref #3277).
This commit fixes `jv_mem_calloc` callers to avoid undefined behavior.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -757,9 +757,12 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
+    if (len == 0) {
+      jv_free(input);
+      return jv_string("");
+    }
     size_t decoded_len = (3 * (size_t)len) / 4; // 3 usable bytes for every 4 bytes of input
     char *result = jv_mem_calloc(decoded_len, sizeof(char));
-    memset(result, 0, decoded_len * sizeof(char));
     uint32_t ri = 0;
     int input_bytes_read=0;
     uint32_t code = 0;
--- a/src/compile.c
+++ b/src/compile.c
@@ -1,5 +1,4 @@
 #include <assert.h>
-#include <math.h>
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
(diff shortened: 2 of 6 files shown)


src/builtin.c
```diff
@@ -757,9 +757,8 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
-    size_t decoded_len = (3 * (size_t)len) / 4; // 3 usable bytes for every 4 bytes of input
+    size_t decoded_len = (3 * (size_t)len) / 4 + 1; // 3 usable bytes for every 4 bytes of input
     char *result = jv_mem_calloc(decoded_len, sizeof(char));
-    memset(result, 0, decoded_len * sizeof(char));
```
wader, line 762: The code below seems to not rely on result being null terminated so this was not really needed?

src/builtin.c
```diff
@@ -757,9 +757,8 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
-    size_t decoded_len = (3 * (size_t)len) / 4; // 3 usable bytes for every 4 bytes of input
+    size_t decoded_len = (3 * (size_t)len) / 4 + 1; // 3 usable bytes for every 4 bytes of input
```
wader: Alternative i guess could be to do a early return if input is an empty string?

wader (approved):

src/builtin.c
itchyny: After calloc, allocated memory is filled by zero.

wader: Ah yes 👍

src/builtin.c
itchyny: Fixed.

wader (approved):

src/builtin.c
itchyny: No! The fix was not correct. I should have remembered why I did +1 here.
```
 $ jq -n '"=" | @base64d' 
""

 $ ./jq -n '"=" | @base64d'
Assertion failed: (nemb > 0 && sz > 0), function jv_mem_calloc, file jv_alloc.c, line 154.
```

PR #3286, opened 2025-03-05
Fix calloc error on `@base64d` format (ref #3280)

itchyny:
This commit fixes a regression of a8ce2ff, e.g. `"=" | @base64d`.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -734,11 +734,7 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
-    if (len == 0) {
-      jv_free(input);
-      return jv_string("");
-    }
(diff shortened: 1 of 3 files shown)


src/builtin.c
```diff
@@ -734,11 +734,7 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
     input = f_tostring(jq, input);
     const unsigned char* data = (const unsigned char*)jv_string_value(input);
     int len = jv_string_length_bytes(jv_copy(input));
-    if (len == 0) {
-      jv_free(input);
-      return jv_string("");
-    }
-    size_t decoded_len = (3 * (size_t)len) / 4; // 3 usable bytes for every 4 bytes of input
+    size_t decoded_len = MAX((3 * (size_t)len) / 4, 1); // 3 usable bytes for every 4 bytes of input
```
wader, line 737: Ah doh sorry, yeah my suggetion wont work 👍 think i've convinced myself that below code in the case of `len` < 4 will not be able to write outside `result`

wader (approved):

PR #3292, opened 2025-03-10
Improve error messages to show problematic position in the filter

itchyny:
This commit improves the parser and compiler error messages, making
it more easier to understand where in the filter the problem is.
Like modern compilers, jq will show the line in the filter where the
error occurred, will display squiggly line under the invalid token.

     $ ./jq -n '1 + 2 ( 3'    
    jq: error: syntax error, unexpected '(', expecting end of file at <top-level>, line 1, column 7:
        1 + 2 ( 3
              ^
    jq: 1 compile error

     $ ./jq -n '1 + $foo + 2'
    jq: error: $foo is not defined at <top-level>, line 1, column 5:
        1 + $foo + 2
            ^^^^
    jq: 1 compile error

### DIFF
--- a/src/parser.y
+++ b/src/parser.y
@@ -139,15 +139,7 @@ struct lexer_param {
 void yyerror(YYLTYPE* loc, block* answer, int* errors,
              struct locfile* locations, struct lexer_param* lexer_param_ptr, const char *s){
   (*errors)++;
-  if (strstr(s, "unexpected")) {
-#ifdef WIN32
-      locfile_locate(locations, *loc, "jq: error: %s (Windows cmd shell quoting issues?)", s);
-#else
-      locfile_locate(locations, *loc, "jq: error: %s (Unix shell quoting issues?)", s);
-#endif
-  } else {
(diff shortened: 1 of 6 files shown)


src/parser.y
```diff
@@ -139,15 +139,7 @@ struct lexer_param {
 void yyerror(YYLTYPE* loc, block* answer, int* errors,
              struct locfile* locations, struct lexer_param* lexer_param_ptr, const char *s){
   (*errors)++;
-  if (strstr(s, "unexpected")) {
-#ifdef WIN32
-      locfile_locate(locations, *loc, "jq: error: %s (Windows cmd shell quoting issues?)", s);
-#else
-      locfile_locate(locations, *loc, "jq: error: %s (Unix shell quoting issues?)", s);
-#endif
```
wader, line 147: Removed as it might be more confusing than useful to show for all parse errors?

wader (approved):

src/parser.y
itchyny: Yeah, it's less likely an issue of quoting issue in my experience.

PR #3306, opened 2025-04-02
Add ssv output format

HenrikHolst:
Currently jq supports output in CSV format, however in many parts of Europe the CSV format is in fact semicolon and not comma separated. So I've added a small patch that adds support for semicolon separated CSV files (SSV) while also clarified in the doc and man page that the CSV version is the comma separated one.

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -2151,9 +2151,15 @@ sections:
 
           * `@csv`:
 
-            The input must be an array, and it is rendered as CSV
-            with double quotes for strings, and quotes escaped by
-            repetition.
+            The input must be an array, and it is rendered as comma 
+            separated CSV with double quotes for strings, and quotes 
+            escaped by repetition.
+
+          * `@ssv`:
+
+            The input must be an array, and it is rendered as semicolon
+            separated CSV with double quotes for strings, and quotes 
+            escaped by repetition.
 
           * `@tsv`:
 
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -2155,7 +2155,7 @@ sections:
             separated CSV with double quotes for strings, and quotes 
             escaped by repetition.
 
-          * `@ssv`:
+          * `@scsv`:
 
             The input must be an array, and it is rendered as semicolon
             separated CSV with double quotes for strings, and quotes 
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -2340,7 +2340,13 @@ The inverse of \fB@uri\fR, applies percent\-decoding, by mapping all \fB%XX\fR s
 \fB@csv\fR:
 .
 .IP
-The input must be an array, and it is rendered as CSV with double quotes for strings, and quotes escaped by repetition\.
+The input must be an array, and it is rendered as comma separated CSV with double quotes for strings, and quotes escaped by repetition\.
+.
+.TP
+\fB@ssv\fR:
+.
+.IP
+The input must be an array, and it is rendered as semicolon separated CSV with double quotes for strings, and quotes escaped by repetition\.
 .
 .TP
 \fB@tsv\fR:
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -558,7 +558,7 @@ static jv f_format(jq_state *jq, jv input, jv fmt) {
   } else if (!strcmp(fmt_s, "text")) {
     jv_free(fmt);
     return f_tostring(jq, input);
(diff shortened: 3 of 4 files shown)


HenrikHolst:
I don't fully understand what went wrong with the man page check, if any one could point me in the correct direction I would be happy to fix it.

itchyny:
I'm afraid ssv is not a common technical term for this format. I think it's one of a [DSV (Delimiter-separated values)](https://en.wikipedia.org/wiki/Delimiter-separated_values) (oh, I've never heard of this but this term seems to be appear in The Art of Unix Programming). Quick googling ssv leads me to [space-separated values](https://docs.rs/ssv/latest/ssv/) or [separator-separated values](https://pypi.org/project/ssv/), or even [something-separated value](https://hackage.haskell.org/package/ssv-0.3). CSV is very common and has RFC specification ([RFC-4180](https://datatracker.ietf.org/doc/html/rfc4180)), we'd like to avoid uncommon terms and specification.

HenrikHolst:
> I'm afraid ssv is not a common technical term for this format. I think it's one of a [DSV (Delimiter-separated values)](https://en.wikipedia.org/wiki/Delimiter-separated_values) (oh, I've never heard of this but this term seems to be appear in The Art of Unix Programming). Quick googling ssv leads me to [space-separated values](https://docs.rs/ssv/latest/ssv/) or [separator-separated values](https://pypi.org/project/ssv/), or even [something-separated value](https://hackage.haskell.org/package/ssv-0.3). CSV is very common and has RFC specification ([RFC-4180](https://datatracker.ietf.org/doc/html/rfc4180)), we'd like to avoid uncommon terms and specification.

I was looking for a term to use since the semicolon variant usually is just named CSV here in Europe, desptite it being a RFC, Microsoft in their magic thinking used semicolon as a separator for CSV from Excel over here due to us using comma as the decimal separator.

First I was thinking about something like "csv-semi" or if perhaps there should be a setting for the delimiter used but then I stumbled across sites like this: [https://www.dell.com/support/manuals/sv-se/dell-opnmang-srvr-admin-v7.4/omsa_cli-v3/semicolon-separated-values-ssv?guid=guid-7ae0aea0-ff58-49aa-b2f5-3931eed1b033&lang=en-us](url) as well as numerous github issues asking various projects to support the SSV format so I thought then and there that SSV although ugly seemed to be a very well known name for the format.

DSV is "something else", SSV is exactly CSV with the sole difference that ; is used over , all other rules like escaping with just " is the same (due to the source being MS Excel once upon a time).

But the name is for me unimportant and I will happily change it to anything, it's the functionality that I'm after.

pkoppstein:
Would it be feasible to support @dsv(delim), so that one could write, for example, @dsv(";")

HenrikHolst:
> Would it be feasible to support @dsv(delim), so that one could write, for example, @dsv(";")

did a quick look at this and there are some issues, for one bash is not happy with () so it has to be inside single quotes like '@dsv()' and once you do jq starts to complain about syntax error since jq does not like the () at all and jq is also not at all happy with any other separator like ;, :, - or , so jq appears to have very strict syntax here (which ofc is good) that makes it very hard to give arguments to any of the output formats.

Which is why I stand by that @ssv is the better option, or say @csv_semi or @csv_ssv. I'm open for all suggestions here.

pkoppstein:
Since `@SSV` is ambiguous (between space or semicolon) and `@DSV(delim)` presents implementation issues, the obvious choice is `@SCSV`.  This is confirmed by ChatGPT's advice:

```
The most consistent and intuitive abbreviation for "semicolon-separated values" in the context of DSV (Delimiter-Separated Values) is:

SCSV

This follows the same pattern as:

CSV – Comma-Separated Values

TSV – Tab-Separated Values

SSV – Space-Separated Values

So:

SCSV – Semicolon-Separated Values

This abbreviation is already in informal use in some data communities and tools, even if it's not as standardized or widespread as CSV/TSV.
```

HenrikHolst:
> Since `@SSV` is ambiguous (between space or semicolon) and `@DSV(delim)` presents implementation issues, the obvious choice is `@SCSV`. This is confirmed by ChatGPT's advice:
> 
> ```
> The most consistent and intuitive abbreviation for "semicolon-separated values" in the context of DSV (Delimiter-Separated Values) is:
> 
> SCSV
> 
> This follows the same pattern as:
> 
> CSV – Comma-Separated Values
> 
> TSV – Tab-Separated Values
> 
> SSV – Space-Separated Values
> 
> So:
> 
> SCSV – Semicolon-Separated Values
> 
> This abbreviation is already in informal use in some data communities and tools, even if it's not as standardized or widespread as CSV/TSV.
> ```

Sounds like a great move forward so I've implemented that now.

PR #3317, opened 2025-05-07
docs: Fix the streaming section's intro example to match the actual usage examples

tsibley:


tsibley:
Alternatively, if the introductory example is thought to be more illustrative, than I can update the usage examples to match it instead.

itchyny:
I don't know `[1,[2]]` is more actual use case than `[0,[1]]`, but I think you're saying the example is a bit confusing with the indexing number. How about changing to `["a",["b"]]` then?

tsibley:
Ah, this isn't about making the examples more like real-world usage. This is about correcting a mistake in the docs.

The introductory prose for the streaming section says all the usage examples (e.g. of `truncate_stream`, etc) in the section use `[0,[1]]` as input, but they actually use `[1,[2]]`. This commit makes the section internally consistent and correct. (and much less confusing!)

If we're going to stick with example input of either `[1,[2]]` or `[0,[1]]`, I find the first less confusing because the values aren't the same as the path indexes. So I corrected the intro to use that and be consistent with the usage examples rather than the other way around. But I don't know the intent of the original author and why they seemingly waffled between two different example inputs, so my "more illustrative" comment was about being open to correcting the docs the other way around.

I agree a better example all together (like `["a",["b"]]`) would be nicer, but I was trying to stick to a minimal correction here.

itchyny:
Thanks for clarifying. Can you fix the dev manual too?

tsibley:
Done.

tsibley:
Updated `jq.1.prebuilt` too.

tsibley:
See also https://github.com/jqlang/jq/pull/3322.

PR #3319, opened 2025-05-10
src: added 'trimstr' function

### LINKED ISSUE #3314: feature: function "trim" for trimming on both sides
Given that we now have both [ltrimstr][1] and [rtrimstr][2], please also add the function `trim` which simply trim a string on both sides, which would be much handy than having to call both.

[1]: https://jqlang.org/manual/#ltrimstr
[2]: https://jqlang.org/manual/#rtrimstr

mihaigmarin:
Closes #3314

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -76,7 +76,7 @@ def fromdate: fromdateiso8601;
 def todate: todateiso8601;
 def ltrimstr($left): if startswith($left) then .[$left | length:] end;
 def rtrimstr($right): if endswith($right) then .[:$right | -length] end;
-def trimstr($val): ltrimstr($val)|rtrimstr($val);
+def trimstr($val): ltrimstr($val) | rtrimstr($val);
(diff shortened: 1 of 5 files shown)


src/builtin.jq
```diff
@@ -76,6 +76,7 @@ def fromdate: fromdateiso8601;
 def todate: todateiso8601;
 def ltrimstr($left): if startswith($left) then .[$left | length:] end;
 def rtrimstr($right): if endswith($right) then .[:$right | -length] end;
+def trimstr($val): ltrimstr($val)|rtrimstr($val);
```
itchyny:
```suggestion
def trimstr($val): ltrimstr($val) | rtrimstr($val);
```

itchyny:
I think adding this function is good. Please add description to the dev manual.

mihaigmarin:
Thank you for your reply, I have updated the dev manual

itchyny:
Thanks. Maintainers, any thoughts?

itchyny (approved):

PR #3322, opened 2025-05-13
docs: Use a better input example in the streaming section

tsibley:
["a",["b"]] seems better than [1,[2]] as the string values won't be
confusable with the indexes in <path>.

Suggested by itchyny <itchyny@cybozu.co.jp> in review commentary.¹

¹ <https://github.com/jqlang/jq/pull/3317#issuecomment-2870889151>

-----

Closes #3317.


itchyny (approved):
Thank you.

PR #3323, opened 2025-05-14
Test daylight time for strflocaltime

sihde:
Add a test using a time and zone where DST is in effect to confirm the offset and zone name are as expected. The other tests happen to use times when DST is not in effect.

This test would fail in the current release version (jq-1.7.1) but passes since PR #3203 fixed this and other datetime-related issues. This test case covers the behavior described in issue #1912 .


itchyny (approved):
Thank you.

PR #3325, opened 2025-05-19
Fix and refactor scan-build workflow to run on the master branch

itchyny:
I fixed the scan-build workflow to run on the master branch. The workflow mistakenly configured as it runs only when the workflow file is updated. I also refactored the workflow a bit. Tested on my fork: https://github.com/itchyny/jq/actions/runs/15124790716.

### DIFF
--- a/.github/workflows/scanbuild.yml
+++ b/.github/workflows/scanbuild.yml
@@ -1,75 +1,47 @@
-name: Linux Clang scan-build Static Analyzer Build
+name: Clang scan-build Static Analyzer Build
 
 on:
-    # Don't do this for PRs.  It eats up minutes.
-    push:
-      branches:
-         - 'master'
-         - 'scan-build'
-      paths:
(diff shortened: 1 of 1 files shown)


wader (approved):
Looks good to me. But is the scan-build so resource extensive or slow so it would now work to run on all PRs?

itchyny:
It takes 7 minutes on my fork. Is that too expensive for PRs? I don't know. The number of PRs is not so many.
Anyways we currently don't notice the scan-build bug reports, and there are 10 suspicious reports now.
If we fixed them, and make fail if it finds any bugs, we may want to enable for PRs.

wader:
The slowest checks currently seems to be in the 4-5min range so maybe ok? but maybe start with fixing the issues and just run for master first?

itchyny:
Yeah, let me fix the issue the workflow did not run for months even in the master branch.

PR #3326, opened 2025-05-21
Fix parser to allow binary operators for binding syntax

### LINKED ISSUE #2446: Likely wrong precedence of the 'as' operator
**Describe the bug**

In the example below the numbers are squared while the output is expected to be the same as input

```
$ jq '.[]|.*3 as $v|.' <<<'[1, 4, 5,3]'
1
16
25
9
```

Similarly input is doubled but should be the same:
```
$ jq '.[]|.+1 as $v|.' <<<'[1, 4, 5,3]'
2
8
10
6
```

 - OS and Version: all platforms including https://jqplay.org/
 - jq version : 1.6

### LINKED ISSUE #3284: The Precedence Issue of `as` and Binary Operations
link https://github.com/01mf02/jaq/issues/261

```
$ jq -n '"x" as $n | "a"+"y" as $s | $n+","+$s'
"ax,y"
```
Like:
```
$ jq -n '"x" as $n | "a"+("y" as $s | $n+","+$s)'
"ax,y"
```
Or (scope has some impact):
```
jq: error: $s is not defined at <top-level>, line 1:
"x" as $n | "a"+("y" as $s | $n)+","+$s
jq: 1 compile error
```

Intuitively, it should be:
```
$ jq -n '"x" as $n | ("a"+"y") as $s | $n+","+$s'
"x,ay"
```

itchyny:
Previously, the precedence of `as` binding was higher than the binary
operators. This results in unexpected behaviors, `1 + 2 as $x | -$x`
was `-1` instead of `-3`. The commit 92daac90f6c3 (PR #3053) changed
the behavior to emit syntax error on this case. Now, this commit allows
binary operators with `as` binding. This fixes #2446, and fixes #3284.


wader (approved):
👍 I'm guessing this will fix more scripts than it breaks?

itchyny:
Yeah, I hope so.

PR #3158, opened 2024-08-09
Increase MAX_PARSING_DEPTH to 512

monperrus:
To, I hit error `parse error: Exceeds depth limit for parsing` on reasonable JSON files in production.

This PR fixes this limitation.

PR #3063, opened 2024-03-11
Make maximum depth configurable via CLI option

mgree:
Addresses #2846. (I didn't see any contributor guidelines/linting info... please let me know if you'd like me to change anything!)

This PR adds state to two places to track maximum parser depth: in `jq_state` and in `jv_parser`.

It might seem like `jv_parser` is enough on its own, but many other parts of the code parser values holding only a `jq_state`. It's not clear to me that these aren't mostly/entirely raw parses (which wouldn't need to track depth), but it seemed easiest to track the state thoroughly.

There are a few helper functions at the end of `jv_parse.c` that have no handle at all---neither a `jq_state` nor a `jv_parser`. I had these just use the `DEFAULT_MAX_PARSING_DEPTH`. They are _mostly_ for fuzzing, it seems, but `jv_parse` gets called in `main` during option processing. It wouldn't be hard to have this use the current `parser_maxdepth`.

PR #3328, opened 2025-05-29
Increase the maximum parsing depth for parsing JSON to 10000

### LINKED ISSUE #2846: increase depth limit
**Describe the bug**
A long minified json seems to exceed the parser limits.

**To Reproduce**
I get this error:
`parse error: Exceeds depth limit for parsing at line 1, column 4381704`
(I can't attach the original file, sorry)

**Expected behavior**
jq works with large files

- OS and Version: Ubuntu 22.04 WLS
- jq version: jq-1.6

itchyny:
This is the same limit as encoding/json in Go and should be sufficient for most use cases.
Closes #2846, closes #3158, and addresses #1214.

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -11,7 +11,7 @@
 typedef const char* presult;
 
 #ifndef MAX_PARSING_DEPTH
-#define MAX_PARSING_DEPTH (256)
+#define MAX_PARSING_DEPTH (10000)
 #endif
 
 #define TRY(x) do {presult msg__ = (x); if (msg__) return msg__; } while(0)


thaliaarchi:
Have you seen https://github.com/jqlang/jq/pull/3063 (Make maximum depth configurable via CLI option)?

itchyny:
Of course. But it's better if you can use for most use cases without having to worry about the depth option.

wader (approved):
Looked up the original issue where @mgree ran into the limit, seems to have been querying some deeply nested AST https://github.com/jqlang/jq/pull/3063#issuecomment-2027603660, but didn't find what limit would have worked in this case. I think it's a good idea to increase the default limit as a start, an option to change it could be done separately later.

itchyny:
Thanks. We could reconsider whether we really need option if the limit is 10000.

liquidaty:
Appreciate the update, thank you. FWIW it is not always better to have a fixed-but-high limit vs a configurable limit. If I'm processing a very large JSON input that should never have a depth > N, I certainly have many use cases where I'd rather quickly disqualify the invalid input than continue to process it.

That said, I agree that enforcing a configurable limit is merely a nice-to-have (esp compared to raising a hardcoded limit that is too low)

mgree:
I could find out what depth would have worked if that would help the PR get merged, but the key thing for us would be the option to turn off the maximum depth, since our JSON is homemade.

(Let me know if it's worth resolving the merge skew on #3063. If it won't help it get merged, though, I won't bother.)

itchyny:
I don't understand why many people think limiting the depth is enough to prevent invalid or huge input. Why you don't need to limit the number of elements in arrays and objects, or string literal length.

liquidaty:
Just because it does not comprehensively address the entire problem space by itself does not make it worthless.

mgree:
I don't know if you're including my use case in "many people", but our AST-parsing use case requires _no_ depth limit, not a fixed one. Changes like this one aren't (in general) going to solve our problem.

PR #3342, opened 2025-06-07
Fix portability of strptime especially for Windows

itchyny:
Close #2071.

PR #3365, opened 2025-07-03
jq.test: drop non-portable %F test

alyssais:
%F is a non-portable GNU extension, not supported by all strptime implementations (for example musl's).

### DIFF
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -1848,10 +1848,6 @@ try ["OK", strflocaltime({})] catch ["KO", .]
 "2015-03-05T23:51:47Z"
 [[2015,2,5,23,51,47,4,63],1425599507]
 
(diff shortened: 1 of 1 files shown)


wader:
@itchyny was there a particular reason to have a %F test https://github.com/jqlang/jq/pull/3342?

@alyssais thanks for the report! working on updating alpine to jq 1.8.1?

Hmm maybe yet another point for using the same strptime implementations on all platforms.

Another solution could be to split out some tests, like we did with some oniguruma test.

alyssais:
> working on updating alpine to jq 1.8.1?

[Nixpkgs](https://github.com/NixOS/nixpkgs/pull/422018), actually. Alpine and co would have the same problem though.

itchyny (approved):
Thanks!

Issue #1225, opened 2016-09-03
Feature request: Option to suppress normal output


fearphage:
I quite often want to confirm that some key is in a dictionary or something similar and I rely on `jq --exit-status` to relay that information. I rarely use the output from the command at that point. It would be convenient to be able to suppress normal output entirely.

## Use case

```shell
is-key-valid() {
  echo "$1" | jq --exit-status '."'$2'"' > /dev/null
}

is-key-valid '{"foo": "bar"}' catpants
``` 

With this option, there'd be no need to pipe `stdout` to `/dev/null`. This is very similar to the `grep` option:

```shell
✗ grep --help | grep quiet
  -q, --quiet, --silent     suppress all normal output
````

**Comment by G-Rath** [comment]
It's been three years - any chance of this?

**Comment by pkoppstein** [comment]
@G-Rath - Apparently not anytime soon.

So, besides redirecting /dev/null (or NUL:), you may also wish to consider using `halt_error/1`, e.g.

    jq -n '{a:1} | if has("a") then null|halt_error(1) else null|halt_error(2) end'
   echo rc=$?

The key here is that `null|halt_error(N)` will emit nothing.

**Comment by socketpair** [comment]
Yes, but in systemd's `ExecCondition=` redirecting stdout to `/dev/null` is not available!

**Comment by themadsens** [comment]
You can do:
`jq -j '""'`
Which will write a "raw" empty string with no trailing newline. ie. nothing

**Comment by wader** [comment]
Can also use `empty` to output nothing
```sh
$ echo 123 | jq '., "abc" | empty'
$
```

itchyny:
@wader That wouldn't work with the intention of this issue; `--exit-status`.

wader:
@itchyny ah sorry 👍 note to self: read whole thread first

leonid-s-usov:
I doubt it's a good idea to change the behavior of `--exit-status`. However, I agree that the request makes sense.

Maybe a solution could be to add `--exit-status-only` variant which will do as the OP suggested: work as the original `--exit-status` but suppress all output, including stderr. We could consider deprecating the old option.

itchyny:
I agree with adding  `-q, --quiet` options, since it covers JSON file validation use case without `--exit-status` option (on validating a JSON file; the content could be just `false`).

emilyyyylime:
I have also just found myself trying `jq -eq` and was surprised to see it doesn't exist; would love to see this implemented

nhed:
[...]

PR #3367, opened 2025-07-04
feat: allow writing to output file

### LINKED ISSUE #2418: Output to file without STDOUT redirection
Is QJ able to output its result directly to a file without using redirection of STDOUT? I need to use it from Windows Registry entry for JSON files context menu.

christf:
This allows to write the output to a file and closes #2418 . When calling jq inside a docker context this will be helpful as it means jq will not have to be called from within a shell with output redirection.

Caveats do apply:

* I have only started looking at the jq codebase today, so the code may be doing something insane

### DIFF
--- a/docs/content/manual/v1.8/manual.yml
+++ b/docs/content/manual/v1.8/manual.yml
@@ -227,6 +227,10 @@ sections:
         like awk's -f option. This changes the filter argument to be
         interpreted as a filename, instead of the source of a program.
 
+      * `-o` / `--output-file filename`:
+
+        Write output to the file called filename instead of stdout.
+
       * `-L directory` / `--library-path directory`:
 
         Prepend `directory` to the search list for modules.  If this
--- a/src/jv_print.c
+++ b/src/jv_print.c
@@ -35,6 +35,8 @@ static const char *colors[] = DEFAULT_COLORS;
 #define COLORS_LEN (sizeof(colors) / sizeof(colors[0]))
 #define FIELD_COLOR (colors[7])
 
+extern FILE *ofile;
+
 static char *colors_buf = NULL;
 int jq_set_colors(const char *code_str) {
   if (code_str == NULL)
@@ -390,7 +392,7 @@ void jv_dumpf(jv x, FILE *f, int flags) {
 }
 
 void jv_dump(jv x, int flags) {
-  jv_dumpf(x, stdout, flags);
+  jv_dumpf(x, ofile, flags);
 }
 
 /* This one is nice for use in debuggers */
--- a/src/main.c
+++ b/src/main.c
@@ -42,6 +42,8 @@ extern void jv_tsd_dtoa_ctx_init();
 
 int jq_testsuite(jv lib_dirs, int verbose, int argc, char* argv[]);
 
+FILE *ofile;
+
 /*
  * For a longer help message we could use a better option parsing
  * strategy, one that lets stack options.
@@ -81,6 +83,7 @@ static void usage(int code, int keep_it_short) {
       "                            each output;\n"
       "  -a, --ascii-output        output strings by only ASCII characters\n"
       "                            using escape sequences;\n"
+      "  -o, --output-file         output to file instead of stdout\n"
       "  -S, --sort-keys           sort keys of each object on output;\n"
       "  -C, --color-output        colorize JSON output;\n"
       "  -M, --monochrome-output   disable colored output;\n"
@@ -179,7 +182,7 @@ static int process(jq_state *jq, jv value, int flags, int dumpopts, int options)
   while (jv_is_valid(result = jq_next(jq))) {
     if ((options & RAW_OUTPUT) && jv_get_kind(result) == JV_KIND_STRING) {
       if (options & ASCII_OUTPUT) {
-        jv_dumpf(jv_copy(result), stdout, JV_PRINT_ASCII);
+        jv_dumpf(jv_copy(result), ofile, JV_PRINT_ASCII);
       } else if ((options & RAW_OUTPUT0) && strlen(jv_string_value(result)) != (unsigned long)jv_string_length_bytes(jv_copy(result))) {
         jv_free(result);
         result = jv_invalid_with_msg(jv_string(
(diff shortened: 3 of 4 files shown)


christf:
Test on linux with utf-8:

✘-INT ~/Programmierung/jq [ofile|✚ 1…1] 
19:42 $ echo '{ "foo": "💩" } '
{ "foo": "💩" } 
✔ ~/Programmierung/jq [ofile|✚ 1…1] 
19:42 $ echo '{ "foo": "💩" } ' | ./jq .
{
  "foo": "💩"
}
✔ ~/Programmierung/jq [ofile|✚ 1…1] 
19:42 $ echo '{ "foo": "💩" } ' | ./jq . -o foo
✔ ~/Programmierung/jq [ofile|✚ 1…1] 
19:43 $ cat foo 
{
  "foo": "💩"
}
✔ ~/Programmierung/jq [ofile|✚ 1…1] 
19:43 $

wader:
There has been talks long time ago about adding adding [`-i`/`--inplace`](https://github.com/jqlang/jq/issues/105) and [jaq](https://github.com/01mf02/jaq) already supports it. Could that be an option? but one thing about how jq handles multiple inputs files, as they were concatenated, probably have be resolved in some way in that case.

```sh
$ echo -n 12 > a.json
$ echo -n 34 > b.json

$ jq . a.json b.json
1234

$ jaq . a.json b.json
12
34
$ jaq -i '.+1' a.json b.json
$ jaq . a.json b.json
13
35
```

christf:
> There has been talks long time ago about adding adding [`-i`/`--inplace`](https://github.com/jqlang/jq/issues/105) and [jaq](https://github.com/01mf02/jaq) already supports it. Could that be an option? but one thing about how jq handles multiple inputs files, as they were concatenated, probably have be resolved in some way in that case.
> 
> ```shell
> $ echo -n 12 > a.json
> $ echo -n 34 > b.json
> 
> $ jq . a.json b.json
> 1234
> 
> $ jaq . a.json b.json
> 12
> 34
> $ jaq -i '.+1' a.json b.json
> $ jaq . a.json b.json
> 13
> 35
> ```

that is an interesting thought. In the use case I am hoping to solve, the output file needs to be in a certain place that is designed to hold only results. This is, what the tekton task specification might look like

```
- image: ghcr.io/jqlang/jq:1.9.0
  name: digest-to-results
  command:
    - jq -j '.[0].value' -o /tekton/results/IMAGE-DIGEST image-digested
```

christf:
How can this best be progressed?

wader:
Hey sorry for the delay. I think the implementation look resonable. I'm mostly concern and reluctant about adding more arguments and features and in this case maybe that we might want inplace edit in the future, bloat to have both?

It would also be nice to get some input from some other maintainers, ping @itchyny @emanuele6 @pkoppstein

christf:
> bloat to have both?

You, of course can make that decision. For the case I outlined above, in-place edit would not solve the problem as tekton separates inputs and outputs, with inputs being accessible read-only.

thaliaarchi:
I don’t agree that this is bloat, even if `--inplace` is also implemented. `--inplace` implies that the entire input be read before any output be written, which is made more difficult by streaming inputs and outputs, and would require a buffer. Although these flags both concern output files, I think they’re rather unrelated. There’s also several flags concerning input files.

wader:
To clarify i menat to have both might be bloaty both in terms of implementation and adding more cli arguments, but don't think i have that strong opinion about more than concern about making the code a lot less maintainable. Maybe a good pitch to do any of this could be a nice refactor? not that i know that much the input/output code

christf:
> To clarify i menat to have both might be bloaty both in terms of implementation and adding more cli arguments, but don't think i have that strong opinion about more than concern about making the code a lot less maintainable. Maybe a good pitch to do any of this could be a nice refactor? not that i know that much the input/output code

Where do you see the refactoring potential in the area of the output code?
What should be the goal of the refactoring? 
Why is a changeset of <50LOC a trigger for a suggestion of refactoring of two different areas of code when one of these is not even touched by the changeset?

thaliaarchi:
@wader I think calling it bloat is definitely reaching too far. It seems useful to me and the code only needs minor fixups.

@christf I'm reviewing it now. I think jq avoids global state so multiple VMs can be instantiated, so the global `FILE *ofile` would need to be made local. The biggest change would be replacing removing `jv_dump` and replacing all uses with `jv_dumpf`. I'm preparing a patch you can apply on your PR with this and some nits.

thaliaarchi:
@christf I ended up making much more significant changes than I expected, so I figured a new PR would be warranted. I have kept your commit, as it is useful as a base, but I squashed a couple tiny changes into it (e.g., consistent ordering and formatting) to reduce the diff. See https://github.com/jqlang/jq/pull/3410.

thaliaarchi:
> To clarify i menat to have both might be bloaty both in terms of implementation and adding more cli arguments, but don't think i have that strong opinion about more than concern about making the code a lot less maintainable. [...] not that i know that much the input/output code

I'm probably the last person to have touched the argument parsing. I don't think this adds an abnormal amount of code for a new flag.

As for in-place editing, I think the work here actually assists in that effort and it probably wouldn't take much effort on top of what's here. Before processing any of the input, it would read the first input file to the end and store it in a buffer. then seek back to the start, change it to write-only, set it to be the `ofile`, then proceed as normal. (Or alternatively, use a temp file for the output, until finished, then overwrite.) That would reuse everything from --output-file.

Also, in my version of --output-file, I clearly delineate between what things write to the primary output, so the JSON output stream would go to --output-file/--inplace and a disassembly or debug trace would still go to stdout. You'd want that for --inplace too.

christf:
Thank you very much for picking up the pieces. Your work goes much beyond what I initially raised as a PR. Also it leaves the codebase in better shape by preparing for a feature. That bit about the debug trace is an aspect that wasn't even discussed in here. Thank you!

wader:
> I'm probably the last person to have touched the argument parsing. I don't think this adds an abnormal amount of code for a new flag.

Sorry i didn't mean to come of as harsh. Maybe just my bad memories of trying to add/fix some IO-related things like https://github.com/jqlang/jq/issues/1225 that ended up quite complex for how to implement and how it would work in streaming mode, with `-e` etc and combinations of it.

But great work and it would be good to get some more maintainers opinon on this before spending much time on it.

Pandapip1:
Looks like this got stalled, although the code looks fine. Any updates?

thaliaarchi:
@Pandapip1 We're close on https://github.com/jqlang/jq/pull/3410

PR #3370, opened 2025-07-09
Fix potential null dereference in jv_sort when sort_items returns NULL

izroxxet:
### Summary

This pull request addresses a memory safety issue in the `jv_sort` function, where a potential null pointer dereference may occur if the result of `sort_items` is not properly validated.

### CWE-476: NULL Pointer Dereference

In the original implementation, `jv_sort` assumes that `sort_items(objects, keys)` always returns a valid pointer. However, when the input array is empty (`length == 0`), `sort_items` returns `NULL`.

Subsequent access to `entries[i]` without checking for NULL may lead to:

- segmentation faults
- undefined behavior
- crashes when sorting empty or malformed input

The following static analysis error was reported:


itchyny:
Tell your AI assistant to read the code more carefully. NULL pointer dereference never happens in this code.

PR #3379, opened 2025-07-27
Fix potential buffer overflow in jv_parse_sized_custom_flags

byteallen:
The `jv_parse_sized_custom_flags` function uses `jv_string_fmt` with the `%s` format specifier, which assumes the provided `string` is null-terminated. If string is not null-terminated, this can lead to undefined behavior such as memory corruption, crashes, or security vulnerabilities due to buffer overflow.


itchyny:
`jv_string_value` returns a NULL-terminated string.

PR #3387, opened 2025-08-08
Add clean_timestamp builtin for readable ISO 8601 formatting

calilkhalil:
## Summary
This PR adds the `clean_timestamp` built-in function to convert ISO 8601 timestamps to human-readable format, "YYYY-MM-DD HH:MM:SS".

## Problem Solved
Currently, formatting ISO timestamps for human consumption requires complex split/join operations. This function provides a simple, intuitive solution for a common data processing need.

## Implementation Details
- Pure jq implementation in `src/builtin.jq`
- Automatic detection of seconds vs milliseconds for Unix timestamps
- Handles various ISO 8601 formats including timezones and microseconds
- Preserves null values and non-timestamp strings unchanged
- Comprehensive documentation with practical examples
- Full backward compatibility maintained

## Testing
- All existing tests pass (`make check`)
- Extensive testing across various input formats
- Examples included in documentation are tested automatically

## Examples
```jq
"2020-01-01T23:43:12.123Z" | clean_timestamp    # "2020-01-01 23:43:12"
1577919792 | clean_timestamp                    # "2020-01-01 23:03:12"

### DIFF
--- a/docs/content/manual/v1.8/manual.yml
+++ b/docs/content/manual/v1.8/manual.yml
@@ -384,33 +384,33 @@ sections:
           pass regardless of whether that option is used.
 
         examples:
-          - program: "."
+          - program: '.'
             input: '"Hello, world!"'
             output: ['"Hello, world!"']
 
-          - program: "."
-            input: "0.12345678901234567890123456789"
-            output: ["0.12345678901234567890123456789"]
+          - program: '.'
+            input: '0.12345678901234567890123456789'
+            output: ['0.12345678901234567890123456789']
 
           - program: '[., tojson] == if have_decnum then [12345678909876543212345,"12345678909876543212345"] else [12345678909876543000000,"12345678909876543000000"] end'
-            input: "12345678909876543212345"
-            output: ["true"]
+            input: '12345678909876543212345'
+            output: ['true']
 
           - program: '[1234567890987654321,-1234567890987654321 | tojson] == if have_decnum then ["1234567890987654321","-1234567890987654321"] else ["1234567890987654400","-1234567890987654400"] end'
-            input: "null"
-            output: ["true"]
+            input: 'null'
+            output: ['true']
 
-          - program: ". < 0.12345678901234567890123456788"
-            input: "0.12345678901234567890123456789"
-            output: ["false"]
+          - program: '. < 0.12345678901234567890123456788'
+            input: '0.12345678901234567890123456789'
+            output: ['false']
 
           - program: 'map([., . == 1]) | tojson == if have_decnum then "[[1,true],[1.000,true],[1.0,true],[1.00,true]]" else "[[1,true],[1,true],[1,true],[1,true]]" end'
-            input: "[1, 1.000, 1.0, 100e-2]"
-            output: ["true"]
+            input: '[1, 1.000, 1.0, 100e-2]'
+            output: ['true']
 
-          - program: ". as $big | [$big, $big + 1] | map(. > 10000000000000000000000000000000) | . == if have_decnum then [true, false] else [false, false] end"
-            input: "10000000000000000000000000000001"
-            output: ["true"]
+          - program: '. as $big | [$big, $big + 1] | map(. > 10000000000000000000000000000000) | . == if have_decnum then [true, false] else [false, false] end'
+            input: '10000000000000000000000000000001'
+            output: ['true']
 
       - title: "Object Identifier-Index: `.foo`, `.foo.bar`"
         body: |
@@ -433,17 +433,17 @@ sections:
           `.foo::bar` does not.
 
         examples:
-          - program: ".foo"
+          - program: '.foo'
             input: '{"foo": 42, "bar": "less interesting data"}'
-            output: ["42"]
+            output: ['42']
 
-          - program: ".foo"
+          - program: '.foo'
             input: '{"notfoo": true, "alsonotfoo": false}'
-            output: ["null"]
+            output: ['null']
 
           - program: '.["foo"]'
             input: '{"foo": 42}'
-            output: ["42"]
+            output: ['42']
 
       - title: "Optional Object Identifier-Index: `.foo?`"
         body: |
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by pkoppstein** [comment]
I'd like to request some discussion about an alternative name for the proposed function, since there are many variants of a timestamp that could reasonably be called "clean".  One possibility would be simply "ymdhms", which has the advantage that it makes quite clear that milliseconds are not included while also anticipating variants such as "ymd" and "ymdhmsm".

**Comment by calilkhalil** [comment]
> I'd like to request some discussion about an alternative name for the proposed function, since there are many variants of a timestamp that could reasonably be called "clean".  One possibility would be simply "ymdhms", which has the advantage that it makes quite clear that milliseconds are not included while also anticipating variants such as "ymd" and "ymdhmsm". 
> 
> 
> 
> 

I like the idea of ymd

**Comment by itchyny** [comment]
We should support timezone offset for `fromdate` (#1053) and parsing (and formatting?) milliseconds for `strptime` (and `strftime`?) (#1409). When these issues are resolved, we don't need to provide the proposed builtin filter. These issues are technically difficult and have not been resolved for years, so I understand the desire to provide feature using this kind of implementation though.

**Comment by calilkhalil** [comment]
> We should support timezone offset for `fromdate` (#1053) and parsing (and formatting?) milliseconds for `strptime` (and `strftime`?) (#1409). When these issues are resolved, we don't need to provide the proposed builtin filter. These issues are technically difficult and have not been resolved for years, so I understand the desire to provide feature using this kind of implementation though.

Thanks for highlighting those related issues. While #1053 and #1409 address important core functionality gaps, I see `clean_timestamp` serving a different purpose.

The proposed function addresses a specific, common use case that exists independently of timezone/millisecond parsing improvements. Even with enhanced `fromdate`/`strptime`, users would still need to chain operations to get human-readable output.

The fact that those issues have remained unresolved for years actually strengthens the case for providing this functionality now. Users shouldn't have to wait indefinitely for complex core improvements when a targeted solution can address their current needs.

I'd be interested in hearing from the maintainers about their perspective on balancing core improvements vs. practical utility functions like this.

**Comment by pkoppstein** [comment]
@calilkhalil - I agree with most of the main points in your most recent post, but I think that to buttress them, there needs to be a comprehensive specification of what exactly the proposed function does.  Perhaps you've already provided these details in your proposed updates to the documentation, but either way, it might be worthwhile discussing them here, in particular:

1) Assuming the output always has the form "year-MM-DD HH:MM:SS", how is rounding or truncation to seconds specified if the input specifies microseconds or some other fractional number of seconds?

2) If the input specifies a timezone, in what timezone is the output?  What happens if the input specifies an unknown timezone?

3) How are years such as 32 or 9032 to be handled?  (Will 'year' always have at least four digits?)

4) How are BC/BCE timestamps to be handled?

**Comment by calilkhalil** [comment]
@pkoppstein -  Thanks for diving into the specifics! Let me clarify how it actually works:

**Fractional seconds**: It just truncates whatever comes after the seconds. No rounding, just drops the extra precision since most people want to see clean seconds in readable output.

```jq
"2020-01-01T23:43:12.999999Z" | clean_timestamp  # "2020-01-01 23:43:12"
```

**Timezones**: The function doesn't do any timezone conversion - it just strips off the timezone part and shows you the time as written. If there's a weird timezone format, it tries to parse what it can and ignores the broken parts.

```jq
"2020-01-01T15:30:00+05:00" | clean_timestamp  # "2020-01-01 15:30:00"
"2020-01-01T15:30:00Z" | clean_timestamp       # "2020-01-01 15:30:00"
```

**Years**: Whatever the input shows is what you get. I kept it simple rather than trying to guess padding requirements.

```jq
"0032-01-01T12:00:00Z" | clean_timestamp  # "0032-01-01 12:00:00"
"9032-01-01T12:00:00Z" | clean_timestamp  # "9032-01-01 12:00:00"
```

**BC dates**: These break the function - it'll just return the input unchanged. ISO 8601 doesn't really handle BC properly anyway, and neither do jq's existing date functions.

```jq
"-0001-01-01T12:00:00Z" | clean_timestamp  # "-0001-01-01T12:00:00Z" (unchanged)
```

The way I see it, this is more of a "make timestamps readable for humans" function than a full date parser. When you're digging through CloudTrail logs or incident data, you just want to quickly see "oh, this happened at 2:30 PM on January 1st" without doing mental gymnastics on ISO strings.

I can add these details to the docs if it helps clarify the expected behavior.

pkoppstein:
@calilkhalil wrote:
>  I can add these details to the docs if it helps clarify the expected behavior.

To avoid wasting your time, you might want to wait until you get some "buy-in" from someone else.
I'm also wondering whether it wouldn't be better to treat "unsupported" or "unexpected" input as an error.
Consider for example an input of a very large-magnitude negative number.

calilkhalil:
@pkoppstein - You're right about the error handling - that's a fair point. The current implementation does silently pass through unexpected inputs like large negative numbers, which isn't ideal behavior for a builtin.

Regarding waiting for buy-in: I understand the concern about adding convenience functions without clear community need. That said, I've seen this pattern come up repeatedly in my work with incident analysis logs (CloudTrail, GCP audit logs, etc.) where readable timestamps are constantly needed.

But I take your point about error handling. A more robust version should probably:

```jq
def clean_timestamp:
  if type == "string" and test("^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}") then
    # existing logic
  elif type == "number" and . >= 0 and . < 253402300799 then  # valid unix range
    # existing numeric logic  
  elif . == null then
    null
  else
    error("clean_timestamp: unsupported input type or value")
  end;
```

Would stricter error handling address your concerns about the implementation? Or do you think the function itself doesn't have sufficient justification regardless of implementation quality?

I'm open to feedback from maintainers on whether this addresses a real need or if effort would be better spent elsewhere.

calilkhalil:
After valuable feedback from the community, I've evolved this proposal into a more robust implementation with strict error handling and additional forensic analysis capabilities.

The new PR ([#3425](https://github.com/jqlang/jq/pull/3425)) includes:
- `toreadable`: Improved version with comprehensive input validation
- `fromwebkit`: New function for browser forensics use cases

Closing this in favor of the updated implementation. Thank you for the feedback that helped shape the improved version.

PR #3407, opened 2025-09-13
Fix year 2038 problem on 32-bit platforms (fix #3369)

### LINKED ISSUE #3369: jq 1.8+ cannot be built on Debian i386
**Describe the bug**

jq 1.8.0+ cannot be built on Debian i386 due to following unit test fails. The following is error log.

```
FAIL: tests/optionaltest
========================

Test #1: 'fromdate' at line number 4
*** Insufficient results for test at line number 6: fromdate
Test #2: 'strftime("%A, %B %e, %Y")' at line number 9
1 of 2 tests passed (0 malformed, 0 skipped)
FAIL tests/optionaltest (exit status: 1)
```

**To Reproduce**

See the build log in https://buildd.debian.org/status/package.php?p=jq.

**Expected behavior**

The build shall be success.

- Debian Trixie, i386
- jq 1.8.0, 1.8.1

**Additional context**

itchyny:
This fixes #3369.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -1,3 +1,6 @@
+#if __SIZEOF_POINTER__==4
+# define _TIME_BITS 64
+#endif
 #ifndef __sun__
 # define _XOPEN_SOURCE
 # define _XOPEN_SOURCE_EXTENDED 1


src/builtin.c
```diff
@@ -1,3 +1,6 @@
+#if __SIZEOF_POINTER__==4
+# define _TIME_BITS 64
+#endif
```
wader, line 3: Is there some downside to always defining this? i noticed this also while reading about `_TIME_BITS`:
>It is closely related to _FILE_OFFSET_BITS and depending on implementation, may require it set.

wader (approved):

PR #3414, opened 2025-09-27
Increase the maximum printing depth from 256 to 10000 (fixes #3413)

### LINKED ISSUE #3413: Deep nesting produces `<skipped: too deep>` artifacts in printing output
**Describe the bug**
`MAX_PARSING_DEPTH` in `src/jv_parse.c` was recently increased from 256 to 10000 in jqlang#3328 , but the corresponding `MAX_PRINT_DEPTH` in `src/jv_print.c` was left untouched. Now `jq` does not fail out on parsing structures deeper than 256 anymore, but it produces invalid json when printing them, inserting the `<skipped: too deep>` literal into the output. 

I think the problem is `MAX_PRINT_DEPTH` not being in sync with `MAX_PARSING_DEPTH`, ideally they should both change together.

**To Reproduce**
```
(
    for i in {0..256}; do echo -n '{"k'$i'":'; done
    echo -n '"VALUE"';
    for i in {0..256}; do echo -n '}'; done
) > print_depth_repro.256.json
```

[print_depth_repro.256.json](https://github.com/user-attachments/files/22572499/print_depth_repro.256.json)

```
$ cat print_depth_repro.256.json | cut -c1939-1954
{"k256":"VALUE"}

$ cat print_depth_repro.256.json | jq -c | jq
jq: parse error: Invalid numeric literal at line 1, column 1955

$ cat print_depth_repro.256.json | jq -c | cut -c1939-1966
{"k256":<skipped: too deep>}
```

**Expected behavior**
JSON data that can be parsed without error, should be possible to print out in a valid form as well.

- OS and Version: macOS/Linux (it really doesn't matter)
- jq version 1.8.1

ishnagy:
Proposed fix for https://github.com/jqlang/jq/issues/3413

### DIFF
--- a/src/jv_print.c
+++ b/src/jv_print.c
@@ -17,7 +17,7 @@
 #include "jv_private.h"
 
 #ifndef MAX_PRINT_DEPTH
-#define MAX_PRINT_DEPTH (256)
+#define MAX_PRINT_DEPTH (10000)
 #endif
 
 #define ESC "\033"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,7 +237,7 @@ jobs:
             --disable-shared \
             --enable-static \
             --enable-all-static \
-            CFLAGS="-O2 -pthread -fstack-protector-all"
+            CFLAGS="-O2 -pthread -fstack-protector-all -Wl,--stack,4194304"
           make -j$(nproc)
           file ./jq.exe
           cp ./jq.exe jq-${{ env.SUFFIX }}.exe
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,7 +237,7 @@ jobs:
             --disable-shared \
             --enable-static \
             --enable-all-static \
-            CFLAGS="-O2 -pthread -fstack-protector-all -Wl,--stack,4194304"
+            CFLAGS="-O2 -pthread -fstack-protector-all -Wl,--stack,8388608"
           make -j$(nproc)
           file ./jq.exe
           cp ./jq.exe jq-${{ env.SUFFIX }}.exe
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -2505,15 +2505,20 @@ strflocaltime("" | ., @uri)
 ""
 
 # regression tests for #3413
-# max parse/print depth == 10000
+# upper range bounds should be in sync with the constants defined at
+#   src/jv_parse.c:#define MAX_PARSING_DEPTH (N)
+#   src/jv_print.c:#define MAX_PRINT_DEPTH (N)
+# (N-1)
 reduce range(9999) as $_ ([];[.]) | tojson | fromjson | flatten
 null
 []
 
+# (N)
 reduce range(10000) as $_ ([];[.]) | tojson | try (fromjson) catch . | (contains("<skipped: too deep>") | not) and contains("Exceeds depth limit for parsing")
 null
 true
 
+# (N+1)
 reduce range(10001) as $_ ([];[.]) | tojson | contains("<skipped: too deep>")
 null
 true
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -2503,3 +2503,17 @@ strflocaltime("" | ., @uri)
 0
 ""
 ""
+
+# regression tests for #3413
+# max parse/print depth == 10000
+reduce range(9999) as $_ ([];[.]) | tojson | fromjson | flatten
+null
+[]
+
+reduce range(10000) as $_ ([];[.]) | tojson | try (fromjson) catch test("Exceeds depth limit for parsing")
+null
+true
+
+reduce range(10001) as $_ ([];[.]) | tojson | try (fromjson) catch test("<skipped: too deep>")
+null
+true
--- a/tests/jq.test
+++ b/tests/jq.test
@@ -2510,10 +2510,10 @@ reduce range(9999) as $_ ([];[.]) | tojson | fromjson | flatten
 null
 []
 
-reduce range(10000) as $_ ([];[.]) | tojson | try (fromjson) catch test("Exceeds depth limit for parsing")
+reduce range(10000) as $_ ([];[.]) | tojson | try (fromjson) catch . | (contains("<skipped: too deep>") | not) and contains("Exceeds depth limit for parsing")
 null
 true
 
-reduce range(10001) as $_ ([];[.]) | tojson | try (fromjson) catch test("<skipped: too deep>")
+reduce range(10001) as $_ ([];[.]) | tojson | contains("<skipped: too deep>")
 null
 true


src/jv_print.c
```diff
@@ -17,7 +17,7 @@
 #include "jv_private.h"
 
 #ifndef MAX_PRINT_DEPTH
-#define MAX_PRINT_DEPTH (256)
+#define MAX_PRINT_DEPTH (10000)
```
wader, line 20: I wonder if it should be defined to be`MAX_PARSING_DEPTH`?

wader:
We should probably add a regression test for this. Maybe something like:
```
reduce range(100000) as $_ ([];[.]) | tojson | fromjson | empty
null

reduce range(100001) as $_ ([];[.]) | tojson | test("skipped: too deep")
null
true
```

src/jv_print.c
ishnagy: I'm not that familiar with the conventions around here (and to be honest, with C either), where would you prefer the common definition to go?
```
$ cat src/jv_parse.c src/jv_print.c | grep '#include' | sort | uniq -d
#include "jv.h"
#include "jv_alloc.h"
#include "jv_dtoa.h"
#include "jv_unicode.h"
#include <assert.h>
#include <stdio.h>
#include <string.h>
```
(my guess would be `jv.h`)

ishnagy:
> We should probably add a regression test for this. Maybe something like:
> 
> ```
> reduce range(100000) as $_ ([];[.]) | tojson | fromjson | empty
> null
> 
> reduce range(100001) as $_ ([];[.]) | tojson | test("skipped: too deep")
> null
> true
> ```

```
$ jq -n 'reduce range(10000) as $_ ([];[.]) | tojson | try (fromjson | flatten) catch test("<skipped: too deep>")'
[]
$ jq -n 'reduce range(10001) as $_ ([];[.]) | tojson | try (fromjson | flatten) catch test("<skipped: too deep>")'
true
```
perhaps?

ishnagy:
which testfile would such regression tests belong?
(`tests/jq.test`?)

ishnagy:
I added three regression tests.

For some reason, there seems to be an off-by-one difference between the handling of the parsing and the printing limits. This turned out to be lucky, this way we can have two separate testcases for the parsing and printing errors.

The windows tests are failing, my best idea for the cause were some CR/LF handling issues on checking out `jq.test`. I tried to play around with `.gitattributes`, but to no avail. Unfortunately, I don't have easy access to a windows machine to troubleshoot the issue.

src/jv_print.c
wader: Aha it's in `jv_parse.h`, hmm think i see two ways.

- Put shared constants in `jv_private.h` as we probably don't want to expose them? otherwise prefix then with `JQ_` i guess?
- A bit more involved: Make the max depth for the parser and printer logic configurable and then pass via the jq state so that the limits can live in the cli code

Other ideas?

ishnagy: I found #3063 on the CLI arg question, last update almost a year ago, but still open. Are there any plans to finish that up?

Wrt. the macro defines, I'm sure we could do something cleaner, but we shouldn't go overboard with it, if we are planning to move this limit to CLI anyway.

I think there are two separate tasks here: (1) fixing the asymmetry between the parsing/printing limits (a bug, IMO) (2) refactoring the configurability of those limits to be a bit more flexible than a hardcoded constant. (improvement)

I opened this ticket/PR with the sole intention of fixing (1). This patch provides the simplest possible delta to achieve "correctness" (in the sense of being symmetric regarding parsing and printing), interfering as little as possible with the rest of the code. (2) already has some proposals open, seems to be a lot more complex, and a lot more contentious on how to do it "right".

I'd be happy to work on a separate patch, figuring out the quirks of moving these to CLI, either by finishing up the work already done, or by coming up with a new proposal. But I would prefer to keep the bugfixing and the impovement goals separated, if that's possible.

WDYT?

ishnagy:
@wader 
do you have any thoughts/pointers on how we could track down the issues with the windows builds?

ishnagy:
okie, new theory.

since `oniguruma/installed` passes, but `oniguruma/disabled` fails, can it happen that using
```
test("...")
```
is causing troubles in some configurations?
(assuming `test` depends on some regexp logic, which is not present in windows builds, or when oniguruma is explicitly disabled)

ishnagy:
well, indeed.

after building with `./configure --with-oniguruma=no`:
```
% ./jq -n 'reduce range(10000) as $_ ([];[.]) | tojson | try (fromjson) catch test("Exceeds depth limit for parsing")'
jq: error (at <unknown>): jq was compiled without ONIGURUMA regex library. match/test/sub and related functions are not available.
```

ishnagy:
are there any oniguruma-independent `test` alternatives we can use?

itchyny:
`contains`

ishnagy:
good news: the `oniguruma/disabled` problem is resolved.
bad news: the problems with the windows builds are caused by something else.

itchyny:
This might hit the maximum recursion depth which depends on the system.

ishnagy:
So, how do we proceed?

I don't think the current asymmetry in parse/print is a good way to leave things, even for the short term.

I see two immediate ways to deal with the situation:
* we can make the limits platform dependent. we determine the limit that can be used on windows safely, set that only in windows builds, but leave the 10000 as the limit on the other platforms.
* if we don't want to introduce platform specific constants, we decrease the current 10000 to a windows safe value on every platform.

itchyny:
~We need to rewrite `jv_dump_term` to a non-recursive function.~

itchyny:
Okay, let's increase the stack size.
```diff
-            CFLAGS="-O2 -pthread -fstack-protector-all"
+            CFLAGS="-O2 -pthread -fstack-protector-all -Wl,--stack,16777216"
```

ishnagy:
> We need to rewrite jv_dump_term to a non-recursive function.

are you proposing to include the full rewrite of `jv_dump_term` in _this_ PR? or would this be more of a longer term goal?

I would suggest to do a quick patchup to return to a consistent state re parse/print limits. I tinkered around with the values on a private branch, and 4000/4000 for parse/print seems to be fairly stable on all platforms. If we want to be more conservative on windows, we can do platform specific limits, reverting back to 256/256 on windows, and have 10000/10000 on any other target.

itchyny:
--stack,4194304 B (4MiB) is enough in my experience.

ishnagy:
I managed to get some limited access to a windows box, so I'll make an attempt to produce a backtrace from the windows crash.

Will keep you posted on the results.

(Should give us more confidence regarding the solution than trial-and-erroring a reasonable stack size.)

itchyny:
Can you apply this patch?

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index c771a92..505964c 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,7 +237,7 @@ jobs:
             --disable-shared \
             --enable-static \
             --enable-all-static \
-            CFLAGS="-O2 -pthread -fstack-protector-all"
+            CFLAGS="-O2 -pthread -fstack-protector-all -Wl,--stack,4194304"
           make -j$(nproc)
           file ./jq.exe
           cp ./jq.exe jq-${{ env.SUFFIX }}.exe
```

ishnagy:
> Can you apply this patch?

I'll give this a try as well.

ishnagy:
```
gdb.exe -ex 'set pagination off' -ex 'set confirm off' -ex run -ex bt -ex q --args ./jq.exe -n 'reduce range(5000) as $_ ([];[.]) | tojson | fromjson | flatten' |& tee jq_segfault.bt.txt
```
[jq_segfault.bt.txt](https://github.com/user-attachments/files/22600740/jq_segfault.bt.txt)

ishnagy:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff67261d425 in jvp_utf8_is_valid (in=0x4041ec "[\022",
    end=0x4041ed "\022") at src/jv_unicode.c:79
79        while ((in = jvp_utf8_next(in, end, &codepoint))) {
#0  0x00007ff67261d425 in jvp_utf8_is_valid (in=0x4041ec "[\022",
    end=0x4041ed "\022") at src/jv_unicode.c:79
#1  0x00007ff67260631b in jv_string_append_buf (a=..., buf=0x4041ec "[\022",
    len=1) at src/jv.c:1508
#2  0x00007ff67261b333 in put_buf (s=0x4041ec "[\022", len=1, fout=0x0,
    strout=0x5fefd0, is_tty=0) at src/jv_print.c:99
#3  0x00007ff67261b4f2 in put_char (c=91 '[', fout=0x0, strout=0x5fefd0, T=0)
    at src/jv_print.c:125
#4  0x00007ff67261c2ad in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=4324, F=0x0, S=0x5fefd0) at src/jv_print.c:289
#5  0x00007ff67261c41d in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=4323, F=0x0, S=0x5fefd0) at src/jv_print.c:300
#6  0x00007ff67261c41d in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=4322, F=0x0, S=0x5fefd0) at src/jv_print.c:300
#7  0x00007ff67261c41d in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=4321, F=0x0, S=0x5fefd0) at src/jv_print.c:300
[...]
#4326 0x00007ff67261c41d in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=2, F=0x0, S=0x5fefd0) at src/jv_print.c:300
#4327 0x00007ff67261c41d in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=1, F=0x0, S=0x5fefd0) at src/jv_print.c:300
#4328 0x00007ff67261c41d in jv_dump_term (C=0x7ab270, x=..., flags=0,
    indent=0, F=0x0, S=0x5fefd0) at src/jv_print.c:300
#4329 0x00007ff67261cf2e in jv_dump_string (x=..., flags=0)
    at src/jv_print.c:406
#4330 0x00007ff672645b3d in f_dump (jq=0x7a8a90, input=...)
    at src/builtin.c:432
#4331 0x00007ff6725f9e81 in jq_next (jq=0x7a8a90) at src/execute.c:917
#4332 0x00007ff6725f1bf8 in process (jq=0x7a8a90, value=..., flags=0,
    dumpopts=513, options=4) at src/main.c:179
#4333 0x00007ff6725f4c1a in umain (argc=3, argv=0x5ffe00) at src/main.c:667
#4334 0x00007ff6725f25a0 in wmain (argc=3, wargv=0x79ec30) at src/main.c:284
#4335 0x00007ff6725f10c9 in __tmainCRTStartup ()
    at D:/W/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:236
#4336 0x00007ff6725f1416 in mainCRTStartup ()
    at D:/W/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:122

**Comment by ishnagy** [comment]
The strange thing is that the top of the backtrace diverges from `jv_dump_term`. In case of a pure stack exhaustion, I would expect the sigsegv happening in the recursive function.

Since the function that crashes deals with unicode, this may be due to some windows quirks in handling multibyte strings. (maybe writing beyond the boundaries of a buffer allocated for unicode data?)

**Comment by ishnagy** [comment]
@itchyny 
increasing the stacksize worked, it prevented the crash in windows builds.

**Comment by ishnagy** [comment]
okie, I did some digging.

It seems that on Linux and MacOS the max stacksize is controlled through `ulimit`
```
$ cat /etc/os-release | grep -B1 ^VERSION=
VERSION_ID="24.04"
VERSION="24.04.3 LTS (Noble Numbat)"
$ ulimit -a | grep stack
stack size                  (kbytes, -s) 8192
$
```
```
% sw_vers
ProductName:            macOS
ProductVersion:         15.6.1
BuildVersion:           24G90

% ulimit -a | grep stack
-s: stack size (kbytes)             8176

%
```
on windows, it is determined from the headers in the executable
```
$ objdump.exe -p ./jq.exe | grep Stack
SizeOfStackReserve      0000000000200000
SizeOfStackCommit       0000000000001000

$
```

Should we set the stack size on windows to match the Linux/MacOS defaults?

itchyny:
I have no objections to setting it to 8MiB, and after asking ChatGPT, I don't see any drawbacks to setting it to this value (especially for single-threaded programs). Any thoughts from other maintainers?

wader:
Fine with me 👍 rewrite to not be recursive should probably be a separate PR anyway

wader (approved):

ishnagy:
Hi
@wader,
@itchyny,
is there anything else needed from my side before this can get merged?

itchyny (approved):
Thank you.

PR #3425, opened 2025-10-21
Add fromwebkit and toreadable builtins for forensic analysis

calilkhalil:
- fromwebkit: Convert WebKit/Chromium timestamps to ISO 8601 Used in browser forensics (Chrome, Safari, Edge artifacts) Handles microseconds since 1601-01-01 with validation

- toreadable: Convert timestamps to human-readable format Auto-detects seconds vs milliseconds for Unix timestamps Strict error handling for invalid/malformed inputs

Both functions include comprehensive input validation for forensic analysis and data processing pipelines.

### DIFF
--- a/docs/content/manual/v1.8/manual.yml
+++ b/docs/content/manual/v1.8/manual.yml
@@ -2298,6 +2298,21 @@ sections:
           some systems. In particular, the `%u` and `%j` specifiers for
           `strptime(fmt)` are not supported on macOS.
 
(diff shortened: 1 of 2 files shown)


calilkhalil:
These functions were originally proposed as PRs to add as built-in functions to jq itself. However, they are better suited as an external module.

The jq maintainers suggested this approach, and it turned out to be the right call - it's easier to maintain, distribute, and use.

https://github.com/calilkhalil/jq-forensics

Issue #3313, opened 2025-05-05
Migrate away from oniguruma?


dcermak:
https://github.com/kkos/oniguruma has been archived recently and will not receive any further updates. I would suggest to move away from it and switch to a different regex engine

mrienstra:
Apologies for posting "AI slop", but I'm auditing Claude Sonnet 4, figured I'd share what it came up with, looks like a decent start. Note that this is not a one-shot, the initial response had some old references (as in not as current as I would like). _Edit: OK, it's really not one-shot, I ended up doing some searches and feeding it newer sources, and now I think I'm about to run https://github.com/BurntSushi/rebar locally to see what kind of benchmarks I get, whoops, stumbled into a rabbithole._

## Top Replacement Options

**PCRE2** is the most logical replacement for several reasons:
- It provides compatibility with Oniguruma syntax[^1], making migration easier
- It's actively maintained and widely adopted[^2]
- Recent comprehensive benchmarks show PCRE2-JIT performing excellently across diverse workloads[^3]
- Has comprehensive Unicode support when configured properly
- It supports some alternative regular expression syntax that does not conflict with the Perl syntax in order to provide some compatibility with regular expressions in Python, .NET, and Oniguruma[^1]

**RE2** is another strong candidate:
- Guarantees linear time complexity, preventing regex denial-of-service attacks
- Recent benchmarks show solid performance across many scenarios[^3]
- However, RE2 can be slower when capturing groups is required, which could be problematic for jq's use cases

**Rust's regex crate** (if interfacing from C is acceptable):
- Shows excellent performance in recent comprehensive benchmarks, ranking first overall in search performance[^3]
- Modern, well-maintained implementation with extensive testing
- The rust regex crate now uses the same SIMD algorithm (Teddy) invented as part of Hyperscan[^4]

## Recent Performance Data

According to the most comprehensive recent regex engine benchmark (rebar, 2023)[^3], the overall rankings by geometric mean of speed ratios across diverse workloads are:

**Search Performance Leaders:**
1. rust/regex - 1.54 ratio (best overall)
2. dotnet/compiled - 1.80 ratio  
3. pcre2/jit - 1.86 ratio
4. dotnet/nobacktrack - 2.33 ratio
5. re2 - 2.48 ratio

**Compile Performance Leaders:**
1. rust/regex/lite - 1.21 ratio (best overall)
2. regress - 1.31 ratio
3. icu - 1.33 ratio
[...]

PR #1325, opened 2017-01-27
Add local oniguruma submodule

erikbrinkman:
This is a test to add oniguruma locally as a submodule. This should allow:

- Control over the oniguruma version
- Regex support in all builds
- The potential to cross compile for windows from TravisCI and move away from appveyor

fixes #1059

PR #3432, opened 2025-11-05
Switch to PCRE2 for regular expressions

wader:
Work on oniguruma (https://github.com/kkos/oniguruma) has sadly been discontinued
so we need to find another regex implementation. PCRE2 seem like a good candidate.

vendor/pcre2 commit is at tag pcre2-10.48

Known TODOs:
- compile-ios.sh
- Needs to be a 1.x release as docs relay on that
- libjq usage? dont builtin and require libpcre2? (seems to be what debian does)
- Update docs
  - Breaking changes
    - Drop "l" modifier?
      - Could possibly reimplement
    - Empty pattern and multi-byte code points behave differntly. I think pcre2 is more correct?
      - onig: jq -n '["🚀" | match(""; "g")] | length' -> 5 (per byte it seems)
      - pcre2: jq -n '["🚀" | match(""; "g")] | length' -> 2 (per code point)

Good references:
- https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2demo.c
Pre pcre2_next_match usage:
- https://github.com/PCRE2Project/pcre2/blob/eb3bd3cf1418cb1a0eabf984b0b1e80b6bdd9314~1/src/pcre2demo.c (pre pcre2_next_match usage)
- pcre2test cli tools is a good Playaround for pcre2

Notes:
".+?\\b" test in onig.test seems to be behaves differently depending on pcre2 version.
I suspect this fix in 10.43:
https://github.com/PCRE2Project/pcre2/commit/0a552803330e6eb4234c887a48a1584cb83c0666

I noticed a clang -fsanitize=memory use-of-uninitialized-valu issue but it seems to go awa with pcre2 master

Related to https://github.com/jqlang/jq/issues/3313


wader:
Looking at the download stats maybe it would be ok to skip some arch? and then we could probably also gets away with using docker for cross builds.

```sh
$ curl -s https://api.github.com/repos/jqlang/jq/releases | jq -c '.[] | select(.tag_name == "jq-1.8.1").assets | map({name,download_count}) | sort_by(-.download_count)[]'
{"name":"jq-linux64","download_count":1557635}
{"name":"jq-linux-amd64","download_count":1175823}
{"name":"jq-windows-amd64.exe","download_count":264340}
{"name":"jq-osx-amd64","download_count":250878}
{"name":"jq-macos-amd64","download_count":218746}
{"name":"sha256sum.txt","download_count":208786}
{"name":"jq-macos-arm64","download_count":150022}
{"name":"jq-1.8.1.tar.gz","download_count":97404}
{"name":"jq-linux-arm64","download_count":88325}
{"name":"jq-win64.exe","download_count":73076}
{"name":"jq-windows-i386.exe","download_count":8321}
{"name":"jq-1.8.1.zip","download_count":3963}
{"name":"jq-linux-i386","download_count":2137}
{"name":"jq-linux-ppc64el","download_count":1778}
{"name":"jq-linux-s390x","download_count":1451}
{"name":"jq-linux-armhf","download_count":294}
{"name":"jq-linux-armel","download_count":194}
{"name":"jq-linux-mips64","download_count":156}
{"name":"jq-linux-mips","download_count":146}
{"name":"jq-linux-mips64el","download_count":140}
{"name":"jq-linux-riscv64","download_count":128}
{"name":"jq-linux-mips64r6el","download_count":110}
{"name":"jq-linux-mipsel","download_count":109}
{"name":"jq-linux-powerpc","download_count":109}
{"name":"jq-linux-mipsr6","download_count":105}
{"name":"jq-linux-mipsr6el","download_count":105}
{"name":"jq-linux-mips64r6","download_count":100}
```
skip `armhf` and below possibly?

.github/workflows/ci.yml
```diff
@@ -15,106 +15,55 @@ jobs:
...
-          cp ./jq jq-${{ env.SUFFIX }}
-      - name: Test
-        # Only run tests for amd64 matching the CI machine arch
-        if: ${{ matrix.arch == 'amd64' }}
-        run: |
-          make check VERBOSE=yes
-          git diff --exit-code
+          docker build --platform ${{ matrix.platform }} -t jq-build-${{ env.BIN_NAME }} .
+          docker cp $(docker create jq-build-${{ env.BIN_NAME }}):/jq ${{ env.BIN_NAME }}
+          file ${{ env.BIN_NAME }}
```
wader: why this got complicated with pcre2 is that we're not vendoring it (i assume we don't want?) so to cross build like before we would have to cross build some version of pcre2 here. the new alternative is that we use docker build with emulation using the our Dockerfile which will end up using pcre2 from debian, that in turn would limit us in archs that can use this approach https://hub.docker.com/_/debian/tags?name=13-slim but maybe it's ok to limit us to whatever debian supports?

keithbowes:
> 
>     * Proper configure options (skip vendor for pcre2?)
>       

It seems ridiculous to me to bundle something as ubiquitous as PCRE. Anyone who's going to compile jq probably already has it.

wader:
> > ```
> > * Proper configure options (skip vendor for pcre2?)
> > ```
> 
> It seems ridiculous to me to bundle something as ubiquitous as PCRE. Anyone who's going to compile jq probably already has it.

Agreed. Looking at the history https://github.com/jqlang/jq/issues/164 https://github.com/jqlang/jq/pull/1325 the reason to vendor seem to have been control of version, less external deps and easier cross compile builds in CI (the thing i ran into also 😄). Sounds right @nicowilliams?

Not sure how valid control of version is for pcre2. I would guess they are serious about backwards comparability and there aren't that many API changes new a days? for cross builds i actually ran into the same thing in this PR, but if we drop support for some, very quite rare, archs in CI i think we can relay on per-packaged pcre2. Maybe we have to make some tests conditional if pcre2 versions behaves differently, like i noticed with some uninited memory and bug fixes.

nicowilliams:
> Agreed. Looking at the history https://github.com/jqlang/jq/issues/164 https://github.com/jqlang/jq/pull/1325 the reason to vendor seem to have been control of version, less external deps and easier cross compile builds in CI (the thing i ran into also 😄). Sounds right @nicowilliams?

Yes, that and to make a .a, but we can totally have a submodule that we use IFF we don't have an installed, acceptable version.

wader:
> Agreed. Looking at the history #164 #1325 the reason to vendor seem to have been control of version, less external deps and easier cross compile builds in CI (the thing i ran into also 😄). Sounds right @nicowilliams?
> Yes, that and to make a .a, but we can totally have a submodule that we use IFF we don't have an installed, acceptable version.

Works for me too, no strong opinion about it. So pcre2 and oniguruma configure would work the same? default "yes" (find installed library?), "builtin" or "no"? both how would the options interact? only ever use oniguruma if explicitly configure "yes" or "builtin", maybe error if both are configured?

nicowilliams:
@wader maybe just remove Oniguruma and switch to PCRE2?  But if you want to have a choice for a while, that's fine too.

wader:
@nicowilliams remove is fine for me 👍 does the changes look ok otherwise? i will try to get some time to fixup the last things

.github/workflows/ci.yml
nicowilliams: Oh yeah, cross-building with dependencies means building them, which is annoying.  I'm doing something similar in other codebases where I support MinGW for cross-building.

nicowilliams (approved):

.github/workflows/ci.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v6
         with:
-          submodules: true
+          submodules: recursive
```
wader, line 80: Need as pcre2 has a sumodule, sljit, not sure used but at least `make dist` will need it as pcre2 dist depends on the sljit files

README.md
```diff
@@ -43,10 +43,10 @@ docker run --rm -i -v "$PWD:$PWD" -w "$PWD" ghcr.io/jqlang/jq:latest '.version'
 #### Instructions
 
 ```console
-git submodule update --init    # if building from git to get oniguruma
-autoreconf -i                  # if building from git
-./configure --with-oniguruma=builtin
-make clean                     # if upgrading from a version previously built from source
+git submodule update --init --recursive # if building from git to get pcre2
+autoreconf -i                           # if building from git
+./configure                             # build with builtin PCRE2
```
wader: `--with-pcre2=builtin` is default so remove it

wader:
Now switch PR to replace oniguruma instead. See TODO things left and of course CI for macos (submodule gets dirty) and windows (i think line ending or msys2 does not play will with pcre2 tests) has issues

wader:
for some reason on macos autoreconf decices to regenerate some files making the repo dirty failing the CI workflow diff test
```
...
2026-01-21T21:18:10.7757010Z autoreconf: configure.ac: adding subdirectory vendor/pcre2 to autoreconf
2026-01-21T21:18:10.7783150Z autoreconf: Entering directory 'vendor/pcre2'
2026-01-21T21:18:10.7791930Z autoreconf: configure.ac: not using Gettext
2026-01-21T21:18:11.7871500Z autoreconf: running: aclocal -I m4
2026-01-21T21:18:12.9170970Z autoreconf: configure.ac: tracing
2026-01-21T21:18:13.3068070Z autoreconf: running: glibtoolize --copy
2026-01-21T21:18:14.8455980Z glibtoolize: putting auxiliary files in '.'.
2026-01-21T21:18:14.8456290Z glibtoolize: copying file './ltmain.sh'
2026-01-21T21:18:14.9095160Z glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
2026-01-21T21:18:14.9248080Z glibtoolize: copying file 'm4/libtool.m4'
2026-01-21T21:18:14.9923450Z glibtoolize: copying file 'm4/ltoptions.m4'
2026-01-21T21:18:15.1428890Z glibtoolize: copying file 'm4/ltversion.m4'
2026-01-21T21:18:15.3492360Z autoreconf: configure.ac: not using Intltool
2026-01-21T21:18:15.3493350Z autoreconf: configure.ac: not using Gtkdoc
2026-01-21T21:18:15.3494090Z autoreconf: running: aclocal -I m4
2026-01-21T21:18:16.2805980Z autoreconf: running: /opt/homebrew/Cellar/autoconf/2.72/bin/autoconf
2026-01-21T21:18:16.7517550Z autoreconf: running: /opt/homebrew/Cellar/autoconf/2.72/bin/autoheader
2026-01-21T21:18:17.1253510Z autoreconf: running: automake --add-missing --copy --no-force
2026-01-21T21:18:17.5641110Z autoreconf: Leaving directory 'vendor/pcre2'
...
```

hmm seems to be because pcre2 adds auto* generated files to tags https://github.com/PCRE2Project/pcre2/blob/release/pcre2-10.47/Makefile.in

wader:
Played around with using pcre2 JIT compiler but didn't notice much difference for common jq use cases and not much for big text input either, maybe i was testing wrongly?

Anyways code is here https://github.com/wader/jq/tree/pcre2-jit and remember to build jq with `--enable-jit` to pass `--enable-jit` to pcre2 configure.

src/main.c
```diff
@@ -302,13 +298,6 @@ int main(int argc, char* argv[]) {
   (void) setlocale(LC_ALL, "");
 #endif
 
-#ifdef HAVE_LIBONIG
-  // use a lower regex parse depth limit than the default (4096) to protect
-  // from stack-overflows
-  // https://github.com/jqlang/jq/security/advisories/GHSA-f946-j5j2-4w5m
-  onig_set_parse_depth_limit(1024);
```
wader, line 309: Just noticed that pcre2 has `pcre2_set_parens_nest_limit` but unsure about pcre2's stack usage during parsing. Also default limit is 250 so maybe not a problem https://github.com/PCRE2Project/pcre2/blob/ac0eb7122a0ac04d6717585f132d82aec3adc8d3/CMakeLists.txt#L344

wader:
Rebased on master and updated to vendor pcre2 10.48

Issue #3447, opened 2025-12-12
optional testcase failure in AIX.


shubhamhii:
In 32 bit mode of AIX optionaltest is failing. Below is the log.

FAIL: tests/optionaltest
========================

Test #1: 'fromdate' at line number 4
*** Insufficient results for test at line number 6: fromdate
Test #2: 'strftime("%A, %B %e, %Y")' at line number 9
1 of 2 tests passed (0 malformed, 0 skipped)
FAIL tests/optionaltest (exit status: 1)

wader:
So if you do
```sh
$ jq -n '"2038-01-19T03:14:08Z" | fromdata'
```
it outputs nothing or an error? nothing seems unlikely as it seems they end up in a builtin that i think can only output error or one value.

Can you try?
```sh
$ jq -n 'try ("2038-01-19T03:14:08Z" | strptime("%Y-%m-%dT%H:%M:%SZ")) catch .'
```
and
```sh
$ jq -nc 'try ([2038,0,19,3,14,8,2,18] | mktime) catch .'
```
and report the results?

shubhamhii:
Hi,

Thanks for you reply. 
Here is the ouput.

jq -nc 'try ([2038,0,19,3,14,8,2,18] | mktime) catch .'
"invalid gmtime representation"

jq -n 'try ("2038-01-19T03:14:08Z" | strptime("%Y-%m-%dT%H:%M:%SZ")) catch .'
[
  2038,
  0,
  19,
  3,
  14,
  8,
  2,
  18
]

wader:
Thanks, do you know if any `mktime` call work? ex:
```sh
$ jq -n 'try (now | debug | gmtime | debug | mktime) catch .'
["DEBUG:",1765581423.203734]
["DEBUG:",[2025,11,12,23,17,3.2037339210510254,5,345]]
1765581423
```
Not sure if any maintainer have access to AIX so it's a bit hard to debug.

itchyny:
Not fixed by #3407?

wader:
@itchyny oh yes that sounds very likely, had forgot about that one

@shubhamhii possible you can try master?

PR #3451, opened 2025-12-15
Fix for AIX optional testcase failure

shubhamhii:
These changes are for fixing #3447 .

Due to 32bit unix time overflow, its going beyond time_t boundary resulting in **invalid gmtime representation** in 32bit mode for AIX.

```
echo '"2038-01-19T03:14:08Z"' | ./jq 'fromdate'
jq: error (at <stdin>:1): invalid gmtime representation

echo '"2038-01-19T03:14:07Z"' | ./jq 'fromdate'
2147483647
```

### DIFF
--- a/tests/optional.test
+++ b/tests/optional.test
@@ -2,8 +2,8 @@
 
 # Regression test for #3276 (fails on mingw/WIN32)
 fromdate
-"2038-01-19T03:14:08Z"
-2147483648
+"2038-01-19T03:14:07Z"
+2147483647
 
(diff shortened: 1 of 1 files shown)


wader:
So the change in https://github.com/jqlang/jq/pull/3407 did not work? i think the point was to make time 64 bit even for 32 bit systems and the test is to make sure that is that case. Is that now possible for AIX?

itchyny:
This is a test case to make sure jq does not have Y2038 issue, so it's not good to change like this.

PR #3459, opened 2026-01-07
chore: update documentation for --binary option

fardarter:
Clarify usage of --binary option for non-Windows users.

This clarification makes it clear that scripters can include -b flag when scripting for cross platform users.

### DIFF
--- a/docs/content/manual/v1.8/manual.yml
+++ b/docs/content/manual/v1.8/manual.yml
@@ -292,7 +292,8 @@ sections:
 
         Windows users using WSL, MSYS2, or Cygwin, should use this option
         when using a native jq.exe, otherwise jq will turn newlines (LFs)
-        into carriage-return-then-newline (CRLF).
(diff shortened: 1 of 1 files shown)


wader:
Maybe phrase it as being a valid option but having no effect instead of being safe? also please update the text in all documention versions

fardarter:
@wader I'm assuming the behaviour is as I've docced based on this: https://github.com/jqlang/jq/pull/3131 / https://github.com/jqlang/jq/commit/5208a4425f24f5c366dfed2eae85f1935199d683

Is it available before 1.8.1?

wader:
Ah sorry now i see, this is about long-option. Since 1.8 if i read things correctly.

thaliaarchi:
I'd write it as "This has no effect on non-Windows platforms."

wader:
Agree with @thaliaarchi and also add it to the dev manual

PR #3460, opened 2026-01-09
Limit number of function parameters and definitions to prevent SEGV

### LINKED ISSUE #3458: NULL pointer dereference in dump_operation
## Summary
SEGV at `bytecode.c:109` when accessing `subfunctions[idx]->debuginfo` without NULL check.

## Root Cause
```c
if (bc->subfunctions[idx]->debuginfo) {
```
`subfunctions[idx]` can be NULL for certain malformed inputs.

## PoC
```python
#!/usr/bin/env python3
params = ";".join(f"$x{i}" for i in range(5000))
args = ";".join("1" for _ in range(5000))
open("poc.jq", "w").write(f"def f({params}):1;\nf({args})\n")
print("Generated poc.jq, run: jq --debug-dump-disasm -n -f poc.jq")
```
- Fuzzer: OSS-Fuzz jq fuzzer
- Sanitizer: AddressSanitizer
- Run: `jq --debug-dump-disasm -n -f poc.jq`

```
==1==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
==1==The signal is caused by a READ memory access.
==1==Hint: address points to the zero page.
    #0 0x561a2504a3b8 in dump_operation /src/jq/src/bytecode.c:109:40
    #1 0x561a25049bda in dump_code /src/jq/src/bytecode.c:52:5
    #2 0x561a25049bda in dump_disassembly /src/jq/src/bytecode.c:77:3
```

## Suggested Fix

```diff
--- a/src/compile.c
+++ b/src/compile.c
@@ -1241,6 +1241,12 @@ static int compile(...)
     if (curr->op == CLOSURE_CREATE) {
       assert(curr->bound_by == curr);
+      // Check subfunction count doesn't overflow into ARG_NEWCLOSURE flag
+      if (bc->nsubfunctions >= ARG_NEWCLOSURE) {
+        locfile_locate(lf, curr->source,
+            "too many subfunctions/closures (max %d)", ARG_NEWCLOSURE - 1);
+        errors++;
+      }
       curr->imm.intval = bc->nsubfunctions++;
     }

@@ -1278,6 +1284,12 @@ static int compile(...)
+        // Check closure count doesn't overflow into ARG_NEWCLOSURE flag
+        if (subfn->nclosures >= ARG_NEWCLOSURE) {
+          locfile_locate(lf, curr->source,
+              "function has too many parameters (max %d)", ARG_NEWCLOSURE - 1);
+          errors++;
+        }
```

OwenSanzas:
…erflow

When closure/subfunction indices overflow into the ARG_NEWCLOSURE flag bit (0x1000 = 4096), it causes:
- NULL pointer dereference in dump_operation() (bytecode.c:109)
- Assertion failure in frame_push() (execute.c:139)

This happens in two scenarios:
1. A single function with >= 4096 parameters
2. Multiple functions with cumulative >= 4096 subfunctions (including lambdas created for each call argument)

This patch adds compile-time checks to reject programs that would overflow:
- Check nclosures per function (parameters)
- Check total nsubfunctions (all defs and lambdas)

Fixes: https://github.com/jqlang/jq/issues/3458

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -1244,7 +1244,7 @@ static int compile(struct bytecode* bc, block b, struct locfile* lf, jv args, jv
       // Prevent subfunction index from overflowing into ARG_NEWCLOSURE flag
       if (bc->nsubfunctions == ARG_NEWCLOSURE) {
         locfile_locate(lf, curr->source,
-            "too many subfunctions/closures (max %d)", ARG_NEWCLOSURE - 1);
+            "too many function parameters or local function definitions (max %d)", ARG_NEWCLOSURE - 1);
         errors++;
       }
       curr->imm.intval = bc->nsubfunctions++;
--- a/src/compile.c
+++ b/src/compile.c
@@ -1280,15 +1280,16 @@ static int compile(struct bytecode* bc, block b, struct locfile* lf, jv args, jv
         for (inst* param = curr->arglist.first; param; param = param->next) {
           assert(param->op == CLOSURE_PARAM);
           assert(param->bound_by == param);
+          // Prevent closure index from overflowing into ARG_NEWCLOSURE flag
+          if (subfn->nclosures == ARG_NEWCLOSURE) {
+            locfile_locate(lf, curr->source,
+                "function has too many parameters (max %d)", ARG_NEWCLOSURE - 1);
+            errors++;
+            break;
+          }
           param->imm.intval = subfn->nclosures++;
           param->compiled = subfn;
           params = jv_array_append(params, jv_string(param->symbol));
-        }
-        // Prevent closure index from overflowing into ARG_NEWCLOSURE flag
-        if (subfn->nclosures == ARG_NEWCLOSURE) {
-          locfile_locate(lf, curr->source,
-              "function has too many parameters (max %d)", ARG_NEWCLOSURE - 1);
-          errors++;
         }
         subfn->debuginfo = jv_object_set(subfn->debuginfo, jv_string("params"), params);
         errors += compile(subfn, curr->subfn, lf, args, env);
--- a/src/compile.c
+++ b/src/compile.c
@@ -1242,10 +1242,11 @@ static int compile(struct bytecode* bc, block b, struct locfile* lf, jv args, jv
     if (curr->op == CLOSURE_CREATE) {
       assert(curr->bound_by == curr);
       // Prevent subfunction index from overflowing into ARG_NEWCLOSURE flag
-      if (bc->nsubfunctions == ARG_NEWCLOSURE) {
+      if (bc->nsubfunctions >= ARG_NEWCLOSURE) {
         locfile_locate(lf, curr->source,
             "too many function parameters or local function definitions (max %d)", ARG_NEWCLOSURE - 1);
         errors++;
+        break;
       }
       curr->imm.intval = bc->nsubfunctions++;
     }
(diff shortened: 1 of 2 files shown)


itchyny:
It would be better to report only once for each function call.

OwenSanzas:
Thanks for the advice. The change is done. Pleas LMK if you have any concern.

Updated to use == instead of >= so the error is reported only once when first hitting the limit.

src/compile.c
```diff
@@ -1278,6 +1284,12 @@ static int compile(struct bytecode* bc, block b, struct locfile* lf, jv args, jv
           param->compiled = subfn;
           params = jv_array_append(params, jv_string(param->symbol));
         }
+        // Prevent closure index from overflowing into ARG_NEWCLOSURE flag
```
itchyny: Shouldn't this check inside the for loop above where nclosures is incremented?

OwenSanzas: Yes you're correct! Pls check the latest commit again. Thanks.

wader:
Maybe nice with two tests that tests the two cases: 

```jq
def f(a0; a1, ..., a4096): empty; f(0; 1; ..., 4096);
```
etc and:
```jq
def f0: empty; ... def f4096: empty;
```

probably sanest to do as a shell test (like `tests/shtest`) and generate the test expressions, like:

```sh
$ ./jq -rn '"def f(\([range(5000) | "f\(.)"] | join(";"))): empty; f(\([range(5000)] | join(";")))"' | ./jq -f /dev/stdin
jq: error: too many subfunctions/closures (max 4095)
jq: 1 compile error
```

```sh
# weirdly i thought this would trigger the def limit case? 
$ ./jq -rn '"\([range(5000) | "def f\(.): 123"] | join(";")); f4999"' | ./jq -nf /dev/stdin
123
```

maybe i misunderstand the the non-arg-closure case?

itchyny:
I'm afraid subfunctions and closures in the error message are not correct terms for jq users.

itchyny:
Hmm, error message here is difficult. The `nsubfunctions` is incremented by the functions used or internal `@lambda` used by operators.

OwenSanzas:
For the terminology: what wording would you prefer? Maybe:                                                                                                                                                                                     
  - "too many function parameters"                                                                                                                                                                                                               
  - "too many nested function definitions"   
  
  Do you have a suggestion?

OwenSanzas:
@wader 

 I investigated this and found that jq only compiles                                                                                                                                                                                     
  functions that are actually called - unused definitions don't                                                                                                                                                                                  
  trigger CLOSURE_CREATE at all.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                 
  Your test `def f0: 123; ... def f4999: 123; f4999` only compiles                                                                                                                                                                               
  f4999 (nsubfunctions=1), so the limit check isn't triggered.                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
  When ALL functions are called, the check works correctly.

wader:
> "too many function parameters"

Ok for me, also correct if my understanding of the difference between argument vs parameter is correct

> "too many nested function definitions"

I wonder if "nested" is correct? maybe just skip it?

> I investigated this and found that jq only compiles functions that are actually called - unused definitions don't trigger CLOSURE_CREATE at all.

Ah good point. So maybe generate calls also in the test then?

OwenSanzas:
I modified the msg and added some tests :).

itchyny:
Sorry for late reply. To report only once, I think using `break` and `>=` looks safe for me. For other points, the patch looks fine.

OwenSanzas:
Great! Pls take a look: @itchyny 

Thanks for the guidance! You're the hero.

wader:
> I modified the msg and added some tests :).

Great, thanks. I think the PR looks fine now.

wader (approved):

wader:
@itchyny merge?

itchyny:
Updated the title. Fine?

wader:
Yeap good 👍

itchyny (approved):

wader:
@OwenSanzas thanks for the report and fix!

Issue #2966, opened 2023-12-07
[Feature]  Add ~/.jq.d to default module search path


systemmonkey42:
I have a number of "safe" macro's I load in `~/.jq` because it will autoload and is super convenient for interactive use.
Unfortunately this makes it impossible to have a fixed location in `$HOME` for dynamically loaded macros.

If I use the `-L` option, the standard search path is cleared, so its not as helpful as it sounds.

Can the directory `~/.jq.d` be added to the start of the default module search path, so that fixed location is available for
dynamic loading which does not require root privileges or conflict with autoloaded macros?

Also, but less important, is it possible to add `$XDG_CONFIG_HOME` support?  Something like adding `~/.config/jq/jq` (yeah, its a mess, configs in `$XDG_CONFIG_HOME` tend to drop the leading `.`) as an autoading module, and adding `~/.config/jq` or `~/.configs/jq/lib` to the module search path?  I try to move configs to `~/.config` whenever possible to reduce clutter in `$HOME`.

Thanks.

emanuele6:
`~/.jq` can be a directory with `.jq` files in it instead of a file.

systemmonkey42:
> `~/.jq` can be a directory with `.jq` files in it instead of a file.

My apologies for not being explicit with my meaning.

`~/.jq` is BOTH an "on-startup-macro-file" and an "on-demand-module-directory".  Having a single name for two entirely different use cases is a problem.

I would like a number of macros autoloaded on startup (hence `~/.jq` the file), and a number of modules available "on demand" when they are needed (hence `~/.jq` the directory).

Unless I'm mistaken, I cannot create a file and a directory of the same name, so I'm suggesting adding `~/.jq.d` to the module search path so both features can be made available. (~/.jq as a directory should continue working if its used)

Sorry for not simply saying that in the beginning.

wader:
> `~/.jq` can be a directory with `.jq` files in it instead of a file.

[...]

PR #3464, opened 2026-01-18
Support $XDG_CONFIG_HOME in module search path

nsfisis:
# Overview

Add support for the `$XDG_CONFIG_HOME` environment variable in module search path, following the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/).

Related issue: #2966 

# Changes

Expand `$JQ_CONFIG_HOME` in module search path like `$ORIGIN`.

The actual value is resolved as follows:

1. `$XDG_CONFIG_HOME/jq` if the environment variable `$XDG_CONFIG_HOME` is set and the directory exists
2. Non-Windows platforms: `$HOME/.config/jq` if the directory exists
3. `$HOME/.jq`

In addition, the default module search path is changed:

- `~/.jq` → `$JQ_CONFIG_HOME` (changed)
- `$ORIGIN/../lib/jq`
- `$ORIGIN/../lib`

## Original Changes

> [!note]
> This section describes the original changes before code review.

Expand `$XDG_CONFIG_HOME/` in module search path to the value of the environment variable `$XDG_CONFIG_HOME` like `$ORIGIN`.

If `$XDG_CONFIG_HOME` is unset or empty,

- Non-Windows: falls back to `$HOME/.config`, which is suggested by XDG Base Directory Specification
- Windows: removes the path from search path list

In addition, add `$XDG_CONFIG_HOME/jq` to the default module search path. Search order is as follows:

- `~/.jq`
- `$XDG_CONFIG_HOME/jq` (added)
- `$ORIGIN/../lib/jq`
- `$ORIGIN/../lib`

`~/.jq` takes priority over `$XDG_CONFIG_HOME/jq` to keep compatibility.

# References

- [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -220,6 +220,8 @@ EXTRA_DIST = $(DOC_FILES) $(man_MANS) $(TESTS) $(TEST_LOG_COMPILER)     \
         tests/modules/a.jq tests/modules/b/b.jq tests/modules/c/c.jq    \
         tests/modules/c/d.jq tests/modules/data.json                    \
         tests/modules/home1/.jq tests/modules/home2/.jq/g.jq            \
+        tests/modules/home3/.config/jq/cfg.jq                           \
+        tests/modules/home3/.jq/priority_test.jq                        \
         tests/modules/lib/jq/e/e.jq tests/modules/lib/jq/f.jq           \
         tests/modules/shadow1.jq tests/modules/shadow2.jq               \
         tests/modules/syntaxerror/syntaxerror.jq                        \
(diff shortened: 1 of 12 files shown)


wader:
I wonder if it would make more sense to skip search path changes and introduce a `get_config_home` function etc that do something like this:

- if `$XDG_CONFIG_HOME` is set and is not empty use `$XDG_CONFIG_HOME/jq`
- if `$HOME/.config/jq` exist and is a directory use it
- use `$HOME`

and use it in `load_program`?

(note that i skipped thinking about windows here)

nsfisis:
Thank you for review!

I introduced `$JQ_CONFIG_HOME`, which is expanded to the resolved config home.

The value is determined as follows:

1. `$XDG_CONFIG_HOME/jq` if the environment variable is set and the directory exists
1. Non-Windows: `$HOME/.config/jq` if the directory exists
1. `$HOME/.jq` (original jq config home)

If a user has `$XDG_CONFIG_HOME/jq` or `~/.config/jq`, the scripts under the directories will be loaded. Otherwise, the `~/.jq` directory is loaded as before.

wader:
Thanks. And i think it makes sense that this should not affect `include "~/..."`, that would be confusing?

What do other maintainers thinks? it seems reasonable that jq should support XDG_CONFIG_HOME somehow

PR #3465, opened 2026-01-20
Fix reversed pointer subtraction in `f_env` bounds check

itchyny:
This fixes a mistake of boundary check in env key length. Unlikely to hit a security problem IMO.


wader:
To be a problem size environ would have to be huge? looking around it seems like args+env size limit is usually around 1-2MB by default so no issue i guess? ... but i wonder if there was a reason to skip instead of throw error or abort in this case?

wader (approved):

PR #3487, opened 2026-02-26
Fix memory leaks and double frees

itchyny:


src/execute.c
```diff
@@ -1234,9 +1243,10 @@ int jq_compile_args(jq_state *jq, const char* str, jv args) {
   int nerrors = load_program(jq, locations, &program);
   if (nerrors == 0) {
     nerrors = builtins_bind(jq, &program);
-    if (nerrors == 0) {
+    if (nerrors == 0)
       nerrors = block_compile(program, &jq->bc, locations, args2obj(args));
```
wader, line 1247: `args2obj` does not seem to consume args in the case it's a object

src/jv.c
```diff
@@ -1746,6 +1746,7 @@ static int jvp_object_write(jv* object, jv key, jv **valpp) {
     slot->value = jv_invalid();
   } else {
     if (!jvp_object_rehash(object)) {
+      jvp_string_free(key);
```
wader, line 1749: just curious, any idea why not `jv_free` is used in this function?

src/jv_aux.c
```diff
@@ -91,19 +91,15 @@ jv jv_get(jv t, jv k) {
...
       double didx = jv_number_value(k);
-      if (jvp_number_is_nan(k)) {
+      if (didx < INT_MIN) didx = INT_MIN;
+      if (didx > INT_MAX) didx = INT_MAX;
+      int idx = (int)didx;
+      if (idx < 0)
+        idx += jv_array_length(jv_copy(t));
+      v = jv_array_get(t, idx);
+      if (!jv_is_valid(v)) {
+        jv_free(v);
```
wader, line 101: was there a leak fix here or just that the second `if (jvp_number_is_nan(k)) {` will never happen?

src/jv.c
itchyny: Not sure, but object related functions seem to use this because the type is obvious?

src/jv_aux.c
itchyny: In the previous code, `t` is leaked in the branch. But `jvp_number_is_nan(k)` is checked in L89.

wader:
Looks good to me. Just have to decide on how args2obj should behave

src/execute.c
itchyny: Because it returns the value, it looks good.

wader: Ah now i see what you mean, yeap make sense, it is "consumed" by returning it

wader (approved):

itchyny:
Sorry, I found another leaks in time functions.

itchyny:
Ah, tm2jv doesn't create literal number, so false positive.

wader (approved):

Issue #704, opened 2015-02-18
In-place edit (`-i`) issues (was: assertion failure in main.c line 518 & coredump)


alanbur:
This behaviour seems a little severe:
````
$ jq --version
jq-1.5rc1-31-g3b8d081
$ cat a
[1,2,3]
$ cat b
{"a":4,"b":5,"c":6}
$ jq '.' -i a b
Assertion failed: first_file != 0 && !strcmp(first_file, "-"), file main.c, line 518, function main
Abort (core dumped)
````

nicowilliams:
Thanks for the report!

nicowilliams:
Fixed.  This time with a test.

alanbur:
Don't think it's quite right - although I don't get the assert, the first file ends up containing the merged content of both the first and second files, and the first file contains all the syntax-colouring escape codes even though the output is a file.

nicowilliams:
The coloring is clearly a bug (fix coming up), but `-i` is intended to put all the output in the first file.  If you don't want this, use `--argfile`.

What would the alternative be?  To in-place edit every file?  I suppose.  What do others thing?

pkoppstein:
@nicowilliams asked:
> What would the alternative be? To in-place edit every file?

Yes.

The inspiration (if not model) for jq's -i option seems to be sed's -i option, and sed's behavior is perfectly reasonable in this respect (as in others).  To avoid confusion, I think it would be reasonable for jq to take the view (so to speak) that the -i option is a switch (!) that is intended to change its behavior to be more like (maybe even much more like) sed's.   (Yes, I often use -i.bak)

nicowilliams:
In-place editing of each file is a significant rototill, particularly after the refactoring I did of main.c.  I'll have to make `process()` a utility (fine, I wanted to anyways).  For each file to be processed I'll have to keep track of whether it is to be edited in place, and... what to do about color?

pkoppstein:
@nicowilliams wrote:
> rototill .... what to do about color?

Maybe the appropriate thing to do is to take -i out of 1.5.

As for color -- as I suggested, I think you'll feel greatly liberated if you take the view that -i is a SWITCH!

alanbur:
The documentation for -i says "Edit the (first) file in place" so I guess what it's doing now is (partly) as advertised, what I didn't expect to happen is for all the following files to be inserted into the first one as that doesn't really seem to be an in-place edit. I think there are perhaps 3 choices:
[...]

PR #3488, opened 2026-02-27
jq --in-place semantics

jubr:
Add `--in-place` editing mode to `jq` with `sed -i` semantics.

This PR implements the `--in-place` option, allowing `jq` to modify a single input file directly, similar to `sed -i`. It ensures atomic updates by writing to a temporary file and then renaming it, preserving original file permissions and handling errors gracefully.

# Historical context / references

Original feature request: https://github.com/jqlang/jq/issues/105
Prior attempt: https://github.com/jqlang/jq/pull/2616
Original -i implementation in jq history by @nicowilliams (commit 01fc816, later removed in b82c231)
Follow-up unhappy-path discussion: https://github.com/jqlang/jq/issues/704

# Documentation (from new manual.yml)

> 
> ## `--in-place` / `-i`
>  
> This adds direct-file in-place editing semantics similar to `sed -i`:
>  
> - the input file is also the output target;
> - jq writes to a temporary file in the same directory, then renames into place;
> - exactly one input file argument is required.
>  
> ### Example usage patterns
>  
> #### 1) Update a package version from a shell variable
> ```bash
>  jq -i ".version = ${VER}" package.json
> ```
>  
> #### 2) Compact JSON rewrite in place
> ```bash
> jq -c --in-place '.count += 1' state.json
> ```
>  
> #### 3) Normalize key order in place
> ```bash
> jq -S --in-place '.' config.json
> ```
>  
> #### 4) Slurp + transform + write back to same file
> ```bash
> jq -s --in-place 'map(.enabled = true)' records.json
> ```
>  
> #### 5) Raw-text output in place (intentional non-JSON file replacement)
> ```bash
> jq -r --in-place '.items[]' items.json
> ```
>  
> ### Invalid combinations / unhappy paths (now explicitly handled)
>  
> #### Missing file argument
> ```bash
> jq --in-place '.+1'
> # exits non-zero (usage error)
> ```
>  
> #### stdin target is rejected
> ```bash
> printf '1\n' | jq --in-place '.+1' -
> # exits non-zero (usage error)
> ```
>  
> #### Multiple file arguments are rejected
> ```bash
> jq --in-place '.+1' a.json b.json
> # exits non-zero (usage error)
>```
>  
> #### `--null-input` cannot be combined with `--in-place`
> ```bash
> jq -n --in-place '.+1' a.json
> # exits non-zero (usage error)
> ```
>  
> ### Failure safety guarantees
>  
> If jq fails, the original input file is preserved:
>  
> - compile error (e.g. invalid filter) leaves original file unchanged;
> - runtime error (e.g. `error("boom")`) leaves original file unchanged;
> - temp-file creation failure (e.g. non-writable directory) leaves original file unchanged.
>  
> Temporary output is cleaned up on error.

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -152,6 +152,40 @@ sections:
 
         Like `-r` but jq won't print a newline after each output.
 
+      * `--in-place` / `-i`:
+
+        Edit one file in place by writing output to a temporary file in
+        the same directory, then renaming it over the original.
+
+        This follows `sed -i`-style direct-file usage:
+
+        * the input file is also the output target;
+        * exactly one input file argument is required;
+        * stdin (`-`) is rejected;
+        * using multiple input files is rejected;
+        * combining with `--null-input` / `-n` is rejected.
+
+        On jq compile/runtime/system errors, jq leaves the original file
+        unchanged and removes temporary output.
+
+        Examples:
+
+        * Increment a field and keep output compact:
+
+          `jq -c --in-place '.count += 1' state.json`
+
+        * Normalize key order in-place:
+
+          `jq -S --in-place '.' config.json`
+
+        * Slurp multiple JSON texts from one file, transform once, write back:
+
+          `jq -s --in-place 'map(.enabled = true)' records.json`
+
+        * Produce non-JSON text in-place (valid, but replaces file with raw text):
+
+          `jq -r --in-place '.items[]' items.json`
+
       * `--ascii-output` / `-a`:
 
         jq usually outputs non-ASCII Unicode codepoints as UTF-8, even
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -168,7 +168,11 @@ sections:
         On jq compile/runtime/system errors, jq leaves the original file
         unchanged and removes temporary output.
 
-        Examples:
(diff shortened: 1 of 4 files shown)


jubr:
Going over https://github.com/jqlang/jq/pull/2616#issuecomment-1597527457 I noticed:
> An initial implementation could just refuse to edit-in-place any files not owned by geteuid(), and we can defer writing code to attempt to copy file owner/group/mode/ACL until later. On Windows it would just always go ahead.

I did a quick check:
- **POSIX**
  The code does not check ownership (`st_uid` vs `geteuid()`) before in-place edit.
  In `src/main.c` around the `IN_PLACE` block (~668+), it validates arg shape and sets temp-file mode with `stat/fchmod` (~694-704), but there is no `geteuid()` ownership refusal path.

- **Windows**
  It effectively always goes ahead (no ownership gate), matching the “On Windows it would just always go ahead” part.
  
We also might be able to go a different route: perhaps a cleaner solution would be to at the end fully `fwrite` the tmpfile content into the original sourcefile and truncate, thereby preserving the original owner/mode/acl. If this turns out to be too slow for large files we could even introduce a size threshold above which we fall back to `rename_file()` semantics.

@nicowilliams, @drm, @emanuele6 what are your thoughts on this?

jubr:
Hi @itchyny, @wader any comments from you perhaps?

PR #3489, opened 2026-02-28
Refactor tm2jv to handle fractional seconds removing duplicated code

itchyny:
This refactors `tm2jv` to remove duplicated code. Also avoids confusion of memory leak of `jv_array_get(jv_copy(a), 5)` (this is not an actual leak, but confusing).


wader (approved):
Nice!

PR #3490, opened 2026-02-28
Fix infinite loop and undefined behavior in `del(.[nan])`

itchyny:
This PR fixes the issue that `[1] | del(.[nan])` runs into an infinite loop.


src/jv_aux.c
```diff
@@ -443,13 +445,14 @@ jv jv_getpath(jv root, jv path) {
...
     assert(jv_array_length(jv_array_get(jv_copy(paths), i)) > start);
     int delkey = jv_array_length(jv_array_get(jv_copy(paths), i)) == start + 1;
     jv key = jv_array_get(jv_array_get(jv_copy(paths), i), start);
-    while (j < jv_array_length(jv_copy(paths)) &&
-           jv_equal(jv_copy(key), jv_array_get(jv_array_get(jv_copy(paths), j), start)))
+    int j = i;
+    do
       j++;
+    while (j < jv_array_length(jv_copy(paths)) &&
+           jv_equal(jv_copy(key), jv_array_get(jv_array_get(jv_copy(paths), j), start)));
```
wader, line 455: This finds the shared prefix index range i - j in paths? did this fix anything or just that we don't have to compare i to itself?

itchyny: This is the fix for infinite loop. `jv_equal` returns false for NaN, so in the previous code `j` remains equal to `i` causing the infinite loop.

wader (approved):

PR #3496, opened 2026-03-03
Fix tonumber and toboolean to reject strings with embedded null bytes

itchyny:
This PR fixes some filters with embedded null bytes.

### DIFF
--- a/src/builtin.c
+++ b/src/builtin.c
@@ -446,6 +446,10 @@ static jv f_tonumber(jq_state *jq, jv input) {
   }
   if (jv_get_kind(input) == JV_KIND_STRING) {
     const char* s = jv_string_value(input);
+    int len = jv_string_length_bytes(jv_copy(input));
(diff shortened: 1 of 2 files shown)


src/builtin.c
```diff
@@ -471,10 +475,11 @@ static jv f_toboolean(jq_state *jq, jv input) {
   }
   if (jv_get_kind(input) == JV_KIND_STRING) {
     const char *s = jv_string_value(input);
-    if (strcmp(s, "true") == 0) {
+    int len = jv_string_length_bytes(jv_copy(input));
+    if (len == 4 && memcmp(s, "true", 4) == 0) {
       jv_free(input);
       return jv_true();
-    } else if (strcmp(s, "false") == 0) {
+    } else if (len == 5 && memcmp(s, "false", 5) == 0) {
```
wader: do same len check as in `f_tonumber` as keep strcmp?

wader (approved):

PR #3497, opened 2026-03-04
Fix crash when importing a module with errors twice

### LINKED ISSUE #3126: Including and importing module with importing JSON crashes jq
Suppose that we have two files:

~~~ jq
# mod.jq
import "data" as $data;
.

# main.jq
import "mod" as mod;
include "mod";
.
~~~

Running `jq -n -f main.jq` yields:

~~~
jq: error: jq: error: library should only have function definitions, not a main expression
<unknown location>
jq-linux-amd64-1.7.1: src/compile.c:383: block_bind_library: Assertion `block_has_only_binders(binder, bindflags)' failed.
Aborted (core dumped)
~~~

The same also happens when we write twice `import` or `include`.

**Environment**

- OS and Version: Ubuntu 22.04
- jq version 1.7.1

itchyny:
When a module fails to parse as a library, `load_library` cached the
invalid block. A second import/include of the same module would find
the cached invalid block and pass it to `block_bind_library`,
triggering the assertion failure. Skip caching the block when there
are parse errors to avoid this problem. Fixes #3126.

### DIFF
--- a/src/linker.c
+++ b/src/linker.c
@@ -366,11 +366,13 @@ static int load_library(jq_state *jq, jv lib_path, int is_data, int raw, int opt
       program = block_bind_self(program, OP_IS_CALL_PSEUDO);
     }
   }
-  state_idx = lib_state->ct++;
-  lib_state->names = jv_mem_realloc(lib_state->names, lib_state->ct * sizeof(const char *));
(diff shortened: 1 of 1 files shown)


wader (approved):
Add a test?

itchyny:
The fix is obvious, unlikely to regress, so I'll skip the test.

PR #3498, opened 2026-03-04
Fix use-after-free in args2obj() array argument path

sseal:
Copy `kk` and `vk` before passing to jv_object_get(), which consumes
  its arguments. Without copies, subsequent loop iterations read freed
  memory (CWE-416) and post-loop jv_free() double-frees (CWE-415).

 This bug was introduced in `b279713e` when array processing was hoisted out of
  the loop into locals without accounting for jv_object_get() semantics.

### DIFF
--- a/src/execute.c
+++ b/src/execute.c
@@ -1222,7 +1222,7 @@ args2obj(jv args)
   jv kk = jv_string("name");
   jv vk = jv_string("value");
   jv_array_foreach(args, i, v)
-    r = jv_object_set(r, jv_object_get(jv_copy(v), kk), jv_object_get(v, vk));
+    r = jv_object_set(r, jv_object_get(jv_copy(v), jv_copy(kk)), jv_object_get(v, jv_copy(vk)));
   jv_free(args);
   jv_free(kk);
   jv_free(vk);
--- a/src/jq_test.c
+++ b/src/jq_test.c
@@ -11,6 +11,9 @@
 static void jv_test(void);
 static void run_jq_tests(jv, int, FILE *, int, int);
 static void run_jq_start_state_tests(void);
+static void run_jq_compile_args_tests(void);
+static void run_jq_recompile_tests(void);
+static void run_jq_exhaust_and_reuse_tests(void);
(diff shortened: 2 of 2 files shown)


wader (approved):
If you want to add a lib test you can probably add some code to `src/jq_test.c` but we currently don't have any tests workflow that build with sanitizers. Maybe valgrind build would find it?

sseal:
> If you want to add a lib test you can probably add some code to `src/jq_test.c` 

Sure thing, I'll add some 👍 

> we currently don't have any tests workflow that build with sanitizers. Maybe valgrind build would find it?

Yeah, I'll try out the tests with `valgrind` enabled. Generally, do you prefer `valgrind` over `ASAN` for mem-safety instrumentation? I don't mind either way, just curious.

Since it's `on the way`, I'll also take a look at the `gcov` output, to ensure the above codepath is exercised, and to get a sense of current test suite code coverage

sseal:
update:

1. I added a few new tests to exercise the codepaths relevant to this PR
2. I create coverage reports for the jq_test.c suite before and after these tests were added
3. The new tests added a small amount of additional code coverage ( < 1% total line coverage gained), but crucially, I confirmed the new unit tests exercise previously uncovered `args2obj()` code path in `src/execute.c`.

(I've attached both lcov reports ICYI) 

jq-new-tests-coverage-results

[jq-coverage-before.tar.gz](https://github.com/user-attachments/files/25751828/jq-coverage-before.tar.gz)

[jq-coverage-after.tar.gz](https://github.com/user-attachments/files/25751829/jq-coverage-after.tar.gz)

sseal:
> Yeah, I'll try out the tests with valgrind enabled. 

Update: neither `valgrind` nor `ASAN` detected any memory errors for me when I ran the test suite (which is great 😄)

itchyny (approved):

wader:
> > Yeah, I'll try out the tests with valgrind enabled.
> 
> Update: neither `valgrind` nor `ASAN` detected any memory errors for me when I ran the test suite (which is great 😄)

Nice! did valgrind detect the use after free?

PR #3508, opened 2026-03-09
Fix jv_cmp to treat NaN==NaN as equal for sort and bsearch

cuiweixie:
Previously jv_cmp(nan, nan) did not return 0, causing bsearch to never find NaN and sort to behave inconsistently. Now two NaNs compare equal.

### DIFF
--- a/src/jv_aux.c
+++ b/src/jv_aux.c
@@ -595,7 +595,9 @@ int jv_cmp(jv a, jv b) {
     break;
 
   case JV_KIND_NUMBER: {
-    if (jvp_number_is_nan(a)) {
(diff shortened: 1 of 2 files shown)


itchyny:
To be honest, I'm not sure how to respond as a maintainer. There are certainly many strange things about how jq handles NaN. For example, I really don't like that `nan < nan` is true. At the very least, it should have followed IEEE 754 for comparison operators (but it might be too late). This is not unrelated to `jv_cmp`. This PR seems good because it changes the behavior of `<` to true, but `<=` remains unchanged. I understand the desire to fix `bsearch`, but I can't explain why `index(nan)` should not work but `bsearch(nan)` should.

PR #3509, opened 2026-03-14
Fix crashes and resource leaks in `jq_testsuite`

itchyny:
- Fix segfault when --skip/--take is the last argument.
- Fix FILE handle leak: opened test files were never closed.
- Support running multiple test files in one invocation.


wader (approved):
Looks good. Didn't know --skip/--take existed :)

itchyny:
Thanks.

PR #3511, opened 2026-03-19
docs: add animated terminal demo to README

dobestan:
## Summary

Adds an animated terminal demo near the top of the README. The GIF demonstrates:

- Pretty-printing JSON with `jq '.'`
- Extracting fields with `jq '.[].name'`
- Filtering with `jq '.[] | select(.age > 27)'`

## Why

The README currently has no visual demo. An animated recording lets first-time visitors immediately see what jq does and how intuitive the syntax is — which is jq's biggest selling point.

The GIF is 145KB and stored in `media/`.

Recorded with [VHS](https://github.com/charmbracelet/vhs). Demo hosted on [AgentGIF](https://agentgif.com/@agentgif/jq-pretty/).


wader:
I like the idea. Maybe include the vhs script that produced demo.gif so it can be regenerated/updated?

itchyny:
The README.md is distributed, but its attached media should be distributed, or not?

wader:
> The README.md is distributed, but its attached media should be distributed, or not?

As in `make dist`? yes that makes sense. So add to `dist_doc_DATA`? https://github.com/jqlang/jq/blob/master/Makefile.am#L241

itchyny (changes requested):
Agree with wader, VHS script should be included along with the gif image.

PR #3518, opened 2026-04-07
Support building with --disable-maintainer-mode and source != build dir

### LINKED ISSUE #3371: Make failed when building out-of-source
**Describe the bug**

`Make` failed building out-of-source, because of the disabling rebuilding `src/lexer.c` and `src/parser.c`.

**To Reproduce**

Do this:

```sh
git submodule update --init
autoreconf -i
mkdir build
cd build
../configure --with-oniguruma=builtin
make
```

And get the error message:

```
  YACC     src/lexer.c
NOT building lexer.c!
  CC       src/lexer.lo
cc: error: no such file or directory: 'src/lexer.c'
cc: error: no input file
```

This is because the implicit rules of GNU Make. The existence of `$(srcdir)/src/lexer.l` will disable the VPATH `$(srcdir)/src/lexer.c` to be the input file for `src/lexer.o`. So as `src/parser.c`.

However, if configure without maintainer-mode enabled, `Makefile.am` has a rule to disable `%.l: %.c` and `%.y: %.c` generation.

And if you are not building out-of-source, `src/lexer.c` and `src/parser.c` should be fine as input, instead of VPATH `$(srcdir)/src/lexer.c` and `$(srcdir)/src/parser.c`.

**Expected behavior**

`jq` to be built.

- OS and Version: Ubuntu 24.04.1 LTS
- jq version: jq-1.8.1 (git tag)
- GNU Make: 4.3

Saur2000:
If --disable-maintainer-mode is enabled, then the rules for generating parser.[ch] and lexer.[ch] did nothing. This worked fine if the source and build directories are the same as the pre-generated parser.c and lexer.c files would suffice. However, if the build directory is not the same as the source directory, then the rest of the Make rules expect parser.[ch] and lexer.[ch] to have been created in the build directory if their source files (parser.y and lexer.l) are newer than the target files, which can happen in case the source is fetched using Git.

Avoid the problem by copying the files to the build directory if needed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,9 +41,14 @@ src/lexer.h: src/lexer.c
 else
 BUILT_SOURCES = src/builtin.inc src/config_opts.inc src/version.h
 .y.c:
-	$(AM_V_YACC) echo "NOT building parser.c!"
+	$(AM_V_YACC) [ "$(<D)" = "$(@D)" ] || cp $(<D)/$(@F) $@
+	$(AM_V_YACC) [ "$(<D)" = "$(@D)" ] || cp $(<D)/$(*F).h $*.h
+	$(AM_V_YACC) touch $@ $*.h
+
 .l.c:
-	$(AM_V_LEX) echo "NOT building lexer.c!"
+	$(AM_V_LEX) [ "$(<D)" = "$(@D)" ] || cp $(<D)/$(@F) $@
+	$(AM_V_LEX) [ "$(<D)" = "$(@D)" ] || cp $(<D)/$(*F).h $*.h
+	$(AM_V_LEX) touch $@ $*.h
 endif
 
 # Tell YACC (Bison) autoconf macros that you want a header file created.


itchyny:
This PR introduces GNU make extensions in our Makefile.am for the first time.

Saur2000:
> This PR introduces GNU make extensions in our Makefile.am for the first time.

According to https://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html, the use of variables such as `$(<D)` and `$(@F)` are supported by POSIX Make. And even `automake` itself adds one (from the generated `Makefile`):

```
$(am__depfiles_remade):
        @$(MKDIR_P) $(@D)
        @: >>$@
```

So unless you are referring to something other than those automatic variables, I believe this should be ok.

itchyny:
Ah, sorry. I was wrong.

itchyny:
This fixes  #3371, right?

Saur2000:
> This fixes #3371, right?

Yes, it appears that it does. :)

itchyny:
Thank you.

Issue #524, opened 2014-08-01
Syntactic sugar for "regular" def arguments


nicowilliams:
Something like this:

```
# a and b are "regular"; c is "special"
def f($a; $b; c): [., $a, $b] | first(c);
```

which would be equivalent to:

```
def f(a; b; c):
    a as $a |
    b as $b |
    [.,$a,$b] | first(c);
```

PR #3520, opened 2026-04-14
Bind only `$x`, not `x` in `def f($x): ...`

01mf02:
@wader brought to my attention in https://github.com/01mf02/jaq/issues/428 that jq currently has the following behaviour:

~~~
$ jq -cn 'def f($a): [a],$a; f(1,2)'
[1,2]
1
[1,2]
2
~~~

That means that in a function `f($a)`, both `a` and `$a` are bound. This has already been pointed out by @pkoppstein in @nicowilliams's PR #524, which added "regular" definition arguments (where "regular" means "$..." here).

This PR tries to rectify this, by changing how to compile regular definition arguments. Let us use @wader's example above and see what it rewrites to:

~~~ jq
# before
def f(x0): x0 as a | a as $a | [a], $a;
# after
def f(x0): x0 as @regular | @regular as $a | [a], $a;
~~~

The only change is that it binds any regular argument as `@regular`, which is a variable name that cannot be referenced by the user.

This excludes a few programs that previously compiled (including @wader's example above):

~~~
# before
$ jq -n 'def f($a): a; f(1)'
1
# after
$ jq -n 'def f($a): a; f(1)'
jq: error: a/0 is not defined at <top-level>, line 1, column 12:
    def f($a): a; f(1)
               ^
jq: 1 compile error
~~~

Furthermore, it changes the meaning of some programs:

~~~
# before
$ jq -nc 'def foo(a;$a):[a,$a]; foo(1;2)'
[2,2]
# after
$ jq -nc 'def foo(a;$a):[a,$a]; foo(1;2)'
[1,2]
~~~

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -553,7 +553,9 @@ block gen_function(const char* name, block formals, block body) {
     i->nformals = 0;
     if (i->op == CLOSURE_PARAM_REGULAR) {
       i->op = CLOSURE_PARAM;
-      body = gen_var_binding(gen_call(i->symbol, gen_noop()), i->symbol, body);
+      char *regular = strdup("@regular");
+      body = gen_var_binding(gen_call(regular, gen_noop()), i->symbol, body);
+      i->symbol = regular;
     }
     block_bind_subblock(inst_block(i), body, OP_IS_CALL_PSEUDO | OP_HAS_BINDING, 0);
   }


01mf02:
By the way, a little note on LLM usage: I tested whether LLMs could be useful to do this fix. I gave ChatGPT the following prompts:

> Write a patch for jq that prevents binding variable arguments additionally as regular arguments. That is, today def f($x): x compiles, because it binds its argument both as x and $x. However, with the patch, this should yield an error, while def f($x): $x should continue to work.

It gave me a completely unusable patch, then told me: "If you want, I can refine this into a PR-ready patch against a specific jq version (with exact line numbers and tests).". I replied:

> Refine this into a PR-ready patch against jq master.

Behold its result, headlined boldly by "✅ Patch (ready to submit)":

~~~ diff
index 7c1e5ab..e4d9abc 100644
--- a/src/compile.c
+++ b/src/compile.c
@@ -1308,18 +1308,35 @@ static int compile_funcdef(block* b, struct locfile* lf, jv name, jv args, block
   int nformals = jv_array_length(jv_copy(args));

   for (int i = 0; i < nformals; i++) {
-    jv arg = jv_array_get(jv_copy(args), i);
-    const char* argname = jv_string_value(arg);
-
-    // Existing behavior: bind both positional and variable
-    block_bind_argument(b, argname);
-    block_bind_variable(b, argname);
+    jv arg = jv_array_get(jv_copy(args), i);
+    const char* raw = jv_string_value(arg);
+
+    if (raw[0] == '$') {
+      // Variable argument: strip leading '$'
+      const char* varname = raw + 1;
+
+      if (varname[0] == '\0') {
+        locfile_locate(lf, jv_copy(arg));
+        FAIL(*b, "invalid variable name \"$\"");
+      }
+
+      // Bind ONLY as variable
+      block_bind_variable(b, varname);
+
+      // NOTE: intentionally NOT calling block_bind_argument()
+    } else {
+      // Positional argument: preserve existing semantics
+      block_bind_argument(b, raw);
+      block_bind_variable(b, raw);
+    }
   }

   jv_free(args);

   // Compile function body
   block body_block = gen_noop();
   int r = compile(b, lf, body, &body_block);

   if (r != 0)
     return r;
~~~

Hallucination at its finest. The file exists, but neither does the function `compile_funcdef` exist, nor most of the other things in this diff. So this is complete AI slop.

... However, it brought me on track with the `nformals` reference, which is actually used at the place that I modified in the end. So while using the LLM on its own is completely insufficient, it seems that this can still be useful to discover likely places to modify.

In this case, however, it would have been fastest to just look up the commit that added the "regular argument" functionality (https://github.com/jqlang/jq/commit/32c08d48f32162db1f1b6f3daaecf8f24ed85368). Which is what I ended up doing anyway.

PR #3539, opened 2026-05-08
fix(jv): guard deep structural equality recursion

fuyu0425:
## Summary
- add a depth-limited internal equality helper for recursive structural comparisons
- route `==` and `!=` through that helper so deeply nested equality checks fail with a jq error instead of exhausting the C stack
- add a regression test for deep structural equality next to the existing depth-limit regressions

## Testing
- `make -j"$(nproc)"`
- `tests/jqtest`
- direct CLI replay of deep equality returns `"Equality check too deep"`

## Notes
- the helper stays internal in `jv_private.h`
- the depth limit matches jq's existing `10000`-level recursion guards as a stopgap until structural comparison can be made iterative

### DIFF
--- a/src/jv_aux.c
+++ b/src/jv_aux.c
@@ -16,6 +16,24 @@ static double jv_number_get_value_and_consume(jv number) {
   return value;
 }
 
+#ifndef MAX_CMP_DEPTH
+#define MAX_CMP_DEPTH (10000)
+#endif
+
+struct sort_cmp_state {
+  int too_deep;
+};
+
+#ifdef _MSC_VER
+static __declspec(thread) struct sort_cmp_state sort_cmp_state;
+#else
+#ifdef HAVE___THREAD
+static __thread struct sort_cmp_state sort_cmp_state;
+#else
+static struct sort_cmp_state sort_cmp_state;
+#endif
+#endif
+
 static jv parse_slice(jv j, jv slice, int* pstart, int* pend) {
   // Array slices
   jv start_jv = jv_object_get(jv_copy(slice), jv_string("start"));
@@ -463,7 +481,7 @@ static jv delpaths_sorted(jv object, jv paths, int start) {
     do
       j++;
     while (j < jv_array_length(jv_copy(paths)) &&
-           jv_equal(jv_copy(key), jv_array_get(jv_array_get(jv_copy(paths), j), start)));
+           jv_equal(jv_copy(key), jv_array_get(jv_array_get(jv_copy(paths), j), start)) == 1);
     // if i <= entry < j, then entry starts with key
     if (delkey) {
       // deleting this entire key, we don't care about any more specific deletions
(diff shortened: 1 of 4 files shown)


itchyny:
Does this patch work for sort/unique?

fuyu0425:
I add a fix for sort/unique as well.

itchyny (commented):
Thank you for your contribution.
I'm reading the patch and I don't like the naming a bit. The safe suffix makes it seem like there're unsafe helpers, but there aren't. I was about to suggesting renaming, but noticed the difference of jv_equal and jvp_equal_safe. It's a breaking change, but libjq may want to distinguish the too deep error from the equality. And I did so a week ago, in jv_contains. Release notes should mention but I think it's okay for jv_equal to return -1 for error in the next patch release. jq's APIs should be sophisticated that jq itself can be implemented by the APIs (including error handling). Hmm, what does jv_cmp return for error, then? INT_MIN? Then use it for all jv_equal, jv_contains for consistency. WDYT? (Another option is to change return type to jv, but that looks _too changed_ for a patch release?)
Another small mistake is that `JV_KIND_OBJECT` case in jvp_cmp silently discards the too deep error.

itchyny:
I'm still undecided, but I'm leaning towards returning jv. The APIs will change, but it will convey that error handling will be necessary due to compilation errors (instead of sentinel INT_MIN or -1). Any thoughts? @nicowilliams @wader

itchyny:
I tried modifying the code to return a jv, but it's far too much trouble. Hmm, I gave up.

fuyu0425:
I revised based on your first comment.
Not sure the `jv` one.

src/jv_aux.c
```diff
@@ -704,7 +739,17 @@ static struct sort_entry* sort_items(jv objects, jv keys) {
...
   jv_free(objects);
   jv_free(keys);
+  sort_cmp_state.too_deep = 0;
   qsort(entries, n, sizeof(struct sort_entry), sort_cmp);
+  if (sort_cmp_state.too_deep) {
+    for (int i = 0; i < n; i++) {
+      jv_free(entries[i].key);
+      jv_free(entries[i].object);
+    }
+    jv_mem_free(entries);
```
wader: This is repeated a few times, maybe a helper to free a `struct sort_entry*`?

src/jv_aux.c
```diff
@@ -671,6 +697,10 @@ int jv_cmp(jv a, jv b) {
   return r;
 }
 
+int jv_cmp(jv a, jv b) {
```
itchyny, line 702: Could you add a single line comment for the returned value like how jv_contains does?

wader: Also jv_equal? any other public function that will change behavior?

src/jv_aux.c
fuyu0425: added

src/jv_aux.c
fuyu0425: added

itchyny (approved):
Thank you!

itchyny:
I requested CVE to include in the commit message. Please wait.

carnil:
The CVE assigned appears to be CVE-2026-47770 (and recorded as well in the commit message).

Issue #1928, opened 2019-06-17
Variable assignment with arithmetic


MondayHopscotch:
**Describe the bug**
Arithmetic operations combined with variable assignment leads to unintuitive parsing.

**To Reproduce**
https://jqplay.org/s/EYE60uy1jx

Input
```
["a", "b", "c"]
```

JQ
```
length + 1 as $array_length | $array_length | debug
```

Output
```
["DEBUG:",1]
4
```

The value of `$array_length` is `1`, instead of the expected `4`

**Expected behavior**
The grouping during parsing seems unintuitive and/or incorrect here.

What the current implementation is doing:
```
length + (1 as $array_length | $array_length | debug)
```

What I would expect:
```
(length + 1) as $array_length | $array_length | debug
```

**Environment (please complete the following information):**
- JQ Play, listed as running 1.6 

**Additional context**
This is the simplest example I could create, but in more complicated examples, this creates some unintuitive results:

I ended up here while trying to figure out why something like this example wasn't working:
https://jqplay.org/s/EHrQ49_MDl

Input:
```
{
    "foo": 5
}
```

JQ:
```
def testFunc($input): 
    $input + 5 as $new | 
    {"original": $input, "new": $new}
;

testFunc(.foo)
```

Result:
```
jq: error (at <stdin>:2): number (5) and object ({"original"...) cannot be added
exit status 5
```

I understand I can get around this issue by explicitly adding some parentheses to force the grouping in the way I am wanting, or by removing the `$new` variable declaration and just in-lining it in the object constructor. These are just examples to show the behavior I was dealing with.

nicowilliams:
Yes, jq has some precedence issues.  I don't think they can be fixed now.  Perhaps we could add higher- and lower-precendence versions of various operators, not unlike Haskell, but that's about all we can do.

I.e., something like `length $+ 1 as $foo | ...` having addition bind more tightly than the `as $foo`.

Not sure that it's worth doing.

Comments?

MondayHopscotch:
I think it's something that would be useful eventually to get a more intuitive behavior. I'm not super familiar with the underlying code, so I don't know how much work it would entail. Since I am able to work around the current limitations by adding some `()`'s around pieces to guarantee the grouping, it's nothing urgent.

wtlangford:
I don't like the idea of having otherwise-identical operators with different precedence semantics. I think that's likely to be more confusing long-term. (See the difference between `&&` and `and` in Ruby for an example)

[...]

PR #3542, opened 2026-05-12
Fix: alternate operator (//) precedence (#3507)

### LINKED ISSUE #3507: Strange `x = y // z` precedence
**Describe the bug**
The alternate operator has a higher precedence than the assignment operator, which is more practical, and vice versa is not very reasonable

**To Reproduce**
```sh
$ jq -n '.[0]=null//2'
[
  null
]
$ jq -n '.[0]=(null//2)'
[
  2
]
```

**Expected behavior**
```
[
  2
]
[
  2
]
```

- OS and Version: Arch Linux
- jq version 1.8.0

CodeSasuke:
Fixes #3507

**Description:**
The alternate operator `//` was declared with lower precedence than the assignment operator `=`, causing expressions like `.[0] = null // 2` to be parsed incorrectly as `(.[0] = null) // 2`. 

This PR moves the `%right "//"` directive below the `%nonassoc '='` block in `src/parser.y` to bump the precedence of `//` above assignment.

### DIFF
--- a/src/parser.y
+++ b/src/parser.y
@@ -100,8 +100,8 @@ struct lexer_param;
 %precedence FUNCDEF
 %right '|'
 %left ','
-%right "//"
 %nonassoc '=' SETPIPE SETPLUS SETMINUS SETMULT SETDIV SETMOD SETDEFINEDOR
+%right "//"
 %left OR
 %left AND
 %nonassoc NEQ EQ '<' '>' LESSEQ GREATEREQ


maximilize:
I built this PR in a clean container (base `42d4035`, `--with-oniguruma=builtin`) and compared it against current `master` as baseline. Three things worth flagging before this lands.

### 1. The precedence change does what #3507 asks — once the parser is actually regenerated

`.[0] = null // 2` : `[null]` → `[2]`, and the natural idiom now works the intuitive way:

```
echo '{"y":null}' | jq '.x = .y // 5'
# master : {"y":null,"x":null}     (parsed (.x=.y)//5)
# this PR: {"y":null,"x":5}        (parsed .x=(.y//5))
```

`make check` stays green and Bison reports **no new conflicts** (`%expect 0` still holds).

### 2. As submitted, the PR is inert in a normal build (and in CI)

`src/parser.c` / `src/parser.h` are **committed** generated files, and `configure.ac` has `AM_MAINTAINER_MODE([disable])`. Neither a default `./configure && make` nor CI regenerates them — `.github/workflows/ci.yml` configures without `--enable-maintainer-mode`, then runs `make check` + `git diff --exit-code`. So:

- A plain build of this branch still parses with the **old** committed `parser.c` → `.[0]=null//2` still returns `[null]`. I only got the new behavior after `--enable-maintainer-mode` + removing the shipped `parser.c` to force Bison to re-run.
- CI's `make check` therefore exercises the **old** grammar — it's green but does not test this change at all.

To actually ship the fix, the regenerated `src/parser.c` (and `parser.h`) need to be committed alongside the `parser.y` edit.

### 3. This is a silent breaking change across the whole assignment/update family

The swap moves `//` above the entire `%nonassoc '=' SETPIPE SETPLUS SETMINUS SETMULT SETDIV SETMOD SETDEFINEDOR` group, not just `=`. Any existing `EXPR // path OP value` regroups from `EXPR // (path OP value)` to `(EXPR // path) OP value`, with **no error** — just different output:

```
echo '{"a":2,"b":1}' | jq '.a // .b = 5'    # master: 2   →  this PR: {"a":5,"b":1}
echo '{"a":2,"b":1}' | jq '.a // .b |= 5'    # master: 2   →  this PR: {"a":5,"b":1}
echo '{"a":2}'       | jq '.a // .b += 5'    # master: 2   →  this PR: {"a":7}
```

Assignment as the lowest-precedence operator is the more consistent grammar and the direction #1928 wanted — but this is a silent, backward-incompatible change, so it needs a NEWS entry and a maintainer decision on whether the break is acceptable. A `tests/jq.test` regression triple should land with it.

PR #3545, opened 2026-05-17
docs: fix typo this is is -> this is

### LINKED ISSUE #3544: Typo on man page *is is*
> It is also possible to define functions in jq, although this * _**is is**_ * a feature whose biggest use is defining jq's standard library (many jq functions such as map and select are in fact written in jq).

https://jqlang.org/manual/#advanced-features

PS thanks for making jq... I love it!

linbo328:
Fixed: this is is -> this is in man page

Closes #3544

### DIFF
--- a/jq.1.prebuilt
+++ b/jq.1.prebuilt
@@ -3116,7 +3116,7 @@ Variables are an absolute necessity in most programming languages, but they\'re 
 In most languages, variables are the only means of passing around data\. If you calculate a value, and you want to use it more than once, you\'ll need to store it in a variable\. To pass a value to another part of the program, you\'ll need that part of the program to define a variable (as a function parameter, object member, or whatever) in which to place the data\.
 .
 .P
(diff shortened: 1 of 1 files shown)


wader:
Hey, thanks. Could you please fix the typo in all `docs/content/manual/*/manual.yml` also?

truffle-dev:
`jq.1.prebuilt` is regenerated from `docs/content/manual/dev/manual.yml` per `Makefile.am:182`, so this patch to the cached manpage will revert on the next docs rebuild. The typo also lives in the source yamls at:

- `docs/content/manual/dev/manual.yml:2817`
- `docs/content/manual/v1.7/manual.yml:2736`
- `docs/content/manual/v1.8/manual.yml:2817`

The v1.7 and v1.8 snapshots are served at `https://jqlang.org/manual/v1.7/` and `/manual/v1.8/`, so fixing those drops the typo from the published older-version docs as well. The fix is the same one-line edit in each (drop the leading `is ` on the second line of the wrapped sentence).

itchyny:
This file should not be edited directly.

PR #3546, opened 2026-05-18
Add options `--rawfd` and `--slurpfd`

howtonotwin:
These allow jq to read data from files that might not have names, make it easier for shell scripts that use jq to avoid races, etc., without resorting to tricks like `/dev/fd` or `/proc/self/fd` that might not always be available (and don't quite behave the same). These options are otherwise very similar to `--rawfile` and `--slurpfile` and share most of their code.

I'm using this on (GNU/)Linux. I'd appreciate feedback about/testing on other platforms. I didn't find a contributors' guidelines file or a style guide, so please tell me if I should do something differently!

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -266,6 +266,20 @@ sections:
         available in the program and has a string whose content is set to the text
         in the file named `bar`.
 
+      * `--slurpfd variable-name fd`:
+
+        This option is just like `--slurpfile` except it reads from the file
+        descriptor `fd` (specified as a decimal integer). The parent process of
+        jq is expected to have opened `fd` for reading and set it up to be
+        inherited by jq.
+
+      * `--rawfd variable-name fd`:
+
+        This option is just like `--rawfile` except it reads from the file
+        descriptor `fd` (specified as a decimal integer). The parent process of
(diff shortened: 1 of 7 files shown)


howtonotwin:
These options *do* work on Windows, but Windows programs that use FDs are subject to the restriction that programs with incompatible C runtimes can't inherit FDs from each other. The `sh` that runs the test script in CI is using the Cygwin/MSYS2 runtime but is calling a `jq` that uses the MSVCRT/UCRT runtime. So the FDs set up by the new test in `sh` don't quite make it to `jq`.

If I use a native Windows (MSVCRT/UCRT) binary like `busybox-w32` to call `jq`, it's more apparent that this PR works:

```sh
busybox> ./jq -n '{$foo, $bar}' 8<tests/modules/data.json 9<>tests/modules/data.json --slurpfd foo 8 --rawfd bar 9
{
  "foo": [
    {
      "this": "is a test",
      "that": "is too"
    }
  ],
  "bar": "{\n  \"this\": \"is a test\",\n  \"that\": \"is too\"\n}\n"
}
```

So my questions are

* Should these options just take handles on Windows from the start?
* Should the test just be skipped on Windows, or should the test be moved to a compiled C program (which will work for the same reason `busybox-w32` works)?

PR #3551, opened 2026-05-29
docs: use \(aq instead of \' in code and pre blocks

aivot-on:
Both groff and mandoc render the `\'` as acute accent `´` making the man pages hard to read and copy from. Replace those with `\(aq` so they will render correctly as apostrophes.


aivot-on:
Seems like #3189 achieves the same result. Hopefully it gets merged in.

PR #3555, opened 2026-06-02
docs: clarify join behavior on objects

### LINKED ISSUE #3541: Clarification on running `join` on objects
**Describe the bug**
The [documentation for `join`](https://jqlang.org/manual/#join) says:

> Joins the array of elements given as input, using the argument as separator

However, one finds that the function works equally on objects:

```bash
$ jq -nr '{"b": 1, "a": 2} | join(",")'
1,2
```

**To Reproduce**
See example above.

**Expected behavior**
I think the documentation needs clarifications.

What exactly is the behaviour when the input is an object? It seems it's joining their values in the order they're given. Notably, the output is the same even if I use the `-S` option, hinting at the fact that the object's keys are not sorted. That's to say, the outputs of `jq -n '{"b": 1, "a": 2}'` and `jq -Sn '{"b": 1, "a": 2}'` ordinarily differ.

The documentation also says:

> Arrays and objects in the input are not supported.

Perhaps a bit pedantic, but I think it meant to say this about the elements _inside_ the input array (or object), not the input itself. Specifically that this won't work:

```
jq -nr '[[1,2], [3,4]] | join(",")'
```

- OS and Version: Linux
- jq version: 1.8.1

ded-furby:
Closes #3541

## Summary
- clarify that `join/1` accepts object input as well as arrays
- explain that object input is processed via the values produced by `.[]`
- tighten the wording so the unsupported case is arrays/objects among the joined values

## Testing
- verified the current behavior locally with `jq -nr '{"b":1,"a":2} | join(",")'` and `jq -nr '[[1,2],[3,4]] | try join(",") catch .'`\n- ran `git diff --check`

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -1865,14 +1865,16 @@ sections:
       - title: "`join(str)`"
         body: |
 
-          Joins the array of elements given as input, using the
+          Joins the values from the input array or object, using the
           argument as separator. It is the inverse of `split`: that is,
           running `split("foo") | join("foo")` over any input string
           returns said input string.
 
           Numbers and booleans in the input are converted to strings.
-          Null values are treated as empty strings. Arrays and objects
-          in the input are not supported.
+          Null values are treated as empty strings. When the input is
+          an object, `join` operates on the values that `.[]`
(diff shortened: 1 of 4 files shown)


itchyny (approved):
Thank you!

PR #3570, opened 2026-06-25
Fix compile crash in transitive include chains

tianon-sso:
Any program that uses `include` (or `import`) to load a module which itself includes or imports another module aborts at compile time with one of two assertion failures, depending on the call pattern:

    jq: src/compile.c:1086: nesting_level: Assertion `bc && target && target->compiled' failed.

    jq: src/compile.c:1158: expand_call_arglist: Assertion `0 && "Unknown function type"' failed.

The program never executes; the crash is purely a compiler artefact and occurs even if the offending call site is unreachable at runtime.

When `load_program` assembles the final block, it joins `lib_state` entries in registration order. Libraries are registered DFS pre-order, so an importer always receives a lower index than its dependencies, placing callers before callees in the joined block.

`block_mark_referenced` traverses the block last-to-first. Its `saw_top` optimisation skips any `CLOSURE_CREATE` not yet marked referenced at the time it is encountered. With callers preceding callees in the block, callees appear earlier in the backward pass -- before the callers' subfunctions have been walked to mark them -- so `block_drop_unreferenced` frees them as unreferenced. Any caller that holds a `bound_by` pointer into the freed instruction is then corrupt. Two assertion failures result:

- `expand_call_arglist`: the default assert fires when `bound_by->op` contains a reallocated opcode rather than `CLOSURE_CREATE` or `CLOSURE_PARAM`.
- `nesting_level`: the assert fires when `bound_by->compiled` is `NULL` because the freed slot was reused by a not-yet-compiled `CLOSURE_CREATE`.

The regression was introduced by commit f58787c41835d9b17795730cb04925fdba25c71c ("Detect circular module imports to prevent stack overflow"), which restructured the dependency loop in `load_program` to add the `loading` flag for circular-import detection; the restructuring left `lib_state` entries joined in registration order rather than the reverse order that `block_mark_referenced` requires.

Fix by iterating `lib_state` in reverse when building `libs`. Reverse registration order places callees before callers in the block; the backward pass encounters callers first, marks their callees referenced, and those callees are retained correctly when the pass reaches them.

Assisted-By: "claude my eyes right out"

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -234,6 +234,8 @@ EXTRA_DIST = $(DOC_FILES) $(man_MANS) $(TESTS) $(TEST_LOG_COMPILER)     \
         tests/modules/test_bind_order2.jq                               \
         tests/modules/cycle_a.jq tests/modules/cycle_b.jq               \
         tests/modules/cycle_self.jq                                     \
+        tests/modules/cross_base.jq tests/modules/cross_gen.jq          \
+        tests/modules/cross_mid.jq tests/modules/cross_user.jq          \
         tests/onig.supp tests/local.supp                                \
         tests/setup tests/torture/input0.json                           \
         tests/optional.test tests/man.test tests/manonig.test           \
--- a/src/linker.c
+++ b/src/linker.c
@@ -459,13 +459,20 @@ int load_program(jq_state *jq, struct locfile* src, block *out_block) {
   }
 
   nerrors = process_dependencies(jq, jq_get_jq_origin(jq), jq_get_prog_origin(jq), &program, &lib_state);
-  block libs = gen_noop();
-  for (uint64_t i = 0; i < lib_state.ct; ++i) {
+  for (uint64_t i = 0; i < lib_state.ct; ++i)
     free(lib_state.entries[i].name);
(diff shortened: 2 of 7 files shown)


tianon:
Oof, since this was a regression in one of the security fixes, https://tracker.debian.org/news/1783358/accepted-jq-171-6deb13u3-source-into-stable-security/ has now made it so this issue has propagated to the version of `jq` in Debian Stable too :sob:

Is there anything I can do to help get this into a state where it's reviewable? :bow:

**Update:** filed a Debian bug at https://bugs.debian.org/1144075

happyaron:
The reported problem is fixed by this PR, but "reverse pre-order" is not a valid chain. This example works on HEAD today and crashes with the PR applied:

```
a.jq: def h: 1;
      def f: h;
b.jq: include "a";
      def g: f;

$ jq -L . -n 'include "b"; include "a"; g'
  HEAD: 1        PR #3570: expand_call_arglist assertion failed
```

tianon-sso:
Closing in favor of https://github.com/jqlang/jq/pull/3603 :+1:

PR #3590, opened 2026-07-19
docs: document fromstream end-of-stream markers

### LINKED ISSUE #3479: Document `fromstream` best practices
`fromstream`'s behavior is generally confusing, see the below issues:

* #1957
* #2464
* #3037

It seems like this filter requires end-of-object markers. How well-supported is the `[[null]]` trick? It would be nice if it were documented somewhere.

Alternatively, it would be nice to have a more intuitive filter that collects a stream into a single object. This way, it doesn't require end-of-object markers, so it never produces empty results.

```jq
def collectstream(f):
    fromstream(
        (f | select(length == 2)),
        [[null]]
    );
```

Solaris-star:
## Summary
`fromstream` is easy to misuse without end-of-stream / container-closing markers. Document the required shape and the common `[[null]]` pattern.

## Changes
- `docs/content/manual/v1.8/manual.yml` and `dev/manual.yml`: expand `fromstream` description with marker semantics, `[[null]]` usage, and a `collectstream` helper sketch
- `jq.1.prebuilt`: regenerated from the v1.8 manual

## How tested
- Regenerated manpage via `docs/build_manpage.py content/manual/v1.8/manual.yml`
- Confirmed new wording appears in `jq.1.prebuilt`

Closes #3479

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -3440,8 +3440,30 @@ sections:
       - title: "`fromstream(stream_expression)`"
         body: |
 
-          Outputs values corresponding to the stream expression's
-          outputs.
+          Reconstructs values from a stream of path-value pairs such as those
+          produced by `tostream`.
+
+          Stream entries look like `[path]` or `[path, value]`. A path-only
+          entry closes the container at that path. `fromstream` needs those
+          closing markers to finish nested objects/arrays; without them the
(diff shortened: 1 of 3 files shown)


Solaris-star:
Closing this proactively to keep one contribution under review in this repository at a time. Sorry for opening a second PR before #3588 had maintainer feedback.

PR #3593, opened 2026-07-20
Split token `?//` to `?` and `//`

### LINKED ISSUE #3589: feature: Equivalent `x?//y` to `x? //y`
Because `?//` only uses after `as`, and `.name? //"unknown"` isn't convenient enough

## Possible Implements

Perhaps we can split the `?//` token into two tokens, `?` and `//`, if there is no ambiguity
So, `x as [$a] ? // {$a}` can also be pass through

Alternatively, add a priority operation `?//` and generate a `? //` tree,
operators with lower priority than `?//` should behave consistently

A4-Tacks:
Close #3589


thatssoheil:
Independent verification of this patch, in case it helps the review.

I rebuilt a tree with the split applied and exercised the parser directly, because the objection a token split usually draws is a grammar conflict:

- bison reports zero conflicts over the resulting grammar. jq's Makefile turns any conflict into a hard error (the grammar carries `%expect 0`), and the split builds clean under it - `Pattern` cannot contain `?` or `//`, so `?//` stays unambiguous once it is lexed as two tokens.
- `make check` passes on the rebuilt split, and the existing lexer/parser tests are unaffected.
- The change is confined to the lexer rule for the token plus the two parser uses of `?//`, which are only reachable after `as`; no other production is affected.

That is everything I could establish from the outside; whether this is the intended reading of `?//` is the maintainers' call, but the patch itself looked safe in the rebuild, and it matches the behaviour described in the issue.

(This analysis was produced with tooling assistance.)

PR #3599, opened 2026-08-02
docs: `in` example use [3,4] instead of [0,1]

A4-Tacks:
The example of jq is excellent, but `[0,1]` may misunderstand the role of 'with key' as the role of 'with value'

### DIFF
--- a/docs/content/manual/v1.8/manual.yml
+++ b/docs/content/manual/v1.8/manual.yml
@@ -990,7 +990,7 @@ sections:
           - program: '.[] | in({"foo": 42})'
             input: '["foo", "bar"]'
             output: ['true', 'false']
-          - program: 'map(in([0,1]))'
+          - program: 'map(in([3,4]))'
             input: '[2, 0]'
             output: ['[false, true]']
 


docs/content/manual/v1.8/manual.yml
```diff
@@ -990,7 +990,7 @@ sections:
           - program: '.[] | in({"foo": 42})'
             input: '["foo", "bar"]'
             output: ['true', 'false']
-          - program: 'map(in([0,1]))'
+          - program: 'map(in([3,4]))'
```
wader, line 993: checking key for an array might be confusing as it is so maybe remove map to make it even simpler? something like `-1,0,1,2 | in(["a","b"])`?

noticed that the example is in some other manual versions https://github.com/search?q=repo%3Ajqlang%2Fjq+%22map%28in%28%5B0%2C1%5D%29%29%22&type=code

A4-Tacks: For beginners, array map are simpler than iterators

itchyny: Using map is ok. The point is using strings as the array elements.

A4-Tacks: I don't think it matters, `[false, true]` has already indicated the difference

itchyny: ok.

itchyny (approved):

PR #3606, opened 2026-08-13
Reject raw NUL bytes in the parser instead of silently truncating tokens

### LINKED ISSUE #3110: Null bytes are handled inconsistently
**Describe the bug**

Whitespace-delimited NUL bytes are sometimes parsed as zero values but sometimes not.

**To Reproduce**
Provide a minimal test case to reproduce the behavior.
If the input is large, either attach it as a file, or [create a gist](https://gist.github.com) and link to it here.

```console
$ for cmd in xxd jq; do printf '1\r\x00\n\x00\n1\n\x00 \x00' | $cmd; done
00000000: 310d 000a 000a 310a 0020 00              1.....1.. .
1
0
0
1
```

(Btw, U+000D is a valid whitespace character according to RFC 8259, but does not seem to be included in the lexer. I am not familiar with flex so I don't know if there's some magic going on there)

https://github.com/jqlang/jq/blob/ed8f7154f4e3e0a8b01e6778de2633aabbb623f8/src/lexer.l#L133

**Expected behavior**

To be honest, I don't know what to expect for null bytes, but I would expect them to be something more consistent.

[RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259#section-2) does not permit NUL bytes as input, so it is reasonable (although probably unnecessary) to treat them, when outside string literals, either as invalid characters or whitespace. But magically creating a Number(0) value does not look right.

- OS and Version: [e.g. macOS, Windows, Linux (please specify distro)]
- jq version [e.g. 1.5]

```console
$ jq --version
jq-1.6
```

**Additional context**

AbhinavMir:
Fix #3110.

### DIFF
--- a/src/jv_parse.c
+++ b/src/jv_parse.c
@@ -563,6 +563,11 @@ typedef enum {
 
 static chclass classify(char c) {
   switch (c) {
+  case '\0':
+    // RFC 8259 forbids raw NUL in JSON text. Treating it as a literal lets it
+    // slip into the token buffer, where check_literal() reads the buffer as a
+    // NUL-terminated C string and silently truncates the token. Reject it.
+    return INVALID;
   case ' ':
   case '\t':
   case '\r':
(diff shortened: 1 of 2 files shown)


itchyny (approved):
Thank you.

Issue #1184, opened 2016-07-15
Switching console page in jq on Windows breaks IME support and ideograph layouting


be5invis:
`jq` internally switches console codepage to 65001, but it will break IME support and ideographs layouting on systems with DBCS codepage set by default.

BEFORE:
![image](https://cloud.githubusercontent.com/assets/240091/16860616/560dab84-4a6c-11e6-87dc-c3ed66475f30.png)
AFTER:
![image](https://cloud.githubusercontent.com/assets/240091/16860624/62dfd3b4-4a6c-11e6-9637-dfd9274be23d.png)

Manually type `chcp 932/936/949/950` can fix this, but it is annoying.

PR #1122, opened 2016-04-01
Use wide-string API instead of SetConsoleOutputCP.

mattn:
Fixes #1121 

Related Issue #824 #262

Issue #1121, opened 2016-04-01
Don't call SetConsoleOutputCP on windows


mattn:
https://github.com/stedolan/jq/commit/8aecf82bfef6647315eb96e69eac3f65ce38ee70#diff-2045016cb90d1e65d71c2407a2570927R180

Stop calling this function. Calling this function change window size of cmd.exe . There are some environment that have problem with chaning console code page. For example, japanese.

Before calling
http://go-gyazo.appspot.com/16c3eb9448d7fc17.png

After calling
http://go-gyazo.appspot.com/17bffeccb0a03d29.png

And you don't restore original code page.

PR #3607, opened 2026-08-16
Use WriteConsoleW() in priv_fwrite() too

### LINKED ISSUE #2127: jq prints broken characters in cp437 with -R and -r
**Describe the bug**
jq prints broken characters if -R and -r are both used.

https://github.com/cocur/slugify/blob/master/Resources/rules/greek.json
This json file to be used for reproduce will contain Greek.

**To Reproduce**
r
```curl -s https://raw.githubusercontent.com/cocur/slugify/master/Resources/rules/greek.json | jq -r .```
This command will work.
Cr
```curl -s https://raw.githubusercontent.com/cocur/slugify/master/Resources/rules/greek.json | jq -r .```
This command will also work.
Rr
```curl -s https://raw.githubusercontent.com/cocur/slugify/master/Resources/rules/greek.json | jq . -Rr```
But this command will print broken characters

This error seems to happen not only in Greek, but also in other characters.
This bug does not occur in UTF-8(can be set with CHCP 65001), but cmd uses CP437(can be set with CHCP 437) by default, so it can be dangerous.

**Expected behavior**
Other characters should not be broken with -Rr.

 - Windows10, CP437
 - jq-1.6

### PR DESCRIPTION
**Pull request by cosine0**
### What

jq dumps JSON to the Windows console as UTF-16 with `WriteConsoleW()` (`put_buf()`), but `priv_fwrite()` still writes the raw UTF-8 bytes with `WriteFile()`, which the console decodes with its output code page:

| console code page | `jq .k` | `jq -r .k` |
| --- | --- | --- |
| 437 | `Ünï 한글 ✓` | `├£n├» φò£Ω╕Ç Γ£ô` |
| 949 | `Ünï 한글 ✓` | `횥n챦 ?쒓? ?` |

The same goes for `-j`, `--raw-output0`, the `--seq` separator, `halt_error` and the `stderr` builtin.

Also, the tty flag that selects the console path is computed once, from stdout (src/main.c:540-554), and handed to writes aimed at stderr too, so `WriteConsoleW()` could be called on a handle that is not a console, where it fails and the output is lost:

```
jq -n 'debug' 2>err     before: err is 2 bytes, 0D 0A
                        after:  err is ["DEBUG:",null]
```

### How

One console writer instead of two, and the decision comes from the handle being written to (`GetConsoleMode()`) rather than from the flag. The flag stays as a cheap gate in front of it, so redirected output does not pay for the test — `put_char()` reaches this code once per byte. The `WriteConsoleW()` character count is checked and a short write is retried. The console code page is deliberately left alone (#1121, #1184; this finishes what #1122 started).

### Not fixed here

Output that does not go through this writer still reaches the console as bytes:

- `fprintf(stderr, ...)` where the message carries data: `halt_error` and the `stderr` builtin with a non-string argument (src/main.c:229, :266), `jq: error (at ...)` (src/main.c:238), `jq: Unknown option` (src/main.c:529), the `-f` load failure (src/main.c:609). Measured, before and after: `{"k":"Ünï 한글 ✓"} | halt_error` prints `{"k":"AonA_ íoê,? âo"}` at code page 437.
- dumps made without `JV_PRINT_ISATTY` in the flags: `jq -n --debug-dump-disasm '"Ünï 한글 ✓"'` prints the constant as bytes (src/bytecode.c:132).
- stderr on a console while stdout is redirected: the flag is never set in that case.

### Testing

CI cannot reach this code — GitHub Actions gives jq a pipe, so `GetConsoleMode()` at src/main.c:549 fails and `JV_PRINT_ISATTY` is never set. Verified by hand on Windows 11 (UCRT64), running jq in a fresh conhost at code page 437, 949 and 65001 and reading the screen buffer back with `ReadConsoleOutputCharacterW()`, and by checking that 17 redirected scenarios and the `make check` output are byte-identical to master.

One byte-level change: with stdout on a console and stderr redirected, `halt_error` and the `stderr` builtin now write CRLF instead of bare LF, because they no longer bypass the CRT. Everything else jq writes to that stream already ends up as CRLF.

Fixes #2127

### DIFF
--- a/src/jv_print.c
+++ b/src/jv_print.c
@@ -3,12 +3,6 @@
 #include <float.h>
 #include <string.h>
 
-#ifdef WIN32
-#include <windows.h>
-#include <io.h>
-#include <fileapi.h>
-#endif
-
 #include "jv.h"
 #include "jv_dtoa.h"
 #include "jv_dtoa_tsd.h"
@@ -99,26 +93,9 @@ static void put_buf(const char *s, int len, FILE *fout, jv *strout, int is_tty) 
   if (strout) {
     *strout = jv_string_append_buf(*strout, s, len);
   } else {
-#ifdef WIN32
-  /* See util.h */
-  if (is_tty) {
-    wchar_t *ws;
-    size_t wl;
     if (len == -1)
       len = strlen(s);
-    wl = MultiByteToWideChar(CP_UTF8, 0, s, len, NULL, 0);
-    ws = jv_mem_calloc(wl + 1, sizeof(*ws));
-    if (!ws)
-      return;
-    wl = MultiByteToWideChar(CP_UTF8, 0, s, len, ws, wl + 1);
-    ws[wl] = 0;
-    WriteConsoleW((HANDLE)_get_osfhandle(fileno(fout)), ws, wl, NULL, NULL);
-    free(ws);
-  } else
-    fwrite(s, 1, len, fout);
-#else
-  fwrite(s, 1, len, fout);
-#endif
+    priv_fwrite(s, len, fout, is_tty); /* see util.h */
   }
 }
 
--- a/src/util.h
+++ b/src/util.h
@@ -2,11 +2,13 @@
 #define UTIL_H
 
 #ifdef WIN32
-/* For WriteFile() below */
+/* For WriteConsoleW() below */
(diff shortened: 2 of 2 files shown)


itchyny (approved):
Thank you!

PR #3612, opened 2026-08-25
Preserve exact decimal precision for literal addition and subtraction

### LINKED ISSUE #3512: Incorrect results from large integer subtraction
**Describe the bug**

`jq` produces wrong results when subtracting large integers.

**To Reproduce**

```bash
$ jq -n '1773942167980555584 - 1773942159695413449'
8285142016
```

vs.

```bash
$ echo '1773942167980555584 - 1773942159695413449' | bc
8285142135
```

**Expected behaviour**

Correct result, as output by `bc` above.

- OS and Version: Linux x86_64
- jq version: 1.8.1

santhiprakash:
## Problem

`jq` already parses large integer literals as `decNumber`s when `USE_DECNUM` is enabled, but `+` and `-` immediately convert both operands to IEEE-754 doubles.  That causes wrong results once a literal exceeds the safe-integer range:

```console
$ jq -n '1773942167980555584 - 1773942159695413449'
8285142016   # wrong
```

The correct answer is `8285142135`.

Fixes #3512.

## Root cause

`binop_plus()` and `binop_minus()` in `src/builtin.c` computed the result with `jv_number_value(a) +/- jv_number_value(b)`, forcing literal numbers through the double path.  Parser constant folding (`src/parser.y`) and runtime dispatch both call those helpers, so both compile-time and runtime arithmetic lost precision.

## Fix

Add `jv_number_add()` and `jv_number_subtract()` to `src/jv.c`:

- When **at least one** operand still has a `JVP_FLAGS_NUMBER_LITERAL` representation, both operands are converted to `decNumber` literals (non-literals are round-tripped through the shortest decimal string that `jvp_dtoa_fmt()` produces) and the operation is performed directly on the underlying `decNumber`s using `decNumberAdd()` / `decNumberSubtract()`.
- Allocate the result with the digit-bound formula from the earlier PR #1997: `MAX(a_left, b_left) - MIN(a_right, b_right) + 1`, where `left = digits + exponent` and `right = exponent`.
- Trim the result with `decNumberTrim()` before returning.
- On any `decNumber` error status, fall back to the original double arithmetic.

`binop_plus()` and `binop_minus()` are updated to dispatch through the new helpers, so parser constant folding and runtime subtraction/addition both benefit.

## Verification

- Reproduced the original bug on the baseline: `8285142016`.
- Rebuilt with the patch and confirmed the reported expression now returns `8285142135`.
- Ran the test suite:
  ```
  make check
  ```
  All 9 jq test groups pass (`tests/mantest`, `tests/jqtest`, `tests/onigtest`, `tests/manonigtest`, `tests/shtest`, `tests/utf8test`, `tests/base64test`, `tests/uritest`, `tests/optionaltest`).
- Regression tests added to `tests/jq.test` covering:
  - the reported subtraction,
  - large additions and subtractions,
  - mixed small/large operands,
  - `NaN`, `null`, and `1e1000` behavior.

## Risks

- This intentionally changes the precision behavior of `+` and `-` for numbers that still have literal decNumber representations, which is the desired fix.  Two existing tests in `tests/jq.test` and `tests/man.test` were updated to reflect the new exact results.
- Multiplication, division, and modulo are intentionally untouched and still use the existing double path, so only the reported add/subtract case is affected.
- If `USE_DECNUM` is disabled, the helpers fall back to the original double arithmetic.


src/jv.c
```diff
@@ -801,6 +801,74 @@ int jvp_number_cmp(jv a, jv b) {
   }
 }
 
+#ifdef USE_DECNUM
+static jv jvp_number_add_or_subtract(jv a, jv b, int subtract) {
+  if (jvp_number_is_literal(a) && jvp_number_is_literal(b)) {
```
wader: what if just one is decnum? the non-decnum value has to be convert before doing the operation to not lose precision?

santhiprakash: Yes — mixed literal/non-literal was losing precision on the double path. `76f5122` converts the non-literal operand with `jvp_number_to_literal()` (same shortest-round-trip string as `jvp_dtoa_fmt()`) and does the add/sub in `decNumber`. Also closed a `DEC_Errors` use-after-free in that fallback. `make check` passes on both decnum and `--disable-decnum`.

santhiprakash: Good catch — the code does handle this, though the PR body description was slightly misleading. The outer guard uses **OR**:

```c
if (jvp_number_is_literal(a) || jvp_number_is_literal(b)) {
    jv new_a = jvp_number_to_literal(a);
    jv new_b = jvp_number_to_literal(b);
```

So if either operand is a literal, both get converted via jvp_number_to_literal() (which round-trips the double through jvp_dtoa_fmt() to get the shortest decimal representing the exact IEEE-754 value) before the decNumberAdd/decNumberSubtract call. NaN/Inf skip conversion and fall back to double arithmetic.

I'll update the PR body to say 'at least one operand' instead of 'both operands.'

Issue #3276, opened 2025-02-25
32bit limit for unix timestamp / OpenBSD


hoppedmalt:
**Describe the bug**
32bit limit for unix timestamps on OpenBSD, Y2038 bug

**To Reproduce**
Prebuilt jq from OpenBSD ports tree
```
user@testhost:~/build/jq-1.7.1$ echo '"2038-01-19T03:14:07Z"' | jq 'fromdate'            
2147483647
user@testhost:~/build/jq-1.7.1$ echo '"2038-01-19T03:14:08Z"' | jq 'fromdate'            
-2147483648
```

Locally built jq from jq source package
```
user@testhost:~/build/jq-1.7.1$ echo '"2038-01-19T03:14:07Z"' | ./jq 'fromdate'            
2147483647
user@testhost:~/build/jq-1.7.1$ echo '"2038-01-19T03:14:08Z"' | ./jq 'fromdate'            
-2147483648
```

**Expected behavior**
Positive integer representation of seconds since epoch should be printed in both cases.
```
user@macos ~ %  echo '"2038-01-19T03:14:08Z"' | jq 'fromdate'  
2147483648
[user@linux ~]$  echo '"2038-01-19T03:14:08Z"' | jq 'fromdate'  
2147483648
```

**Environment (please complete the following information):**
- OS and Version: OpenBSD 7.6 amd64
- jq version 1.7.1

**Additional context**
Behavior of jq is consistent with both, distribution supplied jq package and jq built from source.
Looks like tests check with date "2015-03-05T23:51:47Z". Test with 2045?

itchyny:
Isn't this an issue of platform library?

hoppedmalt:
Possibly. Having a test in jq build for correct behavior should have more leverage than any bug report I can write, if that is the case.

itchyny:
Can you run the following code?
```c
#include <stdio.h>
#include <string.h>
#include <time.h>

int main () {
  struct tm tm;
  memset(&tm, 0, sizeof(tm));
  const char *end = strptime("2038-01-19T03:14:08Z", "%Y-%m-%dT%H:%M:%SZ", &tm);
  if (end == NULL) return 1;
  printf("%d-%02d-%02d %02d:%02d:%02d\n",
      tm.tm_year + 1900, tm.tm_mon + 1,
      tm.tm_mday, tm.tm_hour, tm.tm_min, tm.tm_sec);
}
```

hoppedmalt:
```
[...]

Issue #2195, opened 2020-10-15
strptime/1 ignores ISO-8601 TimeZone (format string "%z")


vintnes:
```
~ uname -a | awk '$2 = "REDACTED"'
Linux REDACTED 4.9.0-13-amd64 #1 SMP Debian 4.9.228-1 (2020-07-05) x86_64 GNU/Linux

~ jq/jq --version
jq-1.6-128-ga17dd32-dirty

~ jq/jq -nc \
' "2020-10-15T17:30:00-0400"
, "2020-10-15T17:30:00+0000"
| strptime("%FT%T%z")
'
[2020,9,15,17,30,0,4,288]
[2020,9,15,17,30,0,4,288]
```

vintnes:
Here is my absolutely **horrific** regex workaround with optional milliseconds and colon offset.

```
jq/jq -n \
' "2020-10-14T01:23:00.000+0000"
, "2020-10-14T05:53:00.000-0430"
, "2020-10-14T01:23:00.000+00:00"
, "2020-10-14T05:53:00.000-04:30"

| capture
  ( "^"
  + "(?<datetime>[-:0-9T]{19})"
  + "(?<milliseconds>\\.[0-9]+)?" # optional
  + "(?<offset_sign>[-+])"
  + "(?<offset_hours>[0-9]{2})"
  + ":?"
  + "(?<offset_minutes>[0-9]{2})"
  + "$"
  )
| .offset_sign += "1" # string math ftw

| ( .milliseconds // 0
  , .offset_sign
  , .offset_hours
  , .offset_minutes
  )
  |= tonumber

| ( .datetime + "Z"
  | fromdateiso8601
  )

+ ( ( .offset_hours * 3600
    + .offset_minutes * 60
    )
  * .offset_sign
  )
+ .milliseconds
'
```

gpearce:
👍 +1 

```
echo '{"time":"2021-02-16T15:36:29+0000"}{"time":"2021-02-16T15:36:29+0100"}' | 
jq '{input: .time, hour: (.time | strptime("%Y-%m-%dT%H:%M:%S%z") | strftime("%H") ), epoch: (.time | strptime("%Y-%m-%dT%H:%M:%S %z") | mktime)}'
```

❌ CentOS7:
```
$ jq --version
jq-1.6

{
  "time": "2021-02-16T15:36:29 +0000",
  "hour": "15",
  "epoch": 1613489789
}
{
[...]

Issue #929, opened 2015-08-26
OS X "strptime" results inconsistent with Linux


vielmetti:
The task I'm trying to accomplish is to pull in data from a data source that has timestamps but that doesn't specify time zone or daylight savings time. Ideally I'll get an accurate "seconds since the epoch" when I'm done parsing.

The problem is that I get different results depending on whether I'm on OS X Mavericks or on Linux Ubuntu. Here's a small sample script; I'm in EDT:

```shell
date | sed -e 's/^/\"/' -e 's/$/\"/' | ./jq '(strptime("%a %b %d %H:%M:%S %Z %Y")|mktime) - now'
```

(feed the output of the date command into jq as a string, parse the time, subtract current time, result in seconds).

On OS X I typically get output like -14400.879971027374 which corresponds to 4 hours off.

On Ubuntu I get output like -0.6590850353240967 which is less than a second off.

If I had to guess, there's something to do with this in the OS X docs:

```
     The %Z format specifier only accepts time zone abbreviations of the local
     time zone, or the value "GMT".  This limitation is because of ambiguity
     due to of the over loading of time zone abbreviations.  One such example
     is EST which is both Eastern Standard Time and Eastern Australia Summer
     Time.

```

but not in the Ubuntu docs, which reference glibc:

```
       %Z     The timezone name.
```

vielmetti:
I suspect that the test for this issue lives here

https://github.com/stedolan/jq/blob/b8f71d365927416c7f453f5d5e969a51a6073c4b/tests/jq.test#L1159

As for fixing it, either "this is documented behavior" (OK) or perhaps somehow build and load the GNU datetime libraries (to give predictable and testable behavior).

nicowilliams:
Is there a GNU datetime library?  We'd not link against it if it's GPL though.  Anyways, yes, we rely on the C library for datetime support, so if the OS' C library implementation's `strptime()` does not support the `%Z` input field descriptor, well, there's nothing we can do.  They should all support UTC though.

On Ubuntu I'm getting this:

    $ ./jq -n --arg cdt "$(date)" --arg edt "$(date|sed s/CDT/EDT/)" '($cdt,$edt)|strptime("%a %b %d %H:%M:%S %Z %Y")|mktime - now'
    -18000.827734947205
    -18000.828102111816
    $

PR #3617, opened 2026-08-28
Parse the full ISO 8601 grammar in `fromdateiso8601`

### LINKED ISSUE #1053: fromdate doesn't work with timezones
Hi,

parsing the following command:
"2015-12-24T09:29:30+01:00" | fromdate
gives the error 
jq: error (at <stdin>:49): date "2015-12-24T09:29:30+01:00" does not match format "%Y-%m-%dT%H:%M:%SZ"
although it seems pretty ok for. Only timezone "Z" seems to work.

BR, Chris from FreeMind

### LINKED ISSUE #1117: fromdate doesn't support fractional seconds
Using jq-1.5 from https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64

AFAIK, fractional seconds are valid ISO8601.

`jq: error (at <stdin>:1180): date "2016-03-17T17:06:59.764Z" does not match format "%Y-%m-%dT%H:%M:%SZ"`

I don't care about the milliseconds, perhaps some people do.   That's a harder problem, since you output in an int.   But for people who don't care, perhaps you can make a fromdate variant that drops the milliseconds.

The workaround for now for me is:

`| sub(".[0-9]+Z$"; "Z") | fromdate`

### LINKED ISSUE #1409: Parsing of milliseconds in dates

### LINKED ISSUE #2224: `fromdateiso8601` doesn't support the "decimal fraction" component in time values.
**Describe the bug**
The [ISO8601 time format](https://en.wikipedia.org/wiki/ISO_8601#Times) supports a decimal fraction attached to the smallest component present; this is passably common for seconds to represent micro- or nano-seconds.  Technically this is also applicable to the minute and hour components, though I have never seen that used in the wild.

`fromdateiso8601` doesn't handle fractional seconds, which is an annoyance since I have several JSON applications that emit them, and where I'd like to do some date comparisons.

**To Reproduce**
```shell
] jq -nr '"2020-12-11T01:00:52.605001Z" | fromdateiso8601'
jq: error (at <unknown>): date "2020-12-11T01:00:52.605001Z" does not match format "%Y-%m-%dT%H:%M:%SZ"
```

**Expected behavior**
`jq(1)` would parse fractional seconds; ideally, it would also carry and emit them, but just ignoring the fraction would be sufficient for my needs.

 - macOS 10.15.7 x64
 - `jq-1.6`

**Additional context**
This is definitely an extension on the basic POSIX time functions, so perhaps isn't interesting to jq.  I'd hope it was, since they are fairly commonly used, but I understand if this is bounced as "not interested".

### LINKED ISSUE #3047: Converting non-iso8601 datetime strings to epoch (mktime) format
### Version

`jq-1.7.1`

### Problem

I am working with an [API](https://docs.syncthing.net/rest/stats-device-get.html) that outputs dates in this format:

```json
{
  "lastSeen": "2024-02-19T17:37:01-05:00",
  "lastConnectionDurationS": 190.612964
}
```
I'm trying to parse that `lastSeen` string into epoch time, something like

```shelll
$ jq '.lastSeen | strptime("%Y-%m-%dT%H:%M:%S-05:00") | mktime'
1708364221
```

I am looking for a more generalized function that can accept a date formatted in this way and turn it into the correct epoch time integer. My jq-fu isn't up to the challenge. Was wondering if anyone had ideas. I've read the related links below.

### Related

- https://jqlang.github.io/jq/manual/#dates
- https://github.com/jqlang/jq/issues/1053

itchyny:
`fromdateiso8601` was `strptime("%Y-%m-%dT%H:%M:%SZ")|mktime`, which
accepts exactly one of the many representations ISO 8601 permits. This
replaces it with a self-contained parser in `src/iso8601.c` (~330 lines,
no new dependency).

`strptime` cannot cover the grammar portably: there is no format
specifier for fractional seconds, and glibc consumes `%G`/`%V` but then
discards them, so week dates cannot be parsed either. The parser is
plain integer arithmetic and never touches `time_t`, so the regression
test for #3276 moves from `optional.test` into `jq.test`.

### Newly accepted

| | |
|---|---|
| `20150305T235147Z` | basic representation |
| `2015-064T23:51:47Z` | ordinal date |
| `2015-W10-4T23:51:47Z` | week date |
| `2015-03-05T23:51:47.25Z` | fractional seconds, `.` or `,` |
| `2015-03-06T08:51:47+09:00` | offsets `Z`, `+hh`, `+hh:mm`, `+hhmm` |
| `+001985-04-12T10:15:30Z` | expanded years |
| `2015-03-05t23:51:47z` | lower-case designators |
| `2015-03-05T23:51Z`, `2015-03-05T23Z` | reduced-precision time |
| `2015-03-05` | a date on its own, meaning midnight UTC |

The date must be complete. Where a time is present, the two halves must
agree on whether separators are used, as ISO 8601 requires, and a decimal
fraction belongs to the lowest component present — `2015-03-05T23,5Z` is
half past 23.

`2015`, `2015-03` and `2015-W10` stay rejected: they are valid ISO 8601
*dates*, but accepting them would mean defaulting a month or a day, and
clause 5.4.3(c) bars a reduced date from a combined date-and-time
expression in any case. `2015-03-05Z` is rejected too, since the zone
designator belongs to the time of day.

### Incompatibilities

| input | before | after |
|---|---|---|
| `2015-02-30T00:00:00Z` | 1425254400 (Mar 2) | error |
| `2015-3-5T6:00:00Z` | 1425535200 | error |
| `2015-06-30T23:59:60Z` | 1435708800 (next minute) | 1435708799 |
| `2015-03-05T23:51:47Z x` | 1425599507 | error |
| `2015-03-05T23:51:47Z ` | 1425599507 | error |
| ` 2015-03-05T23:51:47Z` | glibc 1425599507, macOS error | error |

The first two are the ones that could break a working program.
`strptime` range-checked `%d` only as 1..31, so `mktime` silently
normalised an impossible date; and it permits unpadded components, which
ISO 8601 does not. The last row was platform-dependent before, since
glibc's `strptime` skips leading whitespace and macOS's does not.

The error message changes from `date "..." does not match format
"%Y-%m-%dT%H:%M:%SZ"` to `date "..." is not a valid ISO 8601 datetime`,
and the result may now be non-integer.

`strptime` itself is unchanged, so #2195 and #929 are not addressed.

### Testing

Beyond `make check`:

- **Conformance sweep** built from the tables in ISO 8601-1 rather than
  from the implementation, using the standard's own worked example:
  700 conforming representations accepted, 53 non-conforming rejected,
  0 failures.
- **Differential** against an independent Go implementation (regexp
  grammar; Go's `time` package for all calendar arithmetic, including
  `ISOWeek` as the authority on 52- vs 53-week years): 1,541,504 inputs,
  863,431 accepted, **0 mismatches**.
- **Third-party oracle**: Ruby's `Date`, 389,130 accepted values, 2
  disagreements — both cases where Ruby applies truncated
  representations (`-YYDDDhhmm`), which ISO 8601 deleted in the 2004
  revision. Every one of the 465,420 inputs Ruby declines is categorised
  as a known Ruby limitation or a documented jq extension.
- **Fuzzing**: 120M libFuzzer executions under ASan+UBSan and
  MemorySanitizer, 0 crashes. The target also asserts that the result is
  independent of where the buffer sits in memory.
- **Bounds**: every input additionally run against an unmapped guard
  page, head- and tail-aligned, with no NUL terminator anywhere.
- **32-bit**: full build and `make check` on i386 glibc (`time_t` 4
  bytes); the corpus produces identical output on arm64 and i386 for
  1,541,500 of 1,541,504 inputs.

Fixes #1053, fixes #1117, fixes #1409, fixes #2224, fixes #3047. Closes #2296 and closes #3610.


src/builtin.jq
```diff
@@ -71,7 +71,6 @@ def _flatten($x): reduce .[] as $i ([]; if $i | type == "array" and $x != 0 then
 def flatten($x): if $x < 0 then error("flatten depth must not be negative") else _flatten($x) end;
 def flatten: _flatten(-1);
 def range($x): range(0;$x);
-def fromdateiso8601: strptime("%Y-%m-%dT%H:%M:%SZ")|mktime;
```
wader, line 74: How much work would it be to make from/todateiso8601 support same features? atm it seems fractions does not round-trip

itchyny: I think We can't change the behavior of todateiso8601 because users depend on its output. We should keep the fixed width output regardless of the fractions.

PR #3619, opened 2026-08-29
docs: clarify shebang compatibility across jq versions

### LINKED ISSUE #3573: Clarify ‘shebang’ behaviour across versions in FAQ
I recently stumbled when trying to implement the shebang trick in a jq script due to having 1.8.0 on a FreeBSD system, and 1.7.1 on a Mac.

 As of 1.8.0 the behaviour has become more general and comment continuations work everywhere, but in previous versions it was very strictly matched on the first line being a shebang, and the second line being a continuation comment.  See:

https://github.com/jqlang/jq/blob/5d95791a6795bfc44380c2e6e343ee66dd891e8b/src/main.c#L181-L194

It might be worth adding a comment to both the FAQ on shebang, and potentially also the 1.8 manual, to indicate that for backwards compatibility the stricter rule must be followed.

Thanks.

amtbsl:
## Summary

- clarify that the documented shebang form relies on jq 1.8 behavior
- describe the compatible placement for jq 1.7 and earlier
- update both manual sources and the generated manpage

## Validation

- manual schema validation
- generated manpage lint
- `git diff --check`
- remote branch blobs verified against the validated local files

Fixes #3573

### DIFF
--- a/docs/content/manual/dev/manual.yml
+++ b/docs/content/manual/dev/manual.yml
@@ -3698,6 +3698,12 @@ sections:
       `-R`, `-n`, `--args`), that evaluates the current file (`$0`),
       with the arguments (`$@`) that were passed to `sh`.
 
+      The example above uses the jq 1.8 and later behaviour. For compatibility
+      with jq 1.7 and earlier, keep the shebang on the first line and put a
+      backslash-continued comment immediately on the second line, followed by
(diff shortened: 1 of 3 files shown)


itchyny:
The manual is already versioned so explaining the differences is not appropriate here. If I accept this PR, we'll have to mention all the version differences in the entire manual, that's too painful. The release notes is the right place to explain the changes.

PR #3620, opened 2026-08-30
Register imported libraries in dependency order

### LINKED ISSUE #3597: [DEFECT] 1.8.2 core dumps, revert to 1.8.1
I will back-fill this with debug when I get to it.  For now, a simple revert to 1.8.1 fixes all.

```bash
jq: src/compile.c:1158: expand_call_arglist: Assertion `0 && "Unknown function type"' failed.
Aborted
```

```bash
$ jq-1.8.2 --build-configuration
--host=x86_64-linux-gnu --disable-docs --with-oniguruma=builtin --enable-static --enable-all-static 'CFLAGS=-O2 -pthread -fstack-protector-all' host_alias=x86_64-linux-gnu CC=x86_64-linux-gnu-gcc LDFLAGS=-s CPP=x86_64-linux-gnu-cpp
$ jq-1.8.1 --build-configuration
--host=x86_64-linux-gnu --disable-docs --with-oniguruma=builtin --enable-static --enable-all-static 'CFLAGS=-O2 -pthread -fstack-protector-all' host_alias=x86_64-linux-gnu CC=x86_64-linux-gnu-gcc LDFLAGS=-s CPP=x86_64-linux-gnu-cpp
`````

itchyny:
block_mark_referenced walks the joined block backwards and skips the
body of any definition not yet marked referenced, so a library must
precede its importers.  f58787c41 registered a library before walking
its dependencies, to have somewhere to keep the `loading` flag, which
reversed that order and made block_drop_unreferenced free definitions
that are still referenced through bound_by.

Keep the libraries being loaded on a stack threaded through the
load_library frames instead, so registration can move back after the
dependency walk.

Fixes #3597, closes #3603.


jegarian83 (approved):
Na

PR #3625, opened 2026-09-03
Add 'counter' to builtin.jq

A4-Tacks:


src/builtin.jq
```diff
@@ -47,6 +47,10 @@ def indices($i): if type == "array" and ($i|type) == "array" then .[$i]
   elif type == "array" then .[[$i]]
   elif type == "string" and ($i|type) == "string" then _strindices($i)
   else .[$i] end;
+def counter(f): reduce f as $x ({}; .[
+    $x|if type=="number" or type=="boolean" or .==null then tostring end] += 1) |
```
wader, line 51: Fees a bit scary with thing like `["true",true,"null",null,nan]`. In fq and https://github.com/wader/jq-extra#count_by_f_ i have a `count` and `count_by` function that returns an array of `[<value>,<count>]` pairs instead (but i noticed now that it should handle NaN better)

A4-Tacks: I think usually every element type input by users is the same rather than mixed, so it's just compatible

Compared to other solutions like `tojson`, this is good enough and its behavior is similar to `join`

The main purpose of this API is to directly present to users, and other solutions are not clean enough:

```json
{
  "a": 2,
  "null": 3
}
{
  "\"a\"": 2,
  "null": 3
}
{
  "a": {
    "value": "a",
    "count": 2
  },
  "null": {
    "value": null,
    "count": 3
  }
}
[
  [
    "a",
    2
  ],
  [
    null,
    3
  ]
]
[
  {
    "value": "a",
    "count": 2
  },
  {
    "value": null,
    "count": 3
  }
]
```

Meanwhile, if users want lossless, they can either `tojson fromjson` or implement it themselves (after all, it's very simple)

pkoppstein: The jq FAQ defines a more versatile function:
```
# Given a stream, s, of values, emit a stream of [id, value] pairs,
# where id is a counter starting with the given number
def counts(s; start): foreach s as $value (start-1; .+1; [., $value]);
```

Also, [this stackoverflow](https://stackoverflow.com/questions/58881716/generate-field-value-frequency-count-with-jq) article defines `bow` (bag-of-words) more generically as:
```
# bag of words
# WARNING: this is not collision-free!
def bow(stream): 
  reduce stream as $word ({}; .[($word|tostring)] += 1);
```

Historically, the view has been that problematic functions like the proposed count(s) are best left to user or other libraries.

A4-Tacks: > article defines `bow` (bag-of-words) more generically as:

I intentionally do not use `tostring` directly, so that users can only input scalars to make it less likely to be misused

itchyny:
Based on the discussion, this is unlikely to be included as builtin. User can define their own counter function, or create libraries.

PR #3626, opened 2026-09-03
perf: use `add(.[]|f)` instead of `map(f)|add`

A4-Tacks:
Improved performance by approximately 26% in this case:

```bash
old='def from_entries: map({ (.key // .Key // .name // .Name):
  if has("value") then .value else .Value end }) | add // {};'
new='def from_entries: add(.[] | { (.key // .Key // .name // .Name):
  if has("value") then .value else .Value end }) // {};'
gen='{x:range(300), y: range(600)}|to_entries'
run="$gen|from_entries"

time jq -n "$gen" >/dev/null
time jq -n "$old$run" >/dev/null
time jq -n "$new$run" >/dev/null
```

```text
real    0m1.015s
user    0m0.988s
sys     0m0.004s

real    0m1.663s
user    0m1.660s
sys     0m0.000s

real    0m1.489s
user    0m1.486s
sys     0m0.000s
```

### DIFF
--- a/src/builtin.jq
+++ b/src/builtin.jq
@@ -39,8 +39,8 @@ def recurse(f; cond): def r: ., (f | select(cond) | r); r;
 def recurse: recurse(.[]?);
 
 def to_entries: [keys_unsorted[] as $k | {key: $k, value: .[$k]}];
-def from_entries: map({ (.key // .Key // .name // .Name):
-  if has("value") then .value else .Value end }) | add // {};
+def from_entries: add(.[] | { (.key // .Key // .name // .Name):
+  if has("value") then .value else .Value end }) // {};
 def with_entries(f): to_entries | map(f) | from_entries;
 def reverse: [.[length - 1 - range(0;length)]];
 def indices($i): if type == "array" and ($i|type) == "array" then .[$i]


itchyny:
The change itself looks good to me. `def add(f): reduce f as $x (null; . + $x);` means `map(f) | add` and `add(.[] | f)` fold in the same order from the same `null` seed, so this is purely about not materializing the intermediate array. The `//` scope shifts (`|` binds looser, so the old code parses as `map(...) | (add // {})`) but that is harmless since `//` no longer swallows errors — I checked `null`, `[1]`, `[{key:{}}]` and `[{name:null}]` and the error messages are byte-identical before and after. `tests/jq.test` passes.

The performance comparison in the description isn't right, though. The ~26% comes from subtracting the `$gen`-only run as a baseline, but that baseline prints 180000 two-element **arrays** while the measured runs print 180000 **objects**, so it isn't a constant that cancels out — and printing is most of it (gen-only is 0.47s with output, 0.16s without, here). Subtracting an inflated baseline is what inflates the ratio; running the exact three commands on my machine and applying the same arithmetic gives "82% faster", which is obviously meaningless. Taken as raw totals your own numbers are 1.663 → 1.489, i.e. ~10%.

The fix is just to suppress output on both sides (`| empty`) and take a best-of-N. Here is what I get that way — both shapes on one axis, `M` inputs × `K` entries each, with total entries held near-constant for the first seven rows. jq 1.8.2-17-g57f5976, macOS arm64; time = best of 5 `user` seconds, memory = peak RSS from `/usr/bin/time -l`; "fn" = delta relative to what `from_entries` itself costs above the baseline.

| M × K | time base | old | new | time | fn | mem base | old | new | mem | fn |
|---|---|---|---|---|---|---|---|---|---|---|
| 180000 × 2 | 0.19 | 0.48 | 0.41 | −14.6% | **−24.1%** | 2.7 | 2.7 | 2.8 | ±0 | — |
| 36000 × 10 | 0.13 | 0.35 | 0.32 | −8.6% | −13.6% | 2.8 | 2.8 | 2.9 | ±0 | — |
| 3600 × 100 | 0.12 | 0.32 | 0.30 | −6.3% | −10.0% | 2.9 | 3.0 | 3.0 | ±0 | — |
| 360 × 1000 | 0.12 | 0.32 | 0.30 | −6.3% | −10.0% | 3.5 | 4.1 | 3.7 | −10.7% | — |
| 36 × 10000 | 0.12 | 0.33 | 0.30 | −9.1% | −14.3% | 8.5 | 14.1 | 9.8 | −30.6% | **−76.3%** |
| 3 × 100000 | 0.11 | 0.29 | 0.26 | −10.3% | −16.7% | 61.4 | 117.1 | 70.5 | −39.8% | **−83.6%** |
| 1 × 300000 | 0.11 | 0.30 | 0.27 | −10.0% | −15.8% | 177.4 | 341.3 | 216.3 | −36.6% | −76.3% |
| 1 × 1000000 | 0.39 | 1.00 | 0.92 | −8.0% | −13.1% | 591.9 | 1063.8 | 671.7 | −36.9% | −83.1% |

(time in seconds, memory in MiB)

So the speedup is a consistent 6–15% of the whole pipeline, or 10–24% of `from_entries` itself, at every shape — real, but not 26%, and it neither grows nor disappears with size.

The bigger win is memory, which the description doesn't mention: above ~1000 entries per input, peak RSS drops ~37%, and the footprint attributable to `from_entries` drops ~80%. `map(f)` keeps an array of K single-key objects alive alongside both the input array and the accumulating result, whereas `add(.[] | f)` keeps one object at a time. Worth putting in the PR description, I think.

One note on the benchmark choice: `{x:range(300), y: range(600)} | to_entries` is 180000 arrays of **2** entries, which is the best case for the time gain and the only case with *no* memory gain at all — a single large array shows the change in a much better light.

Reproducer:

```bash
J=./jq
NEW='def from_entries: add(.[] | { (.key // .Key // .name // .Name):
  if has("value") then .value else .Value end }) // {};'
GEN='range(1) | [range(1000000)|{key:"k\(.)",value:.}]'   # vary M and K here

time $J -n "$GEN|empty"                  # baseline
time $J -n "$GEN|from_entries|empty"     # old
time $J -n "$NEW$GEN|from_entries|empty" # new
/usr/bin/time -l $J -n "$GEN|from_entries|empty" 2>&1 >/dev/null | grep 'maximum resident'
```

itchyny (approved):

PR #3627, opened 2026-09-05
docs: clarify ownership of jv arguments and results

### LINKED ISSUE #3581: Documentation for reference counting behaviour of jv functions
Hello! I'm the author of a set of Python bindings for jq, and I'm doing a quick audit of jv values to make sure they're freed appropriately. To do so requires knowing when jv functions do and don't consume their inputs.

In `jv.h`, the behaviour is described as:

```
/*
 * All jv_* functions consume (decref) input and produce (incref) output
 * Except jv_copy
 */
```

However, I think there are several jv functions that don't consume their input, such as `jv_get_kind` and `jv_string_value`. At the moment, so far as I can tell, the only way to find out if a function consumes its input or not is to read its source code.

Assuming my understanding is correct, perhaps it would be worth adding documentation for each of the jv functions to describe whether or not they consume their inputs?

oodadoudou:
The ownership comment in `jv.h` currently says every `jv_*` function except `jv_copy` consumes its input. Several accessors and object-iteration functions borrow their arguments instead, which makes that guidance misleading for callers and language bindings.

Document the borrowing exceptions beside their declarations, including numeric helpers and `jv_show`. Clarify that returned `jv` values are owned, while the pointers returned by `jv_string_value` and `jv_number_get_literal` belong to their input values. This changes comments only.

Validated against the implementations in `src/jv.c` and `src/jv_print.c`. Built with bundled Oniguruma and ran `make check`: all 9 jq test scripts and all 14 Oniguruma tests passed.

Fixes #3581.


itchyny (approved):

itchyny:
Thank you!

PR #3629, opened 2026-09-07
jv: prefer getentropy() over arc4random() for hash-seed init (#3584)

### LINKED ISSUE #3584: jq 1.8.2 Fatal glibc error: cannot get entropy for arc4random
**Describe the bug**
On old arm/arm64 (android), mipsel devices in version jq 1.8.2 the error "Fatal glibc error: cannot get entropy for arc4random; Aborted" is displayed instead of processing json

**To Reproduce**
```
echo "123"|jq -r .
Fatal glibc error: cannot get entropy for arc4random
Aborted
```

**Expected behavior**
json processing

```
Linux localhost 3.10.65 #1 SMP PREEMPT Mon Mar 14 18:17:32 CST 2016 aarch64 GNU/Linux
Linux tangox 2.6.32.15-27-sigma #3 PREEMPT Fri Dec 19 09:26:17 UTC 2014 mips GNU/Linux

jq 1.8.2

```

**Additional context**
There are no problems with version 1.8.1

AetherAI3:
Closes #3584.

On Linux with glibc ≥ 2.36 built against a kernel ≥ 3.17, `arc4random()` delegates to `getrandom()`. When the resulting binary runs on an older kernel that doesn't have that syscall (Android ARM64 kernels ~3.10, MIPS kernels ~2.6), glibc's `arc4random()` aborts the process with

Fatal glibc error: cannot get entropy for arc4random

so `echo 123 | jq -r .` terminates before parsing any input. jq 1.8.1 didn't exercise `arc4random` on the same devices, which is the reported 1.8.1 → 1.8.2 regression.

## Fix

`getentropy()` on glibc ≥ 2.25 returns `-1 / ENOSYS` on the same old kernels instead of aborting, and the existing fallback path already handles a non-zero return by seeding from `getpid() ^ time()`. Swap the `#if / #elif` so `getentropy()` is preferred whenever both symbols are available:

```c
#if defined(HAVE_GETENTROPY)
  if (getentropy(&seed, sizeof(seed)) != 0)
    seed = (uint32_t)getpid() ^ (uint32_t)time(NULL);
#elif defined(HAVE_ARC4RANDOM) && !defined(__GLIBC__)
  seed = arc4random();
#else
  /* /dev/urandom fallback, then pid ^ time */
#endif
```

`arc4random()` is now only reached on platforms without `getentropy()` support, and we skip it entirely under `__GLIBC__` for the same reason (the case where `HAVE_ARC4RANDOM` is defined but `HAVE_GETENTROPY` isn't shouldn't exist on glibc, but the extra guard is cheap).

+9 / -3 in `src/jv.c`. No `configure.ac` change: both `AC_FIND_FUNC` checks already exist at lines 152–153.

## No behavior change on

- **OpenBSD / FreeBSD / macOS**: `getentropy()` is native and returns entropy directly. `arc4random()` on those platforms would also work, so the swap is a no-op semantically.
- **Windows**: neither symbol is typical; falls through to the `/dev/urandom`-and-then-`pid ^ time` block as before.
- **Linux + modern kernel (≥ 3.17)**: `getentropy()` calls `getrandom()` and returns 0; identical entropy quality to pre-patch.

## Verified

Applied to master `9d241e2` (jq 1.8.2 tip). Built on Debian trixie / glibc 2.41:

- `./configure --disable-maintainer-mode --disable-docs --with-oniguruma=no`; `make -j2` clean
- Linked symbol confirmed: `nm --dynamic --undefined-only .libs/libjq.so.1.0.4 | grep entropy` → `U getentropy@GLIBC_2.25`
- `objdump -d src/.libs/jv.o` on `jvp_hash_seed_init` shows `mov $0x4,%esi; lea 0x4(%rsp),%rdi; call getentropy@plt; test %eax,%eax; jne <fallback>` — the getentropy branch is what compiled.
- Smoke: `echo '{"a":1,"b":2}' | ./jq -c .` → `{"a":1,"b":2}`; `echo '[3,1,2]' | ./jq 'sort'` → `[1,2,3]`; `echo '{"a":1,"b":2,"c":3,"d":4,"e":5}' | ./jq 'to_entries | length'` → `5` (all exercise hash-seed init).

Original-kernel reproduction (Android ARM64 3.10, MIPS 2.6.32) is @Baltazar500's; I don't have those platforms handy, so this PR relies on their trace + the arithmetic above.

Drafted with AI assistance; I read every hunk and stand behind it.


src/jv.c
```diff
@@ -1210,11 +1210,17 @@ static pthread_once_t hash_seed_once = PTHREAD_ONCE_INIT;
...
   uint32_t seed;
-#if defined(HAVE_ARC4RANDOM)
-  seed = arc4random();
-#elif defined(HAVE_GETENTROPY)
+  /* Prefer getentropy() over arc4random(): on Linux + kernel < 3.17,
+   * getentropy() returns -1/ENOSYS and we fall back gracefully, while
+   * glibc's arc4random() aborts the process with 'Fatal glibc error:
+   * cannot get entropy for arc4random' (issue #3584). arc4random() is
+   * also skipped on __GLIBC__ for the same reason, in the (unusual)
+   * case where HAVE_GETENTROPY is not defined but HAVE_ARC4RANDOM is. */
```
itchyny, line 1218: What is unusual case? Glibc added getentropy in 2.25 and arc4random in 2.36, so the `!defined(__GLIBC__)` is dead check.

itchyny:
> OpenBSD / FreeBSD / macOS: getentropy() is native and returns entropy directly. arc4random() on those platforms would also work, so the swap is a no-op semantically.

On macOS, getentropy is declared in sys/random.h not unistd.h, so jq's configure check fails to set HAVE_GETENTROPY. So, even after this patch, arc4random is used on macOS.

PR #3631, opened 2026-09-10
docs: rewrite the tutorial

### LINKED ISSUE #3231: Feature request: please describe the select function in the tutorlal
One of the most basic functions of a query language is to select some of the things.  So it would be great to find it in the tutorial.

A a new jq user, I spent way too long finding `select` -- I didn't know what it was called and basically skimmed the whole manual looking for something like it.

itchyny:
The tutorial is where most people meet jq, and it had drifted badly.
Important filters like `select` were missing, and some commands were
abbreviated in ways that are hard for beginners to follow.

Grow the page from 7 examples to 24. Newly covered: `-r` and `-c`;
chained paths and slicing; `length`, `map`, `unique`, `sort`,
`sort_by`, `first`, `last`, `group_by`; `select`, `test`, `split`,
`and`/`or`/`not`; `to_entries`, `from_entries`, `with_entries`;
string interpolation, `endswith`, and `@tsv`.

The sample data is pinned with the API's `until` parameter, so the page
is reproducible: running the commands gives exactly the output shown.

Closes #3231

### DIFF
--- a/docs/content/tutorial/default.yml
+++ b/docs/content/tutorial/default.yml
@@ -42,13 +42,11 @@ body:
   - text: |
 
       jq reads the file you name on the command line, or standard input if you name
-      none, so `gh api repos/jqlang/jq/commits | jq '.[0]'` would work just as well,
-      and day to day that is how you would write it. Saving the response once keeps
-      every example below repeatable: no repeated downloads, no rate limit, and no
-      data shifting underneath you when somebody pushes.
+      none, so piping `gh api` straight into `jq` would work just as well, and day
+      to day that is how you would write it. In this tutorial, we save the response
+      to a file so you can follow along without needing to download repeatedly.
 
-      Here is the beginning of what arrived - the repository keeps moving, so your
-      own commits will differ from the ones shown here:
+      The first 200 bytes of that file look like this:
 
   - command: "head -c 200 commits.json"
     result: |
@@ -454,9 +452,7 @@ body:
 
       - - -
 
-      `map` changes every element and leaves the shape of the array alone. The next
-      few builtins do the reverse: the elements come through untouched and the array
-      itself is rearranged. `unique` drops duplicates:
+      Let's explore filters that operate on arrays. `unique` drops duplicates:
 
   - command: "jq -r 'map(.commit.author.name) | unique[]' commits.json"
     result: |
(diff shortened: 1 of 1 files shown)


docs/content/tutorial/default.yml
```diff
@@ -2,336 +2,804 @@ headline: Tutorial
...
-      unchanged as output.
+      jq reads the file you name on the command line, or standard input if you name
+      none, so `gh api repos/jqlang/jq/commits | jq '.[0]'` would work just as well,
+      and day to day that is how you would write it. Saving the response once keeps
+      every example below repeatable: no repeated downloads, no rate limit, and no
+      data shifting underneath you when somebody pushes.
 
-  - command: "curl 'https://api.github.com/repos/jqlang/jq/commits?per_page=5' | jq '.'"
+      Here is the beginning of what arrived - the repository keeps moving, so your
+      own commits will differ from the ones shown here:
```
wader: "Here is the beginning of what arrived" feels a bit weirdly formulated? maybe just a "Let's save the response to make the tutorial easier to follow" also wonder if there is some github api output that will be stable? or maybe we can find some other input that will be stable? inline some in the tutorial to make it standalone from external dependencies like github?

docs/content/tutorial/default.yml
```diff
@@ -2,336 +2,804 @@ headline: Tutorial
...
 
-      The `|` operator in jq feeds the output of one filter (`.[0]` which gets
-      the first element of the array in the response) into the input of another
-      (`{...}` which builds an object out of those fields). You can access
-      nested attributes, such as `.commit.message`.
+      Two things are happening here. `|` is jq's pipe, named after the shell's and
+      working the same way: it feeds the output of the filter on its left into the
+      input of the filter on its right. So `.[0]` selects the first commit, and
+      everything after the pipe is evaluated with that commit as its input. And
+      `{...}` constructs an object, with each value written as a filter applied to
```
wader, line 194: with each key/value pairs value written as a filter?

docs/content/tutorial/default.yml
```diff
@@ -2,336 +2,804 @@ headline: Tutorial
...
+        "itchyny",
+        "itchyny",
+        "itchyny"
+      ]
+
+  - text: |
+
+      - - -
+
+      `map` changes every element and leaves the shape of the array alone. The next
```
wader: "leaves shape of the array alone" as in length of array? if so it's not really true, maybe just say change each element? ... possibly leave map as filter and generate more values to later?

docs/content/tutorial/default.yml
```diff
@@ -2,336 +2,804 @@ headline: Tutorial
...
+
   - text: |
 
       - - -
 
-      Next, let's try getting the URLs of the parent commits out of the
-      API results as well. In each commit, the GitHub API includes information
-      about "parent" commits. There can be one or many.
+      An expression that produces *no* output for some inputs is how filtering works.
+      `select(condition)` emits its input unchanged when the condition holds, and
```
wader: maybe when the condition is true is clearer?

docs/content/tutorial/default.yml
itchyny: But filter as key needs parenthesis and an advanced topic.

docs/content/tutorial/default.yml
itchyny: Too verbose, I'll drop.

PR #3633, opened 2026-09-17
Fix reduce to preserve its input across init backtracking

### LINKED ISSUE #3227: `reduce` and `foreach` with backtracking init have observable `DUPN`

itchyny:
`reduce f as $x (init; update)` evaluated `f` against a stack slot
that forkpoints created while evaluating `init` restore. Both the
`DUPN` opening the reduction loop and the `LOADVN` retrieving the
result pop that slot destructively, nulling it out, so an `init`
that backtracks (e.g. `0, 0`) made `f` see `null` and fail with
"Cannot iterate over null".

Duplicate the input before the `FORK` instead of inside the loop,
and drop the leftover copy with `POP` once the loop is exhausted.
The slot left live at the `FORK` is then a block pushed after
`init` completed, which no earlier forkpoint can reach, so nulling
it is harmless. This is the same approach `gen_foreach` has used
since c70ae1a.

`DUPN` had no other user, so remove the opcode.

Fixes #3227 and closes #3611, closes #2059.


tests/jq.test
```diff
@@ -2534,7 +2534,13 @@ try ["OK", setpath([[1]]; 1)] catch ["KO", .]
 []
 ["KO","Cannot update field at array index of array"]
 
-# regression test for #3227
+# regression test for #3227 (reduce)
+[] | reduce .[] as $x (0, 0; .)
```
wader: maybe `[] | reduce .[] as $x (0,1,2; .)` to make sure init outputs are used correctly? eg not accidentally reused

wader (approved):

itchyny:
@kanwren What do you think?

kanwren:
Yep, this is right! And I'm happy to see `DUPN` removed.

I do believe that for simple init clauses, this technically now keeps a ref to the input dot until the end of the loop, whereas it did not before. This is, of course, neither problematic nor possible to correctly avoid (without, like, walking the stack or something).

(I had previously thought that doing exactly this fix would reintroduce the issue 7243989c52cea7933ace6535eae914c19eb1210a was addressing, but looking at it now I had simply misunderstood the context around the leak they were describing)

kanwren (approved):

itchyny:
Thank you!

PR #3635, opened 2026-09-18
website: fold older versions on the download page

itchyny:
Only the latest release stays visible in the binary, checksum and
signature, and source tarball lists; older versions move into
collapsible `<details>` sections. Condense the repetitive per-version
checksum and signature sentences into a single list.


wader (approved):

itchyny:
Thanks!

